Issue #23, opened 2022-04-22
Remove all let declarations


samchon:
It's not functional. Let's replace all of the `let` statements to the `const`.

samchon:
Thanks for your special help, @Funniest and @ltnscp9028

PR #24, opened 2022-04-22
1st trial for the #23

samchon:
Is this functional?


samchon (commented):
The 1st trial.

Thanks for special help, @funniest

PR #25, opened 2022-04-22
Refactor: no let RandomGenerator with reduce x2 faster

ltnscp9028:
Refactor: no let RandomGenerator with reduce x2 faster


samchon (approved):
Good idea. I've forgotten the `reduce`.

PR #29, opened 2022-04-23
Fix #28

### LINKED ISSUE #28: When `Object.toJSON()` returns atomic value, it always be the `null`.
Analyzing the `fast-json-stringify` library, I found the reason why those test function occur erros.

The reason was on return type of the `Object.JSON()` methods in those test functions. All of them are returning the atomic value and `fast-json-stringify` does only calls the `Object.JSON()` method when the JSON schema type is object. In such reason, atomic typed JSON schema like below test codes occurs the JSON string conversion error.

Therefore, when `Object.toJSON()` returns an atomic typed value, just use the `JSON.stringify()` function.

  - [test_stringify_to_json_class.ts](https://github.com/samchon/typescript-json/tree/abf3866d32b0aa86e2de471d62e681ca57fc8813/test/features/test_stringify_to_json_class.ts)
  - [test_stringify_to_json_class_closure.ts](https://github.com/samchon/typescript-json/tree/abf3866d32b0aa86e2de471d62e681ca57fc8813/test/features/test_stringify_to_json_class_closure.ts)
  - [test_stringify_to_json_object_closure.ts](https://github.com/samchon/typescript-json/tree/abf3866d32b0aa86e2de471d62e681ca57fc8813/test/features/test_stringify_to_json_object_closure.ts)

samchon:
Perfect supporting for the `Object.toJSON()` method.


samchon (commented):
OK

PR #36, opened 2022-05-02
Close #35

### LINKED ISSUE #35: Naming convension of the `components.schemas`
Name of the `components.schemas` to follow its origin type name and add incremental value only when duplicated type name exists.

However, as the JSON schema pointer needs duplicate definition for both nullable and non-nullable types, the naming convension always include the nullable symbol like `ISystem.ICommit.NonNullable`. Also, when a target type has `<` and `>` syntax because it has generic parameter(s), change the `<` and `>` words to be compatible with URL patterns by replacing keywords.

samchon:
Naming convention rule has been determined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "2.0.9-dev.20220503-3",
+  "version": "2.0.9",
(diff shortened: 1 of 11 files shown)


samchon (commented):
OK, very well.

Issue #42, opened 2022-05-13
Top level `oneOf` property always be considered as a `any` type.


samchon:
Implementing the #41, I've found the bug.

The bug has been reported to the `fast-json-stringify`.

https://github.com/fastify/fast-json-stringify/issues/413

samchon:
No way to solve this bug through `fast-json-stringify`.

The `v3` upate would fix this - #68.

PR #43, opened 2022-05-13
Fix #41 and Report #42

### LINKED ISSUE #41: Use `oneOf` instead of the `anyOf`
When union type comes, there's no difference between them, only within framework of the `JSON.stringify()` function booster.

However, when user of this `typescript-json` imports the `SchemaFactory` by itself, it would be a matter, as `oneOf` is the exact use case for that union type. Therefore, change the `SchemaFactory` to use the `oneOf` statement instead of the `anyOf`

samchon:
Change `anyOf` to the `oneOf`  and report the top level `oneOf` bug to the `fast-json-stringify`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ const stringify = fast({
                         nullable: false,
                     },
                     gender: {
-                        anyOf: [
+                        oneOf: [
                             { type: "string" },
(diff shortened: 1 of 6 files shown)


samchon (commented):
OK

PR #57, opened 2022-05-26
Fix #56

### LINKED ISSUE #56: `IJsonComponent.IObject.required` to be optional
When using `TSON.createSchema()` for the `swagger.json` generation, empty valued `IJsonComponent.IObject.required` becomes an error (warning level). No matter to building entire `swagger.json`, but the error message seems annoying. Therefore, change the `IJsonComponent.IObject.required` type to be optional and make it to be undefined for the empty array.

samchon:
For the 3rd party library, for the swagger generation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "2.1.4",
+  "version": "2.1.5",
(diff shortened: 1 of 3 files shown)


samchon (commented):
Nicely done.

PR #67, opened 2022-06-05
V2.1.6 update

samchon:
v2.1.6 update.

  - Support tuple type
  - Support constant value type
  - Support enumeration type
  - Fix error when template literal type on generic type


samchon (commented):
OK, very well.

PR #106, opened 2022-06-26
v3.0.2, `isFinite()` and `isNaN()`

samchon:
By checking `isFinite()` and `isNaN()`, `typescript-json` becomes not 10x faster but 5x times faster.

However, this patch would make `typescript-json` much stable.


samchon (commented):
No problem on test automation program, therefore accept this.

PR #112, opened 2022-06-26
Close #111 and Complement #109

### LINKED ISSUE #111: Let `assertType()` and `is()` to check `isNaN()` and `isFinite()`
Although `typeof` statement of a variable returns `number`, don't believe it.

Check the `number` value with `isNaN()` and `isFinite()` again.

samchon:
Designing policy for Infinity and NaN values are really difficult.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.0.3",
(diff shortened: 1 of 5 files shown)


samchon (commented):
OK

PR #116, opened 2022-06-27
v3.0.5, have conquered extremely complicate union type

samchon:
This horrible structure would make `typescript-json` much safer.

https://github.com/samchon/typescript-json/blob/55ada286cc95d54866a716468ba425a29c947dc4/test/structures/ComplicateUnion.ts#L3-L30


samchon (commented):
very well, let's conquer the fxxcking union type.

PR #132, opened 2022-07-01
V3.0.7, performance tuning

samchon:
Make main functions faster and add a new property `TypeGuardError.expected`


samchon (commented):
OK, no problem.

PR #134, opened 2022-07-03
Close #133

### LINKED ISSUE #133: Use `boolean` type instead of `false`|`true` type
Although when a `boolean` type has been declared, TypeScript considers it as a `false`|`true` type for performance reason.

Of course, if `typescript-json` follows the rule, `typescript-json` also can take advantage in the performance aspect.

However, it makes tracing `TypeGuardError.path` hard.

Therefore, when `false`|`true` type comes from TypeScript, convert it to a `boolean` type.

samchon:
Consider `false|`true` type as `boolean` type.


samchon (commented):
Test OK, therefore fine

PR #138, opened 2022-07-04
Close #135, Fix #136 and Close #137

### LINKED ISSUE #135: Enhance benchmark program
Let benchmark program to generate markdown documents under `benchmark/resuilts`, with its CPU name.

### LINKED ISSUE #136: Bug on `JsonTypeChecker.isAtomic()`
It always returns `true`.

### LINKED ISSUE #137: Optimize `assertType` function
Instead of throwing `TypeGuardError` exception through `if ~ else` condition, it would better to create a function that throwing the exception with closure. When the throwing exception statement being capsuled in the closure function, I think that enhancing performance would be possible.

samchon:
- Performance of `assertType()` be enhanced.
  - Benchmark program generates much detailed reports with `mermaid` diagram
  - Bug on `JsonTypeChecker` has been fixed, but it's not formally supported feature


samchon (commented):
Nice

PR #145, opened 2022-07-05
Close #142, Close #143 and Fix #144

### LINKED ISSUE #142: Refactor `AssertProgram.generate()` function
For the 3rd party library, change `AssertProgram.generate()` function like below:

## Before
```typescript
export namespace AssertProgrammer {
    export function generate(
        project: IProject, 
        modulo: ts.LeftHandSideExpression
    ): ts.ArrowFunction;
}
```

## After
```typescript
export namespace AssertProgrammer {
    export const generate =
        (project: IProject, modulo: ts.LeftHandSideExpression) =>
        (type: ts.Type) => ts.ArrowFuntion;
}
```

### LINKED ISSUE #143: Runtime type checkers to test `!Array.isArray()` to object type
When there's an object type which has only a `length` property and `Array` comes for the checker, it becomes `true`.

Therefore, when inspecting object, checking `!Array.isArray(input)` is required.

### LINKED ISSUE #144: `TypeGuardError.expected` value is not exact sometimes
When such dynamic string literal property comes, `TypeGuardError.exepcted` value is not exact.

## Structure
```typescript
export interface ISomething {
    "something-interesting-do-you-want?": string;
    ["or-something-crazy-do-you-want?"]: string;
}
```

## TypeGuardError
```typescript
{
  method: 'TSON.assertType',
  path: '$input[something-interesting-do-you-want?]',
  expected: 'string',
  value: {
    'something-interesting-do-you-want?': {},
    'or-something-crazy-do-you-want?': 'nope'
  }
}
```

samchon:
New version is `v3.0.10`. Maybe, next `v3.0.11` would be last version of `v3.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.0.9",
+  "version": "3.0.10",
   "description": "Runtime type checkers and 5x faster JSON.stringify() function",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/src/TypeGuardError.ts
+++ b/src/TypeGuardError.ts
@@ -8,7 +8,7 @@ export class TypeGuardError extends Error {
         // MESSAGE CONSTRUCTION
         super(
             props.message ||
-                `Error on TSON.${props.method}(): invalid type${
+                `Error on ${props.method}(): invalid type${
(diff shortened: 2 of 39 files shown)


samchon (commented):
Good

PR #154, opened 2022-07-15
V3.1, new function `validate()`

samchon:
A new function `validate()` has come.

It archives all type erros into `IValidation.errors`.


samchon (commented):
OK

PR #167, opened 2022-07-16
Test: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz

Yaminyam:
environment: 2018 Macbook Pro 15'


samchon (approved):
Thanks for contribution

PR #185, opened 2022-09-12
Fix #184, validators support unknown type

### LINKED ISSUE #184: `assertType` and `is` should accept `any` or `unknown`
## Feature Request
Screenshot 2022-09-12 at 12 45 44

The runtime type checking methods `assertType` and `is` work fine when the input value is `any` but if you're trying to validate a value that is one of a union, it doesn't work...

I believe the typings should be changed to:
```js
export declare function is<T>(input: unknown): input is T;
```

This is how `typescript-is` behaves.

The workaround with the current type is to manually typecast everything:
```js
assertType<string>(v as string);
```

cc @samchon

samchon:
Suggestion by @paulrostorp, support `unknown type`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.2.2",
+  "version": "3.2.3",
   "description": "Runtime type checkers and 5x faster JSON.stringify() function",
   "main": "lib/index.js",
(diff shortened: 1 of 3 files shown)


samchon (commented):
OK

paulrostorp (approved):

PR #195, opened 2022-09-18
Close #187, support equivalent validator function

### LINKED ISSUE #187: Add new functions `equals`, `assertEquals` and `validateEquals`
## Question

Is there a way to check if a type has no excess properties for example. the following should throw
```typescript
import TSON from "../../../src";

TSON.assertType<{ a: true }>({ a: true, b: false });
```

samchon:
Below functions are newly added:

  - `TSON.equals()`
  - `TSON.assertEquals()`
  - `TSON.validateEquals()`


samchon (commented):
Many things have been changed, but no problem because test program is working well.

PR #198, opened 2022-09-19
Fix #197 - no superfluous property inspection when `undefined` value comes

### LINKED ISSUE #197: Do not inspect superfluous property when `undefined` value comes
```typescript
interface IPerson {
    name: string;
    age: number;
}

const person = {
    name: "Jeongho Nam",
    age: 34,
    account: undefined, // superfluous property
};

TSON.equals<IPerson>(person); // -> false, even block undefined value
```

Implementing equivalent validator functions (#187), I'd not considered undefined valued property.

Even now, it would better to skip the superfluous property inspection when `undefined` value comes.

samchon:
From no on, superfluous property inspection would not check the `undefined` value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.2.6",
+  "version": "3.2.7",
   "description": "Runtime type checkers and 5x faster JSON.stringify() function",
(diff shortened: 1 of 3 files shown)


samchon (commented):
OK

Issue #211, opened 2022-09-25
Support the `@default` tag


Inkdpixels:
## Feature Request

Really awesome work on this! There are just two things that I miss, one is the following feature: Support of the `@default` tag.

As the library already supports quite common JSDoc tags such as `@minimum` etc as per https://github.com/samchon/typescript-json/issues/175 I think it would be a great addition. This way server side validation can be properly done for optional values where the validator supports infering them, e.g. AJV with the `useDefaults` option.

https://ajv.js.org/options.html#usedefaults

```ts
type Spec = {
  /**
    * @default "foo"
    */
  property: string
}
const schema = TSON.application<[Spec], "swagger">();  // { schemas: [{ type: "object", properties: { property: { type: "string", default: "foo" }} }] }
```

samchon:
I've not seen using `"` symbol on jsDoc tag. Therefore, it seems `@default foo` is better than `@default "foo"`.

If I'm taking a mistake, please inform me.

If agree to using `@default foo`, I'll implement this until Wednesday.

samchon:
@Inkdpixels Look at those files, then you may understand the `@default` tag has been exactly implemented.

  - https://github.com/samchon/typescript-json/blob/19bccd9bc8a2cc888fb82a602f22cb3c17506536/test/structures/TagDefault.ts
  - https://github.com/samchon/typescript-json/blob/19bccd9bc8a2cc888fb82a602f22cb3c17506536/test/features/application/test_application_tag_default.ts

PR #213, opened 2022-09-26
feat: support arbitrary @format tag values #212

### LINKED ISSUE #212: Support arbitrary format values on the `@format` tag
## Feature Request
Really awesome work on this! There are just two things that I miss, one is the following: Support of arbitrary format values on the @format tag.

E.g. if one would be generating the json-schema for ajv with this library, the developer should be able to use custom formats.

https://github.com/ajv-validator/ajv-formats

```ts
type Spec = {
  /**
    * @format "date"
    */
  property: string
}
```

Inkdpixels:
Closes #212


src/factories/MetadataTagFactory.ts
```diff
@@ -171,13 +171,9 @@ const PARSER: Record<
     ----------------------------------------------------------- */
     format: (identifier, metadata, value, output) => {
         validate(identifier, metadata, output, "format", "string", ["pattern"]);
-        if (FORMATS.has(value) === false)
```
Inkdpixels, line 174: Since the value can be `arbitrary` from now on, this check is not needed anymore

src/programmers/helpers/ITagFormatValidator.ts
```diff
@@ -0,0 +1,9 @@
+import { IMetadataTag } from "../../metadata/IMetadataTag";
+
+export namespace ITagFormatValidator {
```
Inkdpixels, line 3: Not sure if this is placed right here, just saw that the folder holds some functionality that is being re-used in `internals`.

src/programmers/internal/check_string_tags.ts
```diff
@@ -5,14 +5,18 @@ import { IdentifierFactory } from "../../factories/IdentifierFactory";
...
 
 export const check_string_tags =
     (importer: FunctionImporter) =>
     (input: ts.Expression, tagList: IMetadataTag[]) => {
         const conditions: ts.Expression[] = [];
         for (const tag of tagList)
-            if (tag.kind === "format")
+            if (
+                tag.kind === "format" &&
+                ITagFormatValidator.isCustomFormat(tag) === false
```
Inkdpixels, line 18: This TS check can also be skipped as far as I could understand if the value is not one of the known formats that have a built-in validator

samchon (changes requested):
When not supported format tag, it should not enrolled to `Metadata.metaTags`. Only `Metadata.jsDocTags` must be used.

Also, I think the reason why you want the arbitrary `@format` tag is for JSON schema's `default` property. Therefore, you'd better implement such `@format date` property assignment on `ApplicationProgrammer`.

If you don't have enough time to change those things, I'll accept and refactor your code on Saturday.

Inkdpixels:
@samchon Ahh alright, that totally makes sense. So this also means that the restrictions for `@format` would still be in place and the arbitrary `@format` values would only be supported when using the `TSON.application()` API?

Inkdpixels:
@samchon Just pushed a revisited implementation with the feedback, let me know if I understood it correctly. This change would ignore arbitrary `@format` values in the metadata, but keep them in ...

- ... the JSDoc 
- ... the top level schema (to allow validators such as AJV accessing them)

samchon (approved):
I'll publish your PR with #211 implementation, maybe Saturday. Also, reading your requirement, I think that `typescript-json` needs to supporting all of the `@format` tag of JSON schema. Thanks for your contribution. Thanks for your contribution.

PR #215, opened 2022-10-01
Close #211, support `@default` tag for JSON schema

### LINKED ISSUE #211: Support the `@default` tag

samchon:

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```


samchon (commented):
OK

PR #217, opened 2022-10-02
Fix #216 - json schema's pattern does use `/`

### LINKED ISSUE #216: JSON schema's pattern must not include `/` symbol
https://github.com/samchon/typescript-json/blob/77b54e7386fe3336dad8dbd86f6fb622290f250d/test/features/application/test_application_template_atomic.ts#L18-L60

Looking at JSON schema definition, its pattern does not include the `/` symbol.

Therefore, `ApplicationProgrammer` must be reformed not to generate the `/` symbol.

samchon:
Follow JSON schema's standard rule

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.3.1",
+  "version": "3.3.2",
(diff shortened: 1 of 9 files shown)


samchon (commented):
OK

PR #227, opened 2022-10-07
Fix #226 - consider meta-programmed class type

### LINKED ISSUE #226: Using partial with declare class doesn't work as expected
### Summary
I wanted to validate on a DeepPartial<T>, but the validation doesn't work as expected when using a type via `declare class`. It seems to work as expected if I use an interface. I included a repl that shows the issue. However, I'm trying to use a type from a third party package that defines its types via declare class.

### Code occuring the bug
Reproducible demo:
https://replit.com/@TJHiggins1/TSON-Partial-Bug#index.ts

```ts
import TSON from "typescript-json";

type DeepPartial<T> = T extends object ? {
  [P in keyof T]?: DeepPartial<T[P]>;
} : T;

const input = { a: 'valid' }

console.log('Works with interface:')

interface TestInterface {
  a: string;
}

const res = TSON.validateEquals<TestInterface>(input);
console.log('valid:', res)

const res2 = TSON.validateEquals<Partial<TestInterface>>(input);
console.log('valid:', res2)

const res3 = TSON.validateEquals<DeepPartial<TestInterface>>(input);
console.log('valid:', res3)

console.log('')
console.log('---------------------------')
console.log('')

console.log('Does not work with declare class:')

declare class TestClass {
  a: string;
}

const res4 = TSON.validateEquals<TestClass>(input);
console.log('valid:', res4)

const res5 = TSON.validateEquals<Partial<TestClass>>(input);
console.log('should be valid:', res5)

const res6 = TSON.validateEquals<DeepPartial<TestClass>>(input);
console.log('should be valid:', res6)
```

samchon:
Thanks @tjhiggins

Now, meta-programmed class type like `Partial<SomeClass>` would not occur any bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.3.6",
+  "version": "3.3.7",
   "description": "Runtime type checkers and 5x faster JSON.stringify() function",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/src/factories/internal/emplace_metadata_object.ts
+++ b/src/factories/internal/emplace_metadata_object.ts
@@ -30,6 +30,7 @@ export const emplace_metadata_object =
         const pred: (node: ts.Declaration) => boolean = isClass
             ? (node) => ts.isParameter(node) || ts.isPropertyDeclaration(node)
             : (node) =>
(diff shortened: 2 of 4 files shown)


samchon (commented):
OK

PR #229, opened 2022-10-08
Close #228 - benchmark of `io-ts` and `zod`

### LINKED ISSUE #228: Revise benchmark (add `io-ts`, `zod` and remove `typescript-is`)
`typescript-is` can't use latest TypeScript (over 4.8) because its author stopped maintaining. Therefore, if I don't remove the `typescript-is` from benchmark program of `typescript-json`, this repository never can update `devDependencies` to the latest TypeScript version. It's the reason why removing `typescript-is` from benchmark.

Also, `zod` and `io-ts` is much famous than what I've known. Therefore, add those validators instead of the `typescript-is`.

samchon:
Benchmark has been completed.


samchon (commented):
OK

samchon (commented):
More benchmarks, OK

PR #231, opened 2022-10-08
Close #230 - how to install on `vite`

### LINKED ISSUE #230: Introduce how to install on `vite`
Developing front application, I understood that `vite` is providing the easiest way to custom configuration.

Therefore, `typescript-json` also needs to introduce about it.

samchon:
I think that `vite` is the best solution for front application configuration.


samchon (commented):
OK

PR #235, opened 2022-10-10
Fix #234 - invalid `TypeGuardError.path` bug

### LINKED ISSUE #234: Bug on `TypeGuardError.path` when dynamic propertied array comes
## Bug Report
Note that, the bug you're reporting may have registered in the [issues](https://github.com/samchon/typescript-json/search?type=issues) by another user. Even the bug you're reporting may have been fixed in the `@next` version. In such reasons, I recommend you to check the old [issues](https://github.com/samchon/typescript-json/search?type=issues) and reproduct your code with the `@next` version before publishing the bug reporting issue.

```bash
`npm install --save-dev typescript-json@next`
```

When the same error occurs even in the `@next` version, then please fill the below template:

### Summary
Invalid programming like below when dynamic propertied array comes

image

  - **SDK Version**: 3.3.8
  - **Expected behavior**: Exact programming
  - **Actual behavior**: Wrong programming

The `TypeGuardError.path` value must be combined like below:

```typescript
path + $join(key) + "[" + index9 + "]
```

### Code occuring the bug
```typescript
import TSON from "typescript-json";

interface DynamicArray {
    [key: string]: string[];
}

TSON.assertType<DynamicArray>({});
```

samchon:
Rare case, but emergent fix may required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "3.3.8",
+  "version": "3.3.9",
   "description": "Runtime type checkers and 5x faster JSON.stringify() function",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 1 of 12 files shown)


samchon (commented):
OK, dynamic propertied array was the only one not being covered by test program.

From now on, validation error may not be occured.

PR #277, opened 2022-10-28
Close #268 - refactored entire benchmark program for TypeBox

### LINKED ISSUE #268: Benchmark `equals()` functions
Special benchmark program for TypeBox (@sinclairzx81)

Those strict validation functions would be benchmarked:

  - `equals()`
  - `assertEquals()`
  - `validateEquals()`

samchon:
@sinclairzx81 Refactored entire benchmark program and it would be very suitable for `TypeBox`.

Especially, concentrated on `assertType()` function and make it to occurs type error always. As you know, if assert only one object, measurement result would be weirdo because property access sequence of validator libraries are all different. Therefore, target values for the `assertType()` benchmark are wrapped by an `Array`.


sinclairzx81:
@samchon Awesome!

Yeah, happy to see heaps of error data run for the `assert` test (as that would run invoke both `.Check()` and `.Errors()` and be a more comparable test overall). It think having error cases would also be able to highlight validation disparities across the libraries being run (which is a good REALLY good thing to help encourage validation alignment)

Nice work!

PS: You should totally consider setting up these benchmarks as a standalone project. Currently, only https://github.com/moltar/typescript-runtime-type-benchmarks exists in this space, and it only asserts on simple structures, and only for passing test cases. It also doesn't make distinctions between AOT, JIT and Dynamic validation. So there's room for innovation there.

Would be an awesome reference point for developers.

samchon:
I'll make a new PR and will show how benchmark result has been changed. 

Especially, the new benchmark program is extremely optimized for the JIT environment. 

Therefore, dramatic enhancement on your `TypeBox` would be.

PR #296, opened 2022-11-03
Close #250 - support more JSON schema properties

### LINKED ISSUE #250: Support generic keywords
# Support generic keywords 

##  Description

JSON Schema allows to use into schema definition generic keywords (https://json-schema.org/understanding-json-schema/reference/generic.html?highlight=description)

for example:

- title
- description
- examples
- deprecated

and so on.

i was looking into 

https://github.com/samchon/typescript-json/blob/c709379db511b13b334ae9d4abdc0923033a1797/src/programmers/internal/application_schema.ts#L21

and i noticed that none of the keywords above can be "injected" into the schema genearation using `TSON.application`.

as specified in

https://swagger.io/specification/#schema-object

the supported properties for the schema definition are an extended Subset of `JSON Schema Draft 00`

## Feature Request

would be nice to have the capabilities to generate thoose props directly with the jsDoc interface, right now for example using `@description` the generated schema is the following

interface:

```typescript
export interface User{
    /**
     * @description email
     */
    email: string
}
```

generated:

```javascript
email: {
    type: 'string',
    nullable: false,
    jsDocTags: [
        {
            name: 'description',
            text: [ { text: 'email', kind: 'text' } ]
        }
    ]
}
```

instead should be generated in this way:

```javascript
email: {
    type: 'string',
    nullable: false,
    description: "email"
}
```

samchon:


samchon (commented):
No problem.

Issue #298, opened 2022-11-03
Help me benchmark on your device


samchon:
I want to measure benchmark on multiple devices.

If you want to help this `typescript-json` project, fork, clone and run below commands. After running the benchmark program, a new directory would be created under [benchmark/results](https://github.com/samchon/typescript-json/tree/master/benchmark/results) following your CPU name. Check the newly generated markdown file and send PR with your new commit.

Whe you're running benchmark on notebook, please process the benchmark program on the charging state. Also, all the benchmarked CPUs are listed up into here: [samchon/typescript-json/benchmark/results](https://github.com/samchon/typescript-json/tree/master/benchmark/results). It's not forbidden to sending duplicated CPU benchmark PR, such duplicated CPU benchmark would be accepted only when the new benchmark performance is better than previous.

  1. Fork
  2. Clone
  3. `npm install`
  4. `npm run benchmark`
  5. Send PR

```bash
git clone https://github.com/user-github-account/typescript-json
cd typescript-json
npm install
npm run benchmark
```

For reference, the benchmark takes about 30 minutes.

Below is an example of the benchmark output, measured on [AMD R7 5800H](https://github.com/samchon/typescript-json/tree/master/benchmark/results/AMD%20Ryzen%207%205800H%20with%20Radeon%20Graphics#is)

![Is Function Benchmark](https://github.com/samchon/typescript-json/raw/master/benchmark/results/AMD%20Ryzen%207%205800H%20with%20Radeon%20Graphics/images/is.svg)

PR #303, opened 2022-11-04
Apple M1 Pro 8-core Benchmark

incheon-kim:
M1 Pro 8-core, 32GB ran on node.js (v16.15.1 - ARM64) #298 

If it something is gone wrong, please let me know.


samchon (approved):
Thanks for contribution

PR #304, opened 2022-11-04
Benchmark with 5900x #298

Thinline20:
#298  Benchmark with Ryzen 5900x and 16GB of memory


samchon (approved):
Thanks for contribution.

PR #305, opened 2022-11-04
Add benchmark result for Ryzen 5600X

michaelhur:
Ryzen 5600X, Ubuntu 20.04, 32GB RAM, node v18.9.0


samchon (approved):
Thanks for contribution.

PR #306, opened 2022-11-05
Benchmark with Intel i5-12600K

jongwoo328:
#298


samchon (approved):
Thanks for contribution

PR #309, opened 2022-11-05
benchmark with apple m1 max

heli-os:
motivation:

- #298 

modifications:

add benchmark result with apple m1 max


samchon (changes requested):
https://github.com/samchon/typescript-json/tree/master/benchmark/results/Apple%20M1%20Pro

Your M1 max is much slower than M1 pro (even slower than old intel processors).

I assume that you've runned the benchmark program during heavy programs are running.

May you measure it again? If same result comes, you should repair or refund it.

heli-os:
@samchon
update benchmark after system reboot and cleanup.

Screen Shot 2022-11-05 at 3 50 15 PM

---

as far as i know, the difference is computing performance between m1-pro and m1-max is insignificant.

there is just a difference in GPU performance and RAM bandwidth.

ref: https://www.apple.com/kr/newsroom/2021/10/introducing-m1-pro-and-m1-max-the-most-powerful-chips-apple-has-ever-built/

samchon (approved):
Thanks for contribution. It still slower than M1 pro, but no matter.

PR #310, opened 2022-11-05
Benchmark with Apple M1 Max

redsubmarine:
Detail
======
 - #298
 - CPU: Apple M1 Max
 - Memory: 64 GB
 - OS: macOS 13.0(22A380)
 - TypeScript-JSON version: 3.3.20


samchon (approved):
Thanks for contribution. 

I accept your PR instead, because your M1 max is much faster than before.

PR #311, opened 2022-11-05
etc: BENCHMARK WITH INTEL I5-9300H

yoonjonglyu:
ref #298 

Intel i5-9300H


samchon (approved):
Thanks for contribution.

PR #312, opened 2022-11-05
Benchmark with Intel i5-1135G7

mopil:
# Detail
- https://github.com/samchon/typescript-json/issues/298
- CPU: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
- Memory: 16GB
- OS: Windows 11
- TypeScript-JSON version: 3.3.20


samchon (commented):
I'm sorry for denial. Your benchmark is slower than before.

PR #318, opened 2022-11-07
Benchmark with Intel(R) Core(TM) i7-1165G7

limejuny:
Detail
======
 - #298
 - CPU: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
 - Memory: 16,064 MB
 - OS: win32
 - TypeScript-JSON version: 3.3.21


samchon (approved):
Thx for contribution

PR #319, opened 2022-11-07
Benchmark on 3700x

SH5H:
https://github.com/samchon/typescript-json/issues/298

add benchmark on 3700x


samchon (approved):
Thanks for contribution.

PR #326, opened 2022-11-12
Close #325 - trace exact alias type name

### LINKED ISSUE #325: Trace exact alias type name
I found a way to trace exact alias type name.

From now on, below alias type would not be named as `__type` more.

```typescript
type Alias = { a: number; };
```

samchon:


samchon (commented):
OK

PR #341, opened 2022-11-14
CPU: 11th Gen Intel(R) Core(TM) i7-11390H @ 3.40GHz

enghitalo:

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```


enghitalo:
@samchon I don't know why it is so different in https://github.com/enghitalo/Utils/blob/main/typescript/src/benchmark-JSON-msgpack.md.
What I am doing wrong?

samchon:
The `Array<any>` type makes `typescript-json` extremely slower. I should patch it.

Also, `ts-node` makes `typescript-json` slower too. It would better to run compiled JS file.

## Before
```javascript
"array0":${`[${input.array0.map(elem => undefined !== elem ? JSON.stringify(elem) : "null").join(",")}]`}
```

## After
```javascript
"array0: " + JSON.stringify(input.array0)
```

samchon (approved):
Thanks for contribution.

PR #459, opened 2023-01-07
Fix #458 - prohibit zero length tuple type on swagger

### LINKED ISSUE #458: Prohibit zero length tuple type on `swagger.json`
When generating JSON schema application through `typia.application()` function, zero length tuple type must be prohibitted because the swagger documents spec does not support the tuple type. It is possible to define the zero length tuple type as an Array type and limit its length only to be zero, but it is nonsensible in logic level, because the Array's element type would be any.

Therefore, when zero length tuple type comes with swagger mode, just throw compilation error.

samchon:

```bash
# COMPILE TEST AUTOMATION PROGRAM
npm run build:test

npm run test
```


samchon (commented):
OK

Issue #337, opened 2022-11-14
Help me benchmark on your device


samchon:
By #332 update, `typia` became much faster. Therefore, I want to re-measure benchmark on multiple devices.

If you want to help this `typia` project, fork, clone and run below commands. After running the benchmark program, a new directory would be created under [benchmark/results](https://github.com/samchon/typia/tree/master/benchmark/results) following your CPU name. Check the newly generated markdown file and send PR with your new commit.

Whe you're running benchmark on notebook, please process the benchmark program on the charging state. Also, all the benchmarked CPUs are listed up into here: [samchon/typia/benchmark/results](https://github.com/samchon/typia/tree/master/benchmark/results). It's not forbidden to sending duplicated CPU benchmark PR, such duplicated CPU benchmark would be accepted only when the new benchmark performance is better than previous.

  1. Fork
  2. Clone
  3. `npm install`
  4. `npm run benchmark`
  5. Send PR

```bash
git clone https://github.com/user-github-account/typia
cd typia
npm install
npm run benchmark
```

For reference, the benchmark takes about 30 minutes.

Below is an example of the benchmark output, measured on [Intel i5-1135g7, Surface Pro 8](https://github.com/samchon/typia/tree/master/benchmark/results/11th%20Gen%20Intel(R)%20Core(TM)%20i5-1135G7%20%40%202.40GHz#is)

![Is Function Benchmark](https://github.com/samchon/typia/raw/master/benchmark/results/11th%20Gen%20Intel(R)%20Core(TM)%20i5-1135G7%20%40%202.40GHz/images/is.svg)

kakasoo:
```bash
TSError: ⨯ Unable to compile TypeScript:
benchmark/internal/AssertBenchmarker.ts:36:33 - error TS2538: Type 'undefined' cannot be used as an index type.

36                 (output as any)[elem.name] = elem.count / elem.times.elapsed;
                                   ~~~~~~~~~
```

As a result of executing the command shown above, an error occurs in 'npm run benchmark'. Below is the error code, but I hid the error code because I didn't want my folder structure to be revealed.
If you still want to get the benchmark results, I think you should check it out.

samchon:
@kakasoo No `AssertBenchmarker` class exists more.

I think you're running on previous version of `typia` (maybe 4 months ago). 

What about getting latest commit?

kakasoo:
@samchon 

[...]

PR #463, opened 2023-01-14
Add 12th Gen Intel(R) Core(TM) i7-12700H results

omizha:
Related to https://github.com/samchon/typia/issues/337


samchon (approved):
Thanks for your measurement.

PR #474, opened 2023-02-02
Benchmark 5950X

SK-FComputer:
Created benchmark on AMD Ryzen 9 5950X 16-Core Processor
Background work while running, so not fully clean.

Might do after work


samchon:
Even slower than my surface pro 8 (i5).

Possible to run again in clean env?

SK-FComputer:
Ran again with clean ENV:
Stopped all background/foreground programs, docker containers etc.

Not much gained by that in terms of raw 1:1 performance

samchon (approved):
Thanks for your contribution.

Please understand if some faster 5950x comes, then the benchmark result can be overwritten.

SK-FComputer:
Taking a quick look at current results:
Seems like Typia prefers Windows, and Intel processors

If benchmarks is correct, some fluctuate very much

samchon:
It is not a benchmark with strict variable control, but a benchmark that PR just believes and accepts, so reliability may be somewhat lower.

PR #555, opened 2023-03-16
Fix samchon/nestia#282 - too long type name

samchon:
Issue of nestia, but related.


samchon (commented):
OK, seems no problem.

PR #559, opened 2023-03-18
Benchmark on Apple M1 Pro

kakasoo:
I tested the latest version of typia.


samchon (approved):
Thanks for measurement.

Issue #560, opened 2023-03-20
Support custom validation through comment tags


samchon:
From v3.7 update, `typia` will support custom validation functions through comment tags. 

You can utilize the custom validation like below:

```ts
typia.customValidators.insert("powerOf")("number")(
    (text: string) => {
        const denominator: number = Math.log(Number(text));
        return (value: number) => {
            value = Math.log(value) / denominator;
            return value === Math.floor(value);
        };
    }
);
typia.customValidators.insert("dollar")("string")(
    () => (value: string) => value.startsWith("$"),
);

interface TagCustom {
   /**
    * @powerOf 10
    */
   powerOf: number;

   /**
    * @dollar
    */
   dollar: string;
}
```

samchon:
There had been lots of people who wanted custom validation tools.

However, it was hard thing for me because `typia` constructs validation logics in the compilation level. While other competitive libraries have supported custom validation tools easily by constructing validation logics in the runtime level, `typia` had to consider how to unify both compilation and runtime levels.

In nowadays, @TheYarin had suggested me an idea about this custom validation tool supporting - #550. I couldn't accept the issue because cannot sure which file or statement (validate function call vs adding custom logic) would come earlier. However, I could determine custom validation spec by getting hints from his idea. I specially thanks to @TheYarin.

samchon:
Anyway, I repeat that, above design is just a plan and not fully determined.

If you have any good idea about detailed interface, please let me know by writing a comment under here.

TheYarin:
The API you suggested feels really cumbersome compared to typia's clean API: nested function definitions and calls, hardcoded type name instead of properly referencing the type...
I do like the ability to add new, custom validation tags that take parameters though. That can make the type definition pretty clean, but on the other hand anyone looking at this custom tag won't be able to find it in typia's documentation and will have no idea what it does.
Another problem with that solution is, it does not allow you to define a custom validator for a type, but only for a field within a type.
Also, I'm not sure this solution will handle custom types, as the type name is passed as a string and not as a proper reference. (What if there are multiple types with the same name in the project?)

My personal (**biased**) opinion is that a custom validator function that is attached to a type will be more useful.

Now, you mentioned the problem is you need to know at compile time that a certain type has a custom validator, but you can't know that for sure because of the random order of the files being compiled - typia might be getting a validate call before it gets the custom validator definition. So how can we change that?
One option is to give typia a hint: "the validation definitions are at <path-to-file/s-or-folder-or-glob?>".

[...]

PR #569, opened 2023-04-01
Complement #560 - determine custom validator spec

samchon:
Determine below spec:

```ts
typia.customValidators.insert("powerOf")("number")(
    (text: string) => {
        const denominator: number = Math.log(Number(text));
        return (value: number) => {
            value = Math.log(value) / denominator;
            return value === Math.floor(value);
        };
    }
);
typia.customValidators.insert("dollar")("string")(
    () => (value: string) => value.startsWith("$"),
);

interface TagCustom {
   /**
    * @powerOf 10
    */
   powerOf: number;

   /**
    * @dollar
    */
   dollar: string;
}
```


samchon (commented):
OK, no problem.

PR #582, opened 2023-04-10
Fix #573 - unable to run setup on Macbook + `pnpm`

### LINKED ISSUE #573: Typia generate incorrectly changes relative import path
## Bug Report
Typia generate incorrectly changes relative import path

### Summary
  - **SDK Version**: typia 3.7.1
 
Relative import path is converted to `"interface"` when it should be `"./interface"`.

### Code occuring the bug
```typescript
import typia from "typia";
import { SomeInterface } from "../interface";
```
and
```bash
pnpm typia generate   --input home/src/typia   --output home/src
```
becomes
```typescript
import typia from "typia";
import { SomeInterface } from "interface";
```

samchon:
When run `pnpm` commands on Macbook, it fails.

I've tried lots of alternative ways, but could not fix it.

Therefore, from now on, `npx typia setup` commands cannot be executed alone, and need pre-install like below:

```bash
npm install --save-dev typia
npx typia setup
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Your donation would encourage `typia` development.
 ## Setup
 ### Transformation
 ```bash
+npm install --save typia
 npx typia setup
 ```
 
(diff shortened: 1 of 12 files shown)


samchon (commented):
OK, seems no problem.

PR #585, opened 2023-04-17
Support TypeScript v5

samchon:
Since v5, `typia` utilizes only `ts-patch` transformer, because `ttypescript` had stopped maintaining.

  - https://github.com/nonara/ts-patch/issues/93

Also, since `typia@3.8`, new guide documents (https://typia.io/docs) are provided.


samchon:
Close #587, implemented by previous commit.

nonara:
I see you removed your comment. Did you get it working?

**Edit:**

Never mind. I saw the new one.

samchon:
@nonara Moved to here, because I'd written it in the wrong issue:

https://github.com/nonara/ts-patch/issues/93#issuecomment-1515727954

nonara:
Tsp3 requires >= ts5. But the require failure is unusual. I'll take a look when I can.

If you have a chance, try `3.0.0-beta3`. I'm guessing that will work. 

I will test the non-working commit on this branch when I have a minute

samchon:
Yes, when I fix the `3.0.0-beta3` version, no problem on TS5. 

  - https://github.com/samchon/typia/actions/runs/4751303707/jobs/8440307743

But TS4 still does not work. It can't run the transform even including your `typescript-transform-paths`

nonara:
Thanks for letting me know. V3+ requires ts 5 or higher.

samchon:
@nonara Ah, you mean that not supporting TS 4.9 is spec, got it.

I should configure peerDependencies like that.

PR #601, opened 2023-04-24
Support for tuple type

kakasoo:
Thanks to the hint you gave me, I could solve it easily.
I ran the test automation module with the bash script in the content of the CONTRIBUTING document and it was successful.

```bash
# COMPILE
npm run build

# PREPARE
npm run test:generate

# DO TEST
npm run test
```

### DIFF
--- a/src/Primitive.ts
+++ b/src/Primitive.ts
@@ -58,13 +58,9 @@ type _Primitive<Instance> = _ValueOf<Instance> extends object
         : never // cannot be
     : _ValueOf<Instance>;
 
-type Spread<T extends any[]> = T extends [infer F, ...infer Rest]
-    ? [F, ...Spread<Rest>]
-    : [];
-
 type _PrimitiveObject<Instance extends object> = Instance extends Array<infer T>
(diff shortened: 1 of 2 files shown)


samchon (approved):
Great job

src/Primitive.ts
```diff
@@ -49,7 +59,11 @@ type _Primitive<Instance> = _ValueOf<Instance> extends object
     : _ValueOf<Instance>;
 
 type _PrimitiveObject<Instance extends object> = Instance extends Array<infer T>
-    ? _Primitive<T>[]
+    ? IsTuple<Instance> extends true
+        ? Instance extends [infer F, ...infer Rest]
+            ? [F, ..._PrimitiveObject<Rest>]
```
samchon: It would better to make a new type to reduce vulnerable iteration depths

kakasoo: @samchon 

Thx for your review. But we were going back a long way, leaving an easier way. :)

samchon (approved):
OK, very well.

PR #611, opened 2023-04-26
Support both TS 4.9 and TS 5.0

samchon:
@nonara, author of `ts-patch` said that `ts-patch@3` will support only `typescript@5` version.

Therefore, made setup wizard to install `ts-patch` first, and determine whether to install TS 4 or TS 5, considering version of the `ts-patch`

  - `ts-patch@2`: `typescript@4.9.5`
  - `ts-patch@3`: `typescript@latest`


nonara:
Good work. Thanks for sponsoring!

PR #615, opened 2023-04-28
Setup wizard to configure `strictNullChecks: true`

samchon:
`nest new <directory>` command of `@nestjs/cli` has changed to configures `strictNullChecks: false`. 

In the past, it had configured `strict: false`.

Therefore, changed setup wizard of `typia` to consider `strictNullChecks` option, too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "3.8.1",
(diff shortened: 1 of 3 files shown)


samchon (commented):
OK, no problem.

PR #616, opened 2023-04-28
Use proper vite configuration

zanettea:
Vite by default will still use its eslint parser. We need to disable it and use the forked plugin version to have transformers working

```bash
# COMPILE
npm run build

# PREPARE
npm run test:generate

# DO TEST
npm run test
```

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -125,10 +125,11 @@ Just configure `vite.config.ts` file below, that's all.
 ```typescript copy filename="vite.config.ts"
 import { defineConfig } from 'vite'
 import react from '@vitejs/plugin-react'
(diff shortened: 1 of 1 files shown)


samchon (approved):
Wow, thanks for documentation.

I believe that your suggestion may work.

Therefore, accept and publish yours first, and will check later \o/.

samchon:
If this feature works well, I'll move the vite configuration to be above of generation mode.

  - Current setup sequence
    1. Transform mode
    2. Generation mode
    3. Vite configuration
  - Plan to be
    1. Transform mode
    2. Vite configuration
    3. Generation mode

samchon:
Be changed:

https://typia.io/docs/setup/#vite

PR #626, opened 2023-05-11
Change JSON schema strategy of AJV mode

samchon:
Until previous, `typia` had utilized `nullable` property instead of using union `null` type, because `swagger` (`openapi`) had not supported it. However, since `openapi@3.1` update, it would start supporting the union `null` type. It is not formally supported by Swagger Editor (https://editor.swagger.io), but maybe coming soon.

Therefore, use union typed `null` value first at AJV mode's JSON schema first. About the Swagger mode, I'll support it when `openapi@3.1` be uploaded to the Swagger Editor. Also, remove prefix customization feature, and just assign the fixed value `#/components/schemas`.

### AJV Mode
```json
{
    "oneOf": {
        { "type": "null" },
        { "type": "string" },
        { "$ref": "#/components/schemas/SomeObject" }
    }
}
```

### Swagger Mode
```json
{
    "oneOf": [
        { "type": "string", "nullable": true },
        { "$ref": "$ref": "#/components/schemas/ObjectSomething.Nullable" },
    ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Check out the document in the [website](https://typia.io/docs/):
 ### 🏠 Home
   - [Introduction](https://typia.io/docs/)
   - [Setup](https://typia.io/docs/setup/)
+  - [Pure TypeScript](https://typia.io/docs/pure/)
   
 ### 📖 Features
   - Runtime Validators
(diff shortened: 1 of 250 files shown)


samchon (commented):
OK, well done.

PR #656, opened 2023-06-11
Fix #655 : `typia.random<T>()` problem

### LINKED ISSUE #655: Error : `typia.random<{a : Date | null }>()` return type is inferred `Date | null` type.
### Summary
Write a short summary of the bug in here.

  - **SDK Version**: 4.0.5-dev.20230609
  - **Expected behavior**: 
  - **Actual behavior**: 

Write detailed description in here.
when I use `typia.random<T>()` function, If T has type of `Data | null` property, the function sometimes occurs Error.

스크린샷 2023-06-11 오후 10 01 40

also, the property that has `Date | null` type, is inferred `Date | null`, not `string | null`.

### Code occuring the bug
```typescript
import typia from "typia";

const a = typia.random<{
	a: Date | null;
}>();

console.log(a);

/* Demonstration code occuring the bug you're reporting */
```

8471919:
When user uses `typia.random` such as `typia.random<{a: Date | null}>()`, the type `Date | null` is not distributed as Date and null (naked type) but from now, It is distributed on PrimitiveMain type.
so, user can get `{a: string | null}` return type.


src/Primitive.ts
```diff
@@ -34,7 +34,9 @@ export type Primitive<T> = Equal<T, PrimitiveMain<T>> extends true
 
 type Equal<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
 
-type PrimitiveMain<Instance> = ValueOf<Instance> extends object
+type PrimitiveMain<Instance> = Instance extends [never]
```
kakasoo, line 37: Personally, I think it will be difficult for developers in the distant future to know that this is a part made to respond to union types such as Date | null. Wouldn't it be better to add comments?

samchon (changes requested):
Thanks for contribution. The `[never]` trick seems wonderful.

By the way, can you revert every changes on `test/features` directory?

8471919:
um... I removed test/features directory... is this okay?
@samchon

samchon:
Okay, no problem.

PR #674, opened 2023-06-20
Docs/literals fix jsdoc

### LINKED ISSUE #673: Invalid jsdoc comment for literals
https://github.com/samchon/typia/blob/master/src/module.ts#L950
https://github.com/samchon/typia/blob/master/src/module.ts#L966

```literals<"A" | "B" | 1>; // [1, 2, 3] as const```
should be
```literals<"A" | "B" | 1>; // ["A", "B", 1] as const```

KristjanTammekivi:
Closes #673

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ node_modules/
 
 packages/typescript-json/src/
 package-lock.json
+yarn.lock
 *.log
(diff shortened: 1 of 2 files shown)


KristjanTammekivi:
Also removed as const since newer versions don't appear to add it

samchon (approved):
Thanks for contribution

PR #677, opened 2023-06-21
Static playground website

samchon:
https://[typia.io/playground](https://typia.io/playground)

Succeeded to build and publish static playground website which can see compiled (transpiled) JavaScript codes directly without any delay. Especially thanks for @GoogleFeud that giving me a good hint. From now on, `typia` users can easily test their validation just by the playground website.

By the way, when `typescript` be bundled and published onto the website, it misunderstands `Object` type as `ArrayLike`. Therefore, updated typia version for only such web-hosted compilation environments.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,13 @@
       "source.fixAll.eslint": true
     },
   },
+  "[typescriptreact]": {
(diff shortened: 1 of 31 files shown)


samchon:
image

PR #698, opened 2023-06-30
Make literals work with null

### LINKED ISSUE #685: literals should accept null
## Feature Request
https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDOAzKERwERIqp4DcAUGYWgHQA2wMAplKrQM4A8A5AIJdwAfOADsArrVoA+ABQBKIA

Currently null does not work, the extends for literals does not allow for it and when suppressing the error it just omits the value from the array

KristjanTammekivi:
Closes #685

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -971,7 +971,7 @@ export function literals(): never;
  *
  * @author Jeongho Nam - https://github.com/samchon
  */
-export function literals<T extends Atomic.Type>(): T[];
+export function literals<T extends Atomic.Type | null>(): T[];
 
 /**
  * @internal
--- a/src/programmers/LiteralsProgrammer.ts
+++ b/src/programmers/LiteralsProgrammer.ts
@@ -28,16 +28,18 @@ export namespace LiteralsProgrammer {
                     throw new Error(ErrorMessages.ONLY);
             },
         })(new MetadataCollection())(type);
-        const values: Set<Atomic.Type> = new Set([
+        const values: Set<Atomic.Type | null> = new Set([
             ...ArrayUtil.flat<Atomic.Type>(meta.constants.map((c) => c.values)),
             ...(meta.atomics.filter((type) => type === "boolean").length
                 ? [true, false]
                 : []),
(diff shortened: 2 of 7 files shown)


samchon:
As I have said, null is not a value categorized in atomic, but a type of object.

Therefore, this PR and related issue need to consider about that.

https://github.com/samchon/typia/issues/685#issuecomment-1613383581

samchon (commented):
As I have said, null is not a value categorized in atomic, but a type of object.

Therefore, this PR and related issue need to consider about that.

https://github.com/samchon/typia/issues/685#issuecomment-1613383581

KristjanTammekivi:
Can you explain why that matters in the context of this issue/pr?

samchon:
I repeat #685 issue's comment. 

`typia.literals()` function has been designed to get list of constant literal values.

  - `boolean`
  - `number`
  - `bigint`
  - `string`

The `null` value what you want is categorized in `object` type.` It's the problem.

KristjanTammekivi:
No problem then, null is a literal
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#description

> The value null is written with a literal: null

Also typeof null is considered a bug that was never fixed https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null

And even if it wasn't a literal I would say that that it's an unnecessary self-imposed constraint. I don't think undefined is a literal (although typeof undefined === 'undefined' so going by that standard it might be), but I would argue that literals should work with that as well. Let me talk about my use case.

We use Joi for validation and want to provide a list of valid values for it. Because literals doesn't support null we have to instead do something like the following
```typescript
type ValidValues = 'A' | 'B' | null;
const validValues = [...literals<Exclude<ValidValues, null>>(), null]
```
And this is why I want null to be included

samchon:
Wow, the `null` value to be considered an `object` type was a bug.

Got it. Your PR would be published as a next version `v4.0.5`.

Thanks for contribution.

Issue #702, opened 2023-07-03
Support "exactOptionalPropertyTypes"


Devidian:
**Problem:** in our project we use the tsconfig property [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes). But i noticed today, that this is incompatible with typia. I have an interface that has all properties optional but every input is throwing errors (does not matter which function i use, assert, is, validate, even null) if at least one optional property is missing.

We don't want to have any property defined but undefined in value.

It would be nice if typia could support it. If not it might be good to add this to "List of incompatible tsconfig settings"

samchon:
Same issue with #706 and succeeded to fix this bug. 

Upgrade to `v4.1.4`, then everything would be fine. Thanks for reporting.

PR #707, opened 2023-07-07
Fix #706 - support `exactOptionalPropertyTypes` option

### LINKED ISSUE #706: optional field treated as required
### 📝 Summary
  - **Typia Version**: 4.1.3
  - **Expected behavior**: type guard should validate optional field correctly 
  - **Actual behavior**: optional field is treated as required

When a property is made optional using a question mark (?) after the property name, all functions such as is/assert/validate are treating it as required. I have to force the type of the field to be an union of the desired type and undefined as a workaround.
This behavior seems to happen in other places as well, e.g. generate JSON schema using `application` function will also include optional field into `required` properties list.

### ⏯ Playground Link
https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDOAzKERwERIqp4DcAUGcAHYwCmUGqAxrXAJICytIARvXAG8ycEXGAATAFxwAzjCjUA5uVFxuqYABsA-NLkKqysgF8KTCFTlxFECOK69+AXkHCRE6QHJUPJp4A0oiZmFjIQmrQAdJoQigAUnmwAygEIyGiRwDIAPJzcfFAAfHE2dg4FAJQV5OaW4VEx8Z4AogCKAKoAggAyKYGEGbQAjgCuqJo5eY5FJbb2+fRV5PKIriK1YRHRsQmdSUnNAEoAKqkDqJGoMjL0MLnl9MWl89NLJnBMqDBMABZwcfRsFAKmsPqF6pFAdAAVAgdVgitXBsIY1dvsjscALRJY6HNgAOQA4mwAGIATTO6QuVxusCS8iUwAwiHuCxmzwewPhxg+X1+-yhwNByK2gphcPIphFDR2ngAUkkAPL4uBJADCAAlmhxOqkFcrIvpGcy4udLmAwJpgJ8YMALNkANpTAoAXUCeFQACsAG54YoVfxUEaaTT+ABMbyAA

(I noticed that the generated code in playground seems correct, is it using the lastest typia 4.1.3?)

### 💻 Code occuring the bug
```typescript
import typia from "typia";

interface IMember {
    id: string;
    email?: string;
}

const goodMember = {
   id: 'abc',   
}

console.log('IS', typia.is<IMember>(goodMember));
console.log('EQUALS', typia.equals<IMember>(goodMember));
try {
  console.log('ASSERT', typia.assert<IMember>(goodMember));
} catch (error) {
  console.error(error);
}
try {
console.log('ASSERT-STRINGIFY', typia.assertStringify<IMember>(goodMember));
} catch (error) {
  console.error(error);
}
console.log('JSON SCHEMA', JSON.stringify(typia.application<[IMember], "ajv">(),null,2));
```

samchon:
When configured `exactOptionalPropertyTypes` property to be `true` in `tsconfig.json` file, predication logic of undefindable type from optional symbol (`?`) becomes different. By the way, I haven't known that and it was the reason `typia` could not detect undefindable type when the optional symbol `?` being used like below:

```typescript
interface IMember {
    id: string;
    age?: number; // considered as required when `exactOptionalPropertyTypes` being used
}
```

From now on (`v4.1.4`), `typia` can distinguish it.

### DIFF
--- a/build/test.ts
+++ b/build/test.ts
@@ -123,7 +123,7 @@ async function main(): Promise<void> {
     await TestApplicationGenerator.generate(structures);
 
     // FILL SCHEMA CONTENTS
-    cp.execSync("npm run build:test");
+    cp.execSync("npm run build:test", { stdio: "inherit" });
     await TestApplicationGenerator.schema();
 
     // GENERATE TRANSFORMED FEATURES
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "4.1.3",
+  "version": "4.1.4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/typescript-json/package.json
+++ b/packages/typescript-json/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "4.1.3",
+  "version": "4.1.4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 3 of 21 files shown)


samchon:
Fix #702 too.

PR #715, opened 2023-07-18
doc-fix: Update comment-tags.mdx

elvitin:
Changed incorrectly typed term `nubmer` to correct term `number`.


samchon (approved):
Thanks for emension.

Issue #721, opened 2023-07-27
More information on how to configure Typia in a webpack + ts-loader setup


abrouaux:
## Question
Hi @samchon , I've been closely at Typia, in the last weeks as a candidate to replace zod.

I develop web apps on old browsers (SmartTVs). We need a packager (webpack) and transpiration (babel usually, experimenting with swc at the moment).
We can't use vite as some of the browsers we need to support are older than 2015.

I've seen hints here & there in the typia issues that it's possible to get typia to work with ts-loader, except I can't figure exactly why.
I've added the typia/lib/transform as a transform plugin in tsconfig. 

I've setup a very small repo with web pack, typia, ts-loader and swc as a final ES5 transpiler (I'm aware of the discussion about SWC - it's not used with typia here): https://github.com/abrouaux/typia-webpack.

I've also added the "inspect" web pack plugin to understand what's going on with my builds.
What it says about the typia step:
<img width="1527" alt="image" src="https://github.com/samchon/typia/assets/47028929/5fad55a5-74ef-4174-b318-ce0260e78b0c">

At runtime, I'm getting (obviously given the preprocessing seems to have failed) this  error:
<img width="899" alt="image" src="https://github.com/samchon/typia/assets/47028929/5298a8aa-e1d1-41e1-a90e-0fcb19b3e7b7">

If you could take a look at the repo & tell me what I'm doing wrong, it'd be super helpful! 
thanks

samchon:
Run `npx typia setup` command, then you may be possible.

Issue #688, opened 2023-06-27
Feedback for “Setup”


bogeeee:
Some more feedback ;)
Please give this item a clear name: Either "Transformation mode" or "AOT (Ahead of Time) compilation mode". But you speak of 
"Transformation" in the headline, then "AOT (Ahead of Time) compilation **mode**". and then later link to "Transform **mode**"

Was maxmimum confused by that ;)

samchon:
Wait your PR

PR #724, opened 2023-07-28
Close #689 - detailed setup.mdx

### LINKED ISSUE #689: Feedback for “Setup”
`npx typia setup`

Also, you could tell us, how this works under the hood and what does it modify and how can you do it manually. For me personally, i wouldn't like having some script magically modifying my configs / build chain and i would at least have a manual option.

What about just using the "slow" validator functions like  `is` ? Is that also covered by the transform/AOT mode ? I guess: Yes. But mention that please. Or mention that it in the introduction page.

This sentence under "Generation"

> Instead, you should utilize the generation mode.

is somehow redundant and confused me (what was that mode again that it's woth beeing mentioned now)

Jeah, lot's of feedback but i'm also the author of some lib and know how it is, swimming in my own soup and would be glad someone gave me feedback about my docs.

Also great idea that you cover ts to ts compilation as well when you don't have tsc.

samchon:


samchon:
Close #688 too

samchon:
Close #721 too

Issue #736, opened 2023-08-02
updated fast-json-stringify


Uzlopak:
Can you please update the benchmarks?

Btw. I updated the serializer for asString and I think you are using a similar serializer for strings. Maybe you want to adapt it for more performance.
https://github.com/fastify/fast-json-stringify/commit/17bb4c2430c60a44079ef572766a8150ab70aefc#diff-8e3d45dd0e9ec504195499d7fafe3efc08756c349fce602ef6538c593aa563d8

Also the benchmarks are a little bit unfair. E.g. you are benching for `stringify` but fast-json-stringify is not only stringifying but also does some assertions, like checking for required fields.

samchon:
> https://dev.to/samchon/good-bye-typescript-is-ancestor-of-typia-20000x-faster-validator-49fi
>
> Around the same time, I had made another library named typescript-json.
> 
> It performs AoT (Ahead of Time) compliation like typescript-is, but it was not for runtime validation, but for JSON schema generation. About JSON serialization boosting, typescript-json had utilized fast-json-stringify by automatically generated JSON schema.
> 
> For reference, purpose of typescript-json was to accomplish below nestia, generating Swagger Documents with pure TypeScript type.

Thanks for notification. As `typia` had started from wrapper of `fast-json-stringify` (at that time, package name was `typescript-json`), I'm still using string serializer of `fast-json-stringify`. I'll update string serializer following `fast-json-stringify`. Also, I'll perform benchmark again with latest version of `fast-json-stringify`.

By the way, `typia` has four stringify functions and `assertStringify` is simliar with `fast-json-stringify`. Benchmark result is showing all of those `typia` stringify functions (https://typia.io/docs/json/stringify/#performance), but as normal users don't know the difference, I can understand why you feel it as unfair. About this topic, how about adding a comment that `fast-json-stringify` is similar with `typia.assertStringify()` under the graph? If you want, you can edit https://github.com/samchon/typia/blob/master/website/pages/docs/json/stringify.mdx#L497-L507, and send a PR for that.

  - `stringify`
  - `isStringify`
  - `assertStringify`
  - `validateStringify`

Uzlopak:
Sure, I will provide a PR.

Please check if the patch for serializer for asString makes a perf boost. If you see a performance regression, than I would need to investigate.

samchon:
You can measure benchmark of `typia.stringify<T>()` function by below commands.

```bash
# PREVIOUS SMALL-STRING
[...]

PR #740, opened 2023-08-03
Adapt #736 - optimize JSON serializer of `string` type

samchon:
> https://dev.to/samchon/good-bye-typescript-is-ancestor-of-typia-20000x-faster-validator-49fi

`typia` had started from a helper library of `fast-json-stringify`. At that time, package name was `typescript-json`, and slogan was convenient JSON serialization boosting without JSON schema definition, but with pure TypeScript type. In such reason, `typia` still has same logic with `fast-json-stringify` about `string` type's JSON serialization.

In today, `fast-json-stringify` has changed their `string` type logic, and developer of `fast-json-stringify` informed me the update. I've measured benchmark with him, and could understand that the new logic is efficient than before. Therefore, updated `typia` to follow the `fast-json-stringify`


samchon (commented):
No problem on test program.

PR #745, opened 2023-08-04
Fix #737 - remove cursing from random any type

### LINKED ISSUE #737: Remove cursing for random "any"
Thank you very much for the random generator tool. It has helped a lot when it comes to perform testing. We just noticed that whenever we generate random objects that have properties with the "any" type, they come as a literal string:

> "f***king any type exists..."

We understand the frustration with the "any" type but unfortunately some of it comes from third-party libraries that we are using. Do you think we could remove the cursing here to something like:

> "any type used..."

An angry face emoji 😠 could be included too to express the frustration.

Thank you for considering this

mmamedel:
Removes cursing from the result of a generated `any` type as discussed in the related issue.

### DIFF
--- a/src/programmers/RandomProgrammer.ts
+++ b/src/programmers/RandomProgrammer.ts
@@ -220,7 +220,7 @@ export namespace RandomProgrammer {
             if (meta.any)
                 expressions.push(
                     ts.factory.createStringLiteral(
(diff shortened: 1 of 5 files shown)


samchon (approved):
Will publish with next patch

PR #786, opened 2023-08-27
Complete type tag supporting

samchon:
In previous PR,I've impelemented validator functions of `typia` supports type tags like below. This way would be much safer and convenient than previous comment tags.

At now, such type tags also can be used in every `typia` functions even including protocol buffer related features. Previous comment tags would be also supported and compatible, but its `CustomValidatorMap` being removed due to its compile time unsafety.

For reference, the next v5 version would be published with documentation.

```typescript
type SomeNumber = number & Type<"int32"> & Maximum<24_000>;

export type Type<
    Value extends "int32" | "uint32" | "int64" | "uint64" | "float" | "double",
> = TagBase<{
    target: Value extends "int64" | "uint64" ? "bigint" | "number" : "number";
    kind: "type";
    value: Value;
    validate: Value extends "int32"
        ? `Math.floor($input) === $input && -2147483648 <= $input && $input <= 2147483647`
        : Value extends "uint32"
        ? `Math.floor($input) === $input && 0 <= $input && $input <= 4294967295`
        : Value extends "int64"
        ? {
              number: `Math.floor($input) === $input && -9223372036854775808 <= $input && $input <= 9223372036854775807`;
              bigint: `true`;
          }
        : Value extends "uint64"
        ? {
              number: `Math.floor($input) === $input && 0 <= $input && $input <= 18446744073709551615`;
              bigint: `BigInt(0) <= $input`;
          }
        : Value extends "float"
        ? `-1.175494351e38 <= $input && $input <= 3.4028235e38`
        : `true`;
    exclusive: true;
}>;

export type Maximum<Value extends number | bigint> = TagBase<{
    target: Value extends number ? "number" : "bigint";
    kind: "maximum";
    value: Value;
    validate: `$input <= ${Numeric<Value>}`;
    exclusive: ["maximum", "exclusiveMaximum"];
}>;

type Numeric<Value extends number | bigint> = Value extends number
    ? Value
    : `BigInt(${Value})`;


samchon (commented):
OK, tests passed

PR #796, opened 2023-08-31
Fix typos

Xvezda:
- Fixed parameter name
- Fixed author name


samchon (approved):
Oh my god, thanks for emension.

PR #819, opened 2023-09-26
Close #815 - new module `http` - `query`, `headeres` and `parameter`

### LINKED ISSUE #815: Suggestion - assertParseQueryString
Hello!

We're using your awesome library for input validation with (pretty much) express.js framework. Great job, neat solution! I would like to ask about implementing assertions for query strings.

In express (as well as everywhere else), when passing a query string `?foo=1&bar=2`, `req.query` gets this object:

```json
{
  "foo": "1",
  "bar": "2"
}
```

...which is validated by typia in our case with `assertEquals(req.query)`.

I am now wondering whether we can do a step further and instead of defining

```ts
interface Query {
  foo: string;
  bar: string;
}
```

Make it be able to parse and validate against this interface:

```ts
interface Query {
  foo: string;
  bar: number; // parse as number!
}
```

To ultimately get this output with no validation errors:

```json
{
  "foo": "1",
  "bar": 2
}
```

Feel free to close this request if it feels boring or keep it to collect feedback :) But this feature would allow to pass pretty much any (primitive) type in a query string and properly decode it. We have an automatically generated client library in our project which infers and reuses all types defined in the api, and having to always define strings in query even in case of a number feels like it can be fixed.

At the same time, I understand this can be very complex to implement.

Thanks!

samchon:
New module `http` has been newly added.

```typescript
export namespace http {
    export function parameter<T extends boolean | bigint | number | string>(input: string): T;

    export function headers<T extends object>(input: Record<string, string | string[] | undefined>): Resolved<T>;
    export function isHeaders<T extends object>(input: Record<string, string | string[] | undefined>): Resolved<T> | null;
    export function assertHeaders<T extends object>(input: Record<string, string | string[] | undefined>): Resolved<T>;
    export function validateHeaders<T extends object>(input: Record<string, string | string[] | undefined>): IValidation<Resolved<T>>;

    export function query<T extends object>(input: URLSearchParams): Resolved<T>;
    export function isQuery<T extends object>(input: URLSearchParams): Resolved<T> | null;
    export function assertQuery<T extends object>(input: URLSearchParams): Resolved<T>;
    export function validateQuery<T extends object>(input: URLSearchParams): IValidation<Resolved<T>>;
}
```

For reference, those features had been developed in [`@nestia/core`](https://github.com/samchon/nestia) for a long time. 

The reason why migrating them from [`@nestia/core`](https://github.com/samchon/nestia) is to support `Fastify` plugin like [`nestia`](https://github.com/samchon/nestia) case.


samchon:
Close #815

samchon (commented):

PR #825, opened 2023-09-28
fix `IArray.items` type

hax:
As title


samchon (commented):
The items means `T` type of `Array<T>`, and spec of OpenAPI.

So it is not bug, but spec.

hax:
Oh, maybe I misunderstand it. My real issue is how to narrow the type. When using `jsonApplication.schemas[i].items.map(...)` (assume `schemas[i]` is json schema array)

PR #828, opened 2023-09-29
Fix #827 - change setup wizard logic.

### LINKED ISSUE #827: Setup downgrades from typescript@latest to typescript@4.9.5
## Question

Yesterday I ran setup

```shell
$ npx typia setup
----------------------------------------
 Typia Setup Wizard
----------------------------------------
? Package Manager npm

npm install --save-dev ts-patch@latest
npm install --save-dev ts-node@latest
npm install --save-dev typescript@latest
npm run postinstall
$
```

and everything was like expected. Today I ran setup again (to regenerate `package-lock.json` in an interactive rebase)

```shell
$ npx typia setup
Need to install the following packages:
  typia@5.1.3
Ok to proceed? (y) y
----------------------------------------
 Typia Setup Wizard
----------------------------------------
? Package Manager npm

npm install --save-dev ts-patch@latest
npm install --save-dev ts-node@latest
npm install --save-dev typescript@4.9.5
npm run postinstall
```

and suddenly the typescript version is downgraded. Is this expected?

Looking into all commits (`git reflog -p -G typia -- package.json | grep typia | sort -u`, same for `package-lock.json`), the version installed has always been 5.1.3, so it is not that I used to have 5.0.5 and then updated.

And since currently there are no commits present after the 5.1.3 release (which was done earlier yesterday) there are no logical explanation for the difference in behaviour.

samchon:
No more need to consider `ts-patch@2` version, so removed special setup logic related with it.

From now on, setup wizard of typia no more consider the `ts-patch` verison. 

It always setups the latest `ts-patch` and `typescript`. versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "5.1.3",
+  "version": "5.1.4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/typescript-json/package.json
+++ b/packages/typescript-json/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "5.1.3",
+  "version": "5.1.4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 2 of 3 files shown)


samchon (commented):
OK, logic is simple, so merge after test program be passed.

PR #861, opened 2023-11-05
Fix #853 - `Record<string & tags.Format<"uuid">, T>` case.

### LINKED ISSUE #853: Type validation is not used for Record keys or indexable types
### 📝 Summary

I'd like to be able to provide a type that limits a string to be of a certain format, and then re-use that type in multiple places to enforce the same formatting everywhere.

  - **Typia Version**: 5.2.4 & `typia@next`
  - **Expected behavior**: A type of the form `type T = string & tags.Pattern<'.*'>;` can be used anywhere and have the same validation applied.
  - **Actual behavior**: A type of the above form works when used as the property of a 

It seems like this primarily applies to object keys, where some default test is applied to keys and it ignores any other tests that type should have applied to it. I assumed this was maybe due to the type being an alias of a primitive (or something like that), but the type itself works fine if used in any other instance except as the key of an object.

Notably, this validation _does_ work for maps, but not Records or using indexable types.

This is related to https://github.com/samchon/typia/issues/725, but my type alias is using tags to add extra conditions. I would have assumed that could work?

### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoF+gUzgGEBXAmFQCSAEzgBeOAKjAAdnjgAyXIQB0ABVQwYHKLIA8AcgB6AbQAMAWgCcqSwC8AulgCMANmIASQwD5WLAHoAuAB3aABrAhYAYwhZATgAN1QAG2ARLQ5RCQRkNFVoqA5MoQJ9PgFhER8ACgBKfzkdKDJUaK4AZSoOADVU3i4sFjgRuHSALh5+QRBRVlG4LRgZACNeHUnZXhAV3VZiNiDQiKjY+PhktIydLpBe-q5JenzC4p1S-Vv7lIHahrZOHAvgB5GAAC10fR+j2ww1GEymlVmInmowg4N0AEFtKt1hxNttdlB9odgmEoJEYnEEpd0pkQRioFCBjlnqgCkUSmUGRCmQ8-v5AcyONwICkUhxojBgHEcgAlSXQETlaZVTBfYV+UlwEQQDhEWTo44UqnnJKpOk6UXiyXS2VPPLs15c-TC60SqUy2QCthNXStdpwUG8t1ij122SwhamcIcRCTCozUSOSY8yEPEmBYK6-VwQ3wcnhU00i3XDjB3Tu21e1mOjlvLJlCt86FVz1xAVAA

Comparison with `Map`:
https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoF+gUzgGEBXAmFQCSAEzgBeOAKjAAdnjgAyXIQB0ABVQwYHKLIA8AcgB6AbQAMAWgCcqSwC8AulgCMANmIASQwD5WLAHoAuAB3aABrAnZkNFUAYygOLQ4hAn0AWVQwfT4BYRFMWV4QACNdHx8ACgBKfyC4EQgOIlkIeDCocOiGeMTk1P0AJQ446BEc-kEQUULisqgKmrrgxua4VvaI7tiEpJ0BrBY4Y7hTcI5EAC4eSfzHa6LS3RZiKtqgA

### 💻 Code occuring the bug
```typescript

type CustomId = string & tags.Pattern<'^[0-9a-z]{16}$'>;

// works
const validateId = typia.createIs<CustomId>();

interface SomeValue {
    id: CustomId;
    attribute: number;
}

// works
const validateSomeValue = typia.createIs<SomeValue>();

type SomeOtherValue = {
    id: CustomId;
    otherAttribute: number;
}

// works
const validateSomeOtherValue = typia.createIs<SomeOtherValue>();

type ValueCollection = Record<CustomId, SomeValue>;

// does not work
const validateCollection = typia.createIs<ValueCollection>();

interface OtherValueCollection {
    [key: CustomId]: SomeOtherValue;
}

// does not work
const validateOtherCollection = typia.createIs<OtherValueCollection>();
```

Comparison to `Map`:
```typescript

type CustomId = string & tags.Pattern<'^[0-9a-z]{16}$'>;

// works
typia.createIs<Map<CustomId, number>>();

// does not work
typia.createIs<Record<CustomId, number>>();

// does not work
typia.createIs<{
    [key: CustomId]: number
}>();
```

samchon:
When `Record<Key, T>` type comes, `typia` had not considered Type Tags being used in the `Key` type. 

Therefore, if `Record<string & tags.Format<"uuid">, T>` type comes, `typia` had just recognized it as `Record<string, T>` type.

This PR enhances and fixes it.


samchon (commented):
OK, no problem on the main logic.

PR #862, opened 2023-11-07
Docs: Enhance Yarn Berry Comment

bradennapier:
This comment is misleading and may lead to users assuming it is not possible to use yarn berry at all.  

Instead, link to the github and indicate it can stll be used manually - at the moment this is not linked in docs that i could find and i almost left the site!

Potentially could add the comment below the `yarn add --dev typia` to make it even more clear but honestly did a find and replace - happy to do that as well if agreed

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -20,7 +20,7 @@ pnpm typia setup --manager pnpm
     </Tab>
     <Tab>
 ```bash filename="Terminal" showLineNumbers copy
-# YARN BERRY AUTO SETUP IS NOT SUPPORTED
+# YARN BERRY AUTO SETUP IS NOT SUPPORTED (yarn typia setup ...)
 # YOU MUST SETUP MANUALLY
 # @see [Github Issue: 840](https://github.com/samchon/typia/pull/840)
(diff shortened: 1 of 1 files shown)


samchon (commented):
It is intended word.

As yarn berry is not supporting standard npm command `prepare`, the yarn berry users must run the script manually every time they install the dependency packages. Therefore, the word "not supporting yarn berry" is intended exactly what I wanted.

samchon:
As yarn berry is not suppoorting the standard npm script `prepare`

bradennapier:
It works perfectly fine usin `tspc` or postinstall although `tspc` seems idea.  Seems less than ideal to just push away a whole segment of developers that could benefit frm the package with little difference to others. 

Could also support berry fully by using the `pnpm linker` resolution from what i have seen

bradennapier:
I mean considering I am using it sems reason enough to not just outright tell everyone it cant be used.  Rather it should have a description why its a bit more work and that you can't directly support users that go that path.

PR #863, opened 2023-11-08
Break change from `ts.factory.createNumericLiteral()`.

samchon:
Since TypeScript v5.3 update, it no more allows negative numeric value. 

Instead, it must be combined with `ts.factory.createPrefixUnaryExpression` function like below.

This PR accepted such break change for future TS 5.3 release.

```typescript
ts.factory.createPrefixUnaryExpression(
    ts.SyntaxKind.MinusToken,
    ts.factory.createNumericLiteral(Math.abs(value)),
)
```


samchon (commented):
OK, no problem.

samchon (commented):
Also pre-tested TS 5.3 on playground.

PR #864, opened 2023-11-10
Gather compilation error of custom type tags too.

samchon:
When user defines wrong custom type tag with invalid validation script, it had immediately thrown compilation error. Besides, other user mistakes had been kept by `typia` for reporting them at once. Therefore, the validation script error on the custom type tag may hide every other mistakes.

This PR fixes such immidately throwing exception problem from custom type tag, so that makes user convenient like other user-mistaken based error cases.


samchon (commented):
Test all pass

PR #867, opened 2023-11-12
Add `reflect.metadata()` function. 

samchon:
`typia` has analyzed TypeScript types by constructing `Metadata` class. All of the code generations are also being processed by utilizing the `Metadata` class.

By the way, some users of `typia` is hoping the `Metadata` class to use by themselves.

Therefore, this PR exports the `Metadata` class by providing the new AoT function, `typia.reflect.metadata()` function.


samchon (commented):
Test pass

PR #881, opened 2023-11-23
Separate test program from the main package.

samchon:
As TypeScript v5.3 no more can make them compatible.

I've separated those main and test program to another packages. 

In other words, here `typia` repo becomes monorepo.


samchon (commented):
Succeeded to pass the new testing system

PR #883, opened 2023-11-26
New cli command `npx typia patch` for hacking `defaultJSDocParsingMode` value

samchon:
Since TypeScript v5.3 update, `tsc` no more parses `JSDocComment`s. Therefore, Therefore, `typia` also cannot utilize those `JSDocComment` related features too, especially "Comment Tags" and "JSON schema generator".

However, in relation to this, the upgrade of `ts-patch` continues to be delayed, and I still don't know how long the delay would be continue. Furthermore, there're some `nestia` users urging to resolve the `peerDependencies` of `typia` that blocking the TypeScript v5.3 update. Therefore, before the `ts-patch` being prepared, I've decoded to provide `typia`'s own solution for a while. It is the new CLI command, `npx typia patch` (`npx typia setup` command also performs the same behavior).

Also, if the `defaultJSDocParsingMode` value not being patched, `typia` will generate an warning message of TypeScript compiler API. For reference, as it is an warning message, it does not interrupt the TypeScript compilation like the compilation error case. If there're some `typia` users never using "Comment Tags" or "JSON schema generator" at all, they don't need to run the CLI command. This is not mandatory command, but just optional command.

Of course, when `ts-patch` being updated, this CLI command would be disabled immediately, if the installed `ts-patch` version is the latest one.

Related issues:
  - https://github.com/microsoft/TypeScript/pull/55739
  - https://github.com/nonara/ts-patch/issues/134


samchon (commented):
No problem on the testing programs.

samchon:
Also, need to update `nestia` project, too.

PR #886, opened 2023-11-27
Fix #884 - wrong union predication about constant type

### LINKED ISSUE #884: Unexpected assertion fail for union type with optional fields
### 📝 Summary
Type equality (`assert`, `is`, etc.) with certain union objects with optional fields is parsed as non-equal.

  - **Typia Version**: 5.3.0-dev.20231127-2
  - **Expected behavior**: Typia should select the correct object for matching
  - **Actual behavior**: Typia decides the object does not match the intersection type. 

```typescript
interface A {
	name?: "A";
	someBool: true;
}
interface B {
	name?: "B";
	someBool: boolean | undefined;
}
```

It seems that Typia only validates that the body matches `A` and doesn't check whether it matches `B`. This problem resolves itself if `someBool` is required as non-undefined, or if `name` is required. 

I believe that `A | B` (see below) should match given that TypeScript doesn't have any issues with the following:
```typescript
const body = {
	name: "B",
	someBool: true,
} satisfies A | B;
```

### 💻 Code occuring the bug
```typescript
import typia from "typia";

interface A {
	name?: "A";
	someBool: true;
}
interface A_NonOptional {
	name: "A";
	someBool: true;
}
interface B {
	name?: "B";
	someBool: boolean | undefined;
}
interface C {
	name?: "C";
	someBool: boolean;
}

// Throws (Incorrect; should match B)
const A_and_B = typia.assert<A | B>({
	name: "B",
	someBool: true,
});

// Doesn't throw (Correct)
const A_NonOptional_and_B = typia.assert<A_NonOptional | B>({
	name: "B",
	someBool: true,
});

// Doesn't throw (Correct)
const A_and_C = typia.assert<A | C>({
	name: "C",
	someBool: true,
});
```

samchon:
When there's same type exists in both `atomics` and `constants` and try to predicate union type about that case, `typia` mis-computes whether each types are intersected or not.

This PR fixes the bug, just by adding the special logic about the same typed `atomics` and `constants` are compatible.
md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "5.3.0",
+  "version": "5.3.1-dev.20231127-4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/benchmark/package.json
+++ b/packages/benchmark/package.json
@@ -73,6 +73,6 @@
     "suppress-warnings": "^1.0.2",
     "tstl": "^2.5.13",
     "uuid": "^9.0.1",
-    "typia": "D:\\github\\samchon\\typia\\typia-5.3.0.tgz"
+    "typia": "D:\\github\\samchon\\typia\\typia-5.3.1-dev.20231127-4.tgz"
   }
 }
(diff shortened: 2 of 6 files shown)


samchon (commented):
Passed test programs

PR #891, opened 2023-11-29
Fix length error of `typia.random<T>()` function.

samchon:
When generating random string value from `typia.random<T>()` function with `tags.MaxLength<1>` tag, `typia.random<T>()` function had not returned the one length string value, but had returned the `5` length string. Such bug being repeated until configure the `tags.MaxLength<5>` over tag value.

This PR changes the invalid random length algorith, so that the bug has been fixed.


samchon (commented):
No problem on the test programs.

PR #898, opened 2023-12-06
Adjust nonara/ts-patch#138

samchon:
Since TypeScript v5.3 update, `tsc` no more parses `JSDocComment`s. Therefore, `typia` also cannot utilize those `JSDocComment` related features too, especially "Comment Tags" and "JSON schema generator".

To reflect such TypeScript v5.3 update, `typia` has newly added a  temporary CLI command `npx typia patch` that is hacking the TypeScript module to turning on the `JSDocComment` parsing.

  - samchon/typia#883

In today, `ts-patch` has started supporting the TypeScript v5.3 update by reading `package.json` of transformer libraries. Therefore, the temporary CLI command no more required. Following the `ts-patch` guide, this PR has configured the `package.json` file for `ts-patch`, and `JSDocComment` parsing would be turned on by the `ts-patch`.

For reference, the temporary CLI command `npx typia patch` would be kept for a while, due to there can be some users using previous version of `ts-patch`. It would be deprecated after the next major update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,16 @@
 {
   "name": "typia",
-  "version": "5.3.3",
+  "version": "5.3.4",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "bin": {
     "typia": "./lib/executable/typia.js"
+  },
+  "tsp": {
+    "tscOptions": {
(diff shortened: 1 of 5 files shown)


samchon (commented):
No problem in testing, and working properly even when manual setup.

PR #913, opened 2023-12-28
Fix #912 - protobuf.decode byteOffset

### LINKED ISSUE #912: typia.protobuf.decode ignored byteOffset and byteLength
Calling `typia.protobuf.decode` (or `typia.protobuf.createDecode`) with `Buffer` throws 'buffer overflow'.

minimal code to reproduce:

``` typescript
import typia from 'typia'

type Foo = {
  id: number & typia.tags.Type<'int64'>
}

const decode = typia.protobuf.createDecode<Foo>()

const obj = { id: 12 }
const encoded = typia.protobuf.encode<Foo>(obj)

try {
  const buffer1 = Buffer.from(encoded)
  const decoded1 = decode(buffer1) // throws

  console.log(obj, decoded1)
} catch (e) {
  console.error(e) // Error on typia.protobuf.decode(): buffer overflow
}

// while this works fine.
const buffer2 = Buffer.from(encoded.buffer, encoded.byteOffset, encoded.byteLength)
const decoded2 = decode(buffer2)

console.log(obj, decoded2) // { id: 12 }, { id: 12 }

```

I tried to patch the constructor code in [`$ProtobufReader`](
https://github.com/samchon/typia/blob/0b9e83cc8453df2e98f06d09d0c8acb7b0c94c56/src/functional/%24ProtobufReader.ts#L23C42-L23C42) to:

``` typescript
  public constructor(buf: Uint8Array) {
    this.buf = buf;
    this.ptr = 0;
    this.view = new DataView(buf.buffer, buf.byteOffset, buf.byteLength);
  }
```

Then decode works fine.

ghlin:


samchon (approved):
Thanks for contribution.

PR #914, opened 2023-12-30
`IJsonSchema.IObject` for object literal expression.

samchon:
When object literal expression type being used instead of named object type, `typia` had enrolled it into the `IJsonComponents` considering as a named type with `__type` and `__type.o${number}`.

However, it seems not suitable for purpose of the JSON schema spec. Therefore, I've added new type `IJsonSchema.IObject` and assigned it into the JSON schema values when object literal expression type being used.

> What lilteral expression type means?
>
> ```typescript
> // NAMED TYPE
> export interface IMember {
>   id: string;
> }
>
> // LITERAL EXPRESSION
> const member: { id: string };
> ```


samchon (commented):
Succeeded to pass the test program.

PR #916, opened 2024-01-03
Lighten up JSON Schema

samchon:
The JSON schema generated by the `typia.json.application<Types>()` function was quite extensive and lax. The first reason is that when `IJsonSchema.IAttribute` properties are union or array types, they are assigned equally to nested types. And the second reason is the special properties of `typia` that starts with `x-typia-`.

Until now, there was no reason for the JSON Schema generated by `typia` to be lightweight. However, recently, as LLM models with ChatGPT as a topic have become popular, and they have started to use a model that charges by token (word), a problem has arisen. JSON Schema is entered into the ChatGPT (LLM) API and utilized to create various AI-enabled services, but these characteristics of the `typia.json.application<Types>()` function ultimately resulted in excessive token charges.

Accordingly, this PR first prevented `IJsonSchema.IAttribute` from being excessively allocated. Second, a new generic argument `Surplus` was added to the `typia.json.application<Types, Purpose, Surplus>()` function that can determine whether to use additional properties starting with `x-typia-`. Of course, the default value for this is `false`.

```typescript
/**
 * JSON Schema Application.
 *
 * Creates a JSON schema application which contains both main JSON schemas and
 * components. Note that, all of the named types are stored in the
 * {@link IJsonApplication.components} property for the `$ref` referencing.
 *
 * Also, `typia.json.application()` has two additional generic arguments.
 *
 * The 1st *Purpose* means the purpose determines the JSON schema definition spec.
 * For an example, `ajv` has an extra property "$recursiveRef"  that are not
 * exists in the standard JSON schema definition spec. Otherwise, `swagger`
 * can't identify the tuple definition.
 *
 * The next *Surplus* means whether to allow surplus properties starting with
 * `x-typia-` or not. If `true`, the surplus properties like `x-typia-jsDocTags`
 * would be registered into the JSON schema, otherwise, not.
 *
 * @template Types Tuple of target types
 * @template Purpose Purpose of the JSON schema
 * @template Surplus Allow surplus properties starting with `x-typia-` or not
 * @return JSON schema application
 *
 * @author Jeongho Nam - https://github.com/samchon
 */
export function application<
  Types extends unknown[],
  Purpose extends "ajv" | "swagger" = "swagger",
  Surplus extends boolean = false,
>(): IJsonApplication;
```


samchon (commented):

PR #918, opened 2024-01-09
Fix bug of `IJsonSchema.default`.

samchon:
When dieting JSON Schema (#916), have taken a mistake that even dieting the `IJsonSchema.default` value when the default value has configured by comment tag (no problem when type tag).

This PR fixes the bug, so that revived the comment tag defined default value.


samchon (commented):
Succeeded to pass every testing programs.

PR #922, opened 2024-01-14
`typia.reflect.metadata<Types>()` function.

samchon:
To reducei bundle size, changed its return type to be interface type. 

If you want to convert it to the class type, then call the `MetadataApplication.from()` function by yourself.

The reason why reducing bundle size such like this is making 0 bundle size is my plan to the next major patch.


samchon (commented):
No problem at test programs at all

Issue #752, opened 2023-08-10
Why does Typia initial have a bundle size of 7.6 kB?


fabian-hiller:
Hi Jeongho, my name is Fabian, and I am the author of the schema validation library [Valibot](https://github.com/fabian-hiller/valibot). I am currently investigating different schema libraries for my bachelor's thesis. Typia is interesting because the approach is completely different.

While examining your library, I noticed that regardless of the complexity of a schema, the bundle size starts at 7.6 kB due to the exported code of the `type` package. What exactly is included in this 7.6 kB? I have seen that this code is used, for example, when validating an email.

Shouldn't it be theoretically possible to import only the code that is actually needed by the compile step? This would make the bundle size of Typia start initially at 0 KB.

samchon:
I have not measured exact bundle size, but the reason why not 0kb may be one of them below:

At 1st, lots of types embeded in `typia`. Looking at below files, then you may understand what I mean.

  - https://github.com/samchon/typia/blob/master/src/module.ts
  - https://github.com/samchon/typia/blob/master/src/TypeGuardError.ts
  - https://github.com/samchon/typia/blob/master/src/Primitive.ts
  - https://github.com/samchon/typia/blob/master/src/schemas/IJsonSchema.ts

Also, there're utility functions used in transformed codes.

  - https://github.com/samchon/typia/tree/master/src/functional

At last, `typia` imports [`randexp`](https://www.npmjs.com/package/randexp) library for random regex paterrn data generation.

fabian-hiller:
The types should have no influence on the JavaScript bundle size, and if you export and import the individual functions separately instead of using a `default` export, you can probably reduce the bundle size for most schemas to 0 kB apart from the generated code. This brings a significant advantage when Typia is used in the front end, for example, with forms. However, if that is not your goal, I understand your approach.

samchon:
To accomplish 0kb bundling size, I've to change `typia` to call `require()` function only when needed.

It's not any problem in CommonJS, but such approach would be a critical problem in ESM.

fabian-hiller:
I think that this also works with ESM. When using wildcard imports, tree shaking and code splitting can be applied by a bundler such as esbuild. If you're interested, you can try it out on [Bundlejs](https://bundlejs.com/?q=valibot&treeshake=%5B%7B+email%2CminLength%2Cobject%2Cstring+%7D%5D) with Valibot. Change the import statement to a wildcard and call e.g. the string function.

```ts
export * as v from "valibot";

[...]

PR #928, opened 2024-01-17
[Draft] Improved tree shaking code

EloB:
# Draft

This is a draft to improve tree shaking and dead code elimination when bundling `typia`.

I created this PR to get early feedback.

When I have more time I will provide an example how I use this combined with a esbuild plugin to bundle typia automatically and that it tree shakes everything that it doesn't need.

## Code changes:
- Replaced `namespace name {}` with `export * as namespaceName from "./name"`. Improves dead code elimination.
- Added `/** @__PURE__ */` to some exports that has `Object.assign`.

## Example
I bundled the code below with esbuild and I was able to compress the bundle from 69827 characters down to 5804 (**-92%**). We also need to add `"sideEffects": false` to `ret` and `randexp` package.json files. I've made a PR to `randexp` but have to make another for `ret`.
```ts
import { IValidation, tags, validate } from "typia";

const res: IValidation<IMember> = validate<IMember>({
  id: 5, // wrong, must be string (uuid)
  age: 20.75, // wrong, not integer
  email: "samchon.github@gmail.com",
});

if (!res.success) console.log(res.errors);

interface IMember {
  email: string & tags.Format<"email">;
  age: number
  & tags.Type<"uint32">
  & tags.ExclusiveMinimum<19>
  & tags.Maximum<100>;
}
```


package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "test": "npm run package:tgz",
     "-------------------------------------------------": "",
-    "build": "rimraf lib && tsc --removeComments --declaration false && tsc --emitDeclarationOnly",
+    "build": "rimraf lib && tsc --declaration false && tsc --emitDeclarationOnly",
```
EloB, line 18: I include the comments so it doesn't strip `/** @__PURE__ */`.

samchon:
Thanks for contribution, but bundled size is increased from 11.2kb to 11.5kb.

  - https://bundlephobia.com/package/typia@5.3.11
  - https://bundlephobia.com/package/typia@5.3.12-dev.20240118

However, it seems like you're expert at the package bundling. I'm considering another bundling diet way through `require()` function. What about changing `typia` to generate `require()` function instead of using the namespace like below? If this way works, it may possible to diet near to the 0 byte.

```typescript
// CURRENT
const $guard = typia_1.default.json.createAssertStringify.guard;

// HOW ABOUT THIS ONE?
const $guard = require("typia/lib/functional/$guard").$guard;
```

samchon (commented):
Thanks for contribution, but bundled size is increased from 11.2kb to 11.5kb.

  - https://bundlephobia.com/package/typia@5.3.11
  - https://bundlephobia.com/package/typia@5.3.12-dev.20240118

However, it seems like you're expert at the package bundling. Looking your PR, I got another idea of bundling diet way through the `require()` function. What about changing `typia` to generate `require()` function instead of using the namespace like below? 

If this way works, it may possible to diet near to the 0 byte, so that don't need to consider tree shaking problem.

```typescript
// CURRENT
const $guard = typia_1.default.json.createAssertStringify.guard;

// HOW ABOUT THIS ONE?
const $guard = require("typia/lib/functional/$guard").$guard;
```

EloB:
@samchon That is a even better improvement but requires a lot of rewriting. Maybe take small steps? Make this happen then gradually moving towards that? 🙂

The npm package might got bigger from my changes but when you try to bundle this as a dependency into a single file using tools like esbuild, webpack and rollup than you will see massive improvements. I need this when I write AWS lambda functions to improve performance and keep the file small by dead code elimination.

I think you still need to consider tree shaking when you call functions in the root scope by for instance `Object.assign`.

src/functional/$ProtobufReader.ts
```diff
@@ -189,7 +189,7 @@ export class $ProtobufReader {
   }
 }
 
-const utf8 = new TextDecoder();
-const N01 = BigInt(0x01);
-const N7F = BigInt(0x7f);
-const N80 = BigInt(0x80);
+const utf8 = /** @__PURE__ */ new TextDecoder();
+const N01 = /** @__PURE__ */ BigInt(0x01);
```
EloB, line 193: I think these could be change to not execute functions in the root scope.

```js
const N01 = 0x01n;
const N7F = 0x7fn;
const N80 = 0x80n;
```

samchon:
image

https://bundlephobia.com/package/typia@6.0.0-dev.20240118

Succeeded to adjust the require function calling. 

Bundling size dramatically reduced, and it would be suitable for the tree shaking.

I'll publish your PR as v5.3.12 version after detailed testing, and require statement version to be v6.

Thanks for contribution.

src/functional/$ProtobufReader.ts
EloB: The syntax for `BigInt` without using `BigInt(value)` is to add an "n" (`1n === BigInt(1)`) in the end of a integer.

src/transformers/CallExpressionTransformer.ts
```diff
@@ -92,38 +92,38 @@ import { ReflectMetadataTransformer } from "./features/reflect/ReflectMetadataTr
...
     (project: IProject) =>
-    (expression: ts.CallExpression): ts.Expression | null => {
-      //----
-      // VALIDATIONS
-      //----
-      // SIGNATURE DECLARATION
-      const declaration: ts.Declaration | undefined =
-        project.checker.getResolvedSignature(expression)?.declaration;
-      if (!declaration) return expression;
+      (expression: ts.CallExpression): ts.Expression | null => {
```
EloB: I will revert these. My editor must had formatted them.

EloB:
What do you think to extract the helpers into a separate library? What I read in some of the issues were that you couldn't have ESM output because `ts-patch` had some issues?

According to this comment.
https://github.com/samchon/typia/issues/752#issuecomment-1683044268

Is the `ts-patch` only used for the `transformer`? If so then we could have two seperate packages like `typia` (dependency) containing the helper functions and `typia-plugin-transformer` (devDependency). Then production builds only contains the helpers and not the actual compiler plugin. Making the actual npm bundle smaller :)

One thing that I didn't tell was that I use a ESM output from `typia` but only the helpers. That is the last thing I needed to get dead code elimination to work.

EloB:
This is my draft for esbuild plugin that intercept typescript files containing strings of `"typia"` or `'typia'`. It transpiles them using the normal typescript compiler then passes the result to esbuild so it could do the dead code elimination.

It probably needs some more love before I would consider it stable but it works.

```ts
import * as ts from "typescript";
import * as fs from "fs/promises";
import * as path from "path";

const typiaEsbuildPlugin = (
  {
    tsconfigPath,
    cwd = process.cwd(),
    nodeModulesPath = path.join(cwd, "node_modules"),
  }: TypiaPluginOptions = {}
): esbuild.Plugin => ({
  name: "typia",
  setup(build) {
    let loadedConfig = {};
    let baseDir = cwd;
    const fileName = ts.findConfigFile(
      cwd,
      ts.sys.fileExists,
      tsconfigPath
    );
    if (tsconfigPath !== undefined && !fileName) throw new Error(`Failed to open '${fileName}'`);
    if (fileName) {
      const text = ts.sys.readFile(fileName);
      if (text === undefined) throw new Error(`failed to read '${fileName}'`);

      const result = ts.parseConfigFileTextToJson(fileName, text);

      if (result.error !== undefined) throw new Error(`failed to parse '${fileName}'`);

      loadedConfig = result.config;
      baseDir = path.dirname(fileName);
    }
    const parsedConfig = ts.parseJsonConfigFileContent(
      loadedConfig,
      ts.sys,
      baseDir
    );
    
    // Use the TypeScript files that are still ESM. To improve dead code elimination.
    build.onLoad({ filter: /node_modules\/typia\/lib\/index\.js$/ }, async (args) => {
      const fileName = path.join(nodeModulesPath, "typia", "src", "index.ts");
      const contents = await fs.readFile(fileName, "utf8");
      return { contents, loader: "ts", resolveDir: path.dirname(fileName) };
    });

    // Look for strings in TypeScript containging "typia" or 'typia' and compile them.
    build.onLoad({ filter: /\.([cm]?ts|tsx)$/, namespace: "file" }, async (args) => {
      const source = await fs.readFile(args.path, "utf8");
      if (!/(["'])typia\1/.test(source)) return undefined;
      const program = ts.createProgram([args.path], parsedConfig.options);
      const sourceFile = program.getSourceFile(args.path);
      let contents;
      program.emit(sourceFile, (fileName, compiledSource) => {
        contents = compiledSource
      });
      return { contents, loader: "js" };
    });
  },
});
```

samchon:
If above plugin code works, it would be nice. Thanks for research.

It is okay to publiish by your own package like `typia-esbuild-plugin`, and also okay to sending a PR to making a namespaced module `@typia/esbuild-plugin` (or better name) into the `packages` directory.

> If make your own package, please tell me the setup way or send PR for the guide documents
>
> https://typia.io/docs/setup/

Also in the `esbuild` plugin case, no need to setup the `ts-patch`, so I think ESM problem would not be happened.

EloB:
Yes the plugin works. There might be some issues in watch mode but I will take a look at that. For instance if the "type" that is consumed by Typia is written in another file. I don't think it will rebuild then because it isn't watched right now but the plugin system has ways to tell what files to listen to.

Like this:
```ts
import typia from "typia";
import type { MyResponse } from "./types";

const validate = typia.createIs<MyResponse>();
```

EloB:
It doesn't include the source maps right but I will fix that too.

samchon:
```typescript
const createValidateDecodePure = /** @__PURE__ */ Object.assign<
  typeof createValidateDecode,
  {},
  {}
>(
  createValidateDecode,
  /** @__PURE__ */ Namespace.validate(),
  /** @__PURE__ */ Namespace.protobuf.decode("createValidateDecode"),
);
export { createValidateDecodePure as createValidateDecode };
```
 
@EloB Changed like above to hide namespace nested functions in the declaration files.

It seems all things are ready and if you agree, I'll publish it.

samchon (approved):
Good, I'd missed somethinig.

Now, may I do publish?

samchon (approved):
Everything working fine. Thanks for contribution and waiting for your esbuild plugin.

EloB:
@samchon Thanks mate for such fast feedback, reviewing and merging. I don't know if you noticed my last push/commit there. I made a change to use the correct reference in the `Object.assign`. The dead code elimination in `esbuild` were to aggressive and were removing code that shouldn't be removed. The last commit corrected that so dead code elimination works as expected now.

Also want to thank you for this amazing tool/plugin/library! I'm really impressed what you have achieved.

# Another interesting find
I discovered it's better to use `typia` like this if you want proper dead code elimination.
```ts
// Good, allowed proper dead code elimination.
import { is, tags } from "typia";

// Good, allowed proper dead code elimination.
import * as typia from "typia";

// Bad, was including everything in my bundle. 
import typia from "typia";

// Bad, was including everything in my bundle. 
import typia, { is, tags } from "typia";
```

# Still on my wishlist
I also want typia as ESM output. That is one of the last thing that is needed for dead code elimination.

PR #929, opened 2024-01-18
Complete #752 - reduce bundling size by using `require()` function.

samchon:
Inspired by #928, and suceeded to reduce the bundling size dramatically.

image


samchon (commented):

samchon:
It occurs undefined error when `vite` project being used.

samchon (commented):
It occurs undefined error when `vite` project being used.

samchon:
Will try another way (using `import` statement) at the next major version.

PR #946, opened 2024-02-02
Close #940 - `typia.http.formData<T>()` function.

### LINKED ISSUE #940: Support `FormData` like `query()`
## Feature Request

Very much like #815 or #819, I would like to see the implementation with [FormData](https://developer.mozilla.org/docs/Web/API/FormData).

INPUT:

```json
{
  "input1": "foo",
  "input2": "15",
  "checkbox": "on",
  "arr": "a,b"
}
```
```json
{
  "input1": "foo",
  "input2": "15",
  "arr": "a,b"
}
```

Make it be able to parse and validate against this interface:

```ts
interface MyForm {
  input1: string;
  input2: number;
  checkbox: boolean; // "on" means true, else false
  arr: string[];
}
```

if not

```ts
interface MyForm {
  input1: string;
  input2: string;
  checkbox?: "on";
  arr: string;
}
```

If this is out of scope, feel free to close it.

samchon:
```typescript

interface ObjectHttpFormData {
  strings: string[];
  number: number;
  integers: Array<number & tags.Type<"int32">>;
  blob: Blob;
  blobs: Blob[];
  file: File;
  files: File[];
}
const data: ObjectHttpFormData = typia.http.assertFormData<ObjectHttpFormData>();
```

Converts `FormData` to an object typed instance, including `Blob` and `File` types.


samchon:
Close #940

samchon (commented):
Succeded to pass the test program, but needs node +18 version.

Issue #953, opened 2024-02-06
Explicit validation failure in error report


t3ned:
## Feature Request

> [Note: this issue orginates from nestia, but it's more suited here](https://github.com/samchon/nestia/issues/785)

### Overview
I'm using nestia configured with the `validate` function for runtime validation in my nestjs applications.

I recently discovered nestia, and everything works great! However, one of the caveats is error handling.

Currently, there is no way to identify exactly what validation stages failed without decoding the value of [`IValidation.IError.expected`](https://github.com/samchon/typia/blob/master/src/IValidation.ts#L18).

I would like to map the error messages to be consistent with the API response structure of my applications. However, that is not possible without writing some sort of decoding algorithm which reads the expected type, and generates a human-readable message.

### Current behaviour
Here's an example error output using the [AOTC example from the nestia docs](https://nestia.io/docs/pure/#aot-compilation):
```json
{
  "errors": [
    {
      "path": "$input.title",
      "expected": "string & MinLength<3>",
      "value": "a"
    },
    {
      "path": "$input.body",
      "expected": "string"
    },
    {
      "path": "$input.files",
      "expected": "Array<IAttachmentFile>",
      "value": 1
    }
  ],
  "message": "Request body data is not following the promised type."
}
```

This is okay, but the [`IValidation.IError.expected`](https://github.com/samchon/typia/blob/master/src/IValidation.ts#L18) value requires some decoding to *humanly* identify that:

* >  `$input.title` must be at least `3` characters long
* >  `$input.body` must be a `string`
* >  `$input.files` must be an `array`

### Proposed solution
It would be great to generate the precise details of a validation error so I can map the errors to a desired message. Here's an example of what that might look like with various types of validations:

```json
{
  "errors": [
    {
      "path": "$input.title",
      "expected": "string & MinLength<3>",
      "value": "a",
      "details": {
        "tag": "MinLength",
        "value": 3
      }
    },
    {
      "path": "$input.body",
      "expected": "string",
      "details": {
        "tag": "BaseType",
        "value": "string"
      }
    },
    {
      "path": "$input.files",
      "expected": "Array<IAttachmentFile>",
      "value": 1,
      "details": {
        "tag": "BaseType",
        "value": "Array"
      }
    }
  ],
  "message": "Request body data is not following the promised type."
}
```

samchon:
[...]

PR #962, opened 2024-02-11
Explicit validation failure in error report

t3ned:
I'm looking to address the issue I opened last week (https://github.com/samchon/typia/issues/953)

I'm not overly familar with the codebase, so any suggestions for improvements would be great!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea/
+
 bin/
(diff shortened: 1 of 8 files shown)


samchon:
https://github.com/samchon/typia/issues/953#issuecomment-1939088734

PR #937, opened 2024-01-23
Support Parcel and npm link

andymac4182:
Hi,

This adds a ProgramTransformer to enable Parcel to use `tsc` and typia. This also removes `node_modules` from the location check to enable `npm link` to test in projects while working on the project. 

I am not sure where to put a check to make sure this works with Parcel. Would appreciate some advise on where to put it.

============== TEST TBA ==================

Before submitting a Pull Request, please test your code. 

If you created a new feature, then create the unit test function, too.

```bash
# COMPILE
npm run build

# RE-WRITE TEST PROGRAMS IF REQUIRED
npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```

Learn more about the [CONTRIBUTING](CONTRIBUTING.md)

PR #964, opened 2024-02-15
Take #937 - To simplifying typia path checking.

samchon:
This implementation has been taken from the #937 of @andymac4182.

As that PR had been delayed too much time, I've taken a piece of his good idea, and publish the concept earlier.


andymac4182:
Thanks :) Sorry I am on leave at the moment and got a bunch of family stuff on. 

Thanks for taking this on 👍

PR #973, opened 2024-02-21
Docs: fix documentation for the Restrictions section inside the assert function docs

surya-maximl:
Fix a small typo in the Docs


samchon (commented):
The `must` word is not proper.

PR #982, opened 2024-02-27
Enhance `RandomGenerator.pattern()`.

samchon:
To avoid wrong pattern string value generation by repeating same behavior with validation.

Default random seeder of `typia.random<T>()` function is  `RandomGenerator`, and its `RandomGenerator.pattern()` function is utilizing the `randexp` library for random pattern string value generation. By the way, `randexp` rarely failed to generate proper patten string value, so that it can't pass the validate function of the `typia.is<T>()` function with same type.

This PR avoids such mis-pattern string value generation case, by repeating the random pattern string value generation 10 times, until succeded to pass the pattern validation.

- idea by @AcrylicShrimp


samchon (commented):
Suceeded to pass the test program.

Issue #954, opened 2024-02-06
Function contracts similar to zod


lorefnon:
## Feature Request

Hello, thanks for maintaining and open sourcing typia. I started exploring it recently and it looks great.

Curious if you'd consider adding validated functions similar to zod ([Docs](https://zod.dev/?id=functions)). 

So given a function like: 

```ts
const getFullName = (firstName: string, lastName: string): string => `${firstName} ${lastName}`
```

We'd be able to generate a wrapper that performs validations of input and output but retains the same signature. 

```ts
const getFullNameGuarded = (firstName: string, lastName: string): string => {
    typia.assertEquals<string>(firstName);
    typia.assertEquals<string>(lastName);
    const out =  `${firstName} ${lastName}`;
    typia.assertEquals<string>(out);
    return out;
}
```

In case of async function, we'd await on the promise and validate the resolved values.

samchon:
Which function name do you want about that feature?

May you suggest with function type declaration like one of below?

- https://github.com/samchon/typia/blob/master/src/module.ts
- https://github.com/samchon/typia/blob/master/src/protobuf.ts

lorefnon:
Hi @samchon, thanks for getting back to me. Would something like this be ok?

```ts
function guard<TIn extends unknown[], TOut>(impl: (...args: TIn) => TOut): (...args: TIn) => TOut;

function guardAsync<TIn extends unknown[], TOut>(impl: (...args: TIn) => Promise<TOut>): (...args: TIn) => Promise<TOut>;
```

samchon:
Then which namespace do you want?

`typia.functional.guard(async (value: T1): Promise<O1> => {})` would be okay?

lorefnon:
Sure, that sounds great.

samchon:
Upgrade to v5.5 and use like this:

https://typia.io/docs/validators/functional/

lorefnon:
Thank you 🙂 This is great.

PR #992, opened 2024-03-05
Complement #954 - `isFunction()` and `validateFunction()`.

samchon:
Added more functional checkers.

```typescript
/**
 * Tests a function.
 *
 * Tests a function, by wrapping the function and checking its parameters and
 * return value through {@link is} function. If some parameter or return value
 * does not match the expected type, it returns `null`. Otherwise there's no
 * type error, it returns the result of the function.
 *
 * By the way, if you want is not just testing type checking, but also finding
 * detailed type error reason(s), then use {@link assertFunction} or
 * {@link validateFunction} instead.
 *
 * On the other hand, if you don't want to allow any superfluous properties,
 * utilize {@link equalsFunction}, {@link assertEqualsFunction} or
 * {@link validateEqualsFunction} instead.
 *
 * @template T Target function type
 * @param func Target function to test
 * @returns The wrapper function with type tests
 *
 * @author Jeongho Nam - https://github.com/samchon
 */
function isFunction<T extends (...args: any[]) => any>(
  func: T,
): T extends (...args: infer Arguments) => infer Output
  ? Output extends Promise<infer R>
    ? (...args: Arguments) => Promise<R | null>
    : (...args: Arguments) => Output | null
  : never;

/**
 * Validates a function.
 *
 * Validates a function, by wrapping the function and checking its parameters and
 * return value through {@link validate} function. If some parameter or return value
 * does not match the expected type, it returns {@link IValidation.IError} typed
 * object. Otherwise there's no type error, it returns {@link IValidation.ISuccess}
 * typed object instead.
 *
 * For reference, {@link IValidation.IError.path} would be a little bit different with
 * individual {@link validate} function. If the {@link IValidation.IError} occurs from
 * some parameter, the path would start from `$input.parameters[number]`. Otherwise
 * the path would start from `$input.return`.
 *
 * - `$input.parameters[0].~`
 * - `$input.return.~`
 *
 * By the way, if what you want is not finding every type errors, but just finding
 * the 1st type error, then use {@link assertFunction} instead. Otherwise, what you
 * want is just validating parameters or return value only, you can use
 * {@link validateParameters} or {@link validateReturn} instead.
 *
 * On the other hand, if you don't want to allow any superfluous properties, utilize
 * {@link validateEqualsFunction} or {@link assertEqualsFunction} instead.
 *
 * @template T Target function type
 * @param func Target function to validate
 * @returns The wrapper function with type validations
 *
 * @author Jeongho Nam - https://github.com/samchon
 */
function validateFunction<T extends (...args: any[]) => any>(
  func: T,
): T extends (...args: infer Arguments) => infer Output
  ? Output extends Promise<infer R>
    ? (...args: Arguments) => Promise<IValidation<R>>
    : (...args: Arguments) => IValidation<Output>
  : never;
```


bogeeee:
Hi,
i saw this in the docs already ;) but not yet in the release, so i wondered and came here ;) Nontheless...

I have an improvement suggestion: You should mark all these functional + the [reusable's](https://typia.io/docs/validators/assert/#reusable-functions) as **convenience**. Because these are not essential an be achieved by the user with just max 3 lines of code himself (i.e. `typia.assertEquals<Parameters<myFunction>>(myParameterAsArray)` ) 
i just happened to do this line successfully a few days ago, and was even impressed that Typia followed the `Parameters<...>` type ;)

The docs for the all the validator functions is rather long already and the user is just overwhelmed when thrown all essential and non-essential function-docs over at him

bogeeee:
Also would rename `typia.functional` to `typia.functions`, cause often enough in the IT do people mix up "function oriented programming" with "functional programming" and fans of real functional programming languages like elm, purescript, haskell might get angry about such missuse of the term all the time.

samchon:
Changed guide documents' example codes following your advise.

However, but about the module name, It has been published yet.

Issue #1059, opened 2024-05-22
[Suggest] typia date format follow `RFC3339`


rojiwon123:
## Summary
suggest that the `typia.tags.Format > "date-time" | "date" | "time"` format follow `RFC3339`.

## Problem 
The current `typia.tags.Format<'date-time'>` format follows just ISO8601. it allows time and timezone to be omitted. And the type is represented as `date-time` in swagger when building swagger documentation.
However, the `date-time` format in swagger is defined by the `date-time` format in the `RFC3339` standard, which requires the time and timezone to be indicated.

In conclusion, the current typia definition of `date-time` leads to inaccuracies in documentation and confusion of terms.

Therefore, I suggest the following

### format
- `RFC3339` standard `full-date` format is define `typia.tags.Format<'date'>` 
- `RFC3339` standard `full-time` format is define `typia.tags.Format<'time'>` 
- `RFC3339` standard `date-time` format is define `typia.tags.Format<'date-time'>` 

### code
- [FormatCheatSheet.ts](https://github.com/samchon/typia/blob/master/src/tags/internal/FormatCheatSheet.ts)
```ts
export const FormatCheatSheet = {
// ...
"date-time": RegexCall(`/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T|\\s)([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](?:\\.[0-9]{1,9})?(Z|[+-]([01][0-9]|2[0-3]):[0-5][0-9])$/i`),
"date": RegexCall(`/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])$/`),
"time": RegexCall(`/^([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](?:\\.[0-9]{1,9})?(Z|[+-]([01][0-9]|2[0-3]):[0-5][0-9])$/i`),
// ...
};
```

## Example

```ts
// before
import typia from "typia"
typia.is<string & typia.tags.Format<"date-time">>("2024-05-23"); // true
typia.is<string & typia.tags.Format<"date-time">>("2024-05-23T23:59:59Z"); // true
```

```ts
// after
import typia from "typia"
typia.is<string & typia.tags.Format<"date-time">>("2024-05-23"); // false
typia.is<string & typia.tags.Format<"date-time">>("2024-05-23T23:59:59Z"); // true
```

## Reference

[...]

PR #1060, opened 2024-05-25
Feat: apply RFC3339 standard to typia.tags.Format<"date-time"|"date"|"time"> 

rojiwon123:
this pr is about #1059


samchon:
Great job, would be published to the next version.

samchon (approved):
Great job, would be published to the next version.

By the way, the `tags.mdx` file would be rolled back.

Issue #1063, opened 2024-06-01
`validate` and `assert` functions bundle size is so huge


ryoppippi:
## Question
Hello.
I have found that the bundle size is larger when using `validate` than when using `check` in `typia`.
However, the difference in capacity is much larger than the difference between using `parse` and `safeparse` in valibot.
Why is there such a difference in bundle size?

ryoppippi:
I found that validation logic is huge.
Is there an option to output a simplified version of Validation logic?

Also, I don't need validation logic for `assert` function (looks like the validation logic is used for `errorFactory`)
Is there any options to remove those logic?

samchon:
Can you show me any example playground link, and point somewhere you think that too much huge thing?

I have structured the generated code to reduce the bundling size as much as possible, but if users still feel that it is insufficient, I would like to know points of improvement.

ryoppippi:
Thank you.

So this code:
```ts
import typia, { tags } from "typia";

interface IMember {
  id: string & tags.Format<"uuid">;
  name: string;
  age: number &
    tags.Type<"uint32"> &
    tags.Minimum<20> &
    tags.ExclusiveMaximum<100>;
}

const member = {
  id: '1',
  name: 'a',
  age: 30
}

typia.assertGuard<IMember>(member)

console.log(member)
```

generates the code like below.
<summary>generated code</summary>
<details>
```ts
import typia from "typia";
const member = {
  id: "1",
  name: "a",
  age: 30,
};
((input, errorFactory) => {
  const __is = (input) => {
    return (
      "object" === typeof input &&
      null !== input &&
      "string" === typeof input.id &&
      /^(?:urn:uuid:)?[0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12}$/i.test(
        input.id,
      ) &&
      "string" === typeof input.name &&
      "number" === typeof input.age &&
      Math.floor(input.age) === input.age &&
      0 <= input.age &&
      input.age <= 4294967295 &&
      20 <= input.age &&
      input.age < 100
    );
  };
  if (false === __is(input))
    ((input, _path, _exceptionable = true) => {
      const $guard = typia.assertGuard.guard;
      const $ao0 = (input, _path, _exceptionable = true) =>
        (("string" === typeof input.id &&
          (/^(?:urn:uuid:)?[0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12}$/i.test(
            input.id,
          ) ||
[...]

PR #1064, opened 2024-06-02
feat: Support ESM and CJS builds with esbuild

ryoppippi:
I made a tiny changes for build sources.

- support ESM because it enables tree shaking( solves https://github.com/samchon/typia/issues/1063)
- Build with `esbuild` instead of `tsc`, which makes build time faster

====
This commit introduces changes to support both ESM and CJS builds. The `package.json` file has been updated to include new build scripts and dependencies. Two new `tsconfig` files have been added for ESM and CJS respectively, and the original `tsconfig.json` has been removed. The `module` and `types` fields in `package.json` have been updated to point to the new ESM build.

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```


samchon:
Good idea, I'll try ESM rollup bundling on this branch after my job duty. 

Thanks for good suggestion

https://github.com/samchon/tgrid/blob/master/rollup.config.js

ryoppippi:
@samchon 
Hi!
I'm tring tsup for bundle.
It is so good.
I'll create another PR for my idea.

ryoppippi:
Also, I found some tree-shaking issues.
I'll create some PRs too.

samchon:
Tried, but understood that, this change needs major update.

As `typia`'s internal utility functions are used in the `nestia` and other third party libraries, I have to take a major update.

If not, the `export` property of `package.json` would occur many problems.

Therefore, it would be take long time to adapt this suggestion \o/.

samchon:
By the way, the next v7 update does not utilize namespace strategy more.

Instead, `typia` generates proper `import` statements only when required.

Therefore, please take this issue to me, then it would be fixed in someday.

https://github.com/samchon/typia/tree/features/import

PR #1074, opened 2024-06-05
feat(setup.md): added how to integrate with `typia` w/ `unplugin-typia`

ryoppippi:
related https://github.com/ryoppippi/unplugin-typia/issues/1

I added the guide using typia and unplugin-typia

If you need more information to add the `setup.mdx`, let me know!

Cheers!!

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -417,7 +417,7 @@ export default defineConfig({
 
 
 
-## Webpack
+## webpack
 <Tabs items={['npm', 'pnpm', 'yarn']}>
   <Tab>
 ```bash filename="Terminal" copy showLineNumbers
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -1,4 +1,4 @@
-import { Tabs, Tab } from 'nextra-theme-docs'
+import { Callout, Tabs, Tab } from 'nextra-theme-docs'
 
 import Alert from '@mui/material/Alert';
 import AlertTitle from '@mui/material/AlertTitle';
@@ -78,7 +78,7 @@ If you are using a non-standard TypeScript compiler such as the following, you w
     - [esbuild](https://esbuild.github.io/)
     - [Babel](https://babeljs.io/)
 
-Or, if you are using a bundler such as `webpack`, `vite`, `esbuild`, or `next.js`, consider using the [unplugin-typia](#unplugin-typia).
+Or, if you are using a bundler such as `webpack`, `vite`, `esbuild`, or `next.js`, consider using the [`unplugin-typia`](#unplugin-typia).
 
 ## Transformation
 ### Concepts
@@ -418,7 +418,7 @@ export default defineConfig({
 
 
 
-**Also check out the [unplugin-typia](#unplugin-typia) guide for easier integration.**
+**Also check out the [`unplugin-typia`](#unplugin-typia) guide for easier integration.**
 
 ## webpack
 <Tabs items={['npm', 'pnpm', 'yarn']}>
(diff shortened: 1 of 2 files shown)


ryoppippi:
I force-pushed the commits because the commit messages are odd.
Finally, please review it!!

samchon:
Is there special reason not to configure `compilerOptions.plugins` property of `tsconfig.json` file?

ryoppippi:
There is nothing special reason. 
However, even tho you configure it, it works!!

ryoppippi:
There is no special reason. 
However, even if you don't configure it, it works!

Also, I don't think this might happens, but if you configure it in `tsconfig.json` the transformer runs twice( I haven't check this out ..., and I don't think that happens, but maybe..)

samchon:
Thanks for your contribution.

I did little bit emend on your content. If my way is not right, it is okay to rolling back.

After the `unplugin-typia` becomes stable, then please tell me. I'll make the `unplugin-typia` section to be directly under the `Transformation` section. The other sections `Generation`, `Webpack` and `NX` would be after of yours.

samchon:
> There is no special reason. However, even if you don't configure it, it works!
> 
> Also, I don't think this might happens, but if you configure it in `tsconfig.json` the transformer runs twice( I haven't check this out ..., and I don't think that happens, but maybe..)

Then may I guide just to use the `npx typia setup` command?

It is because even when developing frontend project, I think some developers may run independent `ts` file in the NodeJS environment for miscellaneous reasons.

ryoppippi:
> After the unplugin-typia becomes stable, then please tell me. I'll make the unplugin-typia section to be directly under the Transformation section.

Yes! Thank you
I tested it with several bundlers and it works fine rn.
For me it is stable...
But idk, I don't have confidence for this new plugin because no one hasn't try this one.
It may takes a couple of months maybe


> Then may I guide just to use the npx typia setup command?	

You are right!!
I had not considered using it in conjunction with generation mode.
`npx typia setup` is fine.

So should I fix the document? or will you do?

samchon:
I'll change it and will get review from you.

By the way, if `unplugin-typia` is enough stable for you, then I'll make your section to be upper after two weeks later.

If there be some issue in the two weeks, just tell me please.

And when you want to change the setup guide content, just send me a PR anytime when you want.

ryoppippi:
Nice! Thanks!!

samchon:
@ryoppippi Finished changing your content. Can you take a review? As I'm not familiar with English, there can be some errors.

ryoppippi:
Also, if unplugin-typia is enough stable, we can delete the previous guide for [`webpack`](http://localhost:3000/docs/setup/#webpack) and [`vite`](https://typia.io/docs/setup/#vite)

samchon:
https://nestia.io/docs/setup/#webpack

Only about the `webpack` chapter, I'm careful about it due to backend case like above.

The other sections, if you feel confidence, it is okay to remove and make your content to be upper even now.

samchon:
For reference, if you accept my change and this PR me merged, the website would be automatically published by pipeline

ryoppippi:
@samchon 
I checked your modification.

LGTM

website/pages/docs/setup.mdx
```diff
@@ -31,7 +31,7 @@ yarn typia setup --manager yarn
 
 If you're using standard TypeScript compiler, you can use [transform mode](#transformation).
 
-Assuming you are using [transformation mode] simply run `npx typia setup` which will provide sane defaults.
+Assuming you are using [transformation mode](#transformation) simply run `npx typia setup` which will provide sane defaults.
```
ryoppippi, line 34: Oh! Thanks!!

ryoppippi:
> Only about the webpack chapter, I'm careful about it due to backend case like above.

Nice Thanks!

samchon:
I'll prepare a new PR that making `unplugin-typia` section to be upper at Saturday or Sunday of this week. 

Thanks for interseting and greating plugin.

PR #1076, opened 2024-06-05
fix(website/setup.mdx): add typia installation step for pnpm/yarn

ryoppippi:
Sorry, I missed that.
The part about installing typia with pnpm and yarn was missing, so I added it.


samchon (approved):
Okay, I will publish it right now

Issue #1078, opened 2024-06-07
Bun build support


ryoppippi:
## Feature Request
Thanks for your suggestion. Feel free to just state your idea. Writing the issue, it would better to filling the below items:

  - A description of the problem you're trying to solve.
  - An overview of the suggested solution.
  - Examples of how the suggestion would work in various places.
    - Code examples showing the expected behavior.

`randexp.js` and thier dependency `DRange` is complied by old CJS.
It works fine with node.js, but `bun build` fails bundling ( I know `typia` does not support bun officially)

Related: https://github.com/ryoppippi/unplugin-typia/issues/44

For solving, I'll implement an alternative of `randexp.js` 

(This is just suggestion. If you do not want it, close this issue)

samchon:
If you can fully replace it, I don't matter.

ryoppippi:
@samchon 
No I found another solution.
I'll create a PR.

PR #1079, opened 2024-06-07
feat: Add exports field to package.json/ fix rollup config

ryoppippi:
Related to #1078
In this PR.
- You can set the `exports` filed only when published, so that only the appropriate ones are published.
- The exports are set to publishConfig. This means that internal packages can also be used in test, but direct import is no longer possible once published
- The `exports` filed allows you to read the typescript directly when using `bun`, so that the build does not fail!
- rollup has been re-configured to generate `mjs` files according to the directory structure

related: #1078 

======

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,5 +102,42 @@
     "lib",
     "src"
   ],
-  "private": true
+  "private": true,
+  "publishConfig": {
+    "exports": {
+      ".": {
+        "bun": "./src/index.ts",
+        "node": "./lib/index.js",
+        "import": "./lib/index.mjs",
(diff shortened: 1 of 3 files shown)


ryoppippi:
Maybe I'm done with this PR!
Plz, reveiw it!

ryoppippi:
Maybe I'm done with this PR!
Plz, review it!

ryoppippi:
Ok, I'll split the PR because
sorry about that

PR #1081, opened 2024-06-07
feat(rollup.config): support multiple formats and preserve modules/ split esm bundle files

ryoppippi:
So this PR fixes how to generate `esm` and `cjs`.
This works fine with bun 
```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```


ryoppippi:
If we do not use `exports` property in `package.json` you do not need merge it

samchon:
Got it.

PR #1082, opened 2024-06-07
feat(package.json): add publishConfig for package exports

ryoppippi:
Related to #1078
In this PR.
- You can set the `exports` filed only when published, so that only the appropriate ones are published.
- The exports are set to publishConfig. This means that internal packages can also be used in test, but direct import is no longer possible once published
- The `exports` filed allows you to read the typescript directly when using `bun`, so that the build does not fail!

This should be merged after #1081

======

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,5 +102,39 @@
     "lib",
     "src"
   ],
-  "private": true
+  "private": true,
+  "publishConfig": {
+    "exports": {
+      ".": {
+        "bun": "./src/index.ts",
+        "node": "./lib/index.js",
+        "import": "./lib/index.mjs",
+        "require": "./lib/index.js",
+        "types": "./lib/index.d.ts",
+        "default": "./lib/index.js"
+      },
+      "./lib": {
+        "bun": "./src/index.ts",
+        "node": "./lib/index.js",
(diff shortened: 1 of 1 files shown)


samchon:
This is good PR, but I hope to be careful about `exports` property of `package.json`, because my another libraries like `nestia` are using internal components of `typia` like `src/utils/ExpressionFactory.ts` file case (Even some compiler related commercial projects I've developed are reflected by it).

To adapt this feature, I need major update in both `typia` and `nestia`. Please wait for some months.

samchon (commented):
This is good PR, but I hope to be careful about `exports` property of `package.json`, because my another libraries like `nestia` are using internal components of `typia` like `src/utils/ExpressionFactory.ts` file case (Even some compiler related commercial projects I've developed are reflected by it).

To adapt this feature, I need major update in both `typia` and `nestia`. Please wait for some months.

ryoppippi:
OK! Thanks
Then I'll think about other solutions! Thanks

PR #1086, opened 2024-06-08
`unplugin-typia` to be main feature of setup guide docs

samchon:
@ryoppippi I've decided to make `unplugin-typia` to be main feature of setup guide docs right now.

 The reason why not waiting for its stablization is, the stabitlization can be boosted by feedbacks from `typia` users.

Therefore, needs your review.

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -226,14 +226,15 @@ Of course, if you don't use any ["Comment Tags"](./validators/tags/#comment-tags
 [`unplugin-typia`](https://jsr.io/@ryoppippi/unplugin-typia) is a plugin to integrate `typia` into your bundlers seamlessly.
 
 Currently, `unplugin-typia` supports the following bundlers:
-  - [Vite](https://vitejs.dev/)
-  - [Rollup](https://rollupjs.org/)
-  - [esbuild](https://esbuild.github.io/)
-  - [webpack](https://webpack.js.org/)
+  - [Bun](https://jsr.io/@ryoppippi/unplugin-typia/doc/bun/~/default)
+  - [Esbuild](https://jsr.io/@ryoppippi/unplugin-typia/doc/esbuild/~/default)
+  - [Farm](https://jsr.io/@ryoppippi/unplugin-typia/doc/farm/~/default)
   - [Next.js](https://nextjs.org/)
-  - [Rspack](https://www.rspack.dev/)
-  - [Rolldown](https://rolldown.rs/)
-  - [farm](https://farm-fe.github.io/)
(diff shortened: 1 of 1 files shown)


website/pages/docs/setup.mdx
```diff
@@ -242,183 +200,115 @@ yarn prepare
...
-    null !== input &&
-    "string" === typeof input.id &&
-    $is_uuid(input.id) &&
-    "string" === typeof input.email &&
-    $is_email(input.email) &&
-    "number" === typeof input.age &&
-    19 <= input.age &&
-    100 >= input.age
-  );
+```javascript filename="next.config.mjs" copy showLineNumbers
```
ryoppippi, line 283: I wrote more detailed version of config for Next.js. so please replace it.
https://jsr.io/@ryoppippi/unplugin-typia/doc/next/~/default

website/pages/docs/setup.mdx
```diff
@@ -242,183 +200,115 @@ yarn prepare
...
+Currently, `unplugin-typia` supports the following bundlers:
+  - [Vite](https://vitejs.dev/)
+  - [Rollup](https://rollupjs.org/)
+  - [esbuild](https://esbuild.github.io/)
+  - [webpack](https://webpack.js.org/)
+  - [Next.js](https://nextjs.org/)
+  - [Rspack](https://www.rspack.dev/)
+  - [Rolldown](https://rolldown.rs/)
+  - [farm](https://farm-fe.github.io/)
+
```
ryoppippi, line 238: - [bun.build](https://bun.sh/docs/bundler)

I also added a [partial support for Bun](https://jsr.io/@ryoppippi/unplugin-typia/doc/bun/~).
It works for generating code, but Bun fails to solve the CJS codes including `randexp.js` (like https://github.com/oven-sh/bun/issues/7465)
So It's up to you if you include `Bun` to the official document.

I think this plugin for `Bun.build` will be the easiest way to try typia in the future, but we need to wait for Bun's stablization

ryoppippi (commented):
Thanks! @samchon 
More or less, This is OK.
I think my `unplugin-typia` is getting stable(and actually I implemented cache feature too!)

I wrote a couple of comments, so please check them out!!

website/pages/docs/setup.mdx
ryoppippi: I updated a Docs.
We can use `Bun.build` fully. So please add this !

samchon:
@ryoppippi Updated as you wish. If you want to modify or add more things, please send a new PR.

ryoppippi:
Thank you!!! @samchon

PR #1092, opened 2024-06-09
feat: Replace dax-sh with child_process for check if bun is available

ryoppippi:
I remove `dax-sh` dependencies because it is too much.
I replaced with `child_process`

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```

### DIFF
--- a/deploy/index.ts
+++ b/deploy/index.ts
@@ -1,6 +1,5 @@
 import chalk from "chalk";
 import cp from "child_process";
-import $ from "dax-sh";
 import fs from "fs";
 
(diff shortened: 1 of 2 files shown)


ryoppippi:
This is just a PR for our CI.

samchon:
Changed a little bit because bun checker does not need to be async function.

samchon (changes requested):
Changed a little bit because bun checker does not need to be async function.

PR #969, opened 2024-02-18
fix: merge installation with setup

seo-rii:
Embedded `npm install typia` into `npx typia setup`.

PR #1093, opened 2024-06-09
feat(TypiaSetupWizard): installs typia 

ryoppippi:
now `SetupWizard` installs `typia`, so that you do not have to install `Typia` before executing setup wizard

### DIFF
--- a/src/executable/TypiaSetupWizard.ts
+++ b/src/executable/TypiaSetupWizard.ts
@@ -21,6 +21,7 @@ export namespace TypiaSetupWizard {
     const args: IArguments = await ArgumentParser.parse(pack)(inquiry);
 
     // INSTALL TYPESCRIPT COMPILERS
+    pack.install({ dev: true, modulo: "typia", version: "latest" });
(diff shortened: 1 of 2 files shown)


samchon (commented):
https://github.com/samchon/typia/pull/969

There had been same PR with you.

However, I'd accepted only of `nestia`. This is because `typia`'s setup wizard has some dependencies like `commander` and `inquirer`, and those dependencies make `npx typia setup` command to be failed when the local machine's NodeJS version is lower than v20. In such reason, I'd accepted only the PR of `nestia` (`nestia` requires Node +v20, but `typia` is not).

Also, as `typia`'s setup wizard specializes TypeScript version due to `peerDependencies` problem, I'm careful for about intruducing to run only `npx typia setup` wizard without local install. For reference, TypeScript project is occuring many break changes on compiler API even in the minor and patch updates.

ryoppippi:
@samchon 
OK I see.
Then maybe we need to modify setup tools itself.
Thanks

PR #1134, opened 2024-06-27
feat(build.yml): Add artifact upload step to workflow

ryoppippi:
This commit introduces a new step in the GitHub Actions workflow for uploading build artifacts. The artifacts include the .tgz files, and the src and lib directories. This will help in debugging and preserving the state of the build at each commit.


samchon (approved):
Thanks for contribution.

PR #1146, opened 2024-07-04
feat(package.json): add sideEffects property

ryoppippi:
With experiments at https://github.com/samchon/typia/issues/752 , `rollup.js` unexpectedly bundle `ret.js`.
However, when we add `sideeffect=true` to package.json, rollup.js can remove it.

I think this is a small change, but need it

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
   "stackblitz": {
     "startCommand": "npm install && npm run test"
   },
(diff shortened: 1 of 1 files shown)


ryoppippi:
I think this needs to be merged soon

samchon (approved):
Good. I'll publish it until Saturday.

PR #1147, opened 2024-07-04
feat(rollup.config.js): Enable inline source maps

ryoppippi:
In current rollup configuration, sourcemap is not linked to typescript file.
So I confugired as an inline sourcemap, so it does not conflict with CJS sourcemaps


ryoppippi:
It is not good way.
I wanna add typescript source to ESM sourcemap, but it is a little bit hard rn

PR #1154, opened 2024-07-08
docs(setup.mdx): Correct setup command for unplugin-typia with Bun

HenryWu01:
There is a mistake in the guidance of installing ``unplugin-typia`` with ``Bun`` as the package manager.

Previous:
```
bun add @ryoppippi/unplugin-typia
```

This command does not contain ``jsr`` which means bun will search the package on ``npmjs`` instead of ``jsr``, resulting in an older version of ``unplugin-typia`` causing issues with newer ``typescript`` version.

Further discussion can be seen at: https://github.com/ryoppippi/unplugin-typia/issues/170

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -284,7 +284,7 @@ yarn typia setup --manager yarn
   </Tab>
   <Tab>
 ```bash filename="Terminal" showLineNumbers copy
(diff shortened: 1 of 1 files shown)


ryoppippi:
@HenryWu01 
Thank you for creating this PR.
This is a correct setup guide

[LGTMeow](https://lgtmeow.com)

PR #1167, opened 2024-07-17
feat: add `Equal` type of Comparison of equivalence

kakasoo:
The existing `Equal` type did not work for the union type. For a more complete type comparison, it is recommended to use the Equal type separately as a separate file.

- https://github.com/toss/es-toolkit/pull/206#discussion_r1680242129


samchon (changes requested):
It would better to propert description comment, and add export keyword.

Also, what about importing the `Equal` type in below files?:

- `Primitive`
- `Resolved`
- `CamelCase`
- `PascalCase`
- `SnakeCase`

samchon (approved):
Good type.

PR #1169, opened 2024-07-17
Type resolver refactoring.

samchon:
Refactored type resolvers like `Resolved` and `PascalCase`.

Also supported `RegExp` as one of native classes.


samchon (commented):
Succeeded to pass test programs.

PR #1170, opened 2024-07-18
Avoid bug of bun

samchon:
Bun suddenly died without any error message.

This is the detour code to avoid the bun's own bug.


samchon (commented):
Succeded to avoid.

PR #1174, opened 2024-07-21
Fix #1166: nonsensible intersection type case

### LINKED ISSUE #1166: Bug with intersection types
## Bug Report
First of all, thanks a lot for all the hard work on typia! This bug report revisits a variant of [issue #856](https://github.com/samchon/typia/issues/856) about better support for intersection types.

We recently encountered the problem that typia does not correctly apply the [`is`-validation](https://typia.io/docs/validators/is/) when some specific intersections exist somewhere nested in the type to be checked.

### 💻 Example
Originally our problem concerned the intersection of arrays in some property of an object, but it turns out that even much simpler intersections are also affected. Thus, the following example with template strings shows a very simplified case and can be tested in [typia's playground ⏯ here](https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDOAzKERwERIqp4DcAUGYQKZwCCscAvHAAYAkA3gM4xTAB2AcwC+qWJx58hwluWpwAQmKasJvASIBGYtVJGyK8+jCVQVxuADJFY8mQDGEfjwRUeAZXVCAXHVimVPG1YUgpHZ3goNwBXABsYC2ZCNAA6YC4AHmMAPgAKGDcYTz0ASnJwlyiuOJNlJORU9IzTPIKPL0EyhydKmPiVZNQ0zOMW-MLijS7u5whYqhTYiEFx9r04AGp8OHS4dEkNHf57WOiAE0OAcjEYS7wAGjgqmuMyuBmuOYWllbaijs2212+wBAhO5yuwUu3geTz6tSgbzCPS+i2Wq3+6y2eB2XD2cAOQiO4IuROusEue34ZzglyhMMez3iSKAA):
```typescript
import typia from "typia";
 
type Art = `${string}art${string}`;
type Bar = `${string}bar${string}`;
 
type ArtBar = Art & Bar;
 
const testString: ArtBar = "bart";
 
const resultArt = typia.is<Art>(testString); // true
const resultBar = typia.is<Bar>(testString); // true
const result = typia.is<ArtBar>(testString); // false 🗲
```

### 📝 Problem circumstances
The different types in the intersection, i.e. `Art` and `Bar` in the example above, may have originally come from external sources such as an OpenAPI type export, and are typically not under our control, or we may not even know exactly what they contain. Thus, the ad hoc workarounds mentioned in the original issue #856 - such as using generics or `Omit` - will usually not work here. Moreover, setting the `is`-validation result to false  in the above example clearly contradicts TypeScript logic, since "bart" is successfully assigned to `testString` with type `ArtBar`. Thus, we still have the impression that typia's behavior here is a bug and not "invalid" and not "wontfix" as mentioned in the original issue #856.

### 💡 Suggestion
Consequently, we would suggest as an improvement for typia that for intersection types (especially in cases without other complications like generics): try to somehow _check each part of the intersection_ independently first and then combine the results with logical `AND`.

If this bug fix is not possible quickly, then it would be great if - as a first, maybe temporary fix - typia could somehow detect such problematic cases and then avoid assigning a boolean value to the `is`-validation `result `at all. Instead, maybe temporarily return `undefined `(or even throw an error), so that we typia-users know that we have to try to use other workarounds until this bug is fixed. 

Such a temporary fix might also be necessary for the [`random`-generator](https://typia.io/docs/random/) on intersection types, because this generator seems to be broken on these intersection types, like the `ArtBar` type from above: It returns a complex object which is clearly not a simple string:

image

Thank you very much already in advance. ❤️

samchon:


samchon (commented):
Passed test programs.

PR #1192, opened 2024-08-03
fix: improve SEO

imjlk:
typia.io website is currently displaying the description meta tag `Nextra: the next docs builder` and Nextra's default description and og tags. it's overriding other meta tags

스크린샷 2024-08-03 오후 5 11 37

Applied:
- add description tag
- add default `head` option with empty component

The `head` option, it must be a bug with Nextra, I found [an issue](https://github.com/shuding/nextra/issues/2918) with this but it hasn't been updated yet.

This may be addressed in a future version update, but I think this will not affect anything else.

### DIFF
--- a/website/next.config.js
+++ b/website/next.config.js
@@ -7,7 +7,7 @@ const withNextra = require("nextra")({
 const config = {
   ...withNextra(),
   output: "export",
-  exportTrailingSlash: true,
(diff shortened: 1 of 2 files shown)


samchon (approved):
Thanks for emension.

PR #1197, opened 2024-08-04
Optimize `random()` and `clone()` functions.

samchon:
No change on interfaces.

Just a little bit performance tuning.


samchon (commented):
Everything is fine, so that passed the test program.

PR #1198, opened 2024-08-04
Fix generaton mode's non-initialized `let` variable problem.

samchon:
When non-initialized `let` variable being generated by `typia` in the JavaScript transformation mode, there would not be any problem. However, it occurs TypeScript compile error in the TypeScript generation mode.

This PR avoids the problem by declaring the non-initailized `let` variable as `any` type. Also, during the generation mode testing, fixed many other miscellaneous bugs.

- Return type problem of
  - `typia.misc.validatePrune()`
  - `typia.misc.assertPrune()`
  - `typia.misc.assertPascal()`, ...
  - `typia.misc.validateCamel()`, ...

### DIFF
--- a/benchmark/package.json
+++ b/benchmark/package.json
@@ -72,6 +72,6 @@
     "suppress-warnings": "^1.0.2",
     "tstl": "^3.0.0",
     "uuid": "^9.0.1",
-    "typia": "../typia-6.7.1.tgz"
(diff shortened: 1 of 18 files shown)


samchon (commented):
Tested completed, and everything is fine about generation mode.

PR #1201, opened 2024-08-06
Support `console.info()` in playground

samchon:


samchon (commented):
Tested, and validated.

PR #1206, opened 2024-08-07
Experiment of samchon/nestia#980: `ts.Type` de/serialization.

samchon:
Succeeded to serialized and deserialize `ts.Type` class instnace from the JSON value level.

Also, to accomplish the experiment, refactored `typia` code never to utlize the `ts.Type` methods.


samchon (commented):
No problem on `typia` side, but cannot sure in the `@nestia/sdk` side.

May needs lots of emensions.

samchon (commented):
No need this feature more, because `nestia` has found another solution.

PR #1217, opened 2024-08-11
  Allow `bigint` type in `JsonApplicationProgrammer`.

samchon:
`typia.json.application<[Types]>()` function still does not allow the `bigint` type, but its internal `JsonApplicationProgrammer.write()` function allows it for 3rd party libraries, especially `@nestia/sdk` case.


samchon (commented):
Test OK

PR #1218, opened 2024-08-12
fix : a typo in docs

hyeonjae1122:
Just fix a typo in docs. link as follows. 

https://typia.io/docs/validators/assert/

Screenshot 2024-08-12 at 10 45 05


samchon (approved):
Thanks for emension.

PR #1231, opened 2024-08-20
Adjust samchon/openapi#40 update to playground.

samchon:
In the JSON schema definition, it is nonsensible to inserting null value to enum property array.

However, Github OpenAPI document is actually doing the crazy thing.

This PR supports it by allowing the null value to the enum proeprty array.


samchon (commented):
OK.

PR #1236, opened 2024-08-22
fix(docs): update SWC usage instructions

ryoppippi:
I fixed a statement about SWC.
My friend misunderstood that he cannot use typia on swc+vite environment because of this statement.
I'd like to show that unplugin-typia enables typia running on swc + several boulders.


samchon (approved):
Good

PR #1239, opened 2024-08-22
Close #760: `IJsonApplication<Version, Types>`

### LINKED ISSUE #760: Feature request: typia.application<T>() to return IJsonApplication<T>
## Feature Request

I'm using typia not only to validate data, but also to extract type fields (or union values) to runtime array.
E.g. typia makes it possible to do:

```typescript
 console.log(unionValues(typia.application<[2 | 1 | 3]>()));
 --> [2, 1, 3] // unknown[]
```

however, i cannot properly type result of my "unionValues" function because it accepts IJsonApplication as parameter, so type cannot be extracted from there. Or, I'll have to specify the type multiple times to get something like that:

```typescript
 console.log(unionValues<[2,1,3]>(typia.application<[2 | 1 | 3]>()));
 --> [2, 1, 3]   // number[]
```

My request: let's extend IJsonApplication like this:

```typescript
export interface IJsonApplication<T> {
    schemas: IJsonSchema[];
    components: IJsonComponents;
    purpose: "swagger" | "ajv";
    type?: T | undefined; 
}
```
this "ephemeral" field will be always undefined during the runtime, but will provide a way to extract the original type from IJsonApplication object at compile-time. 

What do you think?

BTW, typia.application can be used as base to create some other useful functions, like `objectKeys`, etc.

samchon:
@ryoppippi Needs your review.

Also, newly added tag must be such dynamic unique symbol type?

May I change it to just a static property like `__types: Types[]`?

### DIFF
--- a/benchmark/package.json
+++ b/benchmark/package.json
@@ -72,6 +72,6 @@
     "suppress-warnings": "^1.0.2",
     "tstl": "^3.0.0",
     "uuid": "^9.0.1",
-    "typia": "../typia-6.9.0.tgz"
+    "typia": "../typia-6.10.0-dev.20240823.tgz"
   }
 }
--- a/errors/package.json
+++ b/errors/package.json
@@ -32,6 +32,6 @@
     "typescript": "^5.3.2"
   },
   "dependencies": {
-    "typia": "../typia-6.9.0.tgz"
+    "typia": "../typia-6.10.0-dev.20240823.tgz"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "6.9.0",
+  "version": "6.10.0-dev.20240823",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/typescript-json/package.json
+++ b/packages/typescript-json/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json",
-  "version": "6.9.0",
+  "version": "6.10.0-dev.20240823",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
@@ -63,7 +63,7 @@
   },
   "homepage": "https://typia.io",
   "dependencies": {
-    "typia": "6.9.0"
+    "typia": "6.10.0-dev.20240823"
   },
   "peerDependencies": {
     "typescript": ">=4.8.0 <5.6.0"
--- a/src/json.ts
+++ b/src/json.ts
@@ -57,7 +57,7 @@ export function application(): never;
 export function application<
   Types extends unknown[],
   Version extends "3.0" | "3.1" = "3.1",
->(): IJsonApplication<Version>;
+>(): IJsonApplication<Version, Types>;
 
 /**
  * @internal
--- a/src/programmers/json/JsonApplicationProgrammer.ts
+++ b/src/programmers/json/JsonApplicationProgrammer.ts
@@ -1,6 +1,6 @@
 import { OpenApi, OpenApiV3 } from "@samchon/openapi";
 
-import { IJsonApplication } from "../../schemas/json/IJsonApplication";
+import { IJsonApplication, type tag as IJSONTag } from "../../schemas/json/IJsonApplication";
 import { Metadata } from "../../schemas/metadata/Metadata";
 
 import { TransformerError } from "../../transformers/TransformerError";
@@ -44,7 +44,7 @@ export namespace JsonApplicationProgrammer {
     const components: OpenApiV3.IComponents = {};
     const generator = (meta: Metadata): OpenApiV3.IJsonSchema | null =>
       application_v30_schema(true)(components)({})(meta);
-    return {
+    const schema = ({
       version: "3.0",
       components,
       schemas: metadatas.map((meta, i) => {
@@ -56,7 +56,8 @@ export namespace JsonApplicationProgrammer {
           });
         return schema;
       }),
-    };
+    } satisfies Omit<IJsonApplication<"3.0">, typeof IJSONTag>); // check types before as assertion
+    return schema as IJsonApplication<"3.0">;
   };
 
   const v31 = (metadatas: Array<Metadata>): IJsonApplication<"3.1"> => {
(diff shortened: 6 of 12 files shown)


ryoppippi (approved):
@samchon 
static symbol is ok!
Thanks

I tested with the latest code, and my library works fine!

Latest release version (no type error): https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAciRVRoG4AoFgegCpO5UATPnD4BTAgGMowMDGAQAdryKo4AKQDKAEQhi4ndi3rC4AeQCuMMObgBebHGB8AXHAIxJcvHABkuQgDoAFWRhAB4aU2A5GABmACYaAD4SVgNgk3NLGFibOxFxSWlZOWdXd08fGHwCPwBZSIAZYQ8YAAsQgEYk4hTDVQJ5NTEW4RBUVDk+ADVUABsHVBl5EICk2ywWODgAK365QeHR5wAlYRhTKDkgsFDDCDIEZDQ-Hfk-VDAwObEFopCAbQCAF1MDRon4AAw0OAAH1oYPaiQSrE2ADdZvMYMJnAAKSKZZymOQAazkEAA7nIAJQ2JL0J4ASWmcz4PyWK1Y3TYZEJYkWCiaBHOwn2I1QAEEJkyMb8VtiwKgoKgQARnCpdiLRuMpuiWXzlglqetNlBTucFPLFcqOWwBUKNeLJTrWXJsUbtuqhqLnHTUM9dm8Pl9nf8zBZzLFAQlsZT0Bs4GjmQssQ8GH4JMIk1LdaFQ5kozGWMRKawgA

This PR(type error as expected): https://stackblitz.com/edit/vitejs-vite-d7vwtr?file=index.ts

samchon:
@ryoppippi static symbol? It means the static property right?

If you think below is okay, it would be published to v6.10.0 with your CLI enhancement.

```typescript
import type { OpenApi, OpenApiV3 } from "@samchon/openapi";

export type IJsonApplication<
  Version extends "3.0" | "3.1" = "3.1",
  Types = unknown[],
> = Version extends "3.0"
  ? IJsonApplication.IV3_0<Types>
  : IJsonApplication.IV3_1<Types>;
export namespace IJsonApplication {
  export interface IV3_0<Types = unknown[]> {
    version: "3.0";
    schemas: OpenApiV3.IJsonSchema[];
    components: OpenApiV3.IComponents;
    __types?: Types;
  }
  export interface IV3_1<Types = unknown[]> {
    version: "3.1";
    components: OpenApi.IComponents;
    schemas: OpenApi.IJsonSchema[];
    __types?: Types;
  }
}
```

samchon:
@SocketSecurity ignore npm/next@14.2.5

@SocketSecurity ignore npm/protobufjs@7.4.0

ryoppippi:
@samchon 
Thank you for your work!

I think this approach works well. so using static property is fine!

However, I'd like to suggest to use `" types"` as a property instead of `__types `.
Using this technique, `" types"` will no longer appear as a candidate for auto-completion, and it is more developer friendly!
Screenshot 2024-08-23 at 10 40 10
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKlC8UDeAoK6oEsAmB+AXFAM7ABOmAdgOYDcaGARFKJEQwVBQK4C2ARhFJ0AvihQBjAPYUSUAG6E4iJMLoo5AOiA

ryoppippi:
Also I didn't know that this enhancement is requested before lol https://github.com/samchon/typia/issues/760

samchon:
@ryoppippi Looking at your library, I think you need an LLM function calling schema generator. 

I made similar one (https://github.com/wrtnio/openai-function-schema) converting from OpenAPI document to LLM function calling schema, but what you need seems like not conversion from the Restful API's OpenAPI document, but just by the TypeScript function directly.

How about suggesting a feature like below? Also, returned schema from `typia.llm.functions<Functions>()` or `typia.llm.application<ClassType>()` would better to contain the function call executor. Of course, it would much better when the executor contains the parameter separator/unifier of both Human and LLM sides. Referencing the https://github.com/wrtnio/openai-function-schema repository document and knowhows what you've experiened in the field, suggest me a graceful interfaces please.

For reference, below namespace and function name is just example. Just naming by your rule.

- `typia.llm.schemas<Types>()`
- `typia.llm.functions<Functions>()`
- `typia.llm.application<ClassType>()`

ryoppippi:
@samchon 
Thank you for your suggestion.
idk it is the right that `typia` includes LLM (more specifically OpenAI) spec, because typia is a library for "general purpose".
	
I agree that it is good if we have that kind of function, but those functions should be in separate package (like `@typia/llm` ?) in my opinion. and if we create the new library, I'm happy to transfer implementations in my library to the new library

samchon:
OpenAPI (not OpenAI) says that they are planning to publish v4 specification that is familiar to the AI (maybe LLM function calling schema). Following the roadmap, I'll move LLM function calling schema from `@wrtnio/openai-function-schema` to `@samchon/openapi` and would refactor it to be compatible with both OpenAPI operation and independent function indicating RPC (Remote Procedure Call).

Also, `typia` had done too much things of non-general. `typia` already supports `protobuf` and `http`, and people are asking for `graphql` support in addition. I have been requested, so that have developed everything that can be done with types. Therefore, I sometimes think that `typia`'s slogan should be changed; not validator but type framework?

Anyway, don't mind of it about the LLM function calling schemas composition. If you and I need, just do it.

As I've developed [`nestia`](https://github.com/samchon/nestia) which can effectively generate Swagger Document with pure TypeScript type, I just thought about LLM function calling schema transformation only by OpenAPI document, but your case seems enough reasonable and I've missed it. If there're some developers who want to perform LLM function call execution not by external Restful API, but by their programs' internal functions, your feature would be helpful.

I think that such internal function's LLM calling can be a killer use-case of `unplugin-typia`.

ryoppippi:
@samchon 
> type framework

I like it! This is a good expression for `typia`. `typia` is not a minimal library, but FW. Sounds exciting.


> I'll move LLM function calling schema from @wrtnio/openai-function-schema to @samchon/openapi 

Got it.

So do you merge this PR, or dealing this issue with other library?

Also, |'m just curious why you use `@samchon/openapi` instead of `@typia/openapi` (just curious). If you think this is a FW, you can use the name of `typia` more(and maybe crating `typia` GH org is a good choice)

samchon:
Accept this PR, but also needs the new feature. Wait for your suggestion.

Also, no special reason for package naming.

ryoppippi:
@samchon Thanks! OK got it!

Thank you as always.

PR #1242, opened 2024-08-23
Add more reserved keywords

samchon:
To avoid more programming reserved keywords:

  - `module`
  - `public`
  - `private`
  - `protected`


samchon (commented):
OK.

PR #1248, opened 2024-08-28
feat(website): migrate from webpack to rspack

ryoppippi:
I modified the `preapare` script in website to use rspack instead of webpack.
I tested in my local environment, and it runs really fast and works perfectly


samchon (approved):
Thanks for configuration.

ryoppippi:
Wait I'm still working on this PR because I haven't modified the options yet.(that is why this is a draft PR!)
could you revert it?
@samchon

ryoppippi:
If it works, that is fine.I'll create a new PR

PR #1259, opened 2024-09-06
`typia.llm.application<T>()` function.

samchon:
Added `typia.llm.application<T>()` function which generates set of LLM function calling schemas from the TypeScript class or interface types.

The type `ILlmApplication` and `ILlmSchema` came from samchon/openapi#44, and requested by @ryoppippi.


samchon (commented):
Test functions all passed.

PR #1269, opened 2024-09-09
CLI detects package manager automatically.

samchon:


ryoppippi:
@samchon 
OK, thanks.
You can delete `cli/index.js`

PR #1272, opened 2024-09-09
feat(cli): read typscript's version from `package.json`

ryoppippi:
This PR makes CLI read typescript version from `typia`'s package.json


samchon (approved):
Good idea.

I'll publish this at next update.

Issue #1234, opened 2024-08-22
[BUG] - .proto -> typescript -> encode format and API doesn't allow for correct field ids and incorrectly assumes auto incrementation.


ShanonJackson:
Typia Version: "@latest 6.9.0"
Expected behavior: "Some way to add correct field ids when using proto serialization"
Actual behavior: "Typia uses invalid field ids when using proto serialization"

I have the following .proto file
```
message TalentGeneFilter {
    string field1 = 1;
    string field2 = 2;
    string field3 = 4; 
}
```
Please note the missing '3'

I convert this to a typescript type.
```
{field1: string, field2: string, field3: string}
```
I encode that typescript type.
```
typia.encode<type>(data);
```

Please note that in the binary output, `field3` should be marked as using `field 4` but that information is "lossy" lost when you convert the .proto -> typescript

This causes a bug for me because a downstream gRPC service expects correct field ids on correct fields (as per spec).

is there any way in typia I can do this?
```
{field1: string & typia.Field<1>, field2: string & typia.Field<2>, field3: string & & typia.Field<4>}
```

Thanks and hope this makes sense.

samchon:
Good suggestion.

To support the `Field` (or something better name) tag, I've to change the current intersection tag system enormously. In current `typia`, it will actually try to validate the `tag` value when a object type case as a optional property. Due to I've afaid of such enourmous change, I had thought about this feature, but could not developed it for a log time.

By the way, I have to do it now because actual protocol buffer feature user has come.

Here is my additonal question for detailed specification:

1. How about union type? It adjusts the `Field` tag too?
2. Also, if user omits the Field tag to some properties, how will you do? Consider it as a compiler error? Or automatically assign the property field index by inference?
3. How about think any better and proper name than `Flag`?

```typescript
{
  id: string;
  etc: (string & tags.Field<3>) | (number & tags.Field<5>) | (boolean & tags.Field<9>;
  additional: Something;
}
```

ShanonJackson:
@samchon 

Thanks for getting back to me so fast.

1: Union type is fine.
2: I think error is best because if we default to "auto incrementing" it can create bugs that are hard to track down, and also rely on users to specify fields in the same order as they are in the spec in typescript.
3: `tags.Field<5>` looks fine to me.

samchon:
How about another word `Radix`? The name field is too general.

samchon:
About this feature, I may try at next week's Saturday (2024-08-31).

[...]

PR #1273, opened 2024-09-10
Refactor #1272: just a little bit refactoring.

samchon:
Will be published with next update.

The next version would be v6.11 with #1234 feature.


samchon (commented):
No problem.

PR #1276, opened 2024-09-13
Optimize `typia.llm.application<App>()` function for bundling.

samchon:
Do not import `typia.llm.application<App>()` embeded function when no parameter argument assigned to reduce bundling size in the frontend application.


samchon (commented):
OK, no problem.

Issue #1255, opened 2024-09-03
Add Provenance Support


elliot-huffman:
### Description
This allows you to publicly establish where a package was built and who published a package, which can increase supply-chain security for your packages.

This is a super easy publish command change.
Only a NPM parameter needs to be added to the existing publish pipeline. (by default)

### Overview
Publish this package using only GitHub actions and enable Provenance to enable the ability cryptographically to attest that the package hasn't been tampered with during build, publish, and transport.

This can be done via `npm audit signatures` automatically.

NPM Docs:
https://docs.npmjs.com/generating-provenance-statements

### Examples

Before: `npm publish`

After: `npm publish --provenance`

Consuming attestation signatures:
https://docs.npmjs.com/generating-provenance-statements#verifying-provenance-attestations

## Implementation Tracker
- [x] Migrate Publish Process to GitHub Actions #1262
- [x] Enable Provenance Attestation in the GH Actions publish workflow

## PR
Requested Change  - #1281

samchon:
https://github.com/samchon/typia/blob/cd6ed78a9815134617405b34d5ee891da56f539a/deploy/internal/DeployRunner.ts#L72

You can accomplish it just by updating the above line.

Will you try PR?

elliot-huffman:
I'm absolutely happy to do a pr, just want to double check, does the above code get executed inside of a GitHub action runner?
If so, which workflow does it run in? 
I need to make sure that the ID token is enabled for attestation in the workflow

ryoppippi:
@elliot-huffman 
I think he publish this package from his local machine.
We need to configure GitHub Actions first.

ryoppippi:
@elliot-huffman 
Also, I'm working on making this repo as a monorepo using pnpm.
https://github.com/samchon/typia/issues/1065

elliot-huffman:
> @elliot-huffman I think he publish this package from his local machine. We need to configure GitHub Actions first.

I can build the GH Actions version of publish.
@samchon, are you ok with me writing a GH Actions pipeline to publish your project when a GH Release is created? (as a PR)
For reference on what I am thinking about making a PR for, take a peek at and tell me what you think:
https://github.com/Software-Hardware-Integration-Lab/ObjectID-Converter/blob/main/.github/workflows/NPM-Publish.yml

elliot-huffman:
Now that provenance is enabled, would I be able to request that you run a deploy/publish from the GH Actions so that we have a new version with the provenance?

[...]

PR #1281, opened 2024-09-18
Enable Provenance

elliot-huffman:
This request is the final step to enabling provenance.

See #1255 for more details.


samchon (approved):
Thanks

PR #1297, opened 2024-09-25
Add `tags.Example` type with `@samchon/openapi` upgrading.

samchon:
Upgraded `@samchon/openapi` to the latest version (`1.0.5`).

`typia` has added `tags.Example` and `tags.Examples` types for following the `@samchon/openapi`'s newly added properties.

https://github.com/samchon/openapi/pull/54


samchon (commented):
OK, no problem.

PR #1304, opened 2024-09-28
Fix samchon/nestia#1043: object literal shorthand property problem.

samchon:
```typescript
import typia from "typia";

const x: number = 3;
typia.is({ x }); // didn't know x property's existance
typia.is({ x: x }); // catch exact type
```

`typia` could not catch exact type iformation from the shorthand property assigned object literal expression.

This PR solves the problem.


samchon (commented):
OK

PR #1312, opened 2024-10-05
feat: disable additionalProperties for llm schema

antoniomdk:
This is a first pass at disabling `additionalProperties` for `typia.llm.schema`. I think in general LLMs are stochastic enough that for most use cases we would probably want to disable generating additional properties (e.g. to lower the output tokens count).

However, I think we should provide a way to enable additional properties back, although I wasn't sure how to do it with the current types. I was looking at the `Metadata` type but couldn't find useful properties that would let us decide for example if the type inherit from Record-like types.

I was thinking maybe we can control the behavior via a flag, but @samchon already proposed to separate schema generate based on LLM provider like: `typia.openai.application<App>(): ILlmApplication<IOpenAiSchema>`, so probably we could offload the decision there.

We know from OpenAI docs that `additionalProperties` is not supported and will raise an error if we passed it to their API, but it's still unclear if other providers fully support OpenAPI 3.1 or just a subset.

For example, OSS solutions like guidance [supports](https://github.com/guidance-ai/guidance/pull/1029) this feature.
Outlines [supports](https://github.com/dottxt-ai/outlines/pull/907) it as well (with some hidden limitations - 2-level depth max). llm-schema-enforcer has [some issues](https://github.com/noamgat/lm-format-enforcer/issues/129) with additionalProperties.

### DIFF
--- a/src/programmers/internal/llm_schema_object.ts
+++ b/src/programmers/internal/llm_schema_object.ts
@@ -98,7 +98,7 @@ export const llm_schema_object = (props: {
 /**
  * @internal
  */
-const join = (extra: ISuperfluous): ILlmSchema | undefined => {
+const join = (extra: ISuperfluous): ILlmSchema | false => {
   // LIST UP METADATA
   const elements: [Metadata, ILlmSchema][] = Object.values(
     extra.patternProperties || {},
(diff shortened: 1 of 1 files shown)


samchon (approved):
Good, will publish to `6.11.2` soon.

Also, dedicated schema for OpenAI would be supported by the next v7 major update.

PR #1313, opened 2024-10-05
`ImportProgrammer` composing import statements by transformation.

samchon:
No more keeps namespaced functions like `typia.assert.guard()`.

`ImportProgrammer` will compose import declarations and statements in the transformation step.

@ryoppippi Wait for one or two weeks, then v7 would be released with zero size bundling.


ryoppippi:
OMG

samchon:
https://github.com/samchon/typia/tree/b890b9c6e2399a34fc3f612fe1382089096a81b8/test/generate/output

@ryoppippi Here is the final generated (transformed) files of v7 update.

samchon (commented):
Now, only `protobuf` feature is left.

ryoppippi:
Thank youuu

PR #1328, opened 2024-10-11
Prepare #1234: refactor metadata schema for `protobuf` sequencing.

### LINKED ISSUE #1234: [BUG] - .proto -> typescript -> encode format and API doesn't allow for correct field ids and incorrectly assumes auto incrementation.

samchon:


samchon (commented):
Separated `MetadataObject` to `MetadataObject` and `MetadataObjectType`.

Also, `MetadataObject.tags` property is newly added to support sequence ordering.

samchon (commented):
Preparation completed

PR #1332, opened 2024-10-17
Develop #1234: `Sequence<N>` tag and its validator on `ProtobufFactory`

samchon:


samchon (commented):
OK, tested passed.

However, have not tested the failure cases yet.

PR #1333, opened 2024-10-18
New metadata `IProtobufProperty` and `IProtobufSchema`

samchon:
Adjusted to `message` and `decode` functions.

The `encode` functions are not yet.


samchon (commented):
Test OK

PR #1336, opened 2024-10-21
`ImportProgrammer.IOptions` for multiple transformers case.

samchon:
When `@nestia/core` calls `AssertProgrammer` like functions, the transformed code occurs duplicated importment statement error. To solve this problem, I've decided to add `ImportTransformer.IOptions.internalPrefix` to avoid such situation.

For reference, the `typia`'s transformer will assign the `typia_transform_` prefix to there, and `@nestia/core` will do `nestia_core_transform_`.


samchon (commented):
Test OK

PR #1339, opened 2024-10-22
Close ryoppippi/unplugin-typia#335: change `$` prefix to `_` for Svelte.

samchon:
As Svelte-Kit prohibits `$` starting instance name, I've changed typia v7 to utilize `_` starting functions internally generated.


samchon (commented):
Test completed

PR #1349, opened 2024-11-06
replace URLSearchParams to IReadableURLSearchParams

miyaji255:
I'm using Typia on Hono, because Hono's query is `Record<string, string[]>`, so I can't use the http module.
This PR allow a thin wrapper as query arguments.
https://github.com/honojs/hono/blob/main/src/request.ts#L160

Thin wrapper example
```ts
const queries = c.req.queries();
const urlSearchParams: IReadableSearchParams = {
   get: queryName => queries[queryName][0] ?? null,
   getAll: queryName => queries[queryName] ?? []
}
```


samchon (approved):
Would be published in today. Thanks for contribution.

samchon:
Merge anyway and fix the bug after that

PR #1350, opened 2024-11-09
Allow LLM model specification

samchon:


samchon (commented):
No problem on `typia` side, but ChatGPT schema may be changed in near future.

PR #1351, opened 2024-11-10
Fix wrong `$ref` address for recursive array type.

samchon:


samchon (commented):
OK, successfully fixed.

Issue #650, opened 2023-06-05
The detail CONTRIBUTING.md


sunrabbit123:
## Feature Request

Your document is excellent.
It will be sufficient for experienced individuals.

However, it is anticipated that there will be significant difficulties for newbies in understanding the structure.
Especially in the area of transformations and coding while grasping the Abstract Syntax Tree (AST), it can be time-consuming.
If there are explanations about the structure or information on where specific functionalities are located, it would reduce the learning curve and make it easier to overcome trial and error.

Thank you for your attention.

samchon:
Welcome your PR

tsengia:
@samchon As I am working with Typia, I am noticing some general quality of life improvements that could be made (ie. comments in source code, more developer/maintainer documentation). I'll make a PR eventually with additional docs.

On a side note, should the `test/generated/output` directory be tracked in git? I ask because I ran the test generation and `git status` is showing me about 807 files modified, which all seem to be brand new output files! Because it's generated during the test procedure, I have a feeling this should be added to `.gitignore`...

AlexRMU:
Also, as an improvement, I would suggest creating a wrapper interface for the ts package to reduce the size and complexity of the code, and create functions with multiple arguments rather than `()=>()=>()=>()=>()=>{}()()()()()`.

PR #1353, opened 2024-11-11
docs: add file structure and convention

sunrabbit123:
add file structure and convention

I wrote it as I understood, so please review it.

## Associated

https://github.com/samchon/typia/issues/650


samchon (approved):
Thanks for documentation.

PR #1359, opened 2024-11-14
`$importInternal` script in the type tag validator.

samchon:


samchon (commented):
OK, most be completed.

PR #1360, opened 2024-11-16
Close #1062: `_isUniqueItems()`, the deepinng comparison.

### LINKED ISSUE #1062: add UniqueItems tag
## A description of the problem you're trying to solve:
Currently, when defining arrays of primitive types, there is no built-in mechanism to enforce that array items are unique.
    
## An overview of the suggested solution:
I propose adding a new tag for unique items in arrays, leveraging the `uniqueItems` property from the OpenAPI specification.

OpenAPI documentation on uniqueItems: [Swagger Data Types - uniqueItems](https://swagger.io/docs/specification/data-models/data-types/#uniqueItems)

I'm unsure if `uniqueItems` should work on object-types (or if it's even possible to implement). Name could be adjusted to indicate that it only works on primitive types (if implementing it for object types is too hard).

## Code examples showing the expected behavior:
```ts
type UniqueItems = typia.tags.TagBase<{
    kind: "uniqueItems";
    target: "array";
    value: undefined;
    validate: `(new Set($input)).size === $input.length`;
    exclusive: true;
    schema: {
        uniqueItems: true;
    };
}>;
  ```

## Examples of how the suggestion would work in various places:
```ts
export interface MyType {
    emails: Array<string & tags.Format<"email">> & tags.MaxItems<5> & tags.UniqueItems;
}
```

samchon:


samchon (commented):
OK, no problem.

PR #1362, opened 2024-11-16
fix(cli): type error

ryoppippi:
I found a type error in the latest `package-manager-detector` library, so I updated the types of CLI


samchon (approved):
Thanks for contribution.

PR #1375, opened 2024-11-24
fix: Prevent type name numbering like ".o1"

rojiwon123:
I prevented values like “.o1” from being appended if they have the same name, except when the typename is “__type”.
This keeps the `ts.type` objects internally distinct enough to use as keys, but I suspect they will be represented as one type when converted to openapi, etc.

However, there are cases where separate ts.Type objects are created for the same ts type, and i didn't prevent this, so this change may need further review.

And I didn't really get to see if the results were what I wanted in the nestia environment. Because I couldn't find a way to test it.

related with [nestia/1016](https://github.com/samchon/nestia/issues/1016)

### DIFF
--- a/src/factories/MetadataCollection.ts
+++ b/src/factories/MetadataCollection.ts
@@ -83,7 +83,7 @@ export class MetadataCollection {
       const str: string = TypeFactory.getFullName(checker)(type);
       return this.options?.replace ? this.options.replace(str) : str;
     })();
-
+    if (name !== "__type") return name;
     const duplicates: Map<ts.Type, string> = MapUtil.take(this.names_)(
       name,
       () => new Map(),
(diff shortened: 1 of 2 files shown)


samchon (changes requested):
No, duplicated name occurs not only by the `__type` case.

Just accept the duplicated name case, because it is inevitable.

1. TypeScript compiler sometimes make `__object` name.
2. Actually duplicated type case
    - File A contains "Something" type
    - File B constains "Something" type too
    - File C imports both of them

In the `nestia`, you are in the second case, I think you hope to avoid it just by hard coding.

By the way, if  `typia` gives up the duplicated name checking, validation and serialization logics are damaged.

Therefore, this PR is not acceptable.

rojiwon123:
@samchon 

I completely agree with you.

But isn't the duplicated type that happens in [this case](https://github.com/samchon/nestia/tree/master/test/features/clone-type-create-duplicate) different from the case you are talking about? Even if it is different, I think this is fundamentally a bug in TSC, but is there anything else we can try in typia? I'd like to think about this a bit more.

samchon:
Just be tolerance about the duplicated name on  the OpenAPI document. No way.

PR #1376, opened 2024-11-24
Add `@samchon/openapi` to `peerDependencies`.

samchon:
As `@samchon/openapi` is preparing the v2 major update, restricted its version by `peerDependencies`.


samchon (commented):
No problem.

PR #1413, opened 2024-12-03
use mkdist for esm instead of rollup

ryoppippi:
This is the patch for build script for mjs.

I experienced and found that

- cjs issue on randexp is solved in Bun build
- we don't need to complicated bundler for mjs. just remove types from mjs files and it is fine.
- keeping structure is important for tree-shaking

So, for this purpose, I picked [`mkdist`](https://github.com/unjs/mkdist), and it works fine in my env.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "7.2.1",
+  "version": "7.3.0",
   "description": "Superfast runtime validators with only one line",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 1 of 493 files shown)


samchon:
Tested the `mkdist --dist='lib' -d` command, but it is not making the source map files.

samchon (commented):
Tested the `mkdist --dist='lib' -d` command, but it is not making the source map files.

ryoppippi:
OK, sourcemap.
I'll investigate it

ryoppippi:
I close this issue and recreate it!

PR #1414, opened 2024-12-03
feat(ci): update node versions

ryoppippi:
In this patch

- use 'lts' version for pkg-pr-new & bun action
- added multiple node version matrix for Node action

for now, `lts/*` means `22.x`, so if you think 
`node-version: [20.x, 22.x, lts/*]`
is redundant, you can fix that part

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: lts/*
       - uses: actions/cache@v4
         with:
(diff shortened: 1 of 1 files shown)


samchon (approved):
Thanks for contribution.

By the way, as `typia` does not have native feature especially requiring NodeJS version up, I think this PR is much suitable for my another project [nestia](https://github.com/samchon/nestia). Would you contribute same PR to the [nestia](https://github.com/samchon/nestia)?

Always thanks for endless contributions.

ryoppippi:
@samchon in that case, do we still need only `lts/` ?

samchon:
Let's just keep this PR. No reason to rolling back

PR #1415, opened 2024-12-05
Support TypeScript v5.7 update.

samchon:
TypeScript v5.7 has occured many break changes on its filesystem, and it had broken the `ts-patch`. And today, @nonara has fixed the problem by upgrading the `ts-patch`. This PR allows to install TypeScript 5.7+ version following the `ts-patch` update.

Also, TypeScript v5.7 has changed another break change on the `Uint8Array` like native class type to have a generic argument `Uint8Array<T>`. This PR also solves the problem of it.


samchon (commented):
Tested on local, and checked working fine.

PR #1421, opened 2024-12-10
Add `IValidation.IError.data` property.

samchon:
Even when the validation is failed, its origin data should be filled. 

This PR fills the input data in the `IValidation.IError.data` property.

```typescript
export type IValidation<T = unknown> =
  | IValidation.ISuccess<T>
  | IValidation.IFailure;
export namespace IValidation {
  export interface ISuccess<T = unknown> {
    success: true;
    data: T;
  }
  export interface IFailure {
    success: false;
    data: unknown;
    errors: IError[];
  }
  export interface IError {
    path: string;
    expected: string;
    value: any;
  }
}
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,13 +1,5 @@
 name: build
 on: 
-  push:
-    paths:
-      - 'src/**'
-      - 'test/**'
-      - 'test-error/**'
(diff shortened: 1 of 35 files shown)


samchon (commented):
No problem.

Issue #1429, opened 2024-12-13
`createValidateParse` in generation mode fails to compile with Next.js


asterikx:
### 📝 Summary
I'm using `typia@7.3.0` in generation mode with `next@15.1.0`. Whenever I use `createValidateParse` the generated code fails to compile with `next build`:

```
Error:   × Expected '=>', got ':'
    ╭─[/Users/asterikx/app/packages/app/app/types/generated/person.ts:42:1]
 39 │         success: true,
 40 │         data: input
 41 │     } as any;
 42 │ }; return (input: string): import("typia").IValidation<import("typia").Primitive<Person>> => __validate(JSON.parse(input)) as any; })();
    ·                          ─
    ╰────

Caused by:
    Syntax Error

Import trace for requested module:
./app/types/generated/person.ts
```

### 💻 Code occuring the bug
Template code:
```ts
import typia from 'typia';

type Person = {
  name: string;
  age: number;
};

export const validateParsePerson = typia.json.createValidateParse<Person>();
```

Generated code:
```ts
import * as __typia_transform__validateReport from "typia/lib/internal/_validateReport.js";
import typia from 'typia';
type Person = {
    name: string;
    age: number;
};
export const validateParsePerson = (() => { const _io0 = (input: any): boolean => "string" === typeof input.name && "number" === typeof input.age; const _vo0 = (input: any, _path: string, _exceptionable: boolean = true): boolean => ["string" === typeof input.name || _report(_exceptionable, {
        path: _path + ".name",
        expected: "string",
        value: input.name
    }), "number" === typeof input.age || _report(_exceptionable, {
        path: _path + ".age",
        expected: "number",
        value: input.age
    })].every((flag: boolean) => flag); const __is = (input: any): input is Person => "object" === typeof input && null !== input && _io0(input); let errors: any; let _report: any; const __validate = (input: any): import("typia").IValidation<Person> => {
    if (false === __is(input)) {
        errors = [];
        _report = (__typia_transform__validateReport._validateReport as any)(errors);
        ((input: any, _path: string, _exceptionable: boolean = true) => ("object" === typeof input && null !== input || _report(true, {
            path: _path + "",
            expected: "Person",
            value: input
        })) && _vo0(input, _path + "", true) || _report(true, {
            path: _path + "",
            expected: "Person",
            value: input
        }))(input, "$input", true);
        const success = 0 === errors.length;
        return success ? {
            success,
            data: input
        } : {
            success,
            errors,
            data: input
[...]

PR #1430, opened 2024-12-13
Test #1429, but no problem

samchon:
Tested, but could not find any problem.

This is just for the github actions execution, not for merging.


samchon:
`SWC` has fixed this problem.

PR #1435, opened 2024-12-15
Add test-nextjs directory related to #1429

asterikx:


samchon (commented):
Thanks for PR, but as this is a bug of `swc`, I think `nextjs-test` directory is not required for the related issue.

Let's wait the `swc`'s bug fixing.

PR #1461, opened 2025-01-16
Docs: fix wrong configured tsconfig.json at setup document

sunrabbit123:
fix typo
strict and strictNullChecks is property of compilerOptions


samchon (approved):
Thanks for emension.

PR #1467, opened 2025-01-27
fix build for esm

ryoppippi:
Hi @samchon 
finally I fixed the rollup script.
I found that we don't have to include cjs conversion in build anymore because bun solved this problem

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -29,8 +29,8 @@ module.exports = {
       }
       return `[name].${ext}`;
     },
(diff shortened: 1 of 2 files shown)


ryoppippi:
why pkg-pr-now doesn't work....

oh, okay, build.yaml is not triggered by rollup.config.js

samchon (approved):
Thanks for contribution.

I'll publish the next minor update after your ESM refactoring(?).

PR #1469, opened 2025-01-27
include tsconfig and rollup in workflow triggers

ryoppippi:
Added 'tsconfig.json' and 'rollup.config.js' to the list of files that trigger the CI workflow. This ensures that changes to these files will initiate the build process.


samchon (approved):
Thanks for contribution. Good point.

PR #1482, opened 2025-02-02
chore(gitignore): add bun.lock

ryoppippi:
https://bun.sh/docs/install/lockfile#text-based-lockfile

This PR includes new bun lockfile in .gitignore


samchon (approved):
Thanks for contribution.

PR #1483, opened 2025-02-02
migrate to eslint v9

ryoppippi:
This PR includes updating eslint v9 flat config
This pull request includes significant changes to the ESLint configuration and several minor code updates to remove unnecessary ESLint disable comments.

### ESLint Configuration Updates:
* `eslint.config.cjs` was removed and replaced with `eslint.config.mjs`, which includes updated plugin imports, parser settings, and rule configurations. [[1]](diffhunk://#diff-ade92dc557e1c37f3e97d3323edfba82ec5ae154ff4325ddd06962631a5c2666L1-L28) [[2]](diffhunk://#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R1-R49)

### Package Configuration Updates:
* Updated `package.json` to include new ESLint dependencies and modify ESLint scripts to use caching for improved performance. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L24-R25) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56-R57) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L66-R68)

### Code Cleanup:
[...]

PR #1484, opened 2025-02-02
enable consistent-type-imports rule

ryoppippi:
After https://github.com/samchon/typia/pull/1483 merged, I'll propose this rule because it make me easier to read a code(and eslint can fix it automatically)


samchon (commented):
Is there special reason distinguishing the `type import`?

I understand that whether add the `type` keyword or not, `tsc` automatically generates proper JS codes. 

So, if add this rule, I worry about that development can be annoying.

ryoppippi:
@samchon 
I'm working on esm build with tsc right now and I found that type import doesn't disappear after transpiling.
When we use type import, we can reduce the bundling size

related: https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/

Also, eslint can fix it automatically, so you don't have to consider it when developing

ryoppippi:
I can add this rule to https://github.com/samchon/typia/pull/1483

PR #1491, opened 2025-02-05
fix(ci): use `npm pkg delete` to remove `private` instead of `jq`

ryoppippi:
just a small change
This pull request includes a change to the `.github/workflows/build.yml` file to simplify the process of removing the `private` field from `package.json`.

* [`.github/workflows/build.yml`](diffhunk://#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L32-R32): Replaced the `jq` command with `npm pkg delete private` to remove the `private` field from `package.json`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,8 +29,7 @@ jobs:
       - run: npm run build
       - name: remove `private=true` from package.json
         run: |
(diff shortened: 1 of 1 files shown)


samchon (approved):
Thanks for contribution.

Issue #1504, opened 2025-02-20
Format<"url"> incorrectly rejects IP addresses as hostname


stephane-alnet:
### 📝 Summary
Format<"url"> incorrectly rejects IP addresses as hostname

  - **Typia Version**: 7.6.4
  - **Expected behavior**: Format<"url"> URL containing IP address is valid
  - **Actual behavior**: URL containing IP address is rejected

Write detailed description in here.

The URL parser should accept IP addresses as hostname, similarly to `new URL()`.

### ⏯ Playground Link
[Link](https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoFgYwgDsD5gCAqlAA2cALwJkaAHTsoAU1Qx5ASQIAeXlGBc8cAGS5C0gGLQQS9TQCuImgD57ACgCUbTjz4EAguxjXUYSFRCScCAC4tHTwXcXtsFjgEKEQEpKSPXjhUcTgueQB3OAEAJQAZMLd0uAV-KC5suABCCS5rYWFEkjh2JXYACzT02tsGskCCeS7iFhmWMmg4J0z4ADc4CDI4AG0umn6YGDBwgHoTgEYAJgB2aQAGe+lz8IAOO7eTmnQ9g6PTk+24XC5wAuq93ndPl0znAAGqBYAAEyU8gQ-XkkzgBWgAGsfodjmd5AAPVDgYTyWRUcEfL57KCoLiIqhwPCoKAAI3w8jpLBBsSwXUyEAp0mEEDwTgABvQ0OE4AASLD8YJOVYuaQwCAAZRg2l0rmkYFQiIAokynABWFykVB+ALCeVK-i+fyBVXqzU6vXRQ3Gs0W62kAC0iqwq1IUrcMyAA)

### 💻 Code exhibiting the bug
```typescript
import typia, { tags } from "typia"
import assert from "node:assert"

const isUrl = typia.createIs<string & tags.Format<"url">>()

assert.ok(isUrl("http://127.0.0.1:8080/"))
assert.ok(isUrl("http://[::1]:8080/"))
```

samchon:
`uri`, url`, and `iri` are different.

PR #1505, opened 2025-02-20
use `new URL` to validate in _isFormatUrl.ts

stephane-alnet:
This is a possible solution for #1504


ryoppippi:
@stephane-alnet 
Hi! Thank you for your contribution.

How about using `canParse` method (https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static)

samchon:
Same answer with the #1504, and `uri`, `url`, and `iri` are different.

Issue #1497, opened 2025-02-11
Comparing data


snatvb:
Hello,
I use `typia` and noticed that it would be great to have able to compare `A` and `B`
How it could be:
```ts
import typia from 'typia'

type User = {
  name: string
  age: number
  address: {
    room?: number
    city: string
    street: string
  }
}

const ivan: User = {
  name: 'ivan',
  age: 19,
  address: {
    street: 'Ivanovskaya',
    city: 'Berlin'
  }
}

let cache: User | null = null

function heavyFunction(user: User) {
  if (t.compare<User>(user, cache)) {
    return 'cached'
  }
  // heavy operation
  cache = user
  return 'non-cached'
}

heavyFunction(ivan) // -> 'non-cached'
heavyFunction(ivan) // -> 'cached'
```

Compare unwraps in:
```ts
if (
 user === cache || (
   user.name === cache.name &&
   user.age === cache.age &&
  (user.address === cache.address || 
      (user.address.room === cache.address.room && user.address.city === cache.address.city && user.address.street === cache.address.street)
  )
)) {
  return 'cached'
}
```

samchon:
https://github.com/samchon/nestia/blob/master/packages/e2e/src/internal/json_equal_to.ts

Not okay with this function? Transformed comparator is really required?

snatvb:
> Not okay with this function? Transformed comparator is really required?

Usually I use deepEquals / shallowEquals. But directly comparing should be faster I think

Maybe I didn't get what you mean, could you rephrase your message? Sorry I am not native english speaker 😅

samchon:
Oops, missed link https://github.com/samchon/nestia/blob/master/packages/e2e/src/internal/json_equal_to.ts

I can design the new function interface what you want, but I am concentrating the AI chatbot feature development, so if you want that feature, you should challenge the implementation by yourself

snatvb:
> Oops, missed link https://github.com/samchon/nestia/blob/master/packages/e2e/src/internal/json_equal_to.ts
> 
> I can design the new function interface what you want, but I am concentrating the AI chatbot feature development, so if you want that feature, you should challenge the implementation by yourself

I can't use the function because it's internal and it can throw exception
I'll investigate how to add it, if you agree that it's good function for your library
But I am not sure that I'll rich success 😅 but I can try

samchon:
```typescript
// src/compare.ts
export namespace compare {
  export function covers<T>(x: T, y: T): boolean;
[...]

PR #1511, opened 2025-02-24
Compare feature

snatvb:
Feature of #1497

# Classes
Added "class" property to detect that object is class

Classes forbidden to compare because there is a lot of deep cases that might leads to bug with prototypes.

# Extends MetadataCollection
Added find method of type for objects. I think it's good idea to have possibility to get `ts.Type` by any `Metadata`.

It requires in my case, because metadata doesn't cover this methods declarations:
```ts
type Foo {
  bar(): void
}
```

I've noticed that it's hard to compare unions:
```ts
type Foo = { foo: number } 
type Bar = { bar?: number }
type Union = Foo | Bar

const a = { foo: 1 }
const b = { foo: 2 }

console.assert(
  a === b || (
    a.foo === b.foo
  ) || (
    a.bar === b.bar
  )
)
```
This will be true, because after fail on `foo` prop, `a.bar === b.bar` because `undefined === undefined`

And this unions a lot of corner cases. I just forbidden it. Left only for primitive types.

# Development & Test Environment

I've faced with problem that tests take too much time. I've added possibility to add flag that will filter tests by folder.
```sh
pnpm dev --folder=compare.equals
```

Also added watch mode for root:
```
pnpm dev
```

### DIFF
--- a/src/compare.ts
+++ b/src/compare.ts
@@ -1,0 +1,19 @@
+/**
+ * Comapre in deep A and B
+ */
+export function equals<T extends object>(a: T, b: T): boolean;
+/**
+ * @internal
+ */
+export function equals<T extends object>(_a: T, _b: T): boolean {
+  halt("equals");
+}
+
+/**
+ * @internal
+ */
+function halt(name: string): never {
+  throw new Error(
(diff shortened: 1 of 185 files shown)


samchon:
Close due to go-lang migration

Issue #900, opened 2023-12-08
Nx and typia setup docs


renesass:
Hey @samchon,

I love typia and desperately want to integrate it with Nx (express app).
I tried all I could find
- https://github.com/samchon/typia/issues/814
- https://github.com/samchon/nestia/issues/319

However, no matter what I do, I always get errors:
- 'rootDir' is expected to contain all source files.
- throw new Error(`Could not find ${fileToRun}. Make sure your build succeeded.`);

Since some people got it running, I would love to see the specific steps on how to set it up.
Here is my current setup:

/apps/api/project.json
```
{
  "name": "api",
  "$schema": "../../node_modules/nx/schemas/project-schema.json",
  "sourceRoot": "apps/api/src",
  "projectType": "application",
  "targets": {
    "build": {
      "executor": "@nx/js:tsc",
      "outputs": ["{options.outputPath}"],
      "defaultConfiguration": "production",
      "options": {
        "target": "node",
        "compiler": "tsc",
        "outputPath": "dist/apps/api",
        "main": "apps/api/src/main.ts",
        "tsConfig": "apps/api/tsconfig.app.json",
        "assets": ["apps/api/src/assets"],
        "isolatedConfig": true,
        "webpackConfig": "apps/api/webpack.config.js"
      },
      "configurations": {
        "development": {},
        "production": {}
      }
    },
    "serve": {
      "executor": "@nx/js:node",
      "defaultConfiguration": "development",
      "options": {
        "buildTarget": "api:build"
      },
      "configurations": {
        "development": {
          "buildTarget": "api:build:development"
        },
        "production": {
          "buildTarget": "api:build:production"
        }
      }
    },
```
(I tried all kind of executors, specifically @nx/js:tsc and @nx/webpack:webpack)

apps/api/webpack.config.js
```
 const path = require('path')
const nodeExternals = require('webpack-node-externals')
const { composePlugins, withNx } = require('@nx/webpack')

module.exports = composePlugins(withNx(), (config) => {
  // Update the webpack config as needed here.
  // e.g. `config.plugins.push(new MyPlugin())`

  // CUSTOMIZE HERE
  config.entry = ['./src/main.ts']
  config.output = {
    path: path.join(__dirname, 'dist'),
    filename: 'main.js',
  }
  config.optimization = {
    minimize: false,
  }

  // JUST KEEP THEM
  config.mode = 'development'
  config.target = 'node'
  config.module = {
    rules: [
      {
        test: /\.ts$/,
[...]

PR #1519, opened 2025-03-01
docs: explicitly mention results of Typia errors being swallowed by NX

MarjanDB:
I've recently been playing around with NX using a codebase which makes use of Typia, where I encountered an issue that sent me down a rabbit hole for a few days.

The issue was that I was getting errors that I did not configure transforms, when in reality I had.
Well, as it turns out, the issue was actually that Typia encountered a nonsensical type and just did not transform the code, resulting in those transform not being applied and getting the error I just mentioned... and NX in all its glory swallowing those errors and leaving me none the wiser.

I did see that the typia setup guide mentions this behavior, but I failed to make the connection that type errors would result in "did not configure" errors, leaving me to dismiss that as a possible cause of my issues and focusing on the wrong thing: Transpiler / Bundler / Setup issues.

I've updated the setup guide for NX to include an explicit mention of this behavior.

I managed to make the connection thanks to this comment on a similar issue: https://github.com/samchon/typia/issues/900#issuecomment-2492677787

This is how the setup guide would look like now:
image

### DIFF
--- a/website/pages/docs/setup.mdx
+++ b/website/pages/docs/setup.mdx
@@ -537,7 +537,7 @@ After installing `typia` like above, and ensuring the `prepare` script is someth
 }
 ```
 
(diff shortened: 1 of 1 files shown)


samchon (approved):
Thanks for contribution.

PR #1521, opened 2025-03-01
feat(rollup): build script ignore external

ryoppippi:
- modify rollup config to generate submodules in esm format
- add `.js` prefix to `@samchon/openapi` because in esm we need to add extension when we import scripts


ryoppippi:
Oh test failed?  I'll test it

ryoppippi:
@samchon TypeScript in v8.0 repo is 5.9, do you mean 5.8?

PR #1539, opened 2025-03-15
fix(rollup): remove external property

ryoppippi:
Now we use auto external plugin for rollup so we don't need to specify this option


samchon (approved):
Thanks for fixing

PR #1546, opened 2025-03-17
chore: build with rslib

ryoppippi:

```bash
# COMPILE
npm run build

npm run test:template

# BUILD TEST PROGRAM
npm run build:test

# DO TEST
npm run test
```


ryoppippi:
it seems like rslib doesn't use esModuleInterop for cjs build.
That is why `esm-test` fails

ryoppippi:
maybe after https://github.com/samchon/typia/pull/1547 ?

ryoppippi:
maybe we don't need it at the moment so I close this
rollup still works fine

PR #1551, opened 2025-03-24
feat: emit .d.mts declaration files

ryoppippi:
Add plugin to copy TypeScript declaration files to .d.mts format to support ES module imports in TypeScript projects that use the "moduleResolution": "node16" or "nodenext" setting.

Related to #1547


samchon (approved):
Approve for v8.0.4 publish

PR #1557, opened 2025-03-30
fix: more accurate target file patterns on `typia generate`

sushichan044:
Files like `foo.cts` or `bar.mtsx` are valid as TypeScript, but are currently ignored when running `typia generate`.

I fixed this behavior.


samchon (approved):
Thanks for contribution. Good point it is.

PR #1560, opened 2025-04-02
fix typo: validatae -> validate

dengkunli:
This PR fixes a minor typo in https://typia.io/docs/validators/validate/

validatae -> validate


samchon (approved):
Thanks for emension

Issue #1555, opened 2025-03-28
Improve Error Message for Non-Object Parameters in `typia.llm.application`


kakasoo:
# Contents

The typia.llm.application method requires the top-level parameter to be an object. This means that even if a string is part of a union with an object, it is still not allowed.

```json
{
  "code": "typia.llm.application",
  "message": "unsupported type detected\n\n- PickNotionServicegetDatabaseInfo.getDatabaseInfo: unknown\n  - LLM application's function (\"getDatabaseInfo\")'s parameter must be an object type.",
  "name": "Error"
}
```

Strictly speaking, this error message is not incorrect, but I wonder if the phrase "must be an object type" might be unclear for those who are not deeply familiar with TypeScript's type system. It may not be immediately obvious that this means "a union with an object is not allowed."

Would you be open to modifying this error message? Or would it be okay if I made the change myself?

# Reference

@luke0408 's code.

- https://github.com/luke0408/connectors/blob/8a186df7260104ddb9082ea80c8c644e85865609/packages/notion/src/notion/NotionService.ts#L738-L799

samchon:
Wait for your contribution.

kakasoo:
@luke0408 how do you think?

luke0408:
Good! I'll fix it

luke0408:
Would a message like `"${prefix}'s parameter must be an object type. Union types (e.g., string | object) are not supported."` be sufficient?

PR #1572, opened 2025-04-16
feat: modify typia.llm.application non-object error message

luke0408:
Change Error Message by #1555 

> eg. before
```json
{
  "code": "typia.llm.application",
  "message": "unsupported type detected\n\n- PickNotionServicegetDatabaseInfo.getDatabaseInfo: unknown\n  - LLM application's function (\"getDatabaseInfo\")'s parameter must be an object type.",
  "name": "Error"
}
```

> eg. after

```json
{
  "code": "typia.llm.application",
  "message": "unsupported type detected\n\n- PickNotionServicegetDatabaseInfo.getDatabaseInfo: unknown\n  - LLM application's function (\"getDatabaseInfo\")'s parameter must be an object type. Union types (e.g., string | object) are not supported.",
  "name": "Error"
}
```

### DIFF
--- a/src/programmers/llm/LlmApplicationProgrammer.ts
+++ b/src/programmers/llm/LlmApplicationProgrammer.ts
@@ -117,37 +117,37 @@ export namespace LlmApplicationProgrammer {
     const prefix: string = `LLM application's function (${JSON.stringify(name)})`;
     if (func.output.size() && func.output.isRequired() === false)
       output.push(
-        `${prefix}'s return type must not be union type with undefined.`,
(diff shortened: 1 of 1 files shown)


samchon (changes requested):
It would better to get name from the `Metadata` instance.

luke0408:
Should I also add a test case for the error being thrown?

samchon (commented):
Well, need to separate error message case

1. None object
2. None object and union
3. Object but union

PR #1500, opened 2025-02-12
feat: implement Standard Schema

sevenc-nanashi:
(Sorry for bad grammar)

This PR adds [standard schema](https://standardschema.dev/) to `createValidate` and `createValidateEquals`.

Example use case, using [up-fetch](https://github.com/L-Blondy/up-fetch) ([which supports standard schema as a validator](https://github.com/L-Blondy/up-fetch?tab=readme-ov-file#%EF%B8%8F-schema-validation)):

<details>
<summary>Details</summary>

```ts
import typia from "typia";
import { up } from "up-fetch";

const upfetch = up(fetch);

const schema = typia.createValidate<{
  userId: number;
  /** @maximum 5 */
  id: number;
  title: string;
  completed: boolean;
}>();

// This passes
console.dir(
  await upfetch("https://jsonplaceholder.typicode.com/todos/1", {
    schema,
  }),
  {
    depth: Infinity,
  },
);
// This fails because `id` is larger than 5
console.dir(
  await upfetch("https://jsonplaceholder.typicode.com/todos/10", {
    schema,
  }),
  {
    depth: Infinity,
  },
);
```

```
{ userId: 1, id: 1, title: 'delectus aut autem', completed: false }
file:///home/sevenc7c/typia/__gi_test/node_modules/.pnpm/up-fetch@1.2.4/node_modules/up-fetch/dist/index.js:1
(... truncated ...)
y [ValidationError]: Validation error
    at b (file:///home/sevenc7c/typia/__gi_test/node_modules/.pnpm/up-fetch@1.2.4/node_modules/up-fetc
h/dist/index.js:1:1247)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///home/sevenc7c/typia/__gi_test/node_modules/.pnpm/up-fetch@1.2.4/node_modules/up-f
etch/dist/index.js:1:2357
    at async file:///home/sevenc7c/typia/__gi_test/dist/src/index.mjs:59:13 {
  issues: [
    {
      message: 'expected number & Maximum<5>, got 10',
      path: [ { key: 'id' } ]
    }
  ],
  data: {
    userId: 1,
    id: 10,
    title: 'illo est ratione doloremque quia maiores aut',
    completed: true
  }
}
```

</details>

(I don't care closing this PR, so please close this PR if you don't like the standard schema)

PR #1589, opened 2025-05-30
docs: document standard schema

sevenc-nanashi:
This PR documents standard schema, implemented in #1500 and #1581.
(As I'm not native English speaker, there are a lot of weird sentences. Sorry.)


sevenc-nanashi:
fyi: I think it's good time to send a PR to add typia in [the list of standard schema compatible libraries](https://standardschema.dev/#:~:text=what%20schema%20libraries%20implement%20the%20spec%3F), as the standard schema is documented.

PR #1670, opened 2025-09-22
Add meta description tag

andreasvirkus:
Google does not seem to recognize the description from OG nor twitter tags, the random description it grabs from the site's contents is really misleading.

Screenshot 2025-09-22 at 10 50 15


andreasvirkus:
Seems Google no longer uses this description tag anyway

PR #1701, opened 2025-12-12
Close #1699: support regex flags in tags.Pattern

### LINKED ISSUE #1699: Feature: Support regex flags in tags.Pattern (for Unicode property escapes)
### Problem

The `tags.Pattern` type does not support regex flags. This prevents using Unicode property escapes (`\p{ID_Start}`, `\p{ID_Continue}`) which require the `u` flag.

### Use Case

Validating ECMAScript identifiers per [UAX #31](https://www.unicode.org/reports/tr31/) requires Unicode property escapes:

```typescript
// Standard JavaScript approach
const pattern = /^[\p{ID_Start}_$][\p{ID_Continue}_$]*$/u;

// With typia - no way to specify the 'u' flag
type Identifier = string & tags.Pattern<"^[\\p{ID_Start}_$][\\p{ID_Continue}_$]*$">;
```

### Current Workaround

Manually expand all Unicode character ranges from [DerivedCoreProperties.txt](https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt):

```typescript
const ID_Start = [
    "\\u0041-\\u005A", // Latin uppercase
    "\\u0061-\\u007A", // Latin lowercase
    "\\u00C0-\\u00D6", // Latin Extended
    // ... 448 ranges total for full BMP coverage
].join("");

type Identifier = string & tags.Pattern<`^[_$${typeof ID_Start}][_$0-9${typeof ID_Start}]*$`>;
```

This workaround:
- Produces unreadable generated code
- Cannot cover astral planes (characters > U+FFFF)
- Requires manual updates when Unicode standard changes

knoan:
## Summary

Adds support for regex flags in `tags.Pattern`, addressing #1699.

### Type tag syntax
```typescript
// Case-insensitive matching
type CaseInsensitive = string & typia.tags.Pattern<"^hello$", "i">;

// Unicode property escapes (requires 'u' flag)
type Identifier = string & typia.tags.Pattern<"^[\\p{ID_Start}_$][\\p{ID_Continue}_$]*$", "u">;
```

### Comment tag syntax
```typescript
/** @pattern /^hello$/i */
caseInsensitive: string;
```

## Changes

- Extended `Pattern<Value, Flags>` type with optional second parameter for flags
- Updated `MetadataCommentTagFactory` to parse regex literal syntax `/pattern/flags`
- Updated `RandomProgrammer` to pass flags to RegExp constructor
- Added `x-pattern-flags` schema extension for JSON schema output
- Added test structures for both type tag and comment tag syntax

Closes #1699

### DIFF
--- a/src/factories/MetadataCommentTagFactory.ts
+++ b/src/factories/MetadataCommentTagFactory.ts
@@ -524,21 +524,56 @@ const PARSER: Record<
       ],
     };
   },
-  pattern: ({ value }) => ({
-    string: [
-      {
-        name: `Pattern<${JSON.stringify(value)}>`,
-        target: "string",
-        kind: "pattern",
-        value: value,
-        validate: `RegExp(${JSON.stringify(value)}).test($input)`,
-        exclusive: ["format"],
-        schema: {
-          pattern: value,
+  pattern: ({ value }) => {
+    // Support regex literal syntax: /pattern/flags
+    const match = value.match(/^\/(.*)\/([dgimsuy]*)$/);
+    if (match !== null) {
+      const [, pattern, flags] = match;
+      return {
+        string: [
+          {
+            name:
+              flags === ""
+                ? `Pattern<${JSON.stringify(pattern)}>`
+                : `Pattern<${JSON.stringify(pattern)}, ${JSON.stringify(flags)}>`,
+            target: "string",
+            kind: "pattern",
+            value: pattern,
+            validate:
+              flags === ""
+                ? `RegExp(${JSON.stringify(pattern)}).test($input)`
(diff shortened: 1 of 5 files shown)


knoan:
Superseded by #1702 which contains only the CLI file arguments feature.

samchon:
Have to be careful due to JSON schema reason. I will consider the way

samchon (commented):
JSON schema does not define the flags, so I have to be careful.

I'm changing the typia to be monorepo, so if you still want this feature, can you re-request it to the `packages/interface` module?

Issue #1676, opened 2025-10-04
More integer types


jessekrubin:
Howdy @samchon 

you have the following integer types:
- int32
- uint32
- uint64
- int64

Does typia support other <64 bit ints and/or could they be added? 

Would like:
- int8
- uint8
- int16
- uint16

Bonus points (imo) would be 128 bit ints too

samchon:
Current integer types (`uint32` ~ `int64`) are prepared for protobuf supporting.

It would be good to challenge this issue by sending a PR.

samchon:
Implemented in PR #1907. This adds smaller integer Type tags with runtime validators, JSDoc @type support, schema coverage, and boundary fixture tests. This will ship with typia 13 after the official TypeScript-Go release.

samchon:
Resolved by PR #1907 and merged in 26e3189092c0f606169e80fae7ccc628a6650705.

This will ship with the official typia 13 release, which is planned together with the official TypeScript-Go release.

PR #1713, opened 2026-01-21
Include "format" of integer types

fstanis:
Somewhat related to #1676, although this doesn't add new types.

Includes the actual integer type as a `format` in the generated schema; this is valid per [Formats Registry](https://spec.openapis.org/registry/format/).


samchon (commented):
The JSON schema specification does not include the format property on integer type.

PR #1834, opened 2026-05-03
🐛 Fixed hot reloading in Bun

iMrDJAi:
In this PR I have implemented the suggested workaround (https://github.com/oven-sh/bun/issues/4689#issuecomment-2888634558) to restore HMR and file watching functionalities. Currently `--watch`/`--hot` flags don't work for Bun setups that use plugins. I have tested and I confirm that this workaround does indeed fix the issue.

I also rectified the `ITransformOptions` import path in `@typia/unplugin` and restored type intellisense in plugin options.

### DIFF
--- a/packages/unplugin/src/bun.ts
+++ b/packages/unplugin/src/bun.ts
@@ -128,7 +128,7 @@ function bunTypiaPlugin(bunOptions?: BunOptions): BunPlugin {
         build.onLoad({ filter }, async (args) => {
           const id = wrap<ID>(args.path);
 
-          const source = wrap<Source>(await Bun.file(id).text());
+          const source = wrap<Source>((await import(`${args.path}?`, { with: { type: 'text' } })).default);
 
           const code = await taggedTransform(id, source, unpluginRaw);
 
--- a/packages/unplugin/src/core/options.ts
+++ b/packages/unplugin/src/core/options.ts
@@ -1,5 +1,5 @@
 import type { FilterPattern } from "@rollup/pluginutils";
-import type { ITransformOptions } from "@typia/core";
+import type { ITransformOptions } from "../../../core/lib/context/ITransformOptions.js";
 import { createDefu } from "defu";
 import type { OverrideProperties, RequiredDeep } from "type-fest";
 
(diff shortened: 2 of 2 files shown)


iMrDJAi:
@samchon Since you're actively working on the project right now, could you please take a look?

packages/unplugin/src/core/options.ts
```diff
@@ -1,5 +1,5 @@
 import type { FilterPattern } from "@rollup/pluginutils";
-import type { ITransformOptions } from "@typia/core";
+import type { ITransformOptions } from "../../../core/lib/context/ITransformOptions.js";
```
samchon: Rolback please

packages/unplugin/src/bun.ts
```diff
@@ -128,7 +128,7 @@ function bunTypiaPlugin(bunOptions?: BunOptions): BunPlugin {
         build.onLoad({ filter }, async (args) => {
           const id = wrap<ID>(args.path);
 
-          const source = wrap<Source>(await Bun.file(id).text());
+          const source = wrap<Source>((await import(`${args.path}?`, { with: { type: 'text' } })).default);
```
samchon, line 131: Can you make a test function about this one?

@ryoppippi How do you think about this edition?

samchon (changes requested):

packages/unplugin/src/core/options.ts
iMrDJAi: Sure. But this needs to be solved in one way or another. That line is completely missing from `options.d.ts` since `@typia/core` package cannot be solved at built time.

image

packages/unplugin/src/bun.ts
iMrDJAi: @samchon How would you like to make it? Do you want a function to compare the output of both `Bun.file()` and `import()`?

samchon: As I am not using, I don't know which way is proper.

But as you are saying it worked properly, it seems no problem. I will publish 12.1.1 tomorrow if the unplugin test passed.

PR #1874, opened 2026-06-01
perf(native): hoist cleanup regexes out of the per-file hot path

samchon:
## Summary

The native transform's `adapter` cleanup pass runs **once per emitted file**, but it recompiled the same regular expressions on every call. Regex *compilation* (not matching) dominated the cost — `CleanupTransformedText` spent **~907µs and 1398 allocations** processing a ~600-byte file body. On a project with thousands of files this is pure, avoidable overhead on the serial emit path.

> Context: with the recent ttsc major update, tsgo parses/checks/emits in parallel, but plugin call-site collection and emit rewriting stay serial on a single shared checker (and the `WriteFile` callback is mutex-serialized). So the right lever for typia is reducing per-file / per-call-site work, not introducing goroutines.

## Changes

- **`adapter/cleanup.go`**
  - Hoist the 11 static cleanup patterns to package-level `var`s so they compile once instead of per file — including the pattern that was being recompiled inside a `ReplaceAllStringFunc` callback on *every matched import line*.
  - Replace `aliasStillReferenced`'s dynamic `\balias\b` regex with an allocation-free word-boundary scan. The alias is always a `\w+` token, so the scan reproduces Go regexp's `\b` semantics exactly.
  - Add a substring fast-reject plus a build-wide compile cache (`sync.Map`) to `runtimeImportAlreadyExists`, which previously compiled **four** alias/module-parameterized patterns per alias.
- **`core/programmers/llm/LlmApplicationProgrammer.go`** — drop the per-call `^[0-9]` regex in name validation for a byte comparison; hoist the identifier pattern to a package var.

Behavior is unchanged.

## Benchmark (new, `cleanup_bench_test.go` — regression guard)

| function (called per emitted file) | before | after | improvement |
|---|---|---|---|
| `CleanupTransformedText` | 907µs · 393KB · **1398 allocs** | 75µs · 4.8KB · **54 allocs** | **~12× faster, ~26× fewer allocs** |
| `CleanupTypeScriptTransformText` | 147µs · 50KB · 352 allocs | 60µs · 7KB · **33 allocs** | **~2.5× faster** |

## Verification

- `go build` / `go vet` clean; custom 2-space Go formatter applied.
- Public-API Go test suite: **16/16 pass, 0 fail**, including the cleanup correctness tests that directly exercise `CleanupTransformedText` / `CleanupTypeScriptTransformText`.

## Notes

- **Version bump**: native source changed, so the `typia` package version needs a bump per `AGENTS.md` §2.4. Left to the maintainer's release tooling (`package:next` date-stamp format).
- **Follow-up**: the deeper per-call-site serial bottleneck is the metadata path (`MetadataFactory` → repeated `checker.TypeToString`, linear object-name search, visited-map reallocation). Tuning it safely needs a ttsc-driver-based pprof harness to measure with a real checker — kept out of this PR to stay a reviewable slice.

### DIFF
--- a/packages/typia/native/adapter/adapter.go
+++ b/packages/typia/native/adapter/adapter.go
@@ -3,7 +3,6 @@ package adapter
 import (
   "fmt"
   "os"
-  "regexp"
   "runtime/debug"
   "strings"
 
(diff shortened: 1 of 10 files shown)


samchon:
## End-to-end confirmation (real wall-clock, not just the micro-benchmark)

Ran the full `transform` stage over the heaviest workload — **`tests/test-typia-automated` (2096 `.ts` files)** — with two binaries built from identical trees except the two touched source files. Interleaved A/B, 6 pairs, warm, stdout discarded:

| | before (`master`) | after (this PR) |
|---|---|---|
| mean | **7.69s** | **5.49s** |
| range | 6.83 – 9.00s | 4.93 – 6.17s |
| stdev | 0.67 | 0.40 |

**→ 2.20s faster · 28.6% reduction · 1.40× speedup · ~1.05 ms/file saved.**

The after-time still includes all the tsgo parse/check/emit work this PR does **not** touch, so the cleanup-pass regex recompilation was burning ~2.2s of pure overhead on this project. The before-min (6.83s) and after-max (6.17s) don't overlap — the signal is well outside noise, and the 1.05 ms/file end-to-end saving matches the micro-benchmark (907µs→75µs per `CleanupTransformedText` call).

Method: built `before`/`after` `ttsc-typia` binaries from this branch and from `master`'s versions of the two files, ran `transform --tsconfig tsconfig.json` over the workspace 6× each interleaved.

samchon:
## Follow-up commits (same branch)

Continued the perf pass with a CPU profile + 2096-file byte-identical diff as the safety net. Three more commits:

1. **`perf(native): guard cleanup regex passes and scan arrows manually`** — substring guards skip each cleanup `ReplaceAll` when its required literal is absent; `parenthesizeSingleParameterArrows` regex replaced by a manual byte scan (its per-call-site cost dropped ~640ms → ~10ms on the workload).
2. **`fix(native): make protobuf nested message order deterministic`** — `write_hierarchy` ranged the `Children` map directly, so emitted `.proto` nested-message order varied run-to-run. Now emitted in explicit insertion order. (Found while building a deterministic diff to verify the perf work — non-deterministic transform output is its own bug.)
3. **`perf(native): memoize type full-name reconstruction per collection`** — the Set/Map iterators recomputed `checker.TypeToString` for every explored type (twice each) just to test a name prefix. Memoized per analysis: `metadata_type_full_name` allocations 3.03M → 1.62M objects, total allocation bytes **2003MB → 1772MB (−11.5%)**.

### Cumulative effect on the 2096-file `test-typia-automated` transform

| metric | baseline (`master`) | this branch |
|---|---|---|
| total CPU samples | ~10.2s | ~6.8s (**−33%**) |
| total allocation | ~2.0GB | ~1.77GB |
| determinism | non-deterministic `.proto` order | deterministic |

Every step verified **byte-identical** output across all 2096 files (except the protobuf commit, which intentionally fixes the ordering) and the full Go test suite (16/16) stays green. Still iterating — the intersection-path collection clone is the next allocation target.

> Reminder for the maintainer: native source changed across these commits, so the `typia` package version still needs its bump at release time.

Issue #1841, opened 2026-05-07
Boolean literal discriminant collapses to single value in tagged unions


wildduck2:
## What happens

When a discriminated union uses a boolean literal as the tag (ok: true | false), the generated SDK output emits ok: true on both branches. The false literal is lost.

## Versions

- typia: 11.0.3
- @nestia/sdk: 10.0.2
- nestia: 10.0.2
- typescript: 5.9.3

## Repro

Source controller return type:

```
| { ok: false; error: { code: string } }
| { ok: true; code: string; data: { id: string } }
```

Generated SDK output:

```
| { ok: true; error: { code: string } }   // wrong, should be ok: false
| { ok: true; code: string; data: { id: string } }
```

Both branches end up with ok: true. Client-side narrowing via if (res.ok) is broken because both branches are now structurally compatible on the discriminator.

## Minimal sample

NestJS controller:

```ts
@TypedRoute.Get('me')
async getMe(): Promise<
  | { ok: false; error: { code: 'X' } }
  | { ok: true; code: 'X'; data: number }
> {
  return { ok: true, code: 'X', data: 1 }
}
```

Run nestia sdk. Inspect generated me/index.ts. Both Output union members will have ok: true.

## Workaround

String tag survives. Replace boolean discriminator with string:

```
| { kind: 'error'; ok: false; ... }
| { kind: 'success'; ok: true; ... }
```

kind round-trips fine. Only ok: true | false collapses.

Or use property existence on the client ('data' in res) instead of res.ok.

## Where I think the bug lives

I have not traced it all the way, but the symptom looks like boolean literal information being dropped during metadata flattening. String literals in the same position keep both members. So the issue seems specific to how boolean atomics are stored or re-emitted.

Happy to send a PR if someone can point at the right file.

samchon:
PR #1906에서 현재 TypeScript-Go 기반 typia 13 transform 경로를 검증하는 회귀 테스트를 추가했습니다. 해당 테스트는 \
eflect.schema\, \	ypia.json.schema\, \	ypia.json.application\ 산출물을 실제 Node 런타임에서 실행하여 \ok: false\와 \ok: true\가 모두 보존되는지 확인합니다. 이 처리는 TypeScript-Go 정식 출시 이후 typia 13 정식 출시 때 함께 반영될 예정입니다.

PR #1876, opened 2026-06-01
test(typia): cover boolean-literal discriminant in tagged unions

hexbinoct:
## Summary

While looking into #1841 (a boolean-literal discriminant reportedly collapsing to a single value), I checked the current `reflect.schema` output and the metadata is correct — but our test suite doesn't actually pin that behavior. `test_reflect_schema_constant` only exercises a bare `<true>` literal; there is no coverage for `<false>` nor for a boolean-keyed discriminated union.

This adds that missing regression coverage so a future change can't silently collapse a union's `false` branch onto `true` (which would break discriminated-union narrowing) without a test going red.

## What it checks

- a bare `false` literal reflects to a `boolean` constant with value `false`
- the discriminated union from #1841 keeps **both** `false` and `true` across its branch objects:

```ts
type U =
  | { ok: false; error: { code: string } }
  | { ok: true; code: string; data: { id: string } };
```

## Test

- New case `test_reflect_schema_boolean_literal_union` follows the existing `reflect.schema` suite shape (one case per file, `TestValidator`).
- `pnpm run build` passes locally; the reflected output for the new case was verified to contain both branch values (`[false, true]`).

Refs #1841

### DIFF
--- a/tests/test-typia-schema/src/features/reflect.schema/test_reflect_schema_boolean_literal_union.ts
+++ b/tests/test-typia-schema/src/features/reflect.schema/test_reflect_schema_boolean_literal_union.ts
@@ -1,0 +1,61 @@
+import { TestValidator } from "@nestia/e2e";
+import typia from "typia";
+
+/**
+ * Verifies a boolean-literal discriminant survives in a tagged union's
+ * metadata.
+ *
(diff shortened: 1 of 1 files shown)


samchon (approved):
AI stupidly made a duplicated PR. I merge this manually. Thanks for contribution.

PR #1880, opened 2026-06-03
fix(native): alias CommonJS imports by per-file emit format

samchon:
## Problem

Under `module: nodenext` (the typia test/config default), `typia.createRandom<ArrayAny>()`, `typia.llm.controller<Calculator>()`, and any transform that references an imported value emitted a **bare identifier** (`ArrayAny`, `Calculator`) instead of the CommonJS require form (`template_1.ArrayAny`). At runtime this throws `ReferenceError: Calculator is not defined`.

This was the root cause of both the `test-langchain` `class_controller_validation` failure and the 256 `test-typia-automated` ArrayAny failures.

## Root cause

`identifierSubstitutionsForEmit` (native/adapter/adapter.go) gated import aliasing on the **project-wide** module option name:

```go
program.ParsedConfig...CompilerOptions.GetEmitModuleKind().String() != "CommonJS"
```

Under `module: nodenext` that string is `"NodeNext"`, so a CommonJS package (no `type`, or `"type": "commonjs"`) skipped aliasing entirely even though every `.ts` file emits as CommonJS per the implied node format.

## Fix

Drive aliasing off the **per-file** emit format via the tsgo API `Program.GetEmitModuleFormatOfFile`, which resolves each file's format from extension + `package.json` `type` + module kind. CommonJS-emitted files (including under `nodenext`) get aliased; ESM-emitted files (`.mts`, `"type": "module"`) do not.

## Tests

White-box test in `native/adapter/imports_test.go` building real `module: nodenext` projects, pinning both directions:
- commonjs package → `Foo` aliased to `foo_1.Foo`
- `"type": "module"` → no substitutions

Red-green verified (test fails against the old module-name check, passes with the per-file API).

End-to-end verified locally:
- `test-langchain --include class_controller_validation`: Success
- `test-typia-automated --include ArrayAny`: all 8 features Success

### DIFF
--- a/packages/typia/native/adapter/adapter.go
+++ b/packages/typia/native/adapter/adapter.go
@@ -110,15 +110,21 @@ func emitCallWithOptions(program *driver.Program, site CallSite, plugin PluginOp
 }
 
 func identifierSubstitutionsForEmit(program *driver.Program, file any) map[string]string {
-  if program == nil ||
-    program.ParsedConfig == nil ||
-    program.ParsedConfig.ParsedConfig == nil ||
-    program.ParsedConfig.ParsedConfig.CompilerOptions == nil ||
-    program.ParsedConfig.ParsedConfig.CompilerOptions.GetEmitModuleKind().String() != "CommonJS" {
+  if program == nil || program.TSProgram == nil {
     return nil
   }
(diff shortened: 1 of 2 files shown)


samchon:

PR #1882, opened 2026-06-04
fix(native): label TypeGuardError method as typia.<fn>, not the bare name

samchon:
## Problem

`typia.createAssertEquals<T>()` (and assert/assertGuardEquals) produced a `TypeGuardError` whose `method` was the **bare** `"createAssertEquals"` instead of `"typia.createAssertEquals"`. typia's own test harness asserts `exp.method === "typia.createAssertEquals"`, so every surplus-property case failed (204 cases in `test-typia-automated`).

## Root cause

`AssertProgrammer.Write` (and the Is/Validate/Json-parse programmers) derive the label from `props.Modulo.Text()`. But `CallExpressionTransformer.TransformKnown` reduced the callee `typia.assertEquals` (a `PropertyAccessExpression`) down to just its name identifier, dropping the `typia.` qualifier. The original TypeScript transformer uses `props.modulo.getText()` on the **full** callee expression.

## Fix

- Keep the full callee expression as the modulo (mirror the TS transformer).
- Render the label through one shared helper, `ModuloMethodText`, that reads the source-span text via `GetTextOfNode` — `Node.Text()` panics on a `PropertyAccessExpression`, which is why the reduction existed. All ~13 method-label sites across the programmers now go through this single helper instead of each calling `modulo.Text()`.

## Verification

`test-typia-automated`: **266 -> 52 errors** locally (the 204 surplus-property failures clear; no regressions). The remaining 52 are unrelated native-transform issues (`failed to understand the X type`, etc.).

### DIFF
--- a/packages/typia/native/core/programmers/AssertProgrammer.go
+++ b/packages/typia/native/core/programmers/AssertProgrammer.go
@@ -182,10 +182,7 @@ func (assertProgrammerNamespace) Decompose(props AssertProgrammer_DecomposeProps
 }
 
 func (assertProgrammerNamespace) Write(props AssertProgrammer_IProps) *shimast.Node {
-  method := ""
-  if props.Modulo != nil {
-    method = props.Modulo.Text()
-  }
+  method := nativehelpers.ModuloMethodText(props.Modulo)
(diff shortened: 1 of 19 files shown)


samchon:

PR #1883, opened 2026-06-04
fix(native): native-transform type-analysis bugs (toward 0 errors)

samchon:
Fixes wrong-logic / stub bugs in the Go-migrated metadata layer that the original TypeScript version did not have, diffed against legacy v12.1.1. Reduces test-typia-automated failures toward 0.

## metadata_is_internal honored

`metadata_is_internal` was a `return false` stub, so `@internal`-tagged properties were never filtered from object metadata. Legacy `emplace_metadata_object.ts` skips them via `symbol.getJsDocTags()`. Implemented by scanning the symbol's JSDoc tags for `"internal"` (reusing `metadata_node_js_doc_tags`). **ObjectInternal cases now pass.**

More metadata fixes (constant JSDoc tags, alias type arguments, recursive alias, nil guards) will be stacked here as each is verified against legacy.


samchon:

PR #1885, opened 2026-06-05
feat(native): route AST node creation through EmitContext.Factory (slice 1)

samchon:
## Intent

First slice of unifying typia native (Go) AST node creation onto `EmitContext.Factory` (`ec.Factory`), per `INSTRUCTION.md`. Nodes built with `ec.Factory` carry original-node tracking and emit metadata (parent links, comments, source maps, const-enum constants, namespace aliases) that the tsgo emit pipeline relies on, instead of depending on post-hoc fixups (`SetParentInChildrenUnset`). `ImportProgrammer` was already the only place using this pattern; this extends it.

## Scope (this PR)

- **`context.EmitFactory` / `EmitFactoryOf`** — single resolver returning `&ec.Factory.NodeFactory` when an emit context is present, else the standalone fallback factory. The embedded `ast.NodeFactory` inside `printer.NodeFactory` carries the emit hooks, so created nodes gain metadata while **printing identically** (token text/structure unchanged).
- **Shared static factories** (`Expression` / `Type` / `Statement` / `Identifier`) now accept an optional emit context and route creation through the resolver. Variadic-optional, so every existing caller stays green; `nil` selects the legacy / analysis-only / test fallback (always retained, never assumed non-nil).
- **`FunctionProgrammer`** threads the real `props.Context.Emit` through all 46 programmer entry points, so every emitted `const` validator declaration is built with `ec.Factory`.

This is a **behavior-invariant** refactor — printed JS/TS output is unchanged; metadata does not affect token text. Validated by build + Go tests green and (locally) the full suite.

## Deferred (follow-up slices)

- Trailing-variadic `ExpressionFactory.SelfCall` (16 callers) and `IdentifierFactory.Access` (104 callers) — need a leading required `ec` param + caller cascade.
- The ~80 per-programmer standalone factories (`core/programmers/**`, `transform/**`).
- §5 measurement: whether `SetParentInChildrenUnset` workarounds can be removed once unification is complete.

## Test plan

- [x] `GOPROXY=off go build ./...` (native) — green
- [x] `pnpm test:go` (native Go tests) — green
- [ ] `pnpm build` + `pnpm test` (full suite) — running locally / CI


samchon:
## Multi-agent Review Cycle — 3 rounds, 18 agent reviews

Ran an adversarial Review Cycle over this PR plus a broad typia audit for monkey-patching / hardcoding / test-gaming. Lead validated every proposal against the code before applying.

### Defects found & fixed in this PR
- **Union-const emit gap** (`FunctionProgrammer`): the original slice threaded `p.emit_` only into the variable-const loop; the two `tuple.name/tuple.value` union loops were missed. Fixed across two commits — round 2 (6/6 verifiers) caught that the first fix only patched `DeclareUnions()` and missed the dominant `Declare()` union loop (a `replace_all` indentation mismatch). All three `StatementFactory.Constant` sites now thread emit. (`ad9a6ae9`, `75e0fce7`)
- **`_httpHeaderReadString` dangling import** (`HttpHeadersProgrammer.decode_array`): the reader helper — registered as an import side effect — was requested *before* the `string` override discarded it, so `string[]` headers emitted an import to a helper that doesn't exist (only Bigint/Boolean/Number do). Reordered to mirror the scalar `decode_value` path. (`7cdadce7`)
- **Fragile enum arithmetic** (`printer.go`): `KindQuestionToken + 1` → exported `shimast.KindColonToken`. (`7cdadce7`)

Round 3: `pr_is_clean = true`, zero in-scope defects remaining.

### Behavior-invariance — precise claim
Printed token text is unchanged (both factories set `Pos == -1`; the printer's text/trivia branches key off `…IsSynthesized`, which is Pos-based). The emit `onCreate` hook only sets `NodeFlagsSynthesized`, which flips `IsParseTreeNode`/`ParseNode` classification **in the correct direction** for these synthetic consts. So "byte-identical output" is **test-verified**, not guaranteed-by-construction for nodes that ever acquire comment/source-map ranges.

### Pre-existing issues surfaced (out of this PR's scope — for tracking, not bundled here)
- `cmd/ttsc-typia/build.go`: regex-scrapes `tsconfig.json` with a hardcoded `"typia/lib/transform"` literal while the structured `--plugins-json` payload is parsed-then-discarded.
- `adapter/visit.go`: workspace-only path-substring module detection (`packages/typia/src/`) — violates the resolved-package-root Work Rule.
- `internal/printer/printer.go::NormalizeSyntheticTokens`: whole-tree post-hoc fixup filling nil `?`/`:` tokens that ~21 programmer call sites leave nil — recommend constructing conditionals well-formed and retiring the walker in a dedicated slice (mission §5 counsels not force-removing workarounds before full unification).
- `MetadataHelper.go`: blanket `recover()` swallowing all `GetTypeArguments` panics.
- Test oracles: `tests/test-utils-automated/.../_test_validate.ts` comments out the spoiler error-path `throw` (`@todo`); `_test_standardSchema_validate.ts` is count-only. (Maintainer `@todo`s — OpenApiValidator/StandardSchema path-format work.)

### Deferred follow-up (mission continuation)
~78 standalone factories remain (incl. `ValueFactory`/`LiteralFactory`); `SelfCall`/`Access` need threading — recommend **required-ec** signatures there rather than variadic-optional, since the variadic-optional convention is exactly what let the union gap slip silently.

### Note on local test env
A local `pnpm test` shows `test-typia-automated` failures (`_httpHeaderReadString`, then Dynamic* assert) that **reproduce at the base commit with emit-routing disabled** and **pass in CI** — a stale native-binary cache / `node_modules` from Jun 1. Not a branch regression; a local `pnpm install` + cache clear refreshes it.

samchon:
### Deferred follow-ups filed as issues

The review surfaced a few **pre-existing, ambiguous** items (working glue / a maintainer `@todo`) that are out of scope for this emit-factory PR and risky to change without TS-level validation. Filed as self-contained issues so they can be picked up independently:

- #1887 — native: `build.go` plugin-options via structured `--plugins-json` (drop regex + hardcoded specifier)
- #1888 — native: resolve typia module by package root, not workspace path substrings (`visit.go`)
- #1889 — native: narrow the blanket `recover()` in `metadata_get_type_arguments`
- #1890 — tests: re-enable disabled validate oracles (utils `OpenApiValidator` spoiler-paths + StandardSchema paths)

All **actual** bugs / monkey-patching found in review were fixed in this PR (union-emit gap, `_httpHeaderReadString` dangling import, `NormalizeSyntheticTokens` post-emit fixup, `KindColonToken` enum arithmetic).

PR #1903, opened 2026-06-09
fix(typia): cover nullable primitive property guards

### LINKED ISSUE #1806: `is<T>` rejects `null` in `number | null` property
## Summary

`typia.is<T>()` generates validation code that only checks for `number` when a property is typed as `number | null`, rejecting `null` values.

- **Typia Version**: 12.0.1 (`@typia/unplugin` 12.0.1)
- **TypeScript Version**: 5.9.3
- **Expected behavior**: `typia.is<Simple>({ value: null })` returns `true`
- **Actual behavior**: Returns `false`

## tsconfig (relevant options)

```json
{
  "compilerOptions": {
    "strict": true,
    "exactOptionalPropertyTypes": true
  }
}
```

## Code to reproduce

```typescript
import typia from "typia";

interface Simple {
  value: number | null;
}

// Expected: true — but returns false
console.log(typia.is<Simple>({ value: null }));

// Returns true
console.log(typia.is<Simple>({ value: 42 }));
```

## Generated code (observed)

```js
const _io0 = (input) => "number" === typeof input.value;
```

### Expected generated code

```js
const _io0 = (input) =>
  null === input.value || "number" === typeof input.value;
```

## Notes

- `JSON.parse` / `JSON.stringify` preserve `null` faithfully, so this causes silent data loss on round-trip validation.
- Clearing the `@typia/unplugin` cache (`node_modules/.cache/unplugin_typia`) does not help — the freshly generated code has the same issue.
- May be related to #519 / #1617 (`exactOptionalPropertyTypes` handling), but this is a **required** property, not an optional one.

## Environment

- Node.js 22
- Windows 11
- `@typia/unplugin/vite` with `{ cache: true }`

samchon:
## Summary
- Adds a native transform regression test for required `primitive | null` object properties in `typia.is()`.
- Verifies number, string, boolean, reversed `null | number`, nested object, and array-entry property guards keep `null === ...` in the same OR expression as the primitive `typeof` branch.
- Wires `pnpm test:go` to run both the public Go suite and native module Go suite through the CI-compatible `packages/typia/test` workspace.
- Confirms the current TypeScript-Go / typia v13 native transform already preserves the nullable branch that was reported missing in v12.

Fixes #1806

## Validation
- `pnpm format`
- `go -C packages/typia/native test ./cmd/ttsc-typia -run TestNullablePrimitivePropertyIsTransform -count=1`
- `pnpm run test:go:native`
- `pnpm test:go`
- `git diff --check`

## Local limitations
- Full `pnpm test` was not run locally; this change is scoped to native Go transform regression coverage, and the relevant native/test Go suites passed.


samchon:
Research Review Round 1 completed. Accepted improvements: strengthened the nullable primitive assertions to check complete null-or-primitive guard expressions, made the fixture single-purpose for typia.is(), and wired pnpm test:go to include the native module test suite so this regression is covered by the go CI job. Local validation after the follow-up commit: pnpm format, pnpm test:go, git diff --check.

samchon:
Rebased onto current `master` after #1904 landed and force-pushed the branch.

Local validation after rebase:

- `pnpm format`
- `go -C packages/typia/test test ./...`
- `go -C packages/typia/native test ./...`
- `git diff --check origin/master...HEAD`

`pnpm format` only produced Windows line-ending worktree noise after formatting; restored those files so the branch remains scoped to #1806.

samchon:
RRR Round 4 follow-up applied.

Accepted improvement:

- Kept the emitted guard sentinel assertions, and added runtime table coverage by executing the transformed `isNullablePrimitive` validator through Node.
- Runtime cases now cover valid primitive values, all-null nullable fields, wrong primitive types, missing required fields, and present `undefined` values at top-level, nested, and array positions.
- The runtime harness rewrites only the unused transformed `require(typia)` to a local empty stub. If the transform failed to inline `typia.is`, the table still fails because the stub has no validator implementation.

Validation after the change:

- `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestNullablePrimitivePropertyIsTransform -count=1 -v`
- `go -C packages/typia/test test ./...`
- `go -C packages/typia/test test ../native/...`
- `pnpm format`
- `git diff --check`

samchon:
RRR Round 5 completed with no surviving proposals.

Review coverage:

- Emitted-output sentinel plus runtime table coverage for nullable primitive properties.
- Runtime harness checked that the local `typia` stub does not mask an untransformed `typia.is` path.
- `test:go` CI wiring still reaches both public and native Go test packages.
- Source alignment confirmed: nullable metadata contributes the `null` branch before primitive atomics, then object checks retain the branch.
- No overlap with #1904 exact optional files, no generated JS artifacts, no `.wiki` or schema converter changes.

CI remains the only merge gate still settling on the latest head.

PR #1943, opened 2026-06-12
feat(typia): add depth-limited type check (`shallow`)

### LINKED ISSUE #1708: Feature: Shallow .is checks or any "cheaper" discriminator check
There are two common use cases for `.is`: validation and discrimination.

Typia is often overkill for discrimination due to the amount of code it can generate for something like a JSON Schema type.  For discrimination, we often don't need to validate the entire object.  If there is a discriminator key, for instance, often all standard TS does is check the value of that key.

Ideally, typia could define a minimal type check to have a discriminator check based on the possible incoming types, but as I think the complexity of that is likely too high, I'm suggesting a `MaxDepth` param on `.is` or similar:

```ts
typia.is<T, N>

// or a new method to avoid over confidence

typia.isLikely<T, N=2> // 2 seems like a good default
typia.matchesToDepth<T, N=2>
```

Osheun:
Fixes #1708

## Summary

Adds `typia.shallow<T, N>()` and `typia.createShallow<T, N>()`, a depth-bounded
variant of `is` aimed at **discrimination** rather than full validation. When you
only need to know which branch of a union an input belongs to, paying for a deep
structural check of every leaf is wasteful — `shallow` checks only the
near-surface, discriminating part of the type.

```ts
typia.shallow<T, N = 2>(input);     // N bounds how deep the checker descends
typia.createShallow<T, N = 2>();    // reusable factory form
```

`N` is a non-negative integer literal (default `2`). Each level of nesting into an
object, array element, or tuple member spends one unit of budget; once it is
exhausted, a value is accepted as long as it is a non-null object. A `true` result
is therefore **not** a full type guarantee — for soundness, use `is`.

## How it works

The budget is a `*int` on `IsProgrammer_IConfig` / `CheckerProgrammer_IConfig`.
`nil` (every pre-existing caller: `is`, `assert`, `validate`, `equals`, `clone`,
`json`, `protobuf`, `llm`, …) means unlimited, so **no existing path changes**.
Only the `shallow` family sets a non-nil value, and the whole depth path is gated
on `Depth != nil`.

In shallow mode, `CheckerProgrammer.Decode_object` inlines the property checks (via
the existing `Joiner.Object`) with a child config whose budget is spent by one,
instead of emitting the shared `_io{index}` function. The same `Config` already
threads through array/tuple descent, so those boundaries decrement too. The finite
budget also bounds recursion — a self-referential type stops inlining once it
reaches 0, where `Decode` collapses it to a structural object guard.

Because shallow inlines objects, the factored `_io` functions would be dead code.
`FeatureProgrammer.WriteDecomposed` skips any composed function the functor never
referenced (`HasLocal(key) == false`). This only affects shallow — every other
programmer calls `UseLocal` for the functions it emits, so their output is
unchanged (verified against `is`/`equals` emit).

## Emitted output (depth budget in action)

For `interface Deep { kind: "deep"; a: { b: { c: { value: string } } } }`:

- `shallow<Deep, 0>` → `typeof input === "object" && input !== null`
- `shallow<Deep, 1>` → checks `kind` and that `a` is an object, stops there
- `shallow<Deep, 3>` → descends to `a.b.c` (object), stops before `value`
- `is<Deep>` → unchanged, full descent into `a.b.c.value`

Recursive types terminate: `shallow<Tree, 2>` on
`interface Tree { value: string; children: Tree[] }` inlines two levels then
accepts structurally.

## Tests

- `shallow_depth_limit_transform_test.go` — e2e: depth-0 structural collapse,
  default-depth discriminant check, runtime execution on Node incl. union
  discrimination.
- `shallow_nested_depth_transform_test.go` — pins the budget both ways (depth 1
  stops before `input.inner.value`, depth 2 descends into it).
- `shallow_depth_runtime_test.go` — runs the guards on Node incl. a recursive
  `Tree` and a linked `Node | null`, proving termination and correct verdicts.
- `shallow_depth_argument_error_transform_test.go` — pins the `N` validation
  branches (non-literal / negative rejected, not silently defaulted).
- `shallow_factory_depth_runtime_test.go` — runs the `createShallow` factory guard
  end to end (depth bound + recursion termination).
- `feature_transformer_entry_coverage_test.go` — extended for the new
  `ShallowTransformer` / `CreateShallowTransformer` generic-argument guards.

## Docs

Added a `shallow()` guide under `website/src/content/docs/validators/shallow.mdx`,
listed in the validators `_meta.ts` right after `is`. It uses the same
source/compiled `<LocalSource>` tab pattern as the other validator pages, backed
by new `examples/src/validators/{shallow,shallow-depth,createShallow}.ts` fixtures.

## Validation

`go build`, `go vet`, `go test ./...` (public) and
`go test -tags typia_native_internal ../native/...` (full, incl. e2e) all pass.
The `typia` package version is bumped because the native Go source changed.

### DIFF
--- a/benchmark/package.json
+++ b/benchmark/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@typia/benchmark",
-  "version": "13.0.0-dev.20260605.1",
+  "version": "13.0.0-dev.20260612.1",
   "description": "Benchmark program of typia performance",
   "main": "bin/index.js",
   "scripts": {
--- a/config/package.json
+++ b/config/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@typia/config",
-  "version": "13.0.0-dev.20260605.1",
+  "version": "13.0.0-dev.20260612.1",
   "description": "Shared build configuration for typia packages",
   "devDependencies": {
     "@rollup/plugin-commonjs": "catalog:rollup",
--- a/examples/package.json
+++ b/examples/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@typia/examples",
-  "version": "13.0.0-dev.20260605.1",
+  "version": "13.0.0-dev.20260612.1",
   "description": "Example codes for typia website",
   "scripts": {
     "build": "rimraf bin && ttsc && prettier --write --ignore-path /dev/null \"bin/**/*.js\"",
--- a/examples/src/validators/createShallow.ts
+++ b/examples/src/validators/createShallow.ts
@@ -1,0 +1,13 @@
+import typia from "typia";
+
+interface Circle {
+  type: "circle";
+  radius: number;
+}
+interface Square {
+  type: "square";
+  side: number;
+}
+type Shape = Circle | Square;
+
+export const isShape = typia.createShallow<Shape, 1>();
--- a/examples/src/validators/shallow-depth.ts
+++ b/examples/src/validators/shallow-depth.ts
@@ -1,0 +1,16 @@
+import typia from "typia";
+
+interface Deep {
+  kind: "deep";
+  a: { b: { c: { value: string } } };
+}
+
+const input: unknown = {
+  kind: "deep",
+  a: { b: { c: { value: "hello" } } },
+};
+
+typia.shallow<Deep, 0>(input); // typeof input === "object" && input !== null
+typia.shallow<Deep, 1>(input); // also checks `kind` and that `a` is an object
+typia.shallow<Deep, 3>(input); // descends to `a.b.c`, stops before `value`
+typia.is<Deep>(input); // full descent, including `a.b.c.value`
--- a/examples/src/validators/shallow.ts
+++ b/examples/src/validators/shallow.ts
@@ -1,0 +1,16 @@
+import typia from "typia";
+
+interface Circle {
+  type: "circle";
+  radius: number;
+}
+
+const input: unknown = {
+  type: "circle",
+  radius: 5,
+};
+
+// only the discriminator and the top-level shape are checked
+if (typia.shallow<Circle, 1>(input)) {
+  console.log(input.radius); // 5
+}
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@typia/station",
-  "version": "13.0.0-dev.20260605.6",
+  "version": "13.0.0-dev.20260612.1",
   "description": "Typia Station",
   "scripts": {
     "build": "pnpm run build:packages",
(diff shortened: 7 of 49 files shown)


samchon:
Thanks for the work on this! I went ahead and renamed the API before merge.

### `isLikely` → `shallow`

`typia.isLikely<T, N>` / `createIsLikely<T, N>` are now `typia.shallow<T, N>` / `typia.createShallow<T, N>`.

The reason: **"Likely" reads as probabilistic / heuristic**, but the check is fully deterministic — it's exactly `is` down to depth `N`, then a structural object accept past that. "shallow" names the real mechanism (shallow vs deep descent) and the predicate still reads as a type guard.

I applied the rename across the whole surface:
- TS overloads + JSDoc in `module.ts`
- Go `ShallowTransformer` / `CreateShallowTransformer` (files, types, vars), the `"shallow"` / `"createShallow"` functor keys, `shallowTransformer_depth`, and the shallow-mode comments in `CheckerProgrammer` (the neutral `Depth` budget field is untouched)
- the three `shallow_*_test.go` fixtures and identifiers

### Coverage

I also closed two gaps while in there:
- **`shallow_depth_argument_error_transform_test.go`** — pins the `N` validation branches (a non-literal `number` or a negative `-1` is rejected, not silently defaulted). These were untested.
- **`shallow_factory_depth_runtime_test.go`** — `createShallow` was only covered by the no-config panic path; its emitted factory guard was never run. This exercises it end to end (depth bound + recursion termination), bringing the shallow transformers and depth helpers to full reachable-path coverage.

All Go transform/runtime tests pass (`shallow` suites + `feature_transformer_entry_coverage`, internal-tag native sweep green).

### Two follow-ups for you

1. **PR description** still says `isLikely` and lists the old `is_likely_*_test.go` filenames — please refresh it.
2. **No docs page.** There's no guide under `website/src/content/docs/**` for the new API. Per our convention a public-surface addition needs a guide page before release — could you add one (a short section alongside `validators/is`)?

Osheun:
Thanks for the review and for the rename.
I agree that `shallow` is a much better fit for what the API actually does.

I've addressed both follow-ups in the latest commits.

### PR description

Updated to use the new `shallow` naming throughout, including the emitted-output examples and the test references.

### Documentation

Added a guide at `website/src/content/docs/validators/shallow.mdx` and registered it alongside the other validator docs. The page covers:

- the depth budget semantics
- the structural fallback beyond the configured depth
- recursion handling
- `createShallow()` usage

I also added example fixtures so the compiled-output tab is generated the same way as the existing validator documentation.

### One question

Since the compiled examples are generated from `examples/bin/**`, I assumed it's fine for the new page to populate on the next examples build. If you'd prefer different examples or a more compact depth walkthrough, I'm happy to adjust it.

samchon (approved):
Thanks for contribution, I'll merge after CI pass.

Issue #1973, opened 2026-06-24
Go native transform: SIGSEGV panic in GenericTransformer.Scalar on non-specified generic argument


nynj:
## Summary

The Go native transform (`ttsc-typia`) panics with a SIGSEGV when processing a TypeORM/SvelteKit project with ~85 entities that use Typia validation. The panic originates in `GenericTransformer.Scalar()` with "non-specified generic argument", followed by a nil pointer dereference in the `AddDiagnostic` recovery handler.

## Stack trace

```
panic: non-specified generic argument. [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x60 pc=0x1028810e0]

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/ast.(*SourceFile).FileName(...)
	typescript-go@v0.0.0-20260429010842-56ab4af42157/internal/ast/ast.go:2516
main.runTransform.func1(0x3738c1212870)
	cmd/ttsc-typia/transform.go:78 +0x40
github.com/samchon/typia/packages/typia/native/transform.fileTransformer_addDiagnostic(...)
	transform/FileTransformer.go:131
github.com/samchon/typia/packages/typia/native/transform.fileTransformer_try_transform_node.func1()
	transform/FileTransformer.go:114 +0xc4
github.com/samchon/typia/packages/typia/native/transform/internal.genericTransformerNamespace.Scalar({}, ...)
	transform/internal/GenericTransformer.go:159 +0x2d8
github.com/samchon/typia/packages/typia/native/transform/internal.genericTransformerNamespace.decompose({}, ...)
	transform/internal/GenericTransformer.go:111 +0x2d4
github.com/samchon/typia/packages/typia/native/transform/internal.GenericTransformerNamespace_Decompose(...)
	transform/internal/GenericTransformer.go:44 +0x334
github.com/samchon/typia/packages/typia/native/transform/internal.typiaFunctionTransformer_convert(...)
	transform/internal/TypiaFunctionTransformer.go:220 +0x64
```

The initial panic ("non-specified generic argument") is caught by `fileTransformer_try_transform_node`'s recovery handler, which calls `AddDiagnostic`. But the `Diagnostic` passed to the callback has a nil `File()`, so `diag.File().FileName()` in the recovery handler causes a secondary SIGSEGV.

## Reproduction context

- SvelteKit 5 project with ~85 TypeORM entities
[...]

PR #1976, opened 2026-06-26
test(native): guard CLI diagnostics against panic regressions

samchon:
## Intent
Add a dedicated CLI e2e regression for #1973 so the JS workspace matrix verifies the native diagnostic fix at the consumer `pnpm exec ttsc` boundary.

## Scope
- Added `tests/test-typia-cli` as a private test workspace.
- The test creates an OS-temp consumer project with a minimal `typia` package stub and a transform descriptor pointing to `packages/typia/native/cmd/ttsc-typia`.
- It asserts a generic `typia.is<T>` failure preserves `non-specified generic argument`, `error TS(typia.is):`, and `src/generic.ts`.
- It rejects panic/runtime markers, module-resolution noise, and generated JS under `packages/typia/src` before and after the CLI run.

## Deferred
- No `packages/typia/package.json` version bump: this PR is test-only and does not change native package source.
- Full local `pnpm test` is not green because `tests/test-typia-automated` currently reproduces an unrelated TypeScript-Go printer panic: `unexpected Expression: KindBlock`. This branch does not touch that suite or native logic; any CI reproduction will be handled as the next blocker/follow-up.

## Test Plan
- [x] `pnpm exec prettier --write tests/test-typia-cli/src/index.ts`
- [x] `node scripts/with-ttsc-cache.cjs pnpm --filter=./tests/test-typia-cli start`
- [x] `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestTransformProjectDiagnosticsPreserveDetails|TestBuildDiagnosticsPreserveDetails -count=1`
- [x] `pnpm test:go`
- [x] `go -C packages/typia/test test -tags typia_native_internal ./...`
- [ ] `pnpm test` locally fails in unrelated `tests/test-typia-automated` TypeScript-Go printer `KindBlock` panic

Refs #1973

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -688,6 +688,12 @@ importers:
         specifier: catalog:utils
         version: 11.0.0
 
+  tests/test-typia-cli:
+    devDependencies:
+      '@types/node':
+        specifier: catalog:utils
+        version: 25.6.0
+
   tests/test-typia-exact-optional:
     dependencies:
       '@nestia/e2e':
--- a/tests/test-typia-cli/package.json
+++ b/tests/test-typia-cli/package.json
@@ -1,0 +1,26 @@
+{
+  "private": true,
+  "name": "@typia/test-typia-cli",
+  "description": "CLI end-to-end tests for typia.",
+  "scripts": {
+    "start": "ttsx src/index.ts"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/samchon/typia"
+  },
+  "keywords": [
+    "typia",
+    "test",
+    "cli"
+  ],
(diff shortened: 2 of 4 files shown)


samchon:
CI follow-up: the first test workflow failure was caused by the new CLI e2e overriding the shared \TTSC_CACHE_DIR\, which forced a second native plugin build during parallel package tests and hit the local 180s child timeout. Pushed \11b163f4f\ to reuse the parent ttsc cache when present, keep only temp fixtures in the scratch directory, and extend the child timeout for cold builds. Local scoped validation: \
ode scripts/with-ttsc-cache.cjs pnpm --filter=./tests/test-typia-cli start\.

PR #1982, opened 2026-06-29
fix(typia): tune random default lengths

samchon:
## Summary

- change unconstrained random strings to generate length `5..10`
- change non-recursive random arrays to generate length `1..6`, while recursive array props keep a zero-length default path
- pass `recursive: true` through `IRandomGenerator.array` props when the array element graph is recursive
- add a regression test for defaults, explicit zero constraints, and custom recursive array generator props
- keep the typia ttsc descriptor compatible with the current descriptor loader by exporting `createTtscPlugin` as a named export as well as default

## Tests

- `pnpm --filter ./tests/test-typia-schema start -- --include random_defaults`
- `pnpm test:go`
- `pnpm format`
- `pnpm --dir website build`
- `git diff --cached --check`


samchon:
Follow-up after CI: the first run exposed an oversized recursive ArrayRecursive SDK payload. I kept the 
ecursive marker inside the array props as designed, and capped recursive default arrays to

samchon:
Follow-up for recursive random arrays with positive `MinItems`:

- Reject recursive random array types carrying positive `tags.MinItems<N>` at transform time.
- Covered direct self recursion, array alias recursion, root recursive array aliases, mutual graph recursion, and tuple/union nested recursion in `tests/test-error`.
- Added a Go unit test for positive/zero/nonrecursive cases plus schema/value numeric payloads, including defined numeric types.
- Kept non-recursive `MinItems<1>` accepted in `test_random_defaults`.

Local validation:

- `pnpm format`
- `pnpm --filter ./tests/test-error start`
- `pnpm --filter ./tests/test-typia-schema start -- --include random_defaults`
- `pnpm test:go`

CI: all current PR checks pass except `sdk`, which fails in `experiments/nestia/tests/test-sdk` at `clone-and-keyword` with `/arrayRecursive` HTTP 413. Per the explicit instruction to ignore the `nestia.yml`/Nestia failure, I did not change that area in this PR.

PR #1984, opened 2026-07-01
fix(native): reflect interface & type-alias JSDoc as schema description

samchon:
## Intent

A named `interface` used as a nested type exposed only its **property** comments in `json.application` / `json.schema` / `llm.application` schemas — the interface's own `/** ... */` description was silently dropped.

Reproduction (from the report):

```ts
/** Plus operation properties. */
interface IPlusProps {
  /** X coordinate. */
  x: number;
}
```

`IPlusProps` rendered with `x`'s `"X coordinate."` but **no** `"Plus operation properties."` on the component itself.

## Cause

In the Go transform, `MetadataCollection.Emplace` (package `core/schemas/metadata`) fills an object's own `Description` / `JsDocTags` via `metadataCollection_description` / `metadataCollection_jsDocTags` — but those are **stubs returning `nil` / empty**, because the schemas layer cannot import the AST JSDoc helpers (`metadata_node_description`, `metadata_node_js_doc_tags`) that live in `core/factories/internal/metadata` (importing them would be circular). Property-level descriptions were unaffected — they are extracted separately in `emplace_metadata_object_create_property`, which is why property comments worked and masked the gap.

## Scope

- Fill `obj.Description` / `obj.JsDocTags` in `Emplace_metadata_object` (factory layer, which has AST access and the working helpers), on the first-emplacement path.
- Source them from the object's **own** declaring symbol (`props.Type.Symbol()`), **not** the alias name symbol — so utility types such as `Record<K, V>` do not leak their standard-library JSDoc (`"Construct a type with a set of properties K of type T"`).

Deferred: type-alias-level descriptions (`type Id = string /** ... */`) remain unchanged — separate concern, not part of this report.

## Test plan

- New e2e test `tests/test-typia-schema/src/features/json.schema/test_json_schema_object_description.ts` mirrors the reported example and asserts the nested-interface descriptions (`IProps` / `IPlusProps` / `IResult`) alongside the surviving property comment. Verified it **fails** without the fix and **passes** with it.
- `go build ./core/...` and `go test ./...` (native toolchain) green.
- Full `test-typia-schema`, `test-typia-automated`, `test-langchain`, `test-mcp`, `test-vercel` suites green.

> Note: this touches `packages/typia/native`. Per the repo's version convention a `typia` version bump may be expected, but I've left `package.json` untouched — that's the maintainer's call.

### DIFF
--- a/packages/typia/native/core/factories/internal/metadata/MetadataHelper.go
+++ b/packages/typia/native/core/factories/internal/metadata/MetadataHelper.go
@@ -245,6 +245,53 @@ func metadata_node_description(symbol *nativeast.Symbol) *string {
   return nil
 }
 
+// metadata_node_type_description / metadata_node_type_js_doc_tags read the JSDoc
+// of a TYPE-level symbol (an interface, class, or `type` alias), skipping symbols
+// declared in a TypeScript default library file. Standard-library type aliases
+// carry their own JSDoc (e.g. `NonNullable`'s "Exclude null and undefined from
+// T", `Record`'s "Construct a type ...") and, since the type-name symbol of an
+// instantiation such as `NonNullable<...>` resolves to that library declaration,
+// an unfiltered read would leak the library comment into the user's schema. A
+// property's own comment is unaffected — it is read from the property symbol,
+// which is declared in user code.
+func metadata_node_type_description(symbol *nativeast.Symbol) *string {
+  if metadata_symbol_from_default_lib(symbol) {
+    return nil
+  }
+  return metadata_node_description(symbol)
+}
+
+func metadata_node_type_js_doc_tags(symbol *nativeast.Symbol) []schemametadata.IJsDocTagInfo {
+  if metadata_symbol_from_default_lib(symbol) {
+    return []schemametadata.IJsDocTagInfo{}
+  }
+  return metadata_node_js_doc_tags(symbol)
+}
+
+// metadata_symbol_from_default_lib reports whether the symbol's first locatable
+// declaration lives in a TypeScript default library file. Default library files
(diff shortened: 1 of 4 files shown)


samchon:
## Follow-up: extended to type aliases + standard-library leak guard

Self-review surfaced two things the initial commit missed:

1. **Type aliases were still dropped.** A `type Status = "a" | "b"` (union) or `type Foo = { ... }` (object) with its own `/** ... */` did not surface a description. Now filled from the type-name symbol, so aliases behave like interfaces.

2. **Standard-library JSDoc could leak.** Because the type-name symbol of an instantiation such as `Record<K, V>` or `NonNullable<T>` resolves to the *library* alias declaration, naively reading its JSDoc leaked comments like `"Construct a type with a set of properties K of type T"` and `"Exclude null and undefined from T"` into user schemas. Verified reproduction:

   ```ts
   interface IRoot { nn: NonNullable<"a" | "b" | null> }
   // component "NonNullable..." → description: "Exclude null and undefined from T"  ❌
   ```

   Both the object and alias paths now read through `metadata_node_type_description` / `metadata_node_type_js_doc_tags`, which skip declarations from a default library file (`lib.*.d.ts`). Property-level comments are unaffected (they come from the property symbol, always in user code).

The e2e test now also asserts the LLM (`llm.application`) `$defs` description, a user type-alias description, and that `Record` / `NonNullable` components carry **no** description.

Re-verified: `go test ./...` and the `test-typia-schema`, `test-typia-automated`, `test-langchain`, `test-mcp`, `test-vercel` suites all pass. The e2e test fails on the pre-fix tree (non-vacuous).

PR #1986, opened 2026-07-01
feat: LlmJson.validateArguments for MCP-inlined arg coercion (closes #1974)

### LINKED ISSUE #1974: Expose a standalone arg coerce+validate helper for inlined MCP registration
## Context

Building an MCP server in another project (ttsc's `@ttsc/graph`, see samchon/ttsc#275) with `typia.llm.controller`, I inlined the `tools/list` + `tools/call` registration over the controller's `application.functions` rather than using `@typia/mcp`'s `registerMcpControllers`.

The reason was dependency/version surface: a consumer that pins a specific `typia` dev build (e.g. `typia@13.0.0-dev.YYYYMMDD`) then has to keep `@typia/mcp` (and transitively `@typia/interface` / `@typia/utils`) pinned in lockstep. Inlining kept the dependency surface to just `typia` + `@modelcontextprotocol/sdk`. The inlined core is small — list the functions as tools, and on a call run `func.validate(args)` then dispatch.

## The gap

`@typia/mcp` runs `LlmJson.coerce(args, func.parameters)` **before** `func.validate(...)`. My inlined handler calls `func.validate(args)` directly, so it **loses that coercion**: an LLM that sends a coercible argument (e.g. `"12"` for a `number`, or a stringified boolean) is rejected by the inline version but accepted by `@typia/mcp`. So inlining silently degrades LLM tool-calling robustness, which isn't obvious to someone who just wants to avoid the extra dependency.

## Suggestion

Either (or both):

1. **Document** the minimal standalone registration in `@typia/mcp`'s README — including the `LlmJson.coerce` → `func.validate` → execute step — so consumers who inline do it correctly.
2. **Export a small transport-agnostic helper** that bundles coercion + validation for one function, e.g. `prepareLlmArguments(func, rawArgs): IValidation<unknown>` (coerce then validate). Then an inliner gets correct LLM-arg coercion without wiring in the MCP SDK or pinning `@typia/mcp`, and `registerMcpControllers` can use the same helper.

Happy to send a PR if a direction sounds good.

samchon:
Closes #1974.

## Problem

Inlining an MCP-style `tools/list` + `tools/call` registration over `typia.llm.controller`'s `application.functions` (to keep the dependency surface to `typia` + the MCP SDK, instead of pinning `@typia/mcp` / `@typia/interface` / `@typia/utils` in lockstep) is easy to get subtly wrong: `@typia/mcp` runs `LlmJson.coerce(args, func.parameters)` **before** `func.validate(...)`, but an inliner that calls `func.validate(args)` directly loses that coercion. An LLM that sends a coercible argument (`"12"` for a `number`, a stringified boolean) is then rejected by the inline version but accepted by `@typia/mcp` — a silent degradation of tool-calling robustness.

## Solution

Both directions the issue suggested:

1. **`LlmJson.validateArguments(func, args)`** (`@typia/utils`) — bundles coerce → validate and returns `IValidation`. One source of truth for the step.
2. **`typia` re-exports `LlmJson`** — so an inliner reaches both the coercion step (`validateArguments`) and the LLM-feedback formatter (`stringify`) through a single `typia` import, no extra pin.
3. **`McpControllerRegistrar`** now dispatches through `LlmJson.validateArguments`, so an inlined handler and `registerMcpControllers` behave identically.
4. **`@typia/mcp` README** documents the correct inlined registration.

## Test plan

- `test-utils` — `test_llm_validate_arguments`: coercible payload passes `validateArguments` but the bare `func.validate` rejects it; coerced `data` carries corrected types; equivalence with explicit `coerce`→`validate`; non-coercible and omitted inputs fail.
- `test-typia-schema` — `test_llm_application_validate_arguments_reexport`: `LlmJson` reachable via `typia` (named export `===` `typia.LlmJson`); coercion + `stringify` work through the single `typia` pin.
- `test-mcp` — `test_mcp_class_controller_coercion`: a `tools/call` with stringified operands succeeds and executes (coercion preserved through the registrar refactor).
- Ran `test-utils` (llm), `test-typia-schema` (llm), and `test-mcp` in full — all green.

No `typia` version bump (no native change; versioning is handled separately by the maintainer).


samchon:
Final design note: the initial `export { LlmJson } from "typia"` re-export was dropped — it dragged `@typia/utils` source into the `test-typia-exact-optional` workspace's `exactOptionalPropertyTypes` check, surfacing pre-existing violations unrelated to this change. The typia-only inliner path is instead `func.validate(func.coerce(args))` (both are already bound on every `ILlmFunction`), and `LlmJson.validateArguments` remains in `@typia/utils` as the registrar's single-source helper. typia's public surface is unchanged from master.

Issue #1989, opened 2026-07-03
Roadmap: `@typia/mcp` as a type-driven MCP server framework — the whole protocol surface from pure TypeScript types


samchon:
## Roadmap: `@typia/mcp` as a type-driven MCP server framework

#1988 closes the gaps that made `@ttsc/graph` inline its own tool registration. This issue is the step after: stop being a *tool registrar* and become the framework that anyone building an MCP server reaches for. The two are deliberately separate — #1988 is tactical and mergeable now; this is the destination it points at.

### Vision

**Pure TypeScript types as the single source of truth for the entire MCP server surface.** Everything the protocol lets a server *declare* — tool input/output schemas, prompt arguments, resource template variables, completion candidates, elicitation schemas, instructions, descriptions — is derivable from TypeScript types and JSDoc. Everything the wire *delivers* — tool arguments, prompt arguments (string-typed on the wire), URI template variables, elicitation responses, sampling outputs — needs coerce + validate + self-correction feedback. The first half is typia's compile-time reflection; the second half is typia's runtime validators. No other MCP framework has either half; typia has both and currently sells neither.

The official SDK stays underneath. Transports (stdio, Streamable HTTP), sessions, auth, protocol negotiation are `@modelcontextprotocol/sdk`'s job and it does them well. `@typia/mcp` is the declarative layer on top — the relationship NestJS has to Express.

Positioning against the field:

- **Official SDK (`McpServer`)**: imperative registration, every schema hand-written in Zod, schemas duplicate the TypeScript types they describe. Fine plumbing, no reflection.
- **FastMCP-style wrappers**: better ergonomics, same hand-written schema duplication.
- **`@typia/mcp`**: schemas *are* the types. Change a parameter type and the tool schema, validator, coercion, and completion candidates all move together at compile time.

And the spec is moving toward us: revision **2025-11-25** (SEP-1303) now explicitly says input validation errors should be returned as tool execution errors *"to enable model self-correction"* — the exact feedback loop `@typia/mcp` has shipped from day one. What was a typia opinion is now spec-recommended behavior.

### Protocol surface → typia capability map

Verified against `@modelcontextprotocol/sdk@1.29.0` and spec revision 2025-11-25. Each pillar states what MCP defines, what typia already has, and what must be built.

#### P0. Tools — #1988 (filed)

[...]

PR #1990, opened 2026-07-03
feat(mcp): create an MCP server from a single typia controller

### LINKED ISSUE #1988: `@typia/mcp` adoption gaps: why `@ttsc/graph` inlined its own MCP registration, and the fixes
### Context

`@ttsc/graph` (ttsc's checker-resolved architecture-graph MCP server) is exactly the consumer `@typia/mcp` was built for: it reflects a TypeScript interface with `typia.llm.controller`, serves the generated functions as MCP tools, and feeds typia's validation errors back to the model for self-correction. It ships with `typia` as a dependency — and still chose **not** to use `@typia/mcp`. Its `src/server/createServer.ts` re-implements the registration inline and says why:

> Registration is inlined rather than pulled from `@typia/mcp` to keep the dependency surface to `typia` plus the MCP SDK and avoid version-pinning the wider typia ecosystem; the shape it relies on is `typia.llm.controller`'s public output.

Worse, typia itself has institutionalized the workaround. `LlmJson.validateArguments` (#1986) was added for this exact case, and its JSDoc currently reads:

> Use this when inlining an MCP-style `tools/call` handler over an `ILlmFunction` **instead of `registerMcpControllers` from `@typia/mcp`** — it keeps the dependency surface to `typia` while preserving the coercion step.

When the ecosystem's own utils document bypassing the integration package as the recommended pattern, the integration package has a positioning problem. This issue is the gap analysis (verified against `@ttsc/graph@0.16.8` and `@typia/mcp@13.0.0-dev.20260701.1`) and the fix list.

---

#### G1. `zod-to-json-schema` is a hard dependency that only `preserve: true` uses

`packages/mcp/package.json` declares `zod-to-json-schema: >=3.24.0` under `dependencies`, but the only call site is `convertZodToJsonSchema` in `McpControllerRegistrar.registerWithPreserve` — the non-default `preserve: true` path that coexists with `McpServer.registerTool()`'s Zod tools. Every standalone-mode user (the default, and the only mode `@ttsc/graph` needs) pays for a Zod-conversion pipeline they never execute.

The dependency-surface complaint otherwise doesn't hold up: `typia` already hard-depends on both `@typia/interface` and `@typia/utils`, so those add **nothing** to a typia consumer's tree. After demoting `zod-to-json-schema`, default-mode `@typia/mcp` adds zero runtime modules beyond what `typia` already installs.

#### G2. No server-construction API, so `instructions` can't be wired

`@ttsc/graph` constructs its server as:

```typescript
new McpServer({ name: "ttsc-graph", version }, {
  capabilities: { tools: {} },
  // The MCP `instructions` are the interface's JSDoc, which
  // `typia.llm.controller` reflects onto `application.description`.
  instructions: controller.application.description,
});
```

The `application.description` reflection (ab79fc5c43) exists precisely to feed MCP `instructions` — `@ttsc/graph` hangs a ~90-line usage contract off its interface JSDoc and ships it through the handshake. But `registerMcpControllers` only takes an *existing* server and never touches instructions, so the package offers no path from the reflected description to the handshake. Users must know the wiring trick themselves, at which point the remaining value of the package is the ~30 lines they can see in `createServer.ts`.

#### G3. Omitted `params.arguments` always fails validation (bug)

MCP's `CallToolRequest` declares `arguments` optional, and clients legitimately omit it for zero-parameter tools. `McpControllerRegistrar.handleToolCall` passes `request.params.arguments` straight into `LlmJson.validateArguments`; `coerceLlmArguments` passes `undefined` through untouched, and `func.validate(undefined)` fails. Consequences:

- a zero-parameter tool called without `arguments` **cannot succeed**;
- for parameterized tools, the model gets one generic "expected object" error instead of field-level errors it can self-correct from.

`@ttsc/graph` normalizes with `request.params.arguments ?? {}` and comments exactly this rationale. One-line fix plus tests.

#### G4. Hard version pinning to the typia release train

`@typia/interface` and `@typia/utils` are `dependencies` (`workspace:^` → `^13`). If `@typia/mcp` lags one major behind a consumer's `typia`, installs fork the tree into duplicate copies. Since `typia` itself guarantees both packages are present, they can become `peerDependencies` with a wide floor (e.g. `>=13`), so the consumer's typia-provided copies are used and `@typia/mcp` no longer forces lockstep cadence. npm ≥7 / pnpm auto-install peers, so setup friction is nil.

#### G5. No carrot: MCP structured output is sitting unused

`typia.llm.application` already reflects return types onto `ILlmFunction.output` (the parity converter test asserts it). MCP has first-class support for exactly this since spec 2025-06-18: `Tool.outputSchema` and `CallToolResult.structuredContent`. Neither the inline `@ttsc/graph` handler nor `@typia/mcp` uses it — both stringify results into a text block. `@typia/mcp` can advertise `outputSchema` whenever `func.output` is present and return `structuredContent` alongside the spec-recommended text fallback, for free, by construction. This flips the adoption question from "what does the package cost me" to "what does inlining cost me."

---

### Proposal

**1. New `createMcpServer` — the one-call path `@ttsc/graph` actually wanted:**

```typescript
export function createMcpServer(props: {
  name: string;
  version: string;
  controllers: Array<ILlmController | IHttpLlmController>;
  /**
   * MCP handshake instructions.
   *
   * @default the controllers' non-empty `application.description`s
   *          (reflected from class/interface JSDoc), joined by blank lines
   */
  instructions?: string;
  /** Pass-through `ServerOptions`; `capabilities.tools` is always set. */
  options?: ServerOptions;
}): McpServer;
```

Standalone registration only (no private-API dependency); the caller connects the transport. `registerMcpControllers` stays as-is for existing-server integration.

**2. Fix G3:** validate `args ?? {}` in `handleToolCall`.

**3. Structured output (G5):** when `func.output` exists, advertise `outputSchema` in `tools/list` and return `structuredContent` + text fallback in `tools/call`. Needs a peer-floor note or feature guard for pre-`outputSchema` SDK versions.

**4. Dependency surface (G1, G4):**

- `zod-to-json-schema` → optional peer, loaded lazily inside preserve mode, with a clear install-hint error if missing when a Zod-schema'd tool is actually encountered.
- `@typia/interface`, `@typia/utils` → peers with a wide floor (decision point; recommended).

**5. Flip the ecosystem guidance:** revise the `LlmJson.validateArguments` JSDoc so it stops recommending inlining over `@typia/mcp`, and update `website/src/content/docs/utilization/mcp.mdx` + the package README with `createMcpServer` and the slimmed dependency story.

Minor knob worth considering while there: result serialization is `JSON.stringify(result, null, 2)`; `@ttsc/graph` deliberately uses compact `JSON.stringify(result)` to save tokens. Either switch the default to compact or accept a `serialize` option.

### Acceptance — the `@ttsc/graph` litmus test

The package is fixed when `@ttsc/graph`'s `createServer.ts` can be replaced by a `createMcpServer` call with:

1. no new transitive runtime dependency versus its current `typia` + MCP SDK set (`pnpm why zod-to-json-schema` comes back empty in default mode);
2. identical wire behavior: same `tools/list` schemas, same validation-feedback loop, same `instructions` in the handshake;
3. a zero-parameter tool callable without `arguments`;
4. one improvement the inline version lacks: `outputSchema` + `structuredContent`.

### Touched areas / test plan

- `packages/mcp/src/index.ts`, `packages/mcp/src/internal/McpControllerRegistrar.ts`, `packages/mcp/package.json`
- `packages/utils/src/utils/LlmJson.ts` (JSDoc only)
- `tests/test-mcp/src/features/` (function-per-file): `test_mcp_create_server_instructions`, `test_mcp_create_server_standalone`, `test_mcp_class_controller_omitted_arguments`, `test_mcp_class_controller_output_schema`, plus preserve-mode regression against the lazy-loading change
- `website/src/content/docs/utilization/mcp.mdx`, `packages/mcp/README.md`

### Out of scope

- Transport helpers (stdio/HTTP wiring stays the SDK's job; `@ttsc/graph`'s `startServer.ts` is 10 lines and fine).
- Lazy resource construction (`@ttsc/graph` defers graph building via a closure inside its own application class — app-side concern, no package support needed).
- `@typia/langchain` / `@typia/vercel` parity review (separate issue if wanted).

samchon:
## Intent

Make `@typia/mcp` the obvious way to stand up an MCP server from typia — the thing `@ttsc/graph` hand-rolled ~100 lines to avoid (#1988). The whole package is now one function.

```typescript
import { createMcpServer } from "@typia/mcp";
import typia from "typia";

const server = createMcpServer(
  typia.llm.controller<BbsService>("bbs", new BbsService()),
);
await server.connect(new StdioServerTransport());
```

`createMcpServer(controller, version?)` — pass one `typia.llm.controller`, connect a transport. That's the entire public API.

Fixes #1988.

## What the controller gives you, by reflection

- **Tools**: every class method becomes an MCP tool — `inputSchema`, `outputSchema`, and argument validation all from the TypeScript types, no hand-written JSON schema.
- **Server name**: the controller's `name`.
- **Instructions**: the class JSDoc (`application.description`) ships as the MCP handshake instructions — write the usage contract once, as documentation.
- **Structured output**: `ILlmFunction.output` → `outputSchema` in `tools/list` and `structuredContent` in every result, with a compact JSON text fallback.
- **Self-correction**: invalid arguments come back as an in-band tool error with typia's `// ❌` field markers — the loop the MCP spec (2025-11-25, SEP-1303) now recommends for input validation errors. Coercion runs first, so an LLM's `"12"` for a `number` is accepted.

## Fixed

Omitted `params.arguments` now validates as `{}` (the MCP spec allows the omission), so zero-parameter tools are callable and parameterized tools get field-level feedback instead of one generic "expected object".

## Dependencies

`@typia/mcp` now adds **zero** runtime modules beyond what `typia` already installs, plus the MCP SDK peer. The former `preserve` mode — which reached into MCP SDK private internals (`_registeredTools`) and forced `zod-to-json-schema` on every user — is gone, along with the `zod` devDependency.

## Scope

Deliberately narrow, converging on the driving use case (ttsc's graph server is class-based, tools-only):

- **No `registerMcpControllers`** — `createMcpServer` covers it; a second public entry point earned nothing.
- **No prompts / resources / completion** — an earlier draft modeled MCP prompts through `ILlmController`, which is a tool-calling abstraction (wrong fit). These are different MCP primitives and get their own design in #1989 (P1–P3), not a reused controller.
- **Class controllers only** — accepting `IHttpLlmController` would make the MCP server proxy a REST API (mixed concerns) and isn't needed here. Widening the accepted type back later is non-breaking.

## Acceptance — the `@ttsc/graph` litmus test

`@ttsc/graph`'s inline `createServer.ts` (~100 lines) collapses to:

```typescript
export const createServer = (graph: TtscGraphSource, version: string): McpServer =>
  createMcpServer(
    typia.llm.controller<ITtscGraphApplication>("ttsc-graph", new TtscGraphApplication(graph)),
    version,
  );
```

`test_mcp_create_server_lazy_controller` pins the exact pattern this relies on: the handshake and instructions answer immediately while the executor defers its state (the resident graph) until the first tool call, and the reflected result comes back validated. The migration is a strict improvement — it adds coercion and `outputSchema`/`structuredContent` the inline version never had. The one adjustment: name the controller `"ttsc-graph"` (server name = controller name).

## Validation

- `pnpm --filter ./tests/test-mcp start` — 10/10 pass
- `pnpm --filter @typia/mcp build` — ttsc + rollup clean
- `website` `pnpm run build` — MDX/nav clean
- Full `pnpm test` (all workspaces + Go) delegated to CI

## Files

The package is `packages/mcp/src/index.ts` + `internal/McpControllerRegistrar.ts` — two files. Docs: `website/src/content/docs/utilization/mcp.mdx`, README, and JSDoc on `ILlmController` / `IHttpLlmController` / `HttpLlm` / `LlmJson.validateArguments`.

https://claude.ai/code/session_01PGG5HgnRnvqhjjArFSrSUJ


samchon:
Follow-up 5290ff6baf: `registerMcpControllers` removed from the public API per review — `createMcpServer` is the entire surface now. The existing-server path duplicated what the `options` passthrough already covers, so keeping a second public function only preserved a spec nobody needs. The tool registrar remains as an internal module with the same `(server, definitions)` signature as the prompt/resource/completion registrars; all tests, guides, and JSDoc examples (`ILlmController`, `IHttpLlmController`, `HttpLlm.controller`, `@typia/utils` README) now go through `createMcpServer`. Local validation re-run: test-mcp 21/21, `@typia/mcp` build, website build — all green.

samchon:
Force-pushed a full reshape (history rewritten into 3 clean commits). The API converged to a single function: `createMcpServer(controller, version?)`.

Removed from the earlier draft after review: `registerMcpControllers`, prompts/resources/completion (prompts were wrongly modeled through `ILlmController`; deferred to #1989 with a proper abstraction), `IHttpLlmController` support, and the `{name, version, options, instructions}` props object — server name comes from `controller.name`, instructions from the class JSDoc, version defaults to "1.0.0". Added `test_mcp_create_server_lazy_controller` proving the @ttsc/graph lazy-build + JSDoc-instructions pattern. 10/10 tests, package build, and website build all green locally.

samchon:
**Verified end-to-end against a real Claude Code client**, and fixed a latent packaging bug it surfaced.

Added `experiments/mcp` — a real npm consumer that installs the packed tarballs (like `experiments/smoke`) and drives `createMcpServer` over stdio. Two runs:
1. MCP SDK `Client` ↔ server: handshake, tools/list, and tools/call with valid, coerced (`"40"`→`40`), invalid (validation feedback names `$input.x`), and throwing args — all 12 checks pass.
2. The actual `claude` CLI (`--mcp-config .mcp.json`): Claude connected to the server, called `mcp__calc__add {x:40,y:2}`, received `{"value":42}` as **both text and `structuredContent`**, and answered "42" — no hand-written schema anywhere.

**Bug it caught** (fixed in `fix(config)` 5e1b96deae): the shared rollup config left externals as CJS, so ttsc's `require("…/sdk")` was transcoded into a default import — which throws at runtime because `@modelcontextprotocol/sdk` (and `ai`, `@langchain/core/*`, `tinyglobby`) are named-export-only ESM. This broke the published `.mjs` entry of **@typia/mcp, @typia/vercel, and @typia/langchain** for every ESM consumer, latent because the test suites run against source, not the bundle. The fix scopes `esmExternals` to third-party packages (namespace import) while keeping `@typia/*` on the default import; `src/interop.ts` (a randexp-backed `typia.random`) confirms CJS externals stay callable.

samchon:
Ran the full-scope self-review to convergence (repeat until a round yields zero improvements), fixing everything each round:

- **R1:** honest SDK peer floor (`>=1.12.0`, when structured output landed), renamed a stale `standalone` test with a vacuous assertion, added void-return coverage.
- **R2:** dropped stale "prompt call"/"OpenAPI operation" doc references; the validation test now asserts `isError` (the prior clause could never fail); covered method-JSDoc→`tool.description`; tidied step comments.
- **R3:** **unknown tool now raises a JSON-RPC protocol error (`-32602`)** instead of an in-band `isError` result — the MCP spec classifies it as a protocol error and the reference `McpServer` throws `McpError(InvalidParams)`; a model can't self-correct a nonexistent tool (validation failures stay in-band, as before). Docs reframed so the reflection is taught via `typia.llm.application` with the class-JSDoc→`ILlmApplication.description`→MCP-`instructions` chain made explicit both ways; removed MCP from http.mdx's HTTP-controller adapter lists (createMcpServer takes a class controller only).
- **R4:** trimmed the removed-feature migration note from the guide; aligned the e2e comment.
- **R5:** clean — zero findings (fresh core re-read + artifact sweep).

The e2e experiment gained an unknown-tool protocol-error assertion (a real client rejects on it). Every round re-ran test-mcp; the runtime change was re-verified against the rebuilt tarball via the MCP-SDK-client e2e and the real `claude` CLI. Merge-ready once CI is green.

PR #1993, opened 2026-07-03
fix(website): dedupe typescript devDependency to catalog:typescript

samchon:
## Intent

`website/package.json` declared `typescript` **twice** in `devDependencies`:

```json
"typescript": "catalog:typescript",
...
"typescript": "~5.9.3",
```

Since a duplicate JSON key resolves to the **last** occurrence, the website was actually installing TypeScript `5.9.3` instead of the workspace catalog version — silently diverging from every other workspace package.

## Scope

- Remove the duplicate key, keeping a single `"typescript": "catalog:typescript"` in its correct alphabetical position.
- Verified every other workspace member (`packages/*`, `benchmark`, `tests/*`, `website`) already resolves TypeScript via `catalog:typescript`.

## Out of scope

`experiments/mcp` and `experiments/smoke` keep the hardcoded `"typescript": "7.0.1-rc"` on purpose: they live **outside** the pnpm workspace and are installed with plain `npm install` (which does not understand the pnpm-only `catalog:` protocol), so they cannot reference the catalog. Their pinned value already matches the catalog (`7.0.1-rc`).

## Test plan

- [x] `pnpm install` resolves the website to the catalog TypeScript
- [x] No remaining duplicate-key diagnostics in `website/package.json`


samchon:
tsgo(TypeScript 7 RC) 전환이 twoslash(Nextra의 shiki 하이라이터가 내부적으로 로드)와 근본적으로 호환되지 않아 철회합니다. twoslash는 Node ESM 로드 시점에 클래식 `typescript` 패키지의 루트 export를 바로 import하는데, tsgo RC의 package.json에는 루트 `.` export가 없어 `No "exports" main defined`로 실패합니다. webpack alias, pnpm overrides/packageExtensions, pnpm patch 모두 이 실패 지점(로더가 Node에서 직접 동적 import하는 지점)에는 영향을 주지 못해 깔끔한 수정 경로가 없었습니다.

PR #2002, opened 2026-07-04
Document patternProperties schema escape hatch

samchon:
## Summary
- clarify that `x-*` is an OpenAPI-style vendor extension convention, not a JSON Schema reservation
- document `tags.JsonSchemaPlugin` as the schema-only escape hatch for OpenAPI 3.1 / JSON Schema `patternProperties`
- call out that runtime key rejection still needs a validating custom tag or separate key validation

## Discussion
- Addresses discussion #1254

## Validation
- `pnpm format`
- `pnpm run build` from `website/`

### DIFF
--- a/website/src/content/docs/json/schema.mdx
+++ b/website/src/content/docs/json/schema.mdx
@@ -129,7 +129,7 @@ A few JSON-Schema-only comment tags are also recognized — they don't affect ru
 
 ## Custom fields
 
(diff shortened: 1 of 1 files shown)


samchon:
Self-review round 1 completed against the full PR diff.

Reviewed scope: the single `json/schema` docs hunk, the vendor-extension wording change, the `patternProperties` example, and the runtime-validation caveat.

Result: no follow-up changes found. CI is green.

PR #2005, opened 2026-07-07
fix(native): preserve LLM parameter property order

samchon:
﻿## Summary

Fixes `typia.llm.parameters<T>()` object property ordering when parameter schemas pass through tagged atomic properties or object intersections.

The metadata object already preserves declaration order in its `Properties` slice, but JSON schema emission copied properties into an unordered Go map and the LLM schema converter then sorted property keys alphabetically. Tagged examples like `id`, `email`, `age` therefore emitted as `age`, `email`, `id`.

This change carries the metadata property order through JSON schema object emission with `LiteralFactory_OrderedObject`, and has the LLM converter respect that order when converting nested object properties. Plain map fallback behavior remains deterministic by sorting keys.

## Tests

- `pnpm format`
- `pnpm --filter ./tests/test-typia-schema start -- --include llm_parameters_property_order`
- `go -C packages/typia/native test ./core/programmers/llm ./core/programmers/iterate -count=1`
- `go -C packages/typia/native test -tags typia_native_internal ./core/programmers/llm -run TestLlmSchemaProgrammerCoverage -count=1`

Note: full `pnpm test:go` was started earlier but did not complete before interruption/time limit; the related Go packages above passed.

### DIFF
--- a/.github/workflows/nestia.yml
+++ b/.github/workflows/nestia.yml
@@ -17,68 +17,11 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  # Pack the typia tarballs once and share them with every matrix job below,
-  # instead of rebuilding them six times.
-  tarballs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - uses: pnpm/action-setup@v4
-        with:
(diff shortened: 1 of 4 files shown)


samchon:
Updated \.github/workflows/nestia.yml\ to remove the tarball/matrix fan-out and run a single root \pnpm test\ job.

PR #2010, opened 2026-07-07
chore(ttsc): align the @ttsc family and adopt the 0.17 workspace cache

samchon:
## Intent

Two related toolchain chores in one PR:

1. **Dependabot never updated the `@ttsc/*` companions.** The allow list admitted only `typescript` and `ttsc`, so #2004 bumped ttsc to 0.17.x while `@ttsc/playground` / `@ttsc/unplugin` / `@ttsc/wasm` stayed on 0.16.10.
2. **ttsc 0.17 moved its Go plugin build cache to `<workspaceRoot>/node_modules/.cache/ttsc`**, resolved by walking up to `pnpm-workspace.yaml` — which makes the repo's own `TTSC_CACHE_DIR` plumbing redundant and gives CI a well-known, content-keyed directory to cache.

## Scope

- **deps** — align the whole family on `^0.17.3`, including the standalone `experiments/smoke` and `experiments/mcp` pins (`^0.16.0`) that live outside the workspace catalogs.
- **dependabot** — allow `@ttsc/*` and group it with `ttsc` so lockstep releases land as one PR.
- **cache plumbing** — delete `scripts/with-ttsc-cache.cjs` and unwrap the root scripts; ttsc's workspace-local default replaces it (and adds the built-in 30-day / 2 GB GC that pinned locations opt out of). `tests/test-typia-cli` keeps an explicit fallback to the shared cache because its fixture projects live in `os.tmpdir()`, outside the walk-up's reach. `check-file-errors.cjs` keeps its throwaway `--cache-dir` (deliberate isolation).
- **CI** — restore `node_modules/.cache/ttsc` via `actions/cache` in the workflows that actually compile the plugin (test, nestia, experiments, website). Entries are content-keyed by plugin source + ttsc/tsgo versions, so stale restores are safe and still reuse incremental go-build objects. `build.yml` is excluded (`pnpm build` never compiles the plugin — verified empirically); `release.yml` stays cold on purpose.

## Deferred

- `TestIntersectionNonsensibleBackstop` (`packages/typia/native/cmd/ttsc-typia`) runs `runBuild --emit` on a fixture tsconfig without `outDir`. Because typia's `exports` ship TS source, the fixture's program pulls in `packages/typia/src/**.ts` and tsgo emits `.js` **next to typia's own sources**. Pre-existing since #1970 (verified: reproduces with wrapper-era `TTSC_CACHE_DIR` set, and with this test alone); it makes a second consecutive `pnpm test` fail test-typia-cli's no-JS guard and contaminates locally staged tarballs. Fix is one line — add `"outDir": "dist"` to the shared `atomicIntersectionSchemaTSConfig` — but it touches the shipped native tree and therefore requires a typia version bump per the change-integrity rule, so it is deferred to its own PR.
- `typescript` stays outside the dependabot group; ttsc pins its own tsgo line, so grouping them could pair incompatible versions.

## Test plan

- `pnpm install` + `pnpm build` — clean.
- Cold-built the plugin through `examples` with no env override: cache landed in `node_modules/.cache/ttsc/{plugins/<content-key>,go-build}` as expected.
- Full `pnpm test` (TS suites + both Go layers) — green; `test-typia-cli` exercised its new shared-cache fallback.
- `pnpm package:tgz` — tarballs stage cleanly without the env override.


samchon:
Self-review pass over the branch produced three follow-up commits:

- **`ci: seed the ttsc plugin cache from master`** — the four cache consumers run on `pull_request` only, and PR-ref caches are invisible to other PRs, so no shared baseline ever existed (every entry in the caches API sat on `refs/pull/*/merge`). The new `ttsc-cache.yml` rebuilds the cache on relevant master pushes; an exact-key hit skips the save.
- **`ci(dependabot): extend scanning to the experiments manifests`** — `experiments/{mcp,smoke}` are outside the workspace, which is exactly how their `ttsc: ^0.16.0` pins drifted; `directories:` brings them into the same grouped family PR.
- **`chore(ttsc): keep the legacy .ttsc skip in the wasm staleness walk`** — reverts an over-eager cleanup; the walk also covers the sibling ttsc checkout where a leftover `.ttsc` dir would defeat the mtime check.

CI timing with the cache in effect (job durations, latest push vs. recent-run baseline):

| workflow | before | after |
|---|---|---|
| test | 12–15m | ~11m48s |
| nestia | 10–15m | ~9m42s |
| website | 8–9m | ~5m44s |
| experiments | ~4m | ~1m |
| build | ~1m | unchanged (no cache step — `pnpm build` never compiles the plugin) |

Cache entry is ~240MB compressed; restore ≈9s, save ≈6s, save skipped on exact-key hit. The earlier red `test` run was a network flake in `test-utils` (external fetch returned HTML), green on rerun.

PR #2016, opened 2026-07-10
fix(esm): emit genuine ESM through rolldown instead of transcoding CommonJS

samchon:
## Intent

The published `.mjs` of `typia` and every `@typia/*` package is broken for real ESM consumers. The shared rollup config transcoded ttsc's **CommonJS** `lib/**/*.js` output through `@rollup/plugin-commonjs`:

- tslib's runtime `__exportStar` cannot be statically analyzed, so the entry facade lost every named export — `import { assert } from "typia"` throws `SyntaxError` in Node ESM.
- Modules in require cycles were split into `_virtual/` wrappers plus facade chunks whose names collide, producing `index2.mjs`, `functional2.mjs`, … all over `lib/`.
- CI never caught it: the `experiments` workflow only runs the `smoke` project through `ttsx` (source execution), which never touches the built `.mjs`.

## Scope

1. **Build pipeline** — rollup (CJS transcode) is gone. `rolldown -c` consumes the TypeScript sources directly: rolldown transpiles natively, `@ttsc/unplugin` applies the project's ttsc plugins, and `preserveModules` emits a 1:1 `lib/<path>.mjs` twin per module. Named + default exports intact, no facades, no CommonJS interop. Everything outside `src/` is external. Shared config: `config/rolldown.config.mjs`.
2. **Source ESM-safety** — the per-file transpile has no cross-file type info, so: `re-exports.ts` marks type-only re-exports as `export type` (`tags` is the only runtime value in `@typia/interface`); the CLI executables drop ambient `require` / `__dirname` in favor of `createRequire` / `process.argv[1]` anchors; `isolatedModules` is now enforced repo-wide so regressions fail the type-check.
3. **Verification (follow-up commits on this PR)** — a new tgz-installing `experiments/esm` workspace that imports and runs `typia`, `@typia/utils`, `@typia/mcp`, `@typia/langchain`, `@typia/vercel` under real Node ESM, wired into the `experiments` workflow together with the existing `experiments/mcp` stdio e2e.

## Test plan

- `pnpm build` — clean `lib/`: no `*2.mjs`, no `_virtual/`, named exports verified on every entry.
- `pnpm run package:tgz` + `experiments/smoke` + `experiments/mcp` + new `experiments/esm` against the staged tarballs.
- `pnpm test` in CI.

## Deferred

- `typia` CLI wizard (`typia generate`) end-to-end on Windows/global installs — the `process.argv[1]` anchor mirrors the existing `generate/ttsc.ts` behavior shipped in v13.0.0.

https://claude.ai/code/session_01CWKKnrPiTvprpA5sEFNA2Y


samchon:
## Local verification (Windows, node 22)

**Build output** — all five packages, fresh `pnpm build`:
- zero `*2.mjs` facade chunks, zero CommonJS-interop `_virtual` wrappers (the single remaining `_virtual/_rolldown/runtime.mjs` is rolldown's namespace helper — one clean function, kept at its default location per review)
- `lib/index.mjs` of `typia` exports the full named surface + `default`; `transform.mjs` descriptor intact; CLI shebang preserved

**Runtime, tarball-installed real ESM** (`import.meta.resolve("typia")` → `…/lib/index.mjs`, so the bundled ESM is definitively what ran):
- `experiments/esm`: **21/21 checks pass** — typia named+default imports, transformed `is`/`assert`/`validate`, json round-trip, Pattern-tag `random` (randexp CJS external), `@typia/utils` named exports, `createMcpServer`, `toLangChainTools` + `tool.invoke`, `toVercelTools` + `tool.execute`
- `experiments/mcp`: full stdio client/server e2e — all checks pass
- `experiments/smoke`: passes
- `tests/test-typia-cli`: passes (CLI executables were touched for ESM-safety)

For comparison, on v13.0.0 the first line of `experiments/esm` (`import { assert } from "typia"`) dies with `SyntaxError: The requested module 'typia' does not provide an export named 'assert'`.

Issue #2024, opened 2026-07-13
fix(openapi): preserve discriminator metadata through v3.1/v3.2 upgrade


samchon:
## Problem

`OpenApiConverter.upgradeDocument()` silently removes a valid `discriminator` from an OpenAPI 3.1 or 3.2 `oneOf` schema even though the input types and the emended `OpenApi.IJsonSchema.IOneOf` output type all expose the field.

Verified against:

- published `@typia/utils@13.1.0` / `@typia/interface@13.1.0`;
- current local `samchon/typia` master `b9d256c165` (`chore(workspace): bind ttsc catalog family with YAML anchor, bump to ^0.18.4`); and
- Node 22.15.0.

No tsconfig transformer plugin is involved; this is a direct runtime converter call.

## Minimal reproduction

```ts
import { OpenApiConverter } from "@typia/utils";

const components = {
  schemas: {
    Choice: {
      oneOf: [
        { $ref: "#/components/schemas/OptionA" },
        { $ref: "#/components/schemas/OptionB" },
      ],
      discriminator: {
        propertyName: "kind",
        mapping: {
          a: "#/components/schemas/OptionA",
          b: "#/components/schemas/OptionB",
        },
      },
    },
    OptionA: {
      type: "object",
      properties: {
        kind: { const: "a" },
        value: { type: "string" },
      },
      required: ["kind", "value"],
    },
    OptionB: {
      type: "object",
      properties: {
        kind: { const: "b" },
        label: { type: "string" },
      },
      required: ["kind", "label"],
    },
  },
};

for (const openapi of ["3.1.1", "3.2.0"] as const) {
  const upgraded = OpenApiConverter.upgradeDocument({
    openapi,
    paths: {},
    components: structuredClone(components),
  });
  console.log(openapi, upgraded.components.schemas?.Choice);
}
```

Actual result for both versions:

```json
{
  "oneOf": [
    { "$ref": "#/components/schemas/OptionA" },
    { "$ref": "#/components/schemas/OptionB" }
  ]
}
```

Expected result:

```json
{
  "oneOf": [
    { "$ref": "#/components/schemas/OptionA" },
    { "$ref": "#/components/schemas/OptionB" }
  ],
  "discriminator": {
    "propertyName": "kind",
    "mapping": {
      "a": "#/components/schemas/OptionA",
      "b": "#/components/schemas/OptionB"
    }
  }
}
```

## Root cause

[...]

PR #2027, opened 2026-07-15
test(utils): cover v3.0 discriminator topology guards

samchon:
## Summary

- directly exercise the OpenAPI 3.0 document upgrader's discriminator topology guards
- cover enum branch expansion, nullable absorption, and single-branch collapse
- assert each rewritten output omits discriminator metadata while preserving the normalized schema

## Why a follow-up

Recursive self-review after #2026 merged found that its positive 3.0 document case reached `OpenApiV3Upgrader`, but the negative topology controls used the public schema route, which normalizes through `OpenApiV3_1Upgrader`. The production guards were present in both implementations; this PR closes the direct test gap for the separate 3.0 document implementation.

## Scoped validation

- `pnpm --filter @typia/test-utils start -- --include openapi_converter_discriminator`
- targeted Prettier
- `git diff --check`
- targeted `typos`

Refs #2024

### DIFF
--- a/tests/test-utils/src/features/openapi/test_openapi_converter_discriminator.ts
+++ b/tests/test-utils/src/features/openapi/test_openapi_converter_discriminator.ts
@@ -213,6 +213,46 @@ export const test_openapi_converter_discriminator = (): void => {
     Object.hasOwn(clean(nullable), "discriminator") === false,
   );
 
+  const v30RewriteFixture: IV30Fixture = createV30Fixture();
+  TestValidator.equals(
+    "v3.0 document enum rewrite drops discriminator",
+    clean(
+      upgradeV30DocumentSchema(v30RewriteFixture, {
+        oneOf: [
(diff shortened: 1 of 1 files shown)


samchon:
Recursive self-review completed on `64dead42bc`.

- confirmed every added negative case dispatches through an OpenAPI 3.0 document (`openapi: 3.0.4`) and therefore reaches the separate `OpenApiV3Upgrader` implementation;
- confirmed enum expansion, nullable absorption, and one-branch collapse compare the complete normalized `Boundary` schema, so an orphan discriminator or branch-order/content change cannot hide behind a weak predicate;
- reread the full one-file diff after the focused test, targeted formatting, diff check, and typos pass; no further improvement survived.

The first Ubuntu run failed in the unrelated pre-existing random recursion test (`test_random_recursive_terminates`) before this OpenAPI case; the failed job has been rerun as a flaky infrastructure/test retry.

PR #2060, opened 2026-07-16
fix(http): preserve migration request and response contracts

### LINKED ISSUE #2029: HttpMigration loses non-200/201 successes and response metadata
## Problem

`HttpMigration`'s request/response boundary has four independent protocol-loss defects in `packages/utils/src/http/internal/HttpMigrateRouteFetcher.ts` and its response classification in `HttpMigrateRouteComposer.ts`:

1. only status 200/201 is considered successful, so valid 202-299 responses throw `HttpError` and are parsed through the failure branch;
2. a parsed JSON/form/binary failure body is asserted to `string` when constructing `HttpError`, turning objects into `[object Object]` and losing the body that `propagate()` retained;
3. `Set-Cookie` is split at `;`, which separates cookie attributes rather than repeated cookie fields;
4. concatenating a host ending in `/` with a route beginning in `/` produces a double-slash path.

The public contract says execution throws for non-2xx responses, accepts `IHttpConnection.host` as a base URL, and exposes response headers/body through migration results.

## Evidence

At current `master`:

| Case | Current | Expected |
| --- | --- | --- |
| 202, 204, 206, 299 | `Response.ok === true`, but `execute()` throws | success |
| 400 JSON `{ "status": 400 }` | `HttpError.message === "[object Object]"` | structured body preserved by `toJSON()` |
| `sid=abc; Path=/; HttpOnly` | `['sid=abc', 'Path=/', 'HttpOnly']` | one complete cookie string |
| `https://example.com/` + `/users` | `https://example.com//users` | `https://example.com/users` |

The composer likewise chooses a success schema only from 201, then 200, then `default`, and classifies every other explicit 2xx response as an exception. The hard-coded status checks, unsafe body assertion, cookie split, and one-sided slash normalization all originated with the migration code in `a7cbc4f1ae`. Exact open/closed issue and pull-request searches found no implementation claim.

## Consequence surface

- `HttpMigration.execute` and `propagate`, including HEAD/no-content paths;
- `HttpLlm` callers built on migrated routes;
- JSON, text, form, binary, and empty error responses;
- one and multiple `Set-Cookie` fields, especially `Expires` values containing commas;
- origin-only and base-path hosts, path parameters, and query strings.

## Approach

Keep HTTP semantics consistent between composition and fetching: use the 2xx status class, choose/document a deterministic representative schema when OpenAPI declares multiple success statuses, preserve a typed response body separately from the JavaScript `Error.message`, use the platform's repeated-cookie API where available, and normalize exactly the host/path join boundary. Do not globally rewrite URL path contents.

- Cover 200, 201, 202, 204/205, 206, 299, 300, and 400 with declared/default response schemas.
- Cover text, JSON object/array/scalar/null, form, binary, and empty failure bodies; `toJSON()` must preserve the parsed representation.
- Preserve complete single/multiple cookie fields with attributes and `Expires`; ordinary headers remain unchanged.
- Cover all four host/path slash combinations plus a base path and query.
- Run the focused utils/HTTP tests, package build, and the repository suite.

## Coordination

This is fetcher-boundary work. OpenAPI path-parameter association and typia's standalone query parser are separate issues and may proceed in parallel.

### LINKED ISSUE #2030: HttpMigrateRouteComposer cross-wires reverse-ordered path parameters
## Problem

`HttpMigrateRouteComposer` associates OpenAPI path-parameter schemas with URL placeholders by array position instead of parameter name. OpenAPI identifies a parameter by `in: "path"` plus `name`; operation parameter order is not required to match placeholder order.

## Evidence

For `/users/{userId}/posts/{postId}`, reverse the valid OpenAPI parameter array:

```text
postId: integer, description "POST ID"
userId: string,  description "USER ID"
```

The composed route currently reports no errors but maps `userId` to the integer/post metadata and `postId` to the string/user metadata. URL interpolation still reads values by route key, so the request path can look correct while LLM/schema validation and descriptions are cross-wired.

The cause is the positional mapping around `packages/utils/src/http/internal/HttpMigrateRouteComposer.ts`'s `parameterNames` and `pathParameters`: the array is reordered only when counts differ, then `parameterNames[i]` is paired with `pathParameters[i]`. This code dates to `a7cbc4f1ae`; exact issue and pull-request searches found no duplicate.

## Consequence surface

- migrated route schemas and parameter descriptions;
- `HttpLlmApplicationComposer` argument advertisement and validation;
- reversed arrays, same-count wrong names, missing/extra parameters, duplicate declarations, normalized keys, and repeated placeholders.

## Approach

Make placeholder names the canonical order and join operation parameters by name. Preserve the documented synthesis of truly missing path parameters, but diagnose incompatible name sets instead of silently pairing unrelated entries.

- Ordered and reversed parameter arrays compose identical routes and LLM schemas.
- Same-count wrong names and duplicate declarations produce deterministic diagnostics.
- Missing-parameter synthesis continues to work where currently supported.
- Add focused composer and downstream LLM tests, then run the utils build and repository suite.

## Coordination

This may be implemented alongside other `@typia/utils` HTTP work, but its issue/acceptance remains name-based composition rather than fetch behavior.

### LINKED ISSUE #2031: typia.http query decoders discard raw query strings without ?
## Problem

The public `typia.http.query` family accepts a “query string or `URLSearchParams`”, but `_httpQueryParseURLSearchParams` discards a normal raw query string that has no `?`.

## Evidence

`packages/typia/src/internal/_httpQueryParseURLSearchParams.ts` does:

```ts
const index = input.indexOf("?");
input = index === -1 ? "" : input.substring(index + 1);
```

Therefore `"name=typia&count=2"` becomes an empty `URLSearchParams`; only a full URL or a string prefixed with `?` survives. `new URLSearchParams("name=typia&count=2")` is the platform's normal raw-query behavior. Exact issue and pull-request searches found no duplicate.

## Consequence surface

All transformed direct/factory variants of `query`, `assertQuery`, `isQuery`, and `validateQuery`; empty strings, leading `?`, full absolute/relative URLs, fragments, repeated keys, and encoded `?` data.

## Approach

Distinguish raw query text from URL-like input without dropping either form. Preserve the existing `IReadableURLSearchParams` pass-through and avoid treating an encoded question mark as a URL delimiter.

- `a=1&b=2`, `?a=1&b=2`, and a full URL ending in the same query decode identically.
- Empty, repeated, percent-encoded, and fragment-bearing cases are explicit.
- Add regressions across the direct and factory query families and run focused generated HTTP tests plus the repository suite.

## Coordination

This is owned by typia's runtime query parser, not `@typia/utils` migration fetching.

### LINKED ISSUE #2050: HttpMigration drops request metadata and optionality
## Problem

`HttpMigration` preserves OpenAPI request headers on each route but cannot accept or send them, and `HttpLlm` omits them from function arguments. Cookie parameters are discarded during migration. The LLM schema also marks every grouped query and request body required, even when OpenAPI declares them optional.

The route model also discards OpenAPI parameter `style`/`explode`, so query arrays are always serialized as repeated keys even when the operation requires a comma-delimited form value.

As a result, generated function metadata and the request actually sent over the wire do not represent the source operation.

## Evidence

- `IHttpMigrateRoute` exposes a combined `headers` schema, but `HttpMigration.IFetchProps` has no headers field and `HttpMigrateRouteFetcher` builds headers only from the static connection/body content type.
- `HttpLlmApplicationComposer` builds parameters from path, query, and body only; `HttpLlmFunctionFetcher` forwards the same three groups. A required OpenAPI header is absent from both the LLM function and the captured request.
- `HttpMigrateRouteComposer` groups only `header` and `query` parameter locations. OpenAPI `cookie` parameters have no route/fetch representation and disappear without an application error.
- `parameters.required = Object.keys(parameters.properties)` makes optional grouped query and optional request bodies mandatory. The migration body metadata does not retain `requestBody.required`, and the fetcher rejects absence whenever a route body/query descriptor exists.
- Query serialization always appends each array element as a repeated key. For `style: form, explode: false`, `[1, 2]` becomes `ids=1&ids=2` instead of `ids=1,2`; neither `IHttpMigrateRoute` nor its grouped query schema retains the source serialization metadata.

Public in-memory fetch reproduction confirms that a required header is not sent and otherwise-valid calls without optional query/body groups throw. Exact open/closed issue and pull-request searches found no implementation claim.

## Consequence surface

OpenAPI 3.0/3.1 parameter locations, `style`/`explode` combinations, required/optional header and cookie values, grouped object vs primitive parameters, request-body `required`, empty optional query objects, LLM schema/parse/coerce/validate, direct migration execution, controller custom executors, static connection-header precedence, and browser cookie restrictions.

## Approach

Carry request location and requiredness through the route model instead of inferring it from descriptor existence. Expose header/cookie argument groups consistently in the LLM schema and fetch contracts, define precedence against connection headers, and serialize cookies according to the HTTP/OpenAPI contract. If a runtime cannot set a location safely, surface a deterministic composition error rather than silently dropping it.

- Required and optional headers/cookies appear with correct requiredness and reach the captured request.
- Optional query/body groups may be omitted; required groups still fail validation before fetch.
- Object and primitive parameter forms, collisions/casing, all supported style/explode encodings, arrays/objects, special characters, and connection-header precedence are covered.
- `HttpMigration` and `HttpLlm` schemas, validators, execute, and propagate remain in parity.
- Run focused utils HTTP/LLM tests and the full suite.

## Coordination

This owns request metadata. The companion HTTP migration issue owns response status/body/header handling and URL joining; both can share one HTTP batch and test harness.

samchon:
## Intent

Restore the HTTP migration boundary so composed OpenAPI request contracts and observed HTTP responses survive execution without being reordered, dropped, or misclassified.

## Scope

- Treat the complete 2xx class as successful, preserve structured error bodies and repeated cookies, and normalize only the host/path join boundary.
- Associate path parameters by OpenAPI name instead of array position.
- Accept raw query strings, leading-question-mark strings, and full URL query inputs consistently across direct and factory decoders.
- Carry header/cookie parameters, request optionality, and supported OpenAPI query serialization metadata through `HttpMigration` and `HttpLlm`.
- Preserve OpenAPI 3.2 `querystring`, cookie-style pass-through, required delimiter encoding, and reserved-header rules discovered during review.

## Deferred

- No unrelated OpenAPI converter, validator, or non-HTTP runtime changes.
- Repository-wide formatting remains deferred to the campaign Post-Campaign Cleanup pull request.
- The website production build remains excluded under the campaign's recorded packaging prerequisite.

## Verification

- `pnpm --filter @typia/interface build`
- `pnpm --filter @typia/utils build`
- `pnpm --filter typia build`
- `pnpm --filter @typia/test-utils start`
- `pnpm --filter @typia/test-typia-schema start`
- `git diff --check origin/master...HEAD`
- Six complete solo Self-Review rounds, with all accepted findings remediated and the terminal round clean.

Automatic GitHub Actions for campaign commits were deliberately cancelled by exact SHA. Local verification is the implementation gate under the issue-campaign procedure.

Closes #2029
Closes #2030
Closes #2031
Closes #2050

### DIFF
--- a/packages/interface/package.json
+++ b/packages/interface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@typia/interface",
-  "version": "13.1.1",
+  "version": "13.1.2",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "types": "src/index.ts",
--- a/packages/interface/package.json
+++ b/packages/interface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@typia/interface",
-  "version": "13.1.2",
+  "version": "13.1.1",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "types": "src/index.ts",
--- a/packages/interface/src/http/IHttpLlmController.ts
+++ b/packages/interface/src/http/IHttpLlmController.ts
@@ -80,9 +80,11 @@ export interface IHttpLlmController {
          * It is an object of key-value pairs of the API function's parameters.
          * The property keys are composed by below rules:
          *
-         * - Parameter names
-         * - Query parameter as an object type if exists
-         * - Body parameter if exists
+         * - Path parameter names
+         * - Header parameter group if it exists
+         * - Cookie parameter group if it exists
+         * - Query parameter group if it exists
+         * - Body parameter if it exists
          */
         arguments: object;
       }) => Promise<IHttpResponse>);
--- a/packages/interface/src/http/IHttpMigrateRoute.ts
+++ b/packages/interface/src/http/IHttpMigrateRoute.ts
@@ -152,6 +152,9 @@ export namespace IHttpMigrateRoute {
     /** Object properties owned by an object parameter. */
     properties: string[] | null;
 
+    /** Required properties owned by an object parameter. */
+    requiredProperties?: string[] | null;
+
     /** Whether the object parameter accepts undeclared properties. */
     additionalProperties?: boolean;
 
--- a/packages/interface/src/http/IHttpMigrateRoute.ts
+++ b/packages/interface/src/http/IHttpMigrateRoute.ts
@@ -125,6 +125,9 @@ export namespace IHttpMigrateRoute {
     /** Source parameter serialization metadata. */
     parameters?: IHttpMigrateRoute.ISerialization[];
 
+    /** Whole-query media metadata for OpenAPI 3.2 `querystring`. */
+    querystring?: IHttpMigrateRoute.IQuerystring;
+
     /** Returns title. */
     title: () => string | undefined;
 
@@ -136,6 +139,15 @@ export namespace IHttpMigrateRoute {
 
     /** Returns named examples. */
     examples: () => Record<string, any> | undefined;
+  }
+
+  /** Whole-query media metadata. */
+  export interface IQuerystring {
+    /** Normalized media type used to serialize the complete query string. */
+    type: string;
+
+    /** Returns the source media type definition. */
+    media: () => OpenApi.IOperation.IMediaType;
   }
 
   /** Cookie metadata. */
--- a/packages/interface/src/http/IHttpMigrateRoute.ts
+++ b/packages/interface/src/http/IHttpMigrateRoute.ts
@@ -6,7 +6,7 @@ import { OpenApi } from "../openapi/OpenApi";
  * `IHttpMigrateRoute` represents a single API endpoint with all
  * request/response schemas resolved and ready for code generation. Contains
  * {@link parameters} for URL path variables, {@link query} for query strings,
- * {@link headers}, {@link body} for request payload, and
+ * {@link headers}, {@link cookies}, {@link body} for request payload, and
  * {@link success}/{@link exceptions} for responses.
  *
  * @author Jeongho Nam - https://github.com/samchon
@@ -36,7 +36,7 @@ export interface IHttpMigrateRoute {
   headers: IHttpMigrateRoute.IHeaders | null;
 
   /** Combined cookies as single object. Null if none. */
-  cookies: IHttpMigrateRoute.ICookies | null;
+  cookies?: IHttpMigrateRoute.ICookies | null;
 
   /** Combined query parameters as single object. Null if none. */
   query: IHttpMigrateRoute.IQuery | null;
@@ -68,11 +68,11 @@ export namespace IHttpMigrateRoute {
     /** Parameter type schema. */
     schema: OpenApi.IJsonSchema;
 
-    /** Effective serialization style. */
-    style: "matrix" | "label" | "simple";
+    /** Effective serialization style. Defaults to `simple`. */
+    style?: "matrix" | "label" | "simple";
 
-    /** Effective explode behavior. */
-    explode: boolean;
+    /** Effective explode behavior. Defaults to `false`. */
+    explode?: boolean;
 
     /** Returns source parameter definition. */
     parameter: () => OpenApi.IOperation.IParameter;
@@ -90,10 +90,10 @@ export namespace IHttpMigrateRoute {
     schema: OpenApi.IJsonSchema;
 
     /** Whether the combined headers argument is required. */
-    required: boolean;
+    required?: boolean;
 
     /** Source parameter serialization metadata. */
-    parameters: IHttpMigrateRoute.ISerialization[];
+    parameters?: IHttpMigrateRoute.ISerialization[];
 
     /** Returns title. */
     title: () => string | undefined;
@@ -120,10 +120,10 @@ export namespace IHttpMigrateRoute {
     schema: OpenApi.IJsonSchema;
 
     /** Whether the combined query argument is required. */
-    required: boolean;
+    required?: boolean;
 
     /** Source parameter serialization metadata. */
-    parameters: IHttpMigrateRoute.ISerialization[];
+    parameters?: IHttpMigrateRoute.ISerialization[];
 
     /** Returns title. */
     title: () => string | undefined;
@@ -152,8 +152,16 @@ export namespace IHttpMigrateRoute {
     /** Object properties owned by an object parameter. */
     properties: string[] | null;
 
+    /** Whether the object parameter accepts undeclared properties. */
+    additionalProperties?: boolean;
+
     /** Effective serialization style. */
-    style: "form" | "simple" | "spaceDelimited" | "pipeDelimited" | "deepObject";
+    style:
+      | "form"
+      | "simple"
+      | "spaceDelimited"
+      | "pipeDelimited"
+      | "deepObject";
 
     /** Effective explode behavior. */
     explode: boolean;
@@ -174,6 +182,7 @@ export namespace IHttpMigrateRoute {
     type:
       | "text/plain"
       | "application/json"
+      | `application/${string}+json`
       | "application/x-www-form-urlencoded"
       | "multipart/form-data";
 
@@ -181,7 +190,7 @@ export namespace IHttpMigrateRoute {
     schema: OpenApi.IJsonSchema;
 
     /** Whether the request body is required. */
-    required: boolean;
+    required?: boolean;
 
     /** Returns description. */
     description: () => string | undefined;
--- a/packages/interface/src/http/IHttpMigrateRoute.ts
+++ b/packages/interface/src/http/IHttpMigrateRoute.ts
@@ -35,13 +35,16 @@ export interface IHttpMigrateRoute {
   /** Combined headers as single object. Null if none. */
   headers: IHttpMigrateRoute.IHeaders | null;
 
+  /** Combined cookies as single object. Null if none. */
+  cookies: IHttpMigrateRoute.ICookies | null;
+
   /** Combined query parameters as single object. Null if none. */
   query: IHttpMigrateRoute.IQuery | null;
 
   /** Request body metadata. Null if none. */
   body: IHttpMigrateRoute.IBody | null;
 
-  /** Success response (200/201). Null if void return. */
+  /** Representative success response for the declared 2xx class. */
   success: IHttpMigrateRoute.ISuccess | null;
 
   /** Exception responses keyed by status code. */
@@ -65,6 +68,12 @@ export namespace IHttpMigrateRoute {
     /** Parameter type schema. */
     schema: OpenApi.IJsonSchema;
 
+    /** Effective serialization style. */
+    style: "matrix" | "label" | "simple";
+
+    /** Effective explode behavior. */
+    explode: boolean;
+
     /** Returns source parameter definition. */
     parameter: () => OpenApi.IOperation.IParameter;
   }
@@ -79,6 +88,12 @@ export namespace IHttpMigrateRoute {
 
     /** Combined headers schema. */
     schema: OpenApi.IJsonSchema;
+
+    /** Whether the combined headers argument is required. */
+    required: boolean;
+
+    /** Source parameter serialization metadata. */
+    parameters: IHttpMigrateRoute.ISerialization[];
 
     /** Returns title. */
     title: () => string | undefined;
@@ -104,6 +119,12 @@ export namespace IHttpMigrateRoute {
     /** Combined query schema. */
     schema: OpenApi.IJsonSchema;
 
+    /** Whether the combined query argument is required. */
+    required: boolean;
+
+    /** Source parameter serialization metadata. */
+    parameters: IHttpMigrateRoute.ISerialization[];
+
     /** Returns title. */
     title: () => string | undefined;
 
@@ -115,6 +136,30 @@ export namespace IHttpMigrateRoute {
 
     /** Returns named examples. */
     examples: () => Record<string, any> | undefined;
+  }
+
+  /** Cookie metadata. */
+  export interface ICookies extends IHeaders {}
+
+  /** Serialization metadata for a grouped request parameter. */
+  export interface ISerialization {
+    /** OpenAPI parameter name sent on the wire. */
+    name: string;
+
+    /** Property key in the grouped argument, or null for an object parameter. */
+    key: string | null;
+
+    /** Object properties owned by an object parameter. */
+    properties: string[] | null;
+
+    /** Effective serialization style. */
+    style: "form" | "simple" | "spaceDelimited" | "pipeDelimited" | "deepObject";
+
+    /** Effective explode behavior. */
+    explode: boolean;
+
+    /** Returns source parameter definition. */
+    parameter: () => OpenApi.IOperation.IParameter;
   }
 
   /** Request body metadata. */
@@ -134,6 +179,9 @@ export namespace IHttpMigrateRoute {
 
     /** Body type schema. */
     schema: OpenApi.IJsonSchema;
+
+    /** Whether the request body is required. */
+    required: boolean;
 
     /** Returns description. */
     description: () => string | undefined;
--- a/packages/interface/src/http/IHttpMigrateRoute.ts
+++ b/packages/interface/src/http/IHttpMigrateRoute.ts
@@ -161,6 +161,7 @@ export namespace IHttpMigrateRoute {
     /** Effective serialization style. */
     style:
       | "form"
+      | "cookie"
       | "simple"
       | "spaceDelimited"
       | "pipeDelimited"
--- a/packages/interface/src/openapi/OpenApi.ts
+++ b/packages/interface/src/openapi/OpenApi.ts
@@ -242,8 +242,11 @@ export namespace OpenApi {
       /** Parameter location. */
       in: "path" | "query" | "querystring" | "header" | "cookie";
 
-      /** Parameter schema. */
+      /** Normalized parameter schema. */
       schema: IJsonSchema;
+
+      /** Source media types for a content-backed querystring parameter. */
+      content?: IContent;
 
       /** Whether required. */
       required?: boolean;
--- a/packages/interface/src/openapi/OpenApi.ts
+++ b/packages/interface/src/openapi/OpenApi.ts
@@ -248,6 +248,19 @@ export namespace OpenApi {
       /** Whether required. */
       required?: boolean;
 
+      /** OpenAPI parameter serialization style. */
+      style?:
+        | "matrix"
+        | "label"
+        | "form"
+        | "simple"
+        | "spaceDelimited"
+        | "pipeDelimited"
+        | "deepObject";
+
+      /** Whether arrays and objects are exploded during serialization. */
+      explode?: boolean;
+
       /** Parameter description. */
       description?: string;
 
--- a/packages/interface/src/openapi/OpenApi.ts
+++ b/packages/interface/src/openapi/OpenApi.ts
@@ -253,6 +253,7 @@ export namespace OpenApi {
         | "matrix"
         | "label"
         | "form"
+        | "cookie"
         | "simple"
         | "spaceDelimited"
         | "pipeDelimited"
--- a/packages/interface/src/openapi/OpenApiV3.ts
+++ b/packages/interface/src/openapi/OpenApiV3.ts
@@ -221,6 +221,19 @@ export namespace OpenApiV3 {
       /** Whether required. */
       required?: boolean;
 
+      /** Parameter serialization style. */
+      style?:
+        | "matrix"
+        | "label"
+        | "form"
+        | "simple"
+        | "spaceDelimited"
+        | "pipeDelimited"
+        | "deepObject";
+
+      /** Whether arrays and objects are exploded during serialization. */
+      explode?: boolean;
+
       /** Parameter description. */
       description?: string;
 
--- a/packages/interface/src/openapi/OpenApiV3_1.ts
+++ b/packages/interface/src/openapi/OpenApiV3_1.ts
@@ -237,6 +237,19 @@ export namespace OpenApiV3_1 {
       /** Whether required. */
       required?: boolean;
 
+      /** Parameter serialization style. */
+      style?:
+        | "matrix"
+        | "label"
+        | "form"
+        | "simple"
+        | "spaceDelimited"
+        | "pipeDelimited"
+        | "deepObject";
+
+      /** Whether arrays and objects are exploded during serialization. */
+      explode?: boolean;
+
       /** Parameter description. */
       description?: string;
 
--- a/packages/interface/src/openapi/OpenApiV3_2.ts
+++ b/packages/interface/src/openapi/OpenApiV3_2.ts
@@ -240,8 +240,11 @@ export namespace OpenApiV3_2 {
       /** Parameter location. */
       in: "path" | "query" | "querystring" | "header" | "cookie";
 
-      /** Parameter schema. */
-      schema: IJsonSchema;
+      /** Parameter schema for every location except `querystring`. */
+      schema?: IJsonSchema;
+
+      /** Media types for a content-backed `querystring` parameter. */
+      content?: Record<string, IMediaType>;
 
       /** Whether required. */
       required?: boolean;
--- a/packages/interface/src/openapi/OpenApiV3_2.ts
+++ b/packages/interface/src/openapi/OpenApiV3_2.ts
@@ -246,6 +246,19 @@ export namespace OpenApiV3_2 {
       /** Whether required. */
       required?: boolean;
 
+      /** Parameter serialization style. */
+      style?:
+        | "matrix"
+        | "label"
+        | "form"
+        | "simple"
+        | "spaceDelimited"
+        | "pipeDelimited"
+        | "deepObject";
+
+      /** Whether arrays and objects are exploded during serialization. */
+      explode?: boolean;
+
       /** Parameter description. */
       description?: string;
 
--- a/packages/interface/src/openapi/OpenApiV3_2.ts
+++ b/packages/interface/src/openapi/OpenApiV3_2.ts
@@ -251,6 +251,7 @@ export namespace OpenApiV3_2 {
         | "matrix"
         | "label"
         | "form"
+        | "cookie"
         | "simple"
         | "spaceDelimited"
         | "pipeDelimited"
--- a/packages/typia/package.json
+++ b/packages/typia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "13.1.1",
+  "version": "13.1.2",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "exports": {
--- a/packages/typia/package.json
+++ b/packages/typia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "13.1.2",
+  "version": "13.1.1",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "exports": {
--- a/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
+++ b/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
@@ -4,19 +4,30 @@ export const _httpQueryParseURLSearchParams = (
   input: string | IReadableURLSearchParams,
 ): IReadableURLSearchParams => {
   if (typeof input === "string") {
-    if (input.startsWith("?")) input = input.substring(1);
-    else {
+    let url: boolean = false;
+    if (input.startsWith("?")) {
+      input = input.substring(1);
+      url = true;
+    } else {
       const index: number = input.indexOf("?");
-      if (index !== -1 && isUrlPrefix(input.substring(0, index)))
+      if (index !== -1 && isUrlPrefix(input.substring(0, index))) {
         input = input.substring(index + 1);
-      else if (index === -1 && isUrlPrefix(input)) input = "";
+        url = true;
+      } else if (index === -1 && isUrlPrefix(input)) {
+        input = "";
+        url = true;
+      }
     }
-    const fragment: number = input.indexOf("#");
-    if (fragment !== -1) input = input.substring(0, fragment);
+    if (url) {
+      const fragment: number = input.indexOf("#");
+      if (fragment !== -1) input = input.substring(0, fragment);
+    }
     return new URLSearchParams(input);
   }
   return input;
 };
 
 const isUrlPrefix = (prefix: string): boolean =>
-  /^[a-z][a-z\d+.-]*:/i.test(prefix) || prefix.includes("/");
+  /^[a-z][a-z\d+.-]*:/i.test(prefix) ||
+  /^(?:\/\/|\/|\.{1,2}\/)/.test(prefix) ||
+  (prefix.length !== 0 && !/[=&]/.test(prefix));
--- a/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
+++ b/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
@@ -4,9 +4,19 @@ export const _httpQueryParseURLSearchParams = (
   input: string | IReadableURLSearchParams,
 ): IReadableURLSearchParams => {
   if (typeof input === "string") {
-    const index: number = input.indexOf("?");
-    input = index === -1 ? "" : input.substring(index + 1);
+    if (input.startsWith("?")) input = input.substring(1);
+    else {
+      const index: number = input.indexOf("?");
+      if (index !== -1 && isUrlPrefix(input.substring(0, index)))
+        input = input.substring(index + 1);
+      else if (index === -1 && isUrlPrefix(input)) input = "";
+    }
+    const fragment: number = input.indexOf("#");
+    if (fragment !== -1) input = input.substring(0, fragment);
     return new URLSearchParams(input);
   }
   return input;
 };
+
+const isUrlPrefix = (prefix: string): boolean =>
+  /^[a-z][a-z\d+.-]*:/i.test(prefix) || prefix.includes("/");
--- a/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
+++ b/packages/typia/src/internal/_httpQueryParseURLSearchParams.ts
@@ -13,7 +13,7 @@ export const _httpQueryParseURLSearchParams = (
       if (index !== -1 && isUrlPrefix(input.substring(0, index))) {
         input = input.substring(index + 1);
         url = true;
-      } else if (index === -1 && isUrlPrefix(input)) {
+      } else if (index === -1 && isExplicitUrlPrefix(input)) {
         input = "";
         url = true;
       }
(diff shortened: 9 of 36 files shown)


samchon:
Implementation is complete at `651c1cd96c867d84cdded8d24e37db80919abccb`.

Implemented:

- Preserved OpenAPI path/header/cookie/query/body location, requiredness, and `style`/`explode` metadata through route composition, LLM schemas, and fetch execution.
- Joined path parameters by placeholder name, synthesized missing declarations, reused repeated placeholders, and rejected duplicate or wrong-name declarations.
- Treated every `2xx` response as success, handled `204`/`205`/`HEAD` without a body, parsed the runtime response media type, retained structured error bodies, preserved repeated `Set-Cookie`, and joined host/path slash boundaries without rewriting the base path.
- Accepted raw, question-prefixed, and URL-shaped strings across direct and factory HTTP query decoders while excluding literal fragments and preserving percent-encoded question marks.
- Updated public guide text and bumped `@typia/interface`, `@typia/utils`, and `typia` to `13.1.2`.

Verification passed:

- `pnpm --filter ./packages/interface build`
- `pnpm --filter ./packages/typia build`
- `pnpm --filter ./tests/test-utils start -- --include http_migrate_response_contract http_migrate_path_parameter_names http_migrate_request_contract http_llm_application`
- `pnpm --filter ./tests/test-typia-schema start -- --include http_query_raw_strings`
- `git diff --check`

Intentionally skipped:

- The root full suite, because this issue-campaign implementation wave runs package-scoped verification and cancels per-push CI to avoid concurrent shared-cache contention.
- A fresh website build, because its documented prerequisites require the campaign-excluded root `pnpm build` and `pnpm package:tgz` sequence plus an isolated sibling `ttsc` checkout; the available sibling checkout contains unrelated local changes.
- Solo Self-Review and merge, which are the next campaign gates and are intentionally outside this implementation handoff.

The exact-SHA cancellation gate completed after the push; all six observed runs for this SHA finished `cancelled`.

samchon:
Solo Self-Review round 1 found reproducible gaps against #2029, #2030, #2031, and #2050. No remediation edits have been made yet.

## Findings

1. **Raw query values containing URL characters are discarded or truncated.** `_httpQueryParseURLSearchParams("path=/users/1&url=https://example.com")` returns an empty parameter set, and `"url=https://x?y=z"` becomes only `y=z`. The URL classifier currently treats any slash before `?` as a URL prefix. It also strips `#` from raw query text even though `URLSearchParams` treats it as data.
2. **Valid primitive request bodies fail before fetch.** A required `text/plain` string body composes successfully but `HttpMigration.propagate()` throws `body must be an object`; the same guard rejects scalar, array, and `null` JSON bodies. `IFetchProps.body` is likewise narrowed to `object` despite the OpenAPI schema allowing those JSON values.
3. **Multipart requests retain a stale inherited content type.** When the connection supplies `Content-Type: application/json`, a migrated `multipart/form-data` request sends that header with a `FormData` body, preventing the runtime from generating its boundary.
4. **Supported query serialization shapes are rejected or emitted incorrectly.** OpenAPI defines `spaceDelimited` and `pipeDelimited` for arrays and objects with `explode: false`, but the composer rejects object schemas and the fetcher would stringify them as `[object Object]`. Conversely, `spaceDelimited`/`pipeDelimited` with `explode: true` and `deepObject` without `explode: true` are accepted although those combinations are undefined. See the official [OpenAPI 3.1 Parameter Object](https://spec.openapis.org/oas/v3.1.0.html#parameter-object).
5. **Referenced and open object parameter schemas lose request data.** A `$ref` to an array/primitive is omitted from the synthesized LLM group schema (and is rejected for delimited styles), while an object parameter backed by `additionalProperties` serializes no entries. This breaks the requested direct/LLM parity for array/object forms.
6. **JSON-suffix media types are not composed.** `application/merge-patch+json` request bodies and `application/problem+json` success bodies reject the route; an `application/problem+json` exception remains but loses its schema. The fetcher already parses `+json`, so composition and execution disagree.
7. **Cookie group keys can overwrite path arguments.** For `/users/{cookie}` plus a cookie parameter, both the path argument and cookie group retain the key `cookie`; `Object.fromEntries` leaves only the cookie schema and duplicates the required key. `HttpMigrateRouteAccessor` resolves body/header/query collisions but omits cookies.
8. **Raw OpenAPI 3.0/3.1/3.2 duplicate parameters bypass diagnostics.** Each upgrader deduplicates `in:name` entries before the composer sees them. A path with two `id` declarations therefore composes with the last schema and no error, while the new duplicate test only covers an already-emended document.
9. **An optional object parameter promotes its member requirements to the whole group.** Combining optional object `dto` (with required member `nested`) and required primitive `limit` synthesizes `required: ["limit", "nested"]`; callers cannot omit the optional object as OpenAPI permits.
10. **`HttpError` changed an existing public five-argument behavior.** `new HttpError(..., '{"x":1}').toJSON()` returned `{x:1}` before this PR but now returns the string. The fetcher needs to preserve already-parsed text bodies, but manual construction should remain backward compatible.
11. **New route metadata is source-breaking for existing route-shaped values.** `cookies`, `style`, `explode`, `required`, and serialization arrays were added as mandatory interface fields. They can remain guaranteed on newly composed routes while being optional in the public structural type with legacy execution fallbacks.
12. **Public controller documentation is stale.** `IHttpLlmController` still documents only path/query/body inputs and omits the new header/cookie groups.

## Remediation plan

- Classify URL-shaped strings only from their leading URL syntax or an unambiguous relative prefix; preserve raw query delimiters as data.
- Accept `unknown` JSON/text bodies, retain object-only validation for form and multipart encodings, and delete inherited content type before sending `FormData`.
- Resolve parameter references before shape classification; support object delimited serialization, enforce defined style/explode combinations, preserve `additionalProperties`, and keep optional object members optional when their parameter is absent.
- Recognize normalized `application/json`, `application/*+json`, and wildcard JSON media consistently and send the declared JSON-suffix content type.
- Include cookies in accessor collision resolution and preserve invalid same-list duplicates through all three OpenAPI upgraders so the composer can diagnose them.
- Preserve legacy `HttpError` string parsing while marking fetcher-provided bodies as already parsed; make additive route metadata structurally backward compatible with legacy defaults.
- Add focused regressions for each boundary, update the public comments, run package-scoped builds/tests, then commit/push with the exact-SHA CI cancellation gate. After remediation, restart a fresh full-surface Self-Review round.

samchon:
Round 1 findings are remediated in `ea213af2f9ccf12417992607c97e9137e365abef`.

## Resolution evidence

- Raw query classification now uses unambiguous leading URL syntax, so raw path/absolute-URL values and raw `#` data survive while absolute, root-relative, and relative URLs still expose only their query.
- `HttpMigration.IFetchProps.body` accepts `unknown`; JSON/text scalar, array, and `null` payloads pass, form/multipart remain object-only, and multipart deletes inherited `Content-Type` so fetch can supply its boundary.
- Parameter references are resolved before shape checks. Space/pipe object values flatten key/value pairs, open-object entries are retained without duplicating neighboring primitives, and undefined explode combinations receive deterministic errors.
- JSON-suffix media types compose consistently for request, success, and exception schemas; outgoing requests preserve the declared subtype such as `application/merge-patch+json`.
- Cookie groups participate in accessor collision resolution, and direct execution plus LLM schemas now keep colliding path/cookie arguments distinct.
- OpenAPI 3.0, 3.1, and 3.2 upgraders retain invalid same-list duplicates long enough for the composer to diagnose them, while valid operation-level overrides still replace path-level parameters.
- Optional object member requirements are promoted only when the object parameter itself is required; referenced arrays/primitives remain in generated LLM schemas.
- `HttpError` distinguishes fetcher-classified bodies from legacy manual string construction, preserving both text media and the prior lazy JSON parsing behavior.
- New route metadata remains guaranteed on newly composed routes but optional in the public structural interface, with execution defaults for legacy route-shaped values.
- `IHttpLlmController` now documents path, header, cookie, query, and body groups.

## Verification

- PASS: `pnpm --filter @typia/interface build`
- PASS: `pnpm --filter @typia/utils build`
- PASS: `pnpm --filter typia build`
- PASS: focused six-test HTTP migration regression set
- PASS: focused raw-query direct/factory regression
- PASS: full `@typia/test-utils` suite
- PASS: full `@typia/test-typia-schema` suite
- PASS: targeted Prettier check for all 19 remediation files
- PASS: `git diff --check`
- Exact-SHA gate: all six runs for `ea213af2f9ccf12417992607c97e9137e365abef` are cancelled.

The website build remains skipped under the campaign exclusion and the previously recorded dirty isolated `ttsc` prerequisite. I am restarting a fresh full-surface Solo Self-Review from this new HEAD; the PR remains draft and unmerged.

samchon:
## Solo Self-Review round 2: findings and remediation plan

A complete re-read of the current `ea213af2f9ccf12417992607c97e9137e365abef` diff, issue acceptances (#2029, #2030, #2031, #2050), and the interface → converter → composer → accessor → fetcher/runtime → LLM/adapters consequence surface found five additional contract defects. These are reproduced on the current pushed HEAD; no remediation edits have been made yet.

1. **Embedded OpenAPI path templates are rejected and cannot be fetched.** `/files/{name}.{ext}` and `/prefix-{id}` are valid template expressions, but the composer discovers placeholders only when an entire slash segment starts with `{` and ends with `}`. The endpoint normalizer and fetcher likewise convert/replace only whole-segment parameters. The first route reports zero routes with unmatched names, while the second also yields an unsafe accessor namespace if conversion is forced.
   - Remediation: discover every `{name}` expression in source order, replace every expression when composing the emended path, derive accessor namespaces from the static portions of the original path, and interpolate the original `{name}` expressions at fetch time so embedded and repeated placeholders preserve style/explode serialization.
   - Regression: compose and execute embedded/repeated placeholders, verify schemas remain name-associated, the captured path is exact, and accessors contain valid identifiers.

2. **Non-JSON media-type matching is incorrectly case-sensitive.** `Text/Plain`, `Application/X-WWW-Form-Urlencoded`, and `Multipart/Form-Data` all produce unsupported-body composition errors, although media type tokens are case-insensitive. JSON already goes through the normalized matcher.
   - Remediation: normalize every request/response content key once before matching text, form, multipart, and JSON; retain canonical lower-case route media types.
   - Regression: compose and execute mixed-case text, form, and multipart request media types and verify canonical `Content-Type` behavior.

3. **A flattened optional object parameter loses its internal required fields when combined with primitive parameters.** For an optional query DTO `{ nested: string; other?: string }` beside required `limit`, the advertised LLM validator correctly accepts `{ limit: 1 }` but also incorrectly accepts `{ limit: 1, other: x }`. Supplying any DTO-owned field means the optional object is present, so its `nested` requirement must apply.
   - Remediation: represent the flattened conditional with exclusive absent/present object-schema branches: the absent branch permits only sibling primitive fields, while the present branch restores the DTO's required set and additional-property contract.
   - Regression: require omission to pass, partial DTO input to fail, and complete DTO input to pass through the LLM validator.

4. **Required object parameters whose schemas permit `{}` are mistaken for missing parameters.** A required `deepObject` query parameter with only optional properties composes successfully, but direct execution with `query: {}` throws `query filter is required`. The shared extraction helper uses “has serialized entries” as presence, conflating an empty valid object with absence across grouped parameter locations.
   - Remediation: retain the object schema's required-property metadata and treat an empty required object as present only when its schema permits an empty object; keep rejecting empty values when required members are actually absent.
   - Regression: execute a required empty-valid object successfully and preserve the missing-required-member rejection.

5. **The new `http.query` documentation contradicts the raw-query runtime contract.** The guide says every literal `#` starts a fragment, while the parser and regression intentionally preserve `fragment=a#b` when the entire input is raw query text. Fragment stripping applies only after classifying a question-prefixed or URL-shaped input.
   - Remediation: distinguish raw text from URL-shaped/question-prefixed input in the guide and retain the percent-encoding note.

Planned verification after remediation: interface/utils/typia builds; focused migration, LLM, raw-query, and generated-schema tests; full `@typia/test-utils` and `@typia/test-typia-schema` suites; targeted Prettier and `git diff --check`; Lore commit/push; exact pushed-SHA check gate. Per campaign policy, the website build remains excluded in this wave, while the edited MDX will be formatting-checked. After the remediation is pushed, the full Solo Self-Review restarts from the new HEAD.

samchon:
## Solo Self-Review round 2: remediation evidence

All five round-2 findings from https://github.com/samchon/typia/pull/2060#issuecomment-4987944888 are remediated and pushed in `40ad03478eb27f5af6c083820f105ac770be9316`.

1. **Embedded path templates:** every `{name}` expression is discovered in source order, emended anywhere inside a segment, associated by name, and interpolated from the original template. Accessor namespaces retain only meaningful static path portions. Regressions cover `/files/{name}.{ext}` and `/prefix-{id}`, reversed declaration order, exact accessors, and captured paths.
2. **Case-insensitive media types:** text, urlencoded form, and multipart selection now use the same normalized token matching as JSON. Regressions cover mixed-case tokens plus parameters, canonical route/response types, outgoing headers/bodies, and multipart boundary delegation.
3. **Optional flattened object requirements:** the advertised schema has explicit absent/present branches. Because `equals:false` intentionally ignores additional properties, local validation additionally selects the present branch only when an object-owned key is supplied; it does not make unrelated function inputs exact. Direct and adapter-style validation now accept omission, reject partial presence, and accept complete presence.
4. **Valid required empty objects:** grouped serialization retains object required-property metadata. `{}` is treated as present only when a required object's schema permits it; an object with a missing required member still fails before fetch.
5. **Raw-query guide:** the guide now limits literal-`#` fragment stripping to question-prefixed and URL-shaped inputs and explicitly documents preservation in raw query text.

Verification:

- `pnpm --filter @typia/interface build` — pass
- `pnpm --filter @typia/utils build` — pass
- `pnpm --filter typia build` — pass
- focused HTTP/LLM migration suite (application, body media, cookies, serialization edges, path names, request, response) — pass
- focused generated raw-query suite — pass
- full `pnpm --filter @typia/test-utils start` — pass
- full `pnpm --filter @typia/test-typia-schema start` — pass
- targeted TypeScript Prettier — pass
- `git diff --check` — pass

The campaign wave still excludes the website build. Whole-file Prettier for `misc.mdx` retains that file's pre-existing baseline mismatch; the remediation changes exactly one existing prose line and passes diff whitespace checks.

Exact pushed-SHA gate for `40ad03478eb27f5af6c083820f105ac770be9316`: build, test, nestia, website, and experiments were cancelled per campaign policy; spell-check completed successfully before its queued cancellation retry could take effect. The worktree is clean. A full Solo Self-Review now restarts from this new HEAD.

samchon:
## Solo Self-Review round 3 ? findings and remediation plan

I re-reviewed the complete 31-file `origin/master...40ad03478eb27f5af6c083820f105ac770be9316` surface, including interface ? converter ? composer ? accessor ? migration/LLM fetchers ? direct typia runtime ? LangChain/MCP/Vercel adapters and all campaign tests/docs. The worktree is clean and no round-3 edits have been made.

### Confirmed findings

1. **Raw key-only query strings are still discarded.**  
   `_httpQueryParseURLSearchParams` classifies every non-empty string without `?`, `=`, or `&` as a queryless relative URL. Consequently ordinary raw queries such as `flag`, `a%3Fb`, and `flag#tail` become an empty `URLSearchParams` instead of the platform-equivalent empty-valued key. The prior regression covered raw strings containing `=`/`&`, so this branch remained unprotected across all eight direct/factory query decoders.

2. **The cookie serialization contract is incomplete for OpenAPI 3.2 and special-character names.**  
   OpenAPI 3.2 adds `style: "cookie"` for `in: "cookie"`: it uses `; ` delimiters and requires values to pass through without percent-encoding. The new public style unions omit it in both normalized `OpenApi` and `OpenApiV3_2`; the composer rejects it, defaults its `explode` incorrectly, and the fetcher has no style-dependent pass-through path. Separately, `style: "form"` percent-encodes values but leaves parameter names and exploded object keys raw, producing invalid wire data such as `Cookie: session id=a%20b; theme mode=dark`. The OAS 3.2 rules are documented in [?4.12.2?4.12.3](https://spec.openapis.org/oas/latest.html#parameter-object) and [Appendix D.1](https://spec.openapis.org/oas/latest.html#appendix-serialization).

3. **The new path serializer is not RFC 3986/OAS percent-encoding complete.**  
   `serializePath` uses `encodeURIComponent`, which deliberately leaves `!'()*` unescaped. OAS RFC6570-style path expansion permits only the unreserved set for ordinary expansion; for example `Hello World!` must become `Hello%20World%21`. Scalar, array, object-key, and object-value path serialization therefore leak reserved characters introduced through this PR's new style-aware path logic.

### Remediation plan

- Split no-`?` URL recognition into explicit URL prefixes only, keeping ambiguous bare strings as raw query text, and add key-only/percent-encoded-key/fragment-bearing coverage across all eight query decoders.
- Add `cookie` only to the normalized and OpenAPI 3.2 style contracts plus route serialization metadata; accept it in the composer with the correct default `explode: true`.
- Use one RFC 3986 encoder for path expansion and `form` cookie names/keys/values, while preserving `cookie` style data byte-for-byte; retain `; ` cookie-pair composition and inherited-cookie replacement by the actual serialized name.
- Add focused migration/LLM regressions for OpenAPI 3.2 cookie pass-through, form cookie special-character names/keys/values, and scalar/array/object path encoding.
- Rebuild `@typia/interface`, `@typia/utils`, and `typia`; run focused HTTP/LLM/raw-query tests, both full campaign suites, format/diff checks, then create a Lore commit, push, gate the exact SHA, and restart the full Solo Self-Review.

samchon:
## Solo Self-Review round 3 ? converter consequence addendum

A downstream consequence check after the round-3 patch exposed one additional pre-existing-in-this-PR finding. No Swagger 2.0 downgrader edit has been made yet.

4. **Swagger 2.0 downgrade leaks the newly retained OpenAPI 3 serialization fields.**  
   This PR adds `style` and `explode` to normalized parameter metadata. `SwaggerV2Downgrader.downgradeParameter` spreads every non-schema field into the Swagger 2.0 parameter, so downgraded documents now contain unsupported `style`/`explode` keys. The OpenAPI 3.0/3.1 downgrader consequence is version-representable (with OpenAPI 3.2 `cookie` mapped to the older `form` style), but Swagger 2.0 has no corresponding fields and must omit them.

### Addendum remediation

- Destructure `style` and `explode` out of the Swagger 2.0 parameter conversion rather than emitting invalid extension-like fields.
- Extend the OpenAPI 3.2 cookie regression to downgrade the same normalized document to 3.1/3.0 (expect `cookie ? form`) and 2.0 (expect both fields absent).
- Include the focused converter assertion in the existing build/full-suite gate.

samchon:
## Solo Self-Review round 3 ? path-name liveness addendum

The final remediation-diff review found one more path-name defect. No fix for this finding has been made yet.

5. **Non-identifier path parameter names make route composition loop forever.**  
   The new collision loop starts from `EndpointUtil.normalize(parameter.name)` and keeps prefixing `_` while `NamingConvention.variable(key)` is false. Names containing characters that normalization does not remove (for example `coords!`, `user id`, or emoji) remain invalid regardless of how many underscores are prepended, so `HttpMigration.application()` never returns. Matrix serialization would also emit the original parameter name without RFC 3986 encoding if composition could complete.

### Addendum remediation

- Replace non-identifier characters in the generated argument key before the collision loop, preserving the original OpenAPI name separately for template matching and wire serialization.
- RFC 3986-encode the matrix parameter name wherever that style writes `name=value`.
- Add a focused special-name path regression that proves composition terminates, the generated argument key is a valid unique identifier, and the matrix wire path encodes both name and value.

samchon:
Round 3 remediation is complete at `9382f056c07089dec36f342b9d93ada2520d019c`.

Resolved:
- kept key-only raw query strings (`flag`, percent-escaped names, fragments) out of the queryless-URL path
- added OpenAPI 3.2 `style: cookie` composition and pass-through serialization, with `form` cookie RFC 3986 encoding and correct explode defaults
- encoded every RFC 3986 reserved character in path parameter names, keys, and values
- mapped 3.2 cookie style to `form` for OpenAPI 3.0/3.1 and stripped unsupported `style`/`explode` fields from Swagger 2.0 parameters and headers
- separated original path names from sanitized generated argument keys, preventing the invalid-name composition loop

Verification:
- `pnpm --filter @typia/interface build`
- `pnpm --filter typia build`
- `pnpm --filter @typia/utils build`
- focused raw-query, cookie-style, and path-serialization tests
- full `pnpm --filter @typia/test-utils start`
- full `pnpm --filter @typia/test-typia-schema start`
- targeted Prettier and `git diff --check`

Exact-SHA gate: PR head equals the commit above; spell-check and both Socket checks passed. The `build`, `test`, `nestia`, `website`, and `experiments` workflows reached terminal `cancelled` state under the issue-campaign per-push CI policy. The whole-file MDX Prettier baseline remains pre-existing; the edited paragraph itself matches formatter output.

Restarting the full solo review from this head now.

samchon:
## Solo Self-Review round 4: findings and remediation plan

I re-read the complete `origin/master...9382f056c07089dec36f342b9d93ada2520d019c` surface and independently checked query serialization, OpenAPI 3.2 `querystring`, JSON Schema object defaults, and package release history. No round-4 remediation edits have been made yet.

### Confirmed findings

1. **Query styles are re-encoded as HTML form data instead of their OAS/RFC 3986 wire forms.**
   `getQueryPath()` feeds already-serialized style values back through `URLSearchParams`. A current reproduction emits `csv=a%21%2Cb*&pipes=a%21%7Cb*&spaces=a%21+b*&filter%5Br%21%5D=v%7E+%21`: structural comma/pipe/bracket delimiters are escaped, spaces become `+` instead of `%20`, RFC 3986-unreserved `~` is escaped, and reserved `*` is left raw. OAS 3.1's style table requires literal comma/pipe/bracket syntax and `%20` space delimiters: https://spec.openapis.org/oas/v3.1.0.html#style-examples

2. **A valid OpenAPI 3.2 `in: querystring` parameter crashes before composition.**
   The name check itself is correct: OAS 3.2 still requires `name`, although it is not used on the wire. The shape and grouping are not correct. `querystring` MUST use `content`, represents the entire query string, may occur only once, and cannot coexist with `in: query`; `schema`/`style` fields must not be used. The raw and normalized interfaces instead require `schema`, the upgrader calls `convertSchema(undefined)`, and a valid `application/x-www-form-urlencoded` content parameter throws `TypeError: Cannot read properties of undefined (reading 'type')`. Current grouping would also serialize it as an ordinary named query parameter. See https://spec.openapis.org/oas/latest.html#parameter-locations and https://spec.openapis.org/oas/latest.html#parameter-object

3. **Omitted `additionalProperties` silently drops accepted object-parameter entries.**
   OpenAPI 3.1 Schema Objects inherit JSON Schema 2020-12 semantics, where omitting `additionalProperties` behaves as an empty schema and therefore permits unmatched properties. The composer records omission as `additionalProperties: false`; the default LLM validator accepts `{ query: { known: "x", extra: "y" } }`, but direct execution emits only `?known=x`. This affects flattened query, header, and cookie object parameters and also prevents optional-object presence checks from restoring required fields when only an undeclared-but-allowed key is supplied. Sources: https://spec.openapis.org/oas/v3.1.0.html#schema-object and https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#section-10.3.2.3

4. **This development PR pre-bumps three ordinary npm package versions.**
   `packages/interface`, `packages/utils`, and `packages/typia` were changed from `13.1.1` to `13.1.2`. Package history shows ordinary version changes are owned by release commits; the development contract only requires an in-PR `typia` bump when shipped Go source under `packages/typia/native` changes. This PR changes TypeScript/public contracts but no native Go source, so all three bumps are release-state drift and must be removed.

### Remediation plan

- Build style-aware query pairs directly with the shared RFC 3986 encoder, leaving only each style's structural delimiters literal; retain `URLSearchParams` only for true `application/x-www-form-urlencoded` content and legacy route metadata.
- Model OpenAPI 3.2 content-backed `querystring` parameters, normalize their media schema without crashing, enforce single/no-mixed-query cardinality, carry whole-query media metadata into the route, and serialize supported form/JSON/text content as the entire query string. Unsupported media/shape combinations will produce deterministic composition errors rather than runtime corruption.
- Treat every object schema whose `additionalProperties` is not explicitly `false` as open for ownership/serialization, while continuing to exclude sibling parameter keys; cover both wire preservation and optional-object required-field selection.
- Restore the three package versions to `13.1.1`.
- Add RED regressions first for exact query wire text, valid content-backed querystring plus mixed/duplicate diagnostics, and implicit-open object parity; then implement, build the affected packages, run focused and full campaign suites, format/diff checks, Lore commit/push, and exact-SHA gate. A new full Self-Review will restart from the pushed head; the PR remains draft.

samchon:
Round 4 verification follow-up finding: the new OpenAPI 3.2 content-backed querystring branch rejected existing canonical emended documents that use the previously public schema-backed querystring representation. The focused HTTP suite reproduced this in test_http_migrate_empty_required: the route was omitted, and the downstream query assertion dereferenced no route. Remediation: apply whole-query media semantics when content is present, while retaining the established schema/style serialization path when canonical input has no content. The query/querystring coexistence and duplicate guards remain unchanged. A focused legacy regression will be green before the next commit.

samchon:
Round 4 remediation is complete at `cd0b57e863e0bcbe082cbc73eceafb47b9044e74`.

Resolved:
- styled query serialization now emits RFC 3986-encoded names and values without re-encoding OpenAPI's comma, pipe, bracket, or `%20` structural delimiters
- OpenAPI 3.2 content-backed `querystring` parameters compose as one whole-query input, carry their normalized media type, reject duplicate/mixed/invalid declarations deterministically, and execute default form, JSON, and scalar text representations
- canonical schema-backed `querystring` routes keep their established parameter serialization behavior
- object parameters whose schemas omit `additionalProperties` retain unmatched keys and reapply internal requirements when an optional object is present
- the three development-only package version bumps were restored to `13.1.1`

Verification:
- PASS: `pnpm --filter @typia/interface build`
- PASS: `pnpm --filter @typia/utils build`
- PASS: `pnpm --filter typia build`
- PASS: focused `http_migrate_querystring_contract`, `http_migrate_request_contract`, `http_migrate_parameter_serialization_edges`, and `http_migrate_empty_required` regressions
- PASS: full `pnpm --filter @typia/test-utils start`
- PASS: full `pnpm --filter @typia/test-typia-schema start`
- PASS: targeted Prettier check for all changed TypeScript files and `git diff --check`

The exact-SHA cancellation gate settled after four polls. Runs `29474871218`, `29474871246`, `29474871248`, `29474871251`, `29474871270`, and `29474871297` all finished `cancelled` for this SHA. The website production build remains excluded under the previously recorded campaign packaging prerequisites.

Restarting a complete Solo Self-Review from this new head. The PR remains draft and unmerged.

samchon:
## Solo Self-Review round 5: findings and remediation plan

After rebasing onto current `master`, I re-read the complete 33-file `origin/master...fe71da7cd231f08067834da3874d8a882dc83b92` surface and traced the interface, OpenAPI conversion, route composition, accessor, direct/LLM fetch, adapter, tests, documentation, packaging, and platform consequences from scratch. No round-5 remediation edits have been made.

### Confirmed findings

1. **`pipeDelimited` and `deepObject` query delimiters are left as illegal literal URI characters.** `serializeQuery()` emits a literal `|` for `pipeDelimited` and literal `[`/`]` for `deepObject`. The OpenAPI 3.2 style table and percent-encoding guidance require these structural delimiters themselves to be represented as `%7C`, `%5B`, and `%5D`; only comma delimiters remain literal. The current exact-wire regression locks the non-conforming output instead of the OAS oracle.

2. **Reserved OpenAPI header parameters become required caller arguments instead of being ignored.** The newly composed header group includes `Accept`, `Content-Type`, and `Authorization` parameters. OpenAPI requires parameter definitions with those header names to be ignored, while connection headers and request-body media selection continue to own their actual wire values. A required reserved declaration therefore makes the LLM/direct argument contract stricter than the source operation and can compete with the body-selected content type. Header names are case-insensitive, so casing variants have the same defect.

### Remediation plan

- Add RED exact-wire coverage requiring `%7C`, `%5B`, and `%5D` while retaining `%20` and comma behavior.
- Add RED mixed-case reserved-header declarations to the request contract and prove they create no schema requirements, while connection-level authorization and body-selected content type still reach the wire.
- Filter the three reserved header names case-insensitively during group composition, without deleting ordinary header parameters from the normalized source operation.
- Emit the OAS-required encoded query delimiters and update the changed guide sentence to state that structural syntax is preserved with URI-required delimiter encoding.
- Run focused request/serialization tests, all three affected package builds, both full affected workspaces, targeted TypeScript formatting checks, and `git diff --check`; then create a Lore commit, push, run the exact-SHA cancellation gate, and restart a complete solo review from the new head.

samchon:
## Solo Self-Review round 5: remediation evidence

Both round-5 findings are remediated in `f3460ebdbd4ea193190661791e0cfe0b190ea35e`.

- `pipeDelimited` now writes `%7C`, while `deepObject` writes `%5B` and `%5D`; comma and `%20` behavior remains style-correct. The exact request-wire regression now uses the OpenAPI serialization table as its oracle.
- `Accept`, `Content-Type`, and `Authorization` parameter definitions are ignored case-insensitively during header-group composition. Ordinary operation headers remain argument-owned, connection authorization remains on the wire, and request-body serialization retains final `Content-Type` ownership.
- The changed `HttpLlm` guide now distinguishes structural syntax from the URI encoding required for its delimiters.

Verification:

- PASS: RED first — focused request regression failed because reserved headers were required before the fix.
- PASS: focused `http_migrate_request_contract`, `http_migrate_parameter_serialization_edges`, and `http_migrate_querystring_contract` regressions after the fix and after targeted formatting.
- PASS: `pnpm --filter @typia/interface build`.
- PASS: `pnpm --filter @typia/utils build`.
- PASS: `pnpm --filter typia build`.
- PASS: full `pnpm --filter @typia/test-utils start`.
- PASS: full `pnpm --filter @typia/test-typia-schema start`.
- PASS: targeted Prettier check for the three changed TypeScript files and `git diff --check`.

The website production build remains excluded under the campaign packaging prerequisites already recorded in the thread. Exact-SHA gate for this push required seven polls; runs `29476738958`, `29476738963`, `29476738971`, `29476738973`, `29476738978`, and `29476739031` all finished `cancelled`.

A fresh complete solo review restarts from this head. The PR remains draft and unmerged until that round is clean.

samchon:
## Self-Review round 6 — clean

Reviewed PR #2060 at exact head `e6c959bce9c5ecb6f263139ac33abab6efc4b04d` after rebasing onto current `master` `72146543191fc741156e2777f32c2f195f716a44`.

The fresh solo pass re-read the complete 33-file base-to-head surface, traced the route model through OpenAPI conversion, migration composition, LLM validation, request serialization, response classification, runtime query parsing, tests, and documentation, and rechecked the relevant OpenAPI 3.2 parameter and serialization rules. No sound remaining finding was identified.

Verification on the rebased head:

- `pnpm --filter @typia/interface build`
- `pnpm --filter @typia/utils build`
- `pnpm --filter typia build`
- `pnpm --filter @typia/test-utils start`
- `pnpm --filter @typia/test-typia-schema start`
- `git diff --check origin/master...HEAD`

All commands above passed. A supplemental `pnpm exec prettier --check` reported the same formatting warning for both changed files and unchanged baseline files in this Windows checkout, so it was not treated as a change-specific failure; the worktree remained clean. The website production build remains intentionally excluded under the campaign's previously recorded packaging prerequisite.

PR #2062, opened 2026-07-16
fix(utils): preserve Swagger and LLM converter semantics

### LINKED ISSUE #2056: Strict LLM conversion silently drops numeric defaults
## Problem

Strict LLM schema conversion deletes numeric/integer `default` values without preserving them in the description, while the equivalent string conversion retains `@default`. Consumers receive a schema with neither the machine-readable default nor the strict-mode textual fallback.

## Evidence

`OpenApiConstraintShifter.shiftNumeric` gathers numeric constraint tags and immediately writes `schema.description`. Only afterward it appends `@default ...` to the local tag array and deletes `schema.default`, so the new tag is never serialized. `shiftString` appends/deletes default before writing its description.

Public current-build reproduction:

```text
strict number { default: 7 }  -> { type: "number" }
strict integer { default: 7 } -> { type: "integer" }
strict string { default: "x" } -> { type: "string", description: "@default x" }
```

Closed unmerged PR #1864 contains the ordering correction and explicitly records the dropped default, but there is no operative issue or merged implementation.

## Consequence surface

Number/integer schemas, zero/negative/fraction/scientific defaults, existing descriptions and other shifted constraints, nested properties/arrays/unions/references, strict LLM models, OpenAPI-to-LLM conversion, MCP/LangChain/Vercel tool descriptions, and round-trip/inversion behavior.

## Approach

Build the complete tag set before writing the description, using the same ordering and escaping policy across numeric and string shifters. Keep default deletion only when its information has been preserved in the strict target representation.

- Numeric and integer defaults survive strict conversion as deterministic `@default` description tags.
- Existing descriptions and minimum/maximum/multipleOf tags retain ordering/content; falsy defaults such as 0 are covered.
- Nested/reference/union schemas and all supported strict LLM targets agree with string behavior.
- Run focused converter/LLM/MCP adapter tests and the full suite.

## Coordination

Small utils converter issue; compatible with other schema-converter work but independent of validation semantics.

### LINKED ISSUE #2057: Swagger v2 downgrade erases constant schemas
## Problem

Swagger v2 downgrade discards scalar `const` schemas and constant-only `oneOf` unions. The output becomes an unconstrained `{}` instead of a typed Swagger `enum`, losing both type and allowed values.

## Evidence

`SwaggerV2Downgrader.downgradeSchema` defines `visitConstant`, but the constant/oneOf dispatch was accidentally placed inside its nested `insert` function. Because `visitConstant(input)` never calls `insert`, it is a no-op. The dormant `matched` lookup also compares a schema `type` to the value rather than `typeof value`, so simply moving the call without fixing grouping would still split same-type constants incorrectly.

Public reproduction downgrades `{ const: 3 }`, `{ const: "x" }`, and numeric/string constant-only `oneOf` schemas to `{}`. The tracked `test_json_schema_downgrade_v20_enum.ts` oracle expects a typed enum, directly contradicting current source behavior. Closed unmerged PR #1864 contains the complete control-flow/type-comparison correction; no operative duplicate exists, and #2024 concerns later OpenAPI discriminator preservation.

## Consequence surface

String/number/boolean/null constants, same-type and mixed-type constant unions, nullable schemas, constants mixed with ordinary branches/references, component and inline schemas, operation parameters/bodies/responses, discriminator output, and v2 consumers/code generators.

## Approach

Make constant collection an explicit sibling traversal of the ordinary schema visit. Group values by `typeof`/Swagger scalar type into deterministic deduplicated enums, then compose nullable/mixed unions without erasing attributes. Avoid recursive dispatch inside the insertion primitive.

- Scalar constants downgrade to the correct Swagger type plus one-value enum.
- Same-type constant unions combine into one enum; mixed types retain every representable branch.
- Boolean/null, duplicates, nullable, referenced, and mixed ordinary/constant cases are covered.
- The existing v2 enum oracle is active and green; focused converters plus full suites pass.

## Coordination

Swagger v2 converter ownership; independent of OpenAPI 3.1 tuple/schema conversion issues.

### LINKED ISSUE #2058: Swagger v2 conversion loses server and media metadata
## Problem

Swagger v2 upgrade/downgrade loses the document metadata that defines the actual server URL and request/response media types. Base paths and schemes disappear or are folded into `host`, while `consumes`/`produces` are ignored and operations are hardcoded to JSON.

## Evidence

- Upgrade maps `input.host` directly to `servers[0].url` and ignores `basePath`; `{ host: "api.example.com", basePath: "/v1" }` becomes `{ url: "api.example.com" }`.
- `SwaggerV2.IDocument` lacks the standard `schemes` field entirely, so the public representation cannot preserve or convert it.
- Downgrade strips a server scheme with `split("://").pop()` but assigns the entire remainder, including `/v1`, to `host`; it emits neither `basePath` nor schemes.
- Upgrade ignores document/operation `consumes` and `produces`, hardcoding request and response content to `application/json`. Downgrade takes a first content entry but does not restore the corresponding global/operation metadata.

The [OpenAPI 2.0 specification](https://spec.openapis.org/oas/v2.0.html) defines `basePath` relative to `host`, `schemes` as the transfer protocols, and `consumes`/`produces` as MIME-type contracts. The conversion behavior has existed since the converter introduction and exact open/closed issue/PR searches found no duplicate.

## Consequence surface

Swagger v2 interface typing, absolute/relative hosts, http/https and multiple schemes, ports, empty/slashed base paths, server variables/relative OpenAPI servers, global and operation-level consumes/produces overrides, body/form parameters, every response, multi-content OpenAPI documents, round trips, and HTTP migration/LLM consumers that execute converted routes.

## Approach

Represent every standard Swagger v2 document field first. Compose upgrade server URLs from scheme + host + basePath with explicit defaults, and parse downgrade servers into their distinct fields without mixing path into host. Convert global and operation media metadata with defined precedence; report unrepresentable multi-server/content cases rather than silently selecting data.

- Upgrade and downgrade tables cover scheme/host/port/basePath combinations and preserve normalized request base URLs.
- Global and per-operation consumes/produces produce correct request/response content maps, including non-JSON types.
- Round trips preserve every representable field; ambiguous/unrepresentable OpenAPI 3.x inputs follow a documented deterministic policy.
- Add public converter plus downstream HTTP migration tests and run focused utils/interface/full suites.

## Coordination

Same Swagger v2 converter family as the constant-schema issue, but document/media ownership is independently testable and may share a cohesive converter PR.

samchon:
## Overview

Reserve one cohesive implementation batch for strict LLM and Swagger v2 converter correctness:

- preserve numeric defaults when strict LLM schemas shift unsupported constraints into descriptions;
- preserve scalar constants and constant-only unions when downgrading to Swagger v2;
- preserve representable Swagger v2 server and request/response media metadata across upgrade and downgrade.

Closes #2056
Closes #2057
Closes #2058

## Scope

The implementation will remain within the shared interface and converter ownership surfaces, with focused regression coverage and downstream HTTP/LLM consequence checks. Unrelated validator, native-transform, and adapter behavior is deferred.

## Verification

Pending. This draft is an implementation-free campaign claim; focused converter/interface tests, downstream package checks, broader local verification, and solo Self-Review will be recorded before merge.

Repository Actions are deliberately cancelled by exact campaign commit SHA after every push and pull-request event. No workflow or repository Actions setting is disabled; local verification is the campaign gate.


samchon:
Implementation is complete at `566cfcd3c089169dd8548f3478b244706fe89e92`.

Summary:

- Preserve numeric `default` values in strict LLM descriptions by serializing the complete constraint-tag set before removing schema constraints.
- Downgrade OpenAPI scalar constants and constant unions into deterministic, deduplicated, type-correct Swagger 2 enums while retaining ordinary union branches.
- Convert Swagger 2 `schemes` + `host` + `basePath` into complete emended server URLs, honor operation-level `schemes` / `consumes` / `produces`, and preserve request/response media maps.
- Downgrade only exactly representable OpenAPI server/media shapes. The converter throws `TypeError` for server variables, different authorities or base paths, query/fragment or unsupported relative URLs, streaming item schemas, request/named examples, or media entries with different schemas/examples.
- Added matching public interface fields, regression matrices, and website behavior documentation.

Verification:

- `pnpm --filter ./packages/interface build` — passed.
- `pnpm --filter ./tests/test-interface start` — passed.
- `pnpm --filter ./packages/utils build` — passed.
- `pnpm --filter ./packages/utils exec ttsc --noEmit` — passed.
- `pnpm --filter ./tests/test-utils exec ttsc --noEmit` — passed.
- `pnpm --filter ./tests/test-utils start -- --include strict_numeric_default downgrade_v20_enum document_roundtrip_v20_server_media document_downgrade_v20_unrepresentable` — all four focused regressions passed.
- `pnpm --filter ./tests/test-utils start -- --include document_downgrade_v20_unrepresentable` — expanded representability matrix passed.
- Targeted Prettier check for all changed TypeScript files and `git diff --check` — passed.

Deliberate policy and skips:

- Swagger downgrade emits `consumes` / `produces` per operation rather than guessing shared document defaults.
- No dependency or package-version change was required; this patch does not modify Go/native sources, and releases are centrally versioned.
- The full repository suite and website production build were not run during the parallel campaign wave. Focused package builds, interface tests, converter regressions, and typechecks are recorded above.
- Per campaign policy, exact-SHA Actions runs for this implementation push were cancelled after local verification.

samchon:
## Solo Self-Review — round 1 findings and remediation plan

I reviewed the complete nine-file base-to-head diff from scratch against issues #2056, #2057, and #2058, then traced `OpenApiConverter` into `HttpMigration` and checked the changed Swagger behavior against the [OpenAPI 2.0 specification](https://spec.openapis.org/oas/v2.0.html). No remediation has been applied yet.

### Findings

1. **`formData` request payloads are still erased during Swagger upgrade.** Swagger 2 defines form parameters as payload fields when `consumes` is `application/x-www-form-urlencoded` or `multipart/form-data`. The upgrader currently emits them as emended parameters with the invalid location `in: "formData"` and emits no `requestBody`. `HttpMigration` consequently reports no error but produces a route with `body: null` and no parameters. The reverse path also emits a Swagger body parameter for form media instead of representable `formData` fields. This misses #2058's explicit body/form and downstream migration consequence surface.

2. **Empty operation media overrides do not clear document defaults.** OAS 2 explicitly permits empty operation `consumes` and `produces` arrays to clear the global definitions. `selectMediaTypes` tests `.length`, so an empty override falls back to the document list and recreates content entries that the source operation cleared.

3. **Standard Swagger response examples are not represented.** OAS 2 response examples are an `examples` object keyed by produced MIME type. `SwaggerV2.IOperation.IResponse` lacks that field, the upgrader ignores valid MIME-keyed examples, and the downgrader emits a non-standard singular `example`. This loses valid per-media examples and unnecessarily rejects responses whose examples differ by media type.

4. **The exact-representability policy has two false decisions.** The server downgrader accepts OpenAPI server schemes outside Swagger's `http`/`https`/`ws`/`wss` set and accepts credential-bearing authorities, producing invalid Swagger `schemes`/`host`; it also silently drops server descriptions. Conversely, response content maps with identical MIME sets in different object-key order are rejected even though the Swagger `produces` set is representable.

5. **The regression matrices do not yet prove all acceptance branches.** #2056 names nested, union, and referenced numeric defaults, while #2057 names nullable and referenced constants. The implementation paths appear general, but the current tests only pin direct numeric schemas and direct/union constants.

### Remediation plan

1. Extend the Swagger public types with the standard scheme union, standard response `examples`, and a valid file-capable `formData` parameter shape while retaining the legacy singular response example as an input compatibility alias.
2. Upgrade merged form parameters into an object request-body schema for the declared form media types; reject invalid body/form mixtures and missing/non-form media rather than emitting invalid emended parameters.
3. Downgrade exactly representable form request bodies back into `formData` parameters; reject mixed form/non-form media and non-simple form property schemas.
4. Preserve empty media overrides, map response examples by MIME type, compare response MIME sets order-independently, and reject unsupported/credential-bearing/described server alternatives.
5. Expand the existing regression files for form/downstream migration, empty overrides, response examples, server policy, order independence, and the missing numeric/constant acceptance branches.
6. Run focused converter and migration tests, interface/utils builds and typechecks, targeted static checks, then create a Lore follow-up commit, push it, and pass the exact-SHA cancellation gate. Because remediation changes the diff, start another complete solo review round from scratch afterward.

samchon:
Round 1 Self-Review remediation is now pushed in `1a535bde997d1f9f9be12db3117fc66c4681e940`.

Changes:

- upgrade Swagger `formData` fields into object-shaped emended request bodies, including multipart files, and preserve the payload through `HttpMigration`;
- downgrade representable form bodies back to `formData` parameters while rejecting mixed media, complex fields, URL-encoded files, metadata, encryption, and requiredness combinations Swagger cannot express exactly;
- preserve explicit empty operation `consumes` / `produces` overrides and compare response media sets independently of object-key order;
- preserve standard MIME-keyed Swagger response examples and retain legacy singular input compatibility without emitting the non-standard field;
- reject unsupported/credential-bearing/described servers instead of silently changing endpoint semantics;
- expand numeric-default and constant-enum coverage through nested, union, nullable, and referenced schemas;
- document the public Swagger conversion behavior and exact downgrade boundaries.

Verification:

- `@typia/interface`: `pnpm build` passed;
- `@typia/test-interface`: `pnpm start` passed;
- `@typia/utils`: `pnpm exec ttsc --noEmit` and `pnpm build` passed;
- `@typia/test-utils`: `pnpm exec ttsc --noEmit` passed;
- focused tests passed for strict numeric defaults, Swagger constants, server/media/examples, unrepresentable cases, and form-data migration;
- existing `document_upgrade_v20`, `document_downgrade_v20`, and `http_migrate_v20` tests passed;
- targeted Prettier check and `git diff --check` passed.

Skipped as campaign prerequisites: full repository suite and website production build.

Per-push gate: exact SHA `1a535bde997d1f9f9be12db3117fc66c4681e940`; Actions runs `29469175489`, `29469175490`, `29469175492`, `29469175504`, `29469175526`, and `29469175566` all reached `completed/cancelled`.

The required fresh full Self-Review round now restarts from this HEAD.

samchon:
Self-Review round 2 found additional sound improvements on the fresh base-to-`1a535bde997d1f9f9be12db3117fc66c4681e940` diff. I am recording them before changing code, per the review workflow.

1. `SwaggerV2Downgrader.downgradeComponents` inserts a synthetic `definitions.schemas = {}` entry before writing real definitions. Public component/document downgrade therefore emits an unintended schema named `schemas`; this directly affects the referenced-component acceptance surface of #2057.
2. Explicit empty operation `consumes` / `produces` lists combined with a body or response schema upgrade to empty content maps and silently erase the schema. A subsequent downgrade returns a body schema `{}` and no response schema. These schema-bearing empty-media combinations are not representable in the emended content model and must fail instead of claiming a round trip.
3. Form downgrade accepts `additionalProperties: false` on the aggregate object and then drops that constraint when it expands the object into Swagger `formData` parameters. Swagger parameters cannot express the aggregate closed-object constraint, so exact downgrade must reject it.
4. Constant grouping preserves attributes on a scalar/top-level union, but same-type constants with branch-local titles/descriptions/examples are merged into one enum and those branch attributes disappear. Attribute-bearing constants must remain distinct typed enum branches while ordinary unannotated constants may still be grouped and deduplicated.
5. Swagger upgrade accepts `host` / `basePath` values containing query strings or fragments. Depending on the shape, it emits a server URL whose query swallows the base path or even treats `https://` as part of a scheme-relative authority. The converter should reject those invalid Swagger endpoint fields, matching the downgrade-side exactness policy.

Remediation plan:

- remove the synthetic component entry and add a public `downgradeComponents(..., "2.0")` regression;
- reject schema/example-bearing empty media selections on upgrade and replace the current lossy round-trip oracle with explicit failure coverage;
- reject every defined aggregate `additionalProperties` constraint for form downgrade;
- retain branch-local constant attributes by bypassing grouping for attributed union branches;
- validate `host` and `basePath` query/fragment boundaries before composing server URLs;
- update the Swagger conversion guide, run interface/utils builds and focused converter/HTTP migration tests, then push with the exact-SHA Actions cancellation gate;
- restart the full Self-Review from the resulting HEAD.

samchon:
## Self-Review Round 2 remediation

Commit `67d3833ed790f453bcb441275f9838b26a245a78` resolves every finding reported in Round 2:

- removed the synthetic nested `definitions.schemas` entry from Swagger 2 component conversion;
- reject empty `consumes`/`produces` lists when a body or response payload would otherwise be discarded;
- reject aggregate `additionalProperties` constraints when lowering object request bodies to `formData` parameters;
- preserve supported attributes on individual constant union branches instead of collapsing attributed branches into one enum;
- reject query or fragment data embedded in Swagger `host` or `basePath` fields;
- documented the strengthened conversion boundaries.

Verification completed before push:

- `packages/utils`: `pnpm exec ttsc --noEmit`
- `packages/utils`: `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused regression set: numeric defaults, enum/component downgrade, server/media conversion, unrepresentable downgrade, and `formData` round trip
- existing Swagger v2 regression set: document upgrade, document downgrade, and HTTP migration
- targeted Prettier check for all TypeScript files in the PR
- `git diff --check`

All local checks passed. The exact-SHA cancellation gate then completed for the six workflows created by the push: run `29469828317` completed successfully before cancellation, and runs `29469828307`, `29469828329`, `29469828331`, `29469828264`, and `29469828282` were cancelled.

Round 3 restarts the complete base-to-head Self-Review from scratch; this is not yet a clean-round declaration.

samchon:
## Self-Review Round 3 findings and remediation plan

Round 3 restarted the complete review at `67d3833ed790f453bcb441275f9838b26a245a78`: all three claimed issues, all ten current files, every base-to-head hunk, and the public OpenAPI, HTTP migration, and strict LLM consequence paths were inspected again. The round is not clean.

### Findings

1. **Canonical constant branches lose their individual attributes on re-upgrade.** Round 2 correctly preserves descriptions and other supported attributes when OpenAPI constant branches are lowered to Swagger `x-oneOf` enum branches. `SwaggerV2Upgrader.convertSchema`, however, expands each nested enum to bare `{ const }` schemas and derives attributes only from the outer input. A downgrade/re-upgrade reproduction therefore produces canonical `oneOf: [{ const: "a" }, { const: "b" }]` with both descriptions missing, while also retaining the old Swagger `x-oneOf` as a stale vendor extension. Consumers follow the canonical `oneOf`, so the metadata remains effectively lost.

2. **Equivalent root server URLs are rejected as different base paths.** `SwaggerV2Downgrader.normalizeBasePath` represents an absent URL path as `undefined` but a trailing root slash as `"/"`. Consequently, the fully representable pair `https://api.example.com` and `http://api.example.com/` throws `document servers must share one host and basePath` instead of lowering to one host with two schemes. The same mismatch can reject an operation server whose only difference from the document server is root-slash spelling. Swagger 2 defines an omitted `basePath` as serving directly under the host, which is URL-equivalent to `/` here.

### Remediation plan

- Extract Swagger schema attributes per visited branch during upgrade, attach supported attributes to nested enum-derived constants, and omit the structural `x-anyOf`/`x-oneOf`/`x-nullable` keys once they have been converted to canonical emended fields.
- Canonicalize a root-only server path to the same absent base-path representation used by a URL with no path before comparing and emitting Swagger server fields.
- Extend the focused enum and server/media tests with exact downgrade/re-upgrade branch metadata and mixed root-slash multi-scheme/operation-server cases.
- Document typed-enum branch metadata preservation and root URL normalization, then rerun type checks, builds, focused regressions, targeted formatting, and diff checks before a Lore commit and exact-SHA push cancellation.

After remediation, the next review round will again restart from the base rather than assuming these localized fixes make the PR clean.

samchon:
## Self-Review Round 3 remediation

Commit `4ae44c44c213c3565a3f4a70a798169d4722429d` resolves both Round 3 findings:

- nested Swagger enum branches now carry their supported attributes into canonical emended constant branches, while converted `x-anyOf`, `x-oneOf`, and `x-nullable` structural keys no longer remain as stale extensions;
- root-only server paths normalize consistently, so URLs with and without a trailing root slash can lower to one Swagger host/base path and multiple schemes;
- focused round-trip tests cover both canonical branch metadata and document/operation root-server equivalence;
- the public conversion guide records both guarantees.

The new regressions were first run against the previous implementation and failed with the expected missing branch descriptions/stale `x-oneOf` and root-server `TypeError`. After remediation, verification passed:

- `packages/utils`: `pnpm exec ttsc --noEmit`
- `packages/utils`: `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused five-test converter set, including both new regressions
- existing Swagger v2 document upgrade, document downgrade, and HTTP migration regressions
- targeted Prettier check across all TypeScript files in the PR
- `git diff --check`

The exact-SHA cancellation gate completed for all six workflows created by the push: `29470471534`, `29470471510`, `29470471493`, `29470471413`, `29470471443`, and `29470471452` are cancelled.

Round 4 now restarts the full base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Self-Review Round 4 finding and remediation plan

Round 4 restarted the complete review at `4ae44c44c213c3565a3f4a70a798169d4722429d`: all three claimed issues, all ten current files, every base-to-head hunk, and the public OpenAPI, HTTP migration, and strict LLM consequence paths were inspected again. The round is not clean.

### Finding

**Swagger downgrade can generate a body parameter that its own upgrader rejects.** An emended request body with an explicit empty `content` map reaches `downgradeMedia` as `types: []`. `downgradeRequestBody` nevertheless emits both `consumes: []` and a Swagger body parameter with schema `{}`. Round 2 intentionally made `SwaggerV2Upgrader` reject exactly that schema-bearing empty-media combination because the emended content model has no media key under which to retain the payload. The public reproduction therefore succeeds during downgrade but immediately throws `SwaggerV2Upgrader: body parameters require a consumed media type` when the generated document is upgraded. A request body whose `content` is absent has the related silent problem that downgrade invents the default `application/json` representation on re-upgrade. Swagger cannot encode either request body exactly without at least one consumed media type, so downgrade must reject both shapes instead of generating an internally invalid or media-inventing document.

### Remediation plan

- Reject request bodies that do not contain at least one defined media entry before emitting any Swagger parameter.
- Extend the existing unrepresentable-downgrade matrix with both absent and explicitly empty request-body content, first confirming the regressions fail on the current implementation.
- Update the public conversion guide to record the non-empty request-media requirement.
- Run utils and test-utils type checks/builds, focused converter regressions, existing Swagger v2 upgrade/downgrade/migration tests, targeted formatting, and `git diff --check`.
- Create a Lore follow-up commit, push it, pass the exact-SHA Actions cancellation gate, and restart the complete Self-Review from the resulting HEAD.

samchon:
## Self-Review Round 4 addendum: negative tests are false positives

While adding the fail-first cases from the Round 4 finding, the focused test unexpectedly stayed green even though the public runtime reproduction still downgraded both invalid request-body shapes without throwing. The cause is in the installed `@nestia/e2e` test helper: synchronous `TestValidator.error` throws its own “exception must be thrown” assertion inside the same `try` whose `catch` treats every exception as success. A direct probe confirms `TestValidator.error("must throw", () => 42)` returns normally.

This means the negative assertions added or expanded by this PR in the Swagger server/media, form-data, and unrepresentable matrices can all report green when the converter stops rejecting an invalid shape. The positive assertions remain effective, but the claimed exact-representability coverage is not currently material.

Before product remediation I will:

- replace `TestValidator.error` only in the three changed Swagger regression files with a small local synchronous throw assertion backed by `TestValidator.predicate`;
- rerun the new absent/empty request-body cases against the current product source and record the expected red result;
- then apply the already-recorded product fix and continue the full verification/push/cancellation/re-review sequence.

Changing the shared external helper or unrelated repository tests is outside this PR's converter scope; the local assertion keeps this diff reviewable while making every negative case introduced or touched here effective.

samchon:
## Self-Review Round 4 remediation

Commit `8dd3962d98293f3bd570a72c6ccce61a23107a19` resolves both Round 4 findings:

- Swagger downgrade now rejects request bodies whose content is absent or has no defined media entry, instead of emitting `consumes: []` with a body that the upgrader rejects or inventing `application/json` on re-upgrade;
- all synchronous negative assertions in the three changed Swagger regression files now use a local effective throw assertion, avoiding the installed `TestValidator.error` false positive without expanding this converter PR into an external test-helper change;
- the unrepresentable matrix covers both absent and explicitly empty request-body content, and the public guide records the non-empty request-media requirement.

The effective assertion and new cases were run before the product fix and failed at `request body without content`, proving the regression was red. After remediation, verification passed:

- `packages/utils`: `pnpm exec ttsc --noEmit`
- `packages/utils`: `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused strict-numeric, Swagger enum, server/media, form-data, and unrepresentable regressions
- existing Swagger v2 document upgrade, document downgrade, and HTTP migration regressions
- targeted Prettier check across all changed TypeScript files
- `git diff --check`
- public runtime matrix proving absent/empty request media now throws while a defined `application/json` schema still downgrades

The exact-SHA cancellation gate completed for all six workflows created by the push. Run `29471244400` completed successfully before cancellation; runs `29471244418`, `29471244422`, `29471244454`, `29471244403`, and `29471244413` completed cancelled.

Round 5 now restarts the full base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Self-Review Round 5 findings and remediation plan

Round 5 restarted the complete review at `8dd3962d98293f3bd570a72c6ccce61a23107a19`: all three claimed issues, all ten current files, every base-to-head hunk, and the public OpenAPI, HTTP migration, strict LLM, and package-adapter consequence paths were inspected again. The round is not clean.

### Findings

1. **Nullable and single-branch unions erase branch annotations.** Both schema converters build an outer attribute object containing keys whose values are `undefined`, then spread it over a collapsed union branch. That overwrites a branch's defined `title`, `description`, `deprecated`, `example`, or `examples`. The upgrader additionally strips those annotations from the sole non-null branch whenever it materializes `x-nullable`. For example, `{ oneOf: [{ const: "a", description: "First" }, { type: "null" }] }` returns as `{ oneOf: [{ const: "a" }, { type: "null" }] }`, contradicting the documented branch-annotation round-trip guarantee.
2. **Empty or undefined form fields silently erase the request body.** A multipart or URL-encoded object with no defined properties downgrades to `parameters: []`; re-upgrade therefore has no `requestBody` at all. A required property whose value is `undefined` also passes the current `Object.hasOwn` check and is then filtered out, erasing requiredness and potentially the whole body. Swagger form parameters cannot encode a form request body without at least one defined field, so these shapes must be rejected.
3. **A non-form request media entry without a schema invents a body.** Downgrade currently substitutes `{}` for a missing media schema and emits a Swagger body parameter. The original public HTTP-migration route has no body schema, while the downgrade/re-upgrade result has an unconstrained `{}` schema and becomes a real migrated body. This is not an exact representation; request media without a schema must be rejected instead of synthesized.
4. **Equivalent server authorities with different casing are rejected.** `https://API.example.com/v1` and `http://api.example.com/v1` are the same URL authority with two representable Swagger schemes, but downgrade compares the parsed hosts case-sensitively and throws. Host comparison must follow URL's case-insensitive authority semantics while retaining a stable emitted spelling.
5. **Swagger file-form vendor extensions are dropped.** Upgrade's special `type: "file"` branch creates only `{ type: "string", format: "binary", description }`, discarding defined `x-*` fields that the emended schema and the reverse downgrade can represent. A Swagger file parameter with an extension therefore loses it after one round trip.

### Remediation plan

- Stop undefined outer attributes from overwriting collapsed branches, preserve annotations on explicit nullable-union branches, and add nullable/single-branch round-trip regressions.
- Reject form request schemas with no defined fields and validate `required` against the filtered, defined property set.
- Reject every request media entry without a shared schema before emitting a Swagger body or form parameter.
- Compare document and operation server hosts case-insensitively, covering absolute and scheme-relative URLs.
- Carry defined `x-*` fields from Swagger file parameters into binary form-property schemas.
- Update the public conversion guide for the newly enforced exact-representation boundaries and preserved metadata.
- Confirm the new tests fail on the current implementation, implement the fixes, then run utils and test-utils type checks/builds, focused converter regressions, existing Swagger v2 upgrade/downgrade/migration tests, targeted formatting, runtime consequence checks, and `git diff --check`.
- Create a Lore follow-up commit, push it, pass the exact-SHA Actions cancellation gate, and restart the complete Self-Review from the resulting HEAD.

samchon:
## Self-Review Round 5 remediation

Commit `e3f4336b64a0202ea69f433d17a89315221df488` resolves all five Round 5 findings:

- schema attributes are now compacted before composition, so undefined outer fields cannot overwrite a collapsed branch; nullable unions with branch-local annotations remain explicit Swagger `x-oneOf` branches and re-upgrade without moving or erasing either non-null or null annotations;
- downgrade now rejects request media without a schema and form request schemas without at least one defined property, while required names are checked against the filtered defined fields;
- document and operation server authorities compare case-insensitively for both absolute and scheme-relative URLs while retaining the first spelling in emitted Swagger metadata;
- Swagger file-form `x-*` fields remain on the emended binary property and survive re-downgrade;
- the public conversion guide records the exact-representation boundaries and metadata guarantees.

The new regressions were run before the product changes and failed with the expected missing nullable-branch annotation, schema-less request acceptance, and case-sensitive server rejection. After remediation, verification passed:

- `packages/utils`: `pnpm exec ttsc --noEmit`
- `packages/utils`: `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused nullable-enum, unrepresentable, form-data, and server/media regressions
- existing Swagger v2 document upgrade, document downgrade, HTTP migration, and strict LLM numeric-default regressions
- public runtime matrix proving schema-less requests and empty forms now throw, attributed nullable branches round-trip exactly, and differently cased authorities lower to one Swagger endpoint
- targeted Prettier check across all changed TypeScript files
- `git diff --check`

The first exact-SHA gate invocation returned a premature snapshot that still listed the newly created runs as queued. A direct exact-SHA requery and clean gate rerun then settled all six workflows created by the push: run `29472424964` completed successfully before cancellation; runs `29472424992`, `29472424979`, `29472424978`, `29472425024`, and `29472424976` completed cancelled.

Round 6 now restarts the full base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Self-Review Round 6 findings and remediation plan

Round 6 restarted the complete review at `e3f4336b64a0202ea69f433d17a89315221df488`: all three claimed issues, all ten current files, every base-to-head hunk, the public converter surface, existing nullable/form/parameter regressions, and the HTTP migration and strict LLM consequence paths were inspected again. The round is not clean.

### Findings

1. **The downgrader can emit a nullable form parameter that its own upgrader rejects.** An emended nullable string form property downgrades to a Swagger `formData` parameter with `type: "string"` and `x-nullable: true`. Re-upgrade converts that parameter to an emended `oneOf` containing string and null, then the upgrader's simple-form checker rejects every `oneOf`. Nullable scalar and nullable scalar-array fields are representable by the converter's supported nullable extensions, so both form directions must recognize the same simple nullable shape.
2. **A null-only Swagger enum is broadened to its scalar type plus null.** The upgrader deliberately detects null inside Swagger enum arrays, but an enum whose only value is null falls through the ordinary atomic-schema branch after the non-null enum filter becomes empty. `{ type: "string", enum: [null] }` therefore becomes `{ oneOf: [{ type: "string" }, { type: "null" }] }`, admitting every string instead of only null. The null-only case must not synthesize the declared scalar branch.
3. **Response media without a schema or example silently disappears.** An emended response containing `content: { "text/plain": {} }` downgrades to `produces: ["text/plain"]` and an empty Swagger response. Re-upgrade sees no schema or example marker and omits `content`, erasing the response media contract. As with the request-side boundary fixed in Round 5, Swagger cannot attach this empty media object to a particular response exactly, so downgrade must reject it rather than accept a lossy round trip.
4. **Form-property reference annotations are discarded during resolution.** `resolveSchema` returns only the referenced component and drops every sibling attribute on the form property's `$ref`. A scalar form property with `$ref: "#/components/schemas/Text"` and a field-specific `description` therefore emits a parameter without the description even though Swagger can represent it. Reference resolution must overlay supported local annotations on the resolved scalar; the same mechanism should cause aggregate annotations on a root form reference to reach the existing unrepresentable-object check instead of disappearing silently.

### Remediation plan

- Define the simple form scalar predicate once per converter direction, allowing exactly primitive scalars or a one-scalar-plus-null union and arrays of those scalars; add a downgrade/re-upgrade nullable form matrix.
- Treat a non-empty enum containing no non-null values as a null-only schema and add direct Swagger upgrade coverage.
- Reject non-empty response content whose media entries have neither a shared schema nor any MIME-keyed example, and extend the unrepresentable matrix.
- Preserve `$ref` sibling annotations while resolving form schemas, covering field-level description round trips and root-level aggregate rejection.
- Update the public conversion guide for nullable form support and the response-media exactness boundary.
- Confirm the new regressions fail on the current implementation, implement the fixes, then run utils and test-utils type checks/builds, focused converter regressions, existing Swagger v2 upgrade/downgrade/migration tests, targeted formatting, runtime consequence checks, and `git diff --check`.
- Create a Lore follow-up commit, push it, pass the exact-SHA Actions cancellation gate, and restart the complete Self-Review from the resulting HEAD.

samchon:
## Self-Review Round 6 addendum

The RED-test pass exposed one more instance of Finding 1 before any product edit: **standard Swagger form enums are rejected by the same asymmetric simple-form checker.** A `formData` parameter with `type: "string"` and `enum: ["draft", "published"]` upgrades to the canonical emended constant `oneOf`, after which `isFormDataSchema` rejects it. The reverse direction already lowers the same constant union to a typed Swagger enum and accepts it as a simple form field, so the generated representation is again not accepted in both directions.

The remediation predicate will therefore recognize the exact simple shapes produced from Swagger scalar enums as well as nullable scalars: a primitive or scalar constant, a same-scalar-type constant union with at most one null branch, a one-primitive-plus-null union, and an array whose items have one of those scalar shapes. Mixed scalar unions, nested arrays/objects, and binary array items remain rejected. Focused coverage will include scalar and array-item enums alongside nullable fields.

samchon:
## Self-Review Round 6 remediation

Commit `dcd01ee3622b1ae9ad9e0be6fd288d034c29048f` resolves all Round 6 findings and its form-enum addendum:

- both converter directions now recognize the same simple form shapes: primitive scalars, scalar constants and same-type constant unions, nullable scalars, and arrays whose items use those shapes; binary array items and structured/mixed unions remain rejected;
- form `$ref` resolution overlays local sibling annotations on the resolved scalar, while root form annotations reach the existing aggregate-metadata rejection instead of disappearing;
- a Swagger enum containing only null now upgrades to a null-only emended schema instead of widening to its declared scalar type plus null;
- response media without either a shared schema or MIME-keyed example is rejected rather than accepted and erased on re-upgrade;
- the public guide records nullable/enum form support and the response-media exactness boundary.

The new regressions were run before the product edits and failed with the expected form `TypeError`, null-only enum widening, response-media acceptance, and missing reference description. After remediation, verification passed:

- `packages/utils`: `pnpm exec ttsc --noEmit`
- `packages/utils`: `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused nullable/enum/reference form, null-only enum, and unrepresentable-response regressions
- existing Swagger v2 server/media, document upgrade, document downgrade, HTTP migration, parameter-required, empty-required, and strict LLM numeric-default regressions
- public runtime matrix proving nullable and constant-union form fields round-trip, local `$ref` descriptions survive, null-only enums stay null-only, and empty response media now throws
- targeted Prettier check across all changed TypeScript files
- `git diff --check`

The first exact-SHA gate invocation returned a premature empty-run snapshot before Actions indexed the push. A direct exact-SHA requery and clean gate rerun then settled all six workflows: run `29473714781` completed successfully before cancellation; runs `29473714764`, `29473714747`, `29473714771`, `29473714773`, and `29473714786` completed cancelled.

Round 7 now restarts the full base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Self-Review Round 7 findings and remediation plan

Round 7 restarted the complete review at `dcd01ee3622b1ae9ad9e0be6fd288d034c29048f`: all three claimed issues, all ten current files, every base-to-head hunk, the complete server/media/form/nullable matrices, the public converter entry points, and the HTTP migration and strict LLM consequence paths were inspected again. The round is not clean.

### Findings

1. **Nullable `default: null` metadata is erased on downgrade.** Swagger significant schemas support both `x-nullable: true` and `default: null`. Upgrade correctly moves that default to the emended null branch, but downgrade records only the null branch's common annotations and omits its `default`. A Swagger `{ type: "number", "x-nullable": true, default: null }` therefore returns without the default after one round trip. Null-only schemas and nullable reference unions have the same loss. Downgrade must retain the null branch default, place it back on a representable scalar nullable schema, and keep an explicit null branch when a reference cannot carry that default exactly.
2. **An empty response content map incorrectly conflicts with payload responses.** `downgradeMedia` represents an explicit empty response `content: {}` as `types: []`, and operation composition includes that empty array in its all-responses media-equality check. A common operation with a JSON `200` response and an empty-content `204` response therefore throws `response media types must be identical`, even though Swagger can encode it exactly using operation `produces: ["application/json"]`, a schema on `200`, and no schema on `204`. Empty response content is semantically no payload and should not participate in operation media selection; the request-body empty-content rejection remains unchanged.

### Remediation plan

- Preserve `default: null` on converted null branches; restore it onto non-reference `x-nullable` scalar schemas and emit an explicit `x-oneOf` null branch for nullable references when necessary.
- Normalize empty response content to no response media before the operation-level produces comparison, while continuing to reject non-empty response media entries that lack both schema and example.
- Add direct nullable-default, null-only-default, nullable-reference, and mixed payload/empty-response regressions, first confirming the current implementation is RED.
- Update the public guide to distinguish empty response content from non-empty schema-less media.
- Run utils and test-utils type checks/builds, focused converter regressions, existing Swagger v2 upgrade/downgrade/migration and strict LLM tests, targeted formatting, runtime consequence checks, and `git diff --check`.
- Create a Lore follow-up commit, push it, pass the exact-SHA Actions cancellation gate, and restart the complete Self-Review from the resulting HEAD.

samchon:
## Self-Review Round 7 remediation

Commit `55a05310cc0d12ee85c557e8d02494c1c52bacb6` resolves both Round 7 findings:

- Swagger nullable scalars now inline a null-branch `default: null` together with `x-nullable`, null-only schemas keep that default, and nullable references retain it in an explicit `x-oneOf` branch that round-trips exactly;
- an explicit empty response `content` map now normalizes to a response without payload media, so it no longer conflicts with payload-bearing responses when the operation's Swagger `produces` list is selected;
- the public guide distinguishes payload-free empty response content from non-empty response media that still require a shared schema or MIME-keyed example.

The retained RED tests from before the restart exercised nullable scalar, null-only, nullable-reference, and mixed JSON/empty-response cases. The first resumed GREEN attempt exposed a TypeScript error in the remediation itself: the new code called the nonexistent `SwaggerV2TypeChecker.isNull`. It was corrected to the repository's actual narrowing helper, `isNullOnly`, before commit.

Verification passed after that correction:

- `packages/interface`: `pnpm build`
- `tests/test-interface`: `pnpm start`
- `packages/utils`: `pnpm exec ttsc --noEmit` and `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused Swagger/LLM/HTTP regressions: strict numeric defaults, enum/default downgrade, unrepresentable downgrade, form-data round trip, server/media/empty-response round trip, document v2 upgrade/downgrade, HTTP v2 migration, and OpenAPI required/empty-required conversion
- complete `tests/test-utils` suite
- targeted Prettier check for every changed TypeScript file
- `git diff --check`

Skipped as campaign-wide gates: website production build and repository-wide `pnpm test`.

The first exact-SHA cancellation attempt raced run `29474838455` while GitHub was completing its cancellation; an immediate exact-SHA readback showed it completed cancelled. The full gate rerun then settled all six workflows with two quiet polls: run `29474838436` completed successfully before cancellation, and runs `29474838426`, `29474838439`, `29474838438`, `29474838455`, and `29474838429` completed cancelled.

Round 8 now restarts the complete base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Self-Review Round 8 findings and remediation plan

Round 8 restarted the complete review at `55a05310cc0d12ee85c557e8d02494c1c52bacb6`: all three claimed issues, all ten files and every base-to-head hunk, the official Swagger 2 server/media/parameter contracts, the public converter entry points, and the HTTP migration and strict LLM consequence paths were inspected again. The round is not clean.

### Findings

1. **A nullable form array is accepted by downgrade but rejected by the generated Swagger document's own upgrader.** An emended property shaped as `oneOf: [{ type: array, items: { type: string } }, { type: null }]` downgrades to a Swagger `formData` parameter with `type: array` and `x-nullable: true`. On re-upgrade, that parameter becomes the same array-plus-null union, but `SwaggerV2Upgrader.isFormDataSchema` only recognizes a top-level array before nullable expansion and rejects the union. Round 6 covered nullable array *items*, not a nullable array field itself. The simple-form recognizer must be symmetric for nullable arrays and recursively valid Swagger `Items Object` arrays.

2. **Valid URL-encoded Swagger file fields are rejected in both directions.** Swagger 2 explicitly permits `type: file` form parameters when `consumes` is `multipart/form-data`, `application/x-www-form-urlencoded`, or both. The upgrader currently requires every selected media type to be multipart, and the downgrader enables files only when every form media type is multipart. This rejects standard Swagger inputs and emended binary fields that have an exact Swagger representation.

### Remediation plan

- Add effective fail-first coverage for nullable top-level arrays, recursively valid array items, URL-encoded file fields, and operations that advertise both standard form media types.
- Make both simple-form predicates recursive over arrays and allow exactly one nullable wrapper around an otherwise simple field, while continuing to reject objects, mixed scalar unions, binary array items, and other non-simple shapes.
- Accept binary/file form fields for either standard Swagger form media type and their combination, preserving the existing file vendor-extension round trip.
- Correct the public guide's multipart-only statement and document the supported nullable/nested array boundary.
- Run interface/utils builds and type checks, the focused Swagger/LLM/HTTP matrix, complete `test-utils`, targeted Prettier, runtime round-trip probes, and `git diff --check`; then create a Lore follow-up commit, push it, pass the exact-SHA gate, and restart the entire Self-Review from the new HEAD.

samchon:
## Self-Review Round 8 remediation

Commit `65cf59d5c74958b5f5088b5fc6bf7e254fd8ae48` resolves both Round 8 findings:

- both converter directions now accept recursively nested Swagger Items Objects and a nullable wrapper around an otherwise simple scalar or array form field, while objects, mixed scalar unions, and binary array items remain rejected;
- file form parameters and binary form properties now work with `application/x-www-form-urlencoded`, `multipart/form-data`, or both, matching the Swagger 2 Parameter Object contract;
- obsolete multipart-only negative oracles were removed, and dedicated one-case regression files cover form arrays and URL-encoded file fields, vendor extensions, dual media, and HTTP migration;
- the public guide records the supported recursive/nullable array shapes and both valid file media types.

The new URL-encoded-file and nested-array regressions were run before the product edits and failed with the expected multipart-only and non-simple-schema `TypeError` paths. After remediation, verification passed:

- `packages/interface`: `pnpm build`
- `tests/test-interface`: `pnpm start`
- `packages/utils`: `pnpm exec ttsc --noEmit` and `pnpm build`
- `tests/test-utils`: `pnpm exec ttsc --noEmit`
- focused Swagger/LLM/HTTP matrix, including both new form regressions
- complete `tests/test-utils` suite
- targeted Prettier check for every changed TypeScript file
- `git diff --check`

Skipped as campaign-wide gates: website production build and repository-wide `pnpm test`.

The first cancellation invocation accidentally supplied the abbreviated commit ID; because the script compares returned full `headSha` values literally, its empty snapshot was discarded. The immediate full-SHA rerun is the operative gate and completed with two quiet polls: runs `29477530002`, `29477530004`, `29477529981`, `29477529974`, `29477529965`, and `29477529977` all completed cancelled.

Round 9 now restarts the complete base-to-head Self-Review from scratch; this remains a draft and is not yet a clean-round declaration.

samchon:
## Solo Self-Review Round 9 — finding

A fresh full review of all 12 base-to-head files found one remaining representability loss in the Swagger 2 form-file path.

The Swagger 2 Parameter Object permits its standard validation fields on non-body parameters, including file parameters. `convertFormDataSchema` currently keeps only `description` and `x-*` when upgrading `type: file`, so fields such as `minLength`, `maxLength`, `pattern`, `default`, and `enum` disappear. The downgrade path already retains those fields when mapping an emended `string/binary` property back to `type: file`, making the two directions asymmetric and causing the converter to reject its own exact round-trip contract.

Remediation plan:

1. Extend the existing URL-encoded/dual-media file regression with standard file-parameter validation metadata and prove the current upgrade loses it.
2. Preserve the file parameter's remaining standard fields while replacing only Swagger's `type: file` with the emended `type: string, format: binary` representation.
3. Re-run interface/utils builds, type checks, focused form/converter tests, targeted formatting, and diff checks.
4. Restart the exhaustive solo review from the complete changed surface after the fix.

samchon:
## Round 9 remediation and Solo Self-Review Round 10 — clean

Resolved in `ee36599de81e897ba8116975e6c6073da3dda26b`.

- Preserved the remaining Swagger file Parameter Object fields while translating only `type: file` into the emended `type: string, format: binary` marker.
- Extended the URL-encoded and dual-media file regression with `minLength`, `maxLength`, and `pattern`, proving upgrade and downgrade preserve the validation contract together with the existing vendor extension.
- Updated the Swagger conversion guide to state that file validation metadata survives on the binary property schema.
- Confirmed the regression failed before the implementation and passed afterward.

Verification at the final source state:

- `pnpm --filter ./packages/interface build`
- `pnpm --filter ./tests/test-interface start`
- `pnpm --filter ./packages/utils exec ttsc --noEmit`
- `pnpm --filter ./tests/test-utils exec ttsc --noEmit`
- focused strict-default, Swagger enum, unrepresentable, form-array, form-data, server/media, and URL-encoded file tests
- final standalone URL-encoded file regression
- targeted TypeScript formatting and `git diff --check`

After the fix, Round 10 restarted from the complete 12-file base-to-head surface and rechecked every source/test/document hunk, converter symmetry, downstream HTTP migration behavior, and the latest master consequence surface. The only same-file master change adds removal of OpenAPI `content/style/explode` fields during Swagger downgrade; the three-way merge retains both sides without conflict. No sound improvement or remaining issue was found.

Per campaign gates, the website production build and repository-wide `pnpm test` remain intentionally not run.

samchon:
## Rebase resolution and Solo Self-Review Round 11 — clean

Rebased the complete PR onto the latest `origin/master`; final head is `7f16585d75fc3c79bfa0fdceb7ba80b5cecf4cb0`.

The only textual conflict was in `SwaggerV2Downgrader.downgradeResponse`. Resolution keeps this PR's media-aware response conversion and also retains master's removal of OpenAPI `style`/`explode` from downgraded response headers. `git range-diff` shows the other ten commits are content-identical; this conflict commit differs only by the two master-owned removals.

Fresh verification after rebase:

- interface and utils builds passed
- interface test compilation passed
- utils and test-utils type checks passed
- complete `tests/test-utils` suite passed, including every new LLM/Swagger/form/server/media regression
- no conflict markers, clean worktree, and `git diff --check origin/master...HEAD` passed

Round 11 re-read the final 12-file latest-master diff and the resolved consequence surface. No sound improvement remains.

PR #2063, opened 2026-07-16
fix(validation): preserve complete reporting diagnostics

### LINKED ISSUE #1890: Automated validation suites leave exact-result oracles disabled
## Problem

Two automated validation oracles calculate independent expected results but do not enforce them completely:

- `tests/test-utils-automated/src/internal/_test_validate.ts` computes wrong OpenAPI spoiler paths/counts while the throw that fails the suite remains disabled.
- `tests/test-typia-automated/src/internal/_test_standardSchema_validate.ts` compares only the issue count, so the right number of errors at the wrong paths passes; its success-result type assertion is also disabled.

The sibling typia validator oracle already enforces the spoiler path set, and template `SPOILERS` are the independent fixture contract.

## Evidence and provenance

The disabled blocks remain on current `master`; no pull request implemented this closed issue. The 2026-07 campaign also reproduced two path/reporting defects that the restored assertions should expose: identifier-prefix siblings such as `$input.a`/`$input.ab`, and Standard Schema formatting on hostile unknown values. Those defects have dedicated issues; this issue owns the oracle rather than weakening expected results.

## Approach

Restore exact, deterministic path comparison in both suites and restore the success-result type assertion. If enabling the oracle fails, fix the production reporter/adapter in the owning issue. Do not edit fixtures or reduce the assertion to make the suite green.

- The utils OpenAPI spoiler path/count comparison actively fails on mismatch.
- Standard Schema compares sorted paths rather than counts only, and its success-result type assertion compiles.
- The existing independent `SPOILERS`/expected values remain intact.
- Focused utils/typia automated suites and the full suite pass with the oracles enabled.
- During the issue campaign implementation, do not run repository-wide `pnpm format`; formatting is deferred to the campaign cleanup PR.

## Coordination

Reopen rather than duplicate. Reporter prefix and Standard Schema message fixes can land in the same implementation batch so the restored oracle is green from the first non-claim code push.

### LINKED ISSUE #2033: Validation reporters omit identifier-prefix sibling failures
## Problem

Typia's generated validation reporter and `OpenApiValidator` suppress independent sibling failures when one property name is a prefix of another. Results depend on declaration order.

## Evidence

Both `packages/typia/src/internal/_validateReport.ts` and `packages/utils/src/validators/OpenApiValidator.ts` use a raw string-prefix test against the last path. Reporting `$input.ab` and then `$input.a` retains only `$input.ab`; reversing the order retains both. A real OpenAPI object with properties `ab` then `a`, both invalid, reproduces the omission.

`$input.a` is not an ancestor of `$input.ab`; path segment boundaries matter. The same predicate flows through generated validate/equals/parse/stringify/decode compositions, Standard Schema issue arrays, and adapter feedback. Closed unmerged PR #1864 identified but deferred the defect; no open implementation claim exists.

## Consequence surface

Dotted identifiers, quoted property keys, numeric indexes, real parent/child failures, both declaration orders, typia-generated validation, `OpenApiValidator`, Standard Schema, and LLM/MCP/LangChain/Vercel feedback.

## Approach

Use segment-aware ancestor detection or an equivalent boundary-aware invariant. Continue suppressing a true descendant when its ancestor failure makes it redundant, but never suppress identifier-prefix siblings.

- `$input.ab`/`$input.a` and the reverse order both report two errors.
- Genuine parent/child, quoted-key, and indexed cases preserve intended de-duplication.
- Add identical typia and OpenAPI coverage and re-enable the exact spoiler path assertion tracked by #1890.
- Verify downstream Standard Schema/LLM issue lists and run focused plus full suites.

## Coordination

Reopen #1890 and batch its disabled path oracles with this implementation so the regression cannot be re-masked.

### LINKED ISSUE #2034: Standard Schema validation throws while formatting unknown failures
## Problem

Typia's Standard Schema adapter can throw while formatting an ordinary validation failure. `IValidation.IError.value` is `unknown`, but `_createStandardSchema` interpolates it directly into a template literal.

## Evidence

In `packages/typia/src/internal/_createStandardSchema.ts`:

```ts
message: `expected ${error.expected}, got ${error.value}`
```

If the underlying validator returns a normal failure for `Symbol("x")` or `Object.create(null)`, `~standard.validate()` throws `TypeError` instead of returning `{ issues }`. A custom object with throwing coercion hooks has the same result. The line was introduced by PR #1500; exact searches found no issue or fix.

## Consequence surface

`createValidate` and `createValidateEquals` as Standard Schema validators; root and property values; symbols, null-prototype objects, cyclic/BigInt-rich objects, proxies, and hostile coercion hooks. Plain typia validation already returns safely; the adapter is the failure source.

## Approach

Make diagnostic value formatting total for arbitrary `unknown`, with a stable fallback that cannot invoke user coercion indefinitely or throw. Keep path conversion and ordinary readable messages unchanged where possible.

- Root/property Symbol, null-prototype, and throwing-coercion values return a failure result with string messages and paths.
- Ordinary primitives/objects retain useful messages and success behavior is unchanged.
- Restore the Standard Schema success/path oracles in #1890 and run focused generated plus full tests.

## Coordination

This can share a batch with reporter/oracle work but does not depend on the path-prefix fix.

samchon:
## Intent

Make validation reporting complete and deterministic across typia, OpenAPI utilities, and Standard Schema adapters. This draft reserves one campaign batch for the shared reporting/oracle surface before implementation begins.

## Problem

- Validation reporters currently use a raw string-prefix comparison when suppressing redundant descendant errors. Independent sibling properties such as `$input.a` and `$input.ab` can therefore hide one another depending on declaration order.
- Standard Schema diagnostic construction interpolates an arbitrary `unknown` value directly. Symbols, null-prototype objects, and hostile coercion hooks can turn an ordinary validation failure into a thrown `TypeError`.
- The automated OpenAPI and Standard Schema validation oracles do not fully enforce their independently derived path and success expectations, allowing reporting regressions to remain masked.

## Scope

- Make ancestor suppression path-segment aware in both typia-generated validation reporting and `OpenApiValidator`, while preserving de-duplication for genuine parent/child paths.
- Make Standard Schema failure-message value formatting total for arbitrary `unknown` without weakening ordinary readable diagnostics.
- Restore exact deterministic path assertions and the Standard Schema success-result type oracle without changing fixture expectations.
- Add regression coverage for both sibling declaration orders, genuine descendants, quoted/indexed paths, and unsafe diagnostic values, then trace the downstream issue surfaces.

## Deferred

No unrelated reporter redesign, fixture weakening, or repository-wide formatting is included. Campaign-wide formatting remains owned by the dedicated cleanup pull request.

## Verification

Pending implementation. The implementation push will record focused and broader package/test commands here as a chronological comment. Automatic Actions runs for this campaign branch are cancelled by exact commit SHA; local verification and the later Solo Self-Review are the gates.

Closes #2033
Closes #2034
Closes #1890

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/transform_helpers_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform_helpers_test.go
@@ -75,10 +75,14 @@ func ttscTypiaTestRewriteCommonJS(t *testing.T, js string) string {
 }
 
 const ttscTypiaTestValidateReportStub = `module.exports._validateReport = (array) => {
+  const isAncestor = (ancestor, descendant) =>
+    descendant === ancestor ||
+    descendant.startsWith(ancestor + ".") ||
+    descendant.startsWith(ancestor + "[");
   const reportable = (path) => {
     if (array.length === 0) return true;
     const last = array[array.length - 1].path;
-    return path.length > last.length || last.substring(0, path.length) !== path;
+    return !isAncestor(path, last) && !isAncestor(last, path);
   };
   return (exceptable, error) => {
     if (exceptable && reportable(error.path)) {
--- a/packages/typia/src/internal/_createStandardSchema.ts
+++ b/packages/typia/src/internal/_createStandardSchema.ts
@@ -17,7 +17,7 @@ export const _createStandardSchema = <T>(
         } else {
           return {
             issues: result.errors.map((error) => ({
-              message: `expected ${error.expected}, got ${error.value}`,
+              message: `expected ${error.expected}, got ${formatValue(error.value)}`,
               path: typiaPathToStandardSchemaPath(error.path),
             })),
           } satisfies StandardSchemaV1.FailureResult;
@@ -25,6 +25,20 @@ export const _createStandardSchema = <T>(
       },
     },
   } satisfies StandardSchemaV1<T, T>);
+
+const formatValue = (value: unknown): string => {
+  if (typeof value === "object") {
+    if (value === null) return "null";
+    try {
+      if (Array.isArray(value)) return "[object Array]";
+    } catch {
+      // A revoked proxy can throw even for Array.isArray().
+    }
+    return "[object Object]";
+  }
+  if (typeof value === "function") return "[function]";
+  return String(value);
+};
 
 enum PathParserState {
   // Start of a new segment
--- a/packages/typia/src/internal/_validateReport.ts
+++ b/packages/typia/src/internal/_validateReport.ts
@@ -1,10 +1,14 @@
 import { IValidation } from "@typia/interface";
 
 export const _validateReport = (array: IValidation.IError[]) => {
+  const isAncestor = (ancestor: string, descendant: string): boolean =>
+    descendant === ancestor ||
+    descendant.startsWith(`${ancestor}.`) ||
+    descendant.startsWith(`${ancestor}[`);
   const reportable = (path: string): boolean => {
     if (array.length === 0) return true;
     const last: string = array[array.length - 1]!.path;
-    return path.length > last.length || last.substring(0, path.length) !== path;
+    return isAncestor(path, last) === false && isAncestor(last, path) === false;
   };
   return (exceptable: boolean, error: IValidation.IError): false => {
     if (exceptable && reportable(error.path)) {
--- a/packages/utils/src/validators/OpenApiValidator.ts
+++ b/packages/utils/src/validators/OpenApiValidator.ts
@@ -61,11 +61,15 @@ export namespace OpenApiValidator {
   };
 
   const createReporter = (array: IValidation.IError[]) => {
+    const isAncestor = (ancestor: string, descendant: string): boolean =>
+      descendant === ancestor ||
+      descendant.startsWith(`${ancestor}.`) ||
+      descendant.startsWith(`${ancestor}[`);
     const reportable = (path: string): boolean => {
       if (array.length === 0) return true;
       const last: string = array[array.length - 1]!.path;
       return (
-        path.length > last.length || last.substring(0, path.length) !== path
+        isAncestor(path, last) === false && isAncestor(last, path) === false
       );
     };
     return (
--- a/packages/utils/src/validators/internal/OpenApiArrayValidator.ts
+++ b/packages/utils/src/validators/internal/OpenApiArrayValidator.ts
@@ -10,17 +10,6 @@ export namespace OpenApiArrayValidator {
   ): boolean => {
     if (Array.isArray(ctx.value) === false) return ctx.report(ctx);
     return [
-      ctx.value
-        .map((value, i) =>
-          OpenApiStationValidator.validate({
-            ...ctx,
-            schema: ctx.schema.items,
-            value,
-            path: `${ctx.path}[${i}]`,
-            required: true,
-          }),
-        )
-        .every((v) => v),
       ctx.schema.minItems !== undefined
         ? ctx.value.length >= ctx.schema.minItems ||
           ctx.report({
@@ -44,6 +33,17 @@ export namespace OpenApiArrayValidator {
             })
           : true
         : true,
+      ctx.value
+        .map((value, i) =>
+          OpenApiStationValidator.validate({
+            ...ctx,
+            schema: ctx.schema.items,
+            value,
+            path: `${ctx.path}[${i}]`,
+            required: true,
+          }),
+        )
+        .every((v) => v),
     ].every((v) => v);
   };
 }
--- a/packages/utils/src/validators/internal/OpenApiOneOfValidator.ts
+++ b/packages/utils/src/validators/internal/OpenApiOneOfValidator.ts
@@ -93,13 +93,7 @@ export namespace OpenApiOneOfValidator {
     );
     const branches: IDiscriminatorBranch[] = [
       ...(tuples.length === 0 && arrays.length !== 0
-        ? discriminateArrays(
-            ctx,
-            significant.filter(
-              (flat): flat is IFlatSchema<OpenApi.IJsonSchema.IArray> =>
-                OpenApiTypeChecker.isArray(flat.schema),
-            ),
-          )
+        ? discriminateArrays(ctx, arrays)
         : []),
       ...discriminateObjects(
         ctx,
--- a/tests/test-typia-automated/src/internal/_test_standardSchema_validate.ts
+++ b/tests/test-typia-automated/src/internal/_test_standardSchema_validate.ts
@@ -1,7 +1,18 @@
 import { StandardSchemaV1 } from "@standard-schema/spec";
 import { TestStructure } from "@typia/template";
+import { NamingConvention } from "@typia/utils";
 import typia from "typia";
 
+/**
+ * Verifies Standard Schema validation against independent spoiler oracles.
+ *
+ * Issue counts cannot detect a reporter that substitutes or omits paths. This
+ * helper reconstructs typia paths from Standard Schema segments so the same
+ * fixture contract proves both the adapter's success type and exact failures.
+ *
+ * 1. Assert the successful result preserves both its static and runtime value.
+ * 2. Apply every fixture spoiler and compare exact sorted issue paths.
+ */
 export const _test_standardSchema_validate =
   (name: string) =>
   <T>(factory: TestStructure<T>) =>
@@ -16,9 +27,7 @@ export const _test_standardSchema_validate =
       throw new Error(
         `Bug on typia.createValidate["~standard"].validate(): failed to archive the input value.`,
       );
-    // This does not compile.
-    // TODO: Fix this.
-    // typia.assertEquals<StandardSchemaV1.SuccessResult<T>>(valid);
+    valid satisfies StandardSchemaV1.SuccessResult<T>;
 
     const wrong: ISpoiled[] = [];
     for (const spoil of factory.SPOILERS ?? []) {
@@ -34,12 +43,15 @@ export const _test_standardSchema_validate =
       typia.assertEquals(valid);
       expected.sort();
       const issues = [...valid.issues];
+      const actual: string[] = issues.map(issuePath).sort();
 
-      // It's difficult to compare the paths, so we'll just compare the number of issues.
-      if (issues.length !== expected.length)
+      if (
+        actual.length !== expected.length ||
+        actual.every((path, i) => path === expected[i]) === false
+      )
         wrong.push({
           expected,
-          actual: issues.map((e) => e.path?.join(".") ?? ""),
+          actual,
         });
     }
     if (wrong.length !== 0) {
@@ -54,3 +66,13 @@ interface ISpoiled {
   expected: string[];
   actual: string[];
 }
+
+const issuePath = (issue: StandardSchemaV1.Issue): string =>
+  (issue.path ?? []).reduce<string>((path, segment) => {
+    const key: PropertyKey =
+      typeof segment === "object" ? segment.key : segment;
+    if (typeof key === "number") return `${path}[${key}]`;
+    if (typeof key === "string" && NamingConvention.variable(key))
+      return `${path}.${key}`;
+    return `${path}[${JSON.stringify(key)}]`;
+  }, "$input");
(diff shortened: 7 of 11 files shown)


samchon:
Implementation and local verification are complete at `07be4d4457bb606a710ebc2865936d0313d5982e`.

### Implemented invariants

- typia and `OpenApiValidator` now recognize ancestry only at path-segment boundaries (`.` or `[`), and suppress genuine parent/child duplicates in either reporting order. Independent prefixes such as `$input.a` / `$input.ab`, quoted keys, and indexes remain reportable.
- OpenAPI array constraints run before item traversal so a failing container constraint owns the diagnostic; referenced array branches now participate in array discrimination after reference resolution.
- Standard Schema messages format arbitrary `unknown` values without invoking object/function coercion. Symbols, null-prototype objects, cycles with BigInt, hostile coercion hooks, throwing proxies, and revoked proxies all return issues instead of throwing.
- The OpenAPI and Standard Schema automated helpers now enforce independently derived exact canonical path sets/counts, and the Standard Schema helper restores its success-result type oracle.

### Oracle premise adjustments

The fixtures and their `SPOILERS` remain unchanged. Raw spoiler leaf paths are authoritative when OpenAPI can select one branch by runtime type or an object discriminator. For a non-predictable `oneOf`, no leaf branch is authoritative, so both expected and actual paths are canonicalized to the ambiguous union owner before exact set/count comparison. Selecting an arbitrary failed branch was rejected because branch order would determine misleading diagnostics.

The previously commented `typia.assertEquals<StandardSchemaV1.SuccessResult<T>>` cannot be transformed for the helper's unresolved generic `T`. The restored static oracle therefore uses `valid satisfies StandardSchemaV1.SuccessResult<T>`; the existing runtime identity check continues to prove that the successful adapter result archives the original value.

### Verification

Passed:

- `pnpm --filter ./tests/test-typia-schema start -- --include validate_report_path_boundary`
- `pnpm --filter ./tests/test-typia-schema start -- --include standard_schema_unknown_diagnostics`
- `pnpm --filter ./tests/test-utils start -- --include openapi_validator_report_path_boundary`
- full `pnpm --filter ./tests/test-utils-automated start`
- full `pnpm --filter ./tests/test-typia-automated start`
- source `tsc -p` checks for `test-typia-schema`, `test-utils`, `test-utils-automated`, and `test-typia-automated`
- `pnpm --filter ./packages/typia build`
- root `pnpm build`
- root `pnpm test` (all 12 test workspaces plus public/native Go toolchain; 755.9s)
- explicit Prettier check for the ten scoped files and `git diff --check`

The root full-suite result is branch-local additional evidence; it does not replace the campaign's latest-master, isolated integrated full-suite gate after all batches land. Solo Self-Review also remains intentionally pending.

### Exact-SHA Actions gate

The repository helper settled after 3 polls for `07be4d4457bb606a710ebc2865936d0313d5982e`. Runs `29469519981`, `29469519984`, `29469519988`, `29469520024`, `29469520032`, and `29469520040` all reached `completed/cancelled`, followed by the required quiet polls.

samchon:
## Solo Self-Review round 1 findings

The complete `origin/master...07be4d4457bb606a710ebc2865936d0313d5982e` surface is not clean yet.

1. **The restored OpenAPI oracle still masks exact path/count regressions.** `normalizePaths` applies a `Set` to both expected and actual paths, so duplicate actual errors cannot fail the count oracle. It also classifies every multi-array `oneOf` as ambiguous because `selectOneOf` returns before array discrimination. A raw reproduction with the existing `ArrayUnion` fixture showed the first three spoilers and `OpenApiValidator` both report `$input[n][2]`, but the helper canonicalizes both sides to `$input[n]`. That discards an already-authoritative leaf path and would hide an array discriminator regression.
2. **The generated-code command test runtime still embeds the old reporter.** `packages/typia/native/cmd/ttsc-typia/transform_helpers_test.go` copies `_validateReport`, but its stub still uses the raw-prefix predicate. Command-level generated-code tests therefore execute different de-duplication semantics from the shipped runtime and can preserve the exact bug fixed by #2033.
3. **The focused OpenAPI regression does not pin the production branches added while restoring the oracle.** The test comment claims a redundant array parent/child case, but the current `items` property has children only and `duplicates` has a parent constraint only. It also has no referenced-array `oneOf` case, so reverting either array-constraint ordering or `discriminateArrays(ctx, arrays)` is left to broad matrix coverage rather than a direct regression.

### Remediation plan

- Canonicalize ambiguous expected paths separately from actual paths so actual multiplicity remains exact.
- Preserve a multi-array leaf when the fixture value uniquely selects an array branch from its first item, including referenced branches; collapse only genuinely ambiguous unions to their owner.
- Update the native command runtime stub to the segment-aware bidirectional predicate.
- Extend the existing focused OpenAPI test with a parent-constraint-plus-invalid-children case and a referenced array union whose selected branch fails at an indexed leaf.
- Run explicit formatting/diff checks, focused `ArrayUnion` and OpenAPI regressions, the native command package tests, full utils automated tests, relevant package builds, then perform a new complete from-scratch Self-Review round.

samchon:
## Solo Self-Review round 1 resolution

Resolved in `ea099d32ded7541118e3ef0c57e5c2c077031231`.

- Canonicalization now retains a uniquely first-item-discriminated array leaf, including through `$ref` and nested item unions. Only genuinely ambiguous unions collapse to their owner.
- Canonical path grouping collapses different branch leaves that share one owner but preserves the maximum repetition count of any identical raw path, so duplicate reporter output remains count-significant.
- The native command test runtime stub now uses the same segment-aware, bidirectional ancestor predicate as `_validateReport`.
- The focused OpenAPI test directly proves array-constraint ownership over invalid children and referenced-array branch selection to `$input[1]`.

Verification passed:

- `pnpm --filter ./tests/test-utils start -- --include openapi_validator_report_path_boundary`
- `pnpm --filter ./tests/test-utils-automated start -- --include ArrayUnion`
- full `pnpm --filter ./tests/test-utils-automated start`
- `go -C packages/typia/test test ../native/cmd/ttsc-typia`
- plain source typechecks for `test-utils` and `test-utils-automated`
- explicit Prettier, changed-Go formatting, and `git diff --check`

The exact-SHA cancellation chronology for `ea099d32ded7541118e3ef0c57e5c2c077031231` was:

1. The first helper invocation observed run `29470541750` (`spell-check`) active, but `gh run cancel` returned `Cannot cancel a workflow run that is completed` after it won the race and finished. API reread: `completed/success`, updated `2026-07-16T04:06:28Z`.
2. The second invocation hit the same error after observing run `29470541739` (`website`) active. The helper suppresses the run ID in its failed cancel output, so this ID was reconstructed from the exact-SHA update chronology. API reread: `completed/cancelled`, updated `2026-07-16T04:06:51Z`.
3. A separate third helper invocation then settled after 3 polls. Runs `29470541732`, `29470541735`, `29470541739`, `29470541755`, and `29470541769` were `completed/cancelled`; run `29470541750` was `completed/success`. No unrelated run was touched.

A new complete from-scratch Self-Review round follows this comment.

samchon:
## Solo Self-Review round 2 findings

The complete `origin/master...ea099d32ded7541118e3ef0c57e5c2c077031231` surface is not clean yet.

1. **The OpenAPI oracle still normalizes the result under test, so it can accept wrong raw paths and counts.** `normalizePaths` is applied to both the independent spoiler expectation and `valid.errors`. Within an ambiguous `oneOf`, different actual leaves are grouped under one owner and only the maximum repetition of an identical raw leaf survives. Therefore a substituted leaf or additional distinct leaf below that owner can compare equal to the expected owner, contrary to #1890's exact independent path/count contract and its explicit prohibition on reducing the assertion to make the suite green. A full diagnostic with raw actual paths showed why the workaround remained: the expected-side selector does not reproduce OpenAPI's recursive remainder narrowing, so it incorrectly canonicalizes already-authoritative leaves in recursive/implicit object unions. A separate raw-spoiler/raw-actual run showed only the genuinely non-predictable tuple/object/tag union cases need owner canonicalization.
2. **The Standard Schema regression does not pin the ordinary primitive-message acceptance criterion.** The focused test asserts the exact ordinary-object message, but every primitive/hostile case only asserts that the message is a string. A formatter that replaced useful number/boolean/bigint text with a generic token would remain green even though #2034 requires ordinary primitive messages to remain useful.

### Remediation plan

- Canonicalize only the independent expected spoiler paths; compare `valid.errors` as a raw sorted path array with exact multiplicity.
- Make expected-side object-union selection follow the production discriminator's recursive remainder narrowing, retaining leaves once a branch becomes authoritative while collapsing only genuinely non-predictable unions to their owner.
- Add an exact ordinary primitive Standard Schema message assertion alongside the existing object assertion.
- Run focused recursive/implicit/ambiguous union cases, the full utils automated suite, focused Standard Schema tests, relevant typechecks/builds, formatting and diff checks, then commit/push with Lore trailers, settle the exact-SHA gate, and begin another complete from-scratch Self-Review round.

samchon:
## Solo Self-Review round 2 resolution

Resolved in `ba45e452be661cf194cfd3d04d1bc371ac207976`.

- OpenAPI spoiler canonicalization now applies only to the independent expected side; `OpenApiValidator` errors are compared as raw sorted paths with exact multiplicity.
- Expected-side object discrimination now follows recursive remainder narrowing, so recursive and implicit unions retain authoritative leaf paths while genuinely non-predictable tuple/object/tag unions still canonicalize to their owner.
- Standard Schema diagnostics now pin the useful ordinary primitive message `expected string, got 42`.

Verification passed:

- focused `ArrayRecursiveUnionImplicit`, `ObjectUnionComposite`, `TypeTagObjectUnion`, and `ArrayRepeatedUnionWithTuple`
- full `pnpm --filter ./tests/test-utils-automated start`
- full Standard Schema createValidate feature matrix
- focused Standard Schema unknown diagnostics and both reporter boundary suites
- plain source typechecks for `test-utils-automated` and `test-typia-schema`
- explicit Prettier and `git diff --check`

Exact-SHA cancellation chronology:

1. An initial helper invocation used an incorrectly inferred post-amend full SHA (`ba45e452be25ff6051cc3f21aa3ec43ee4a3ed2d`). It settled with zero runs, so it changed no Actions state.
2. The first invocation with the verified PR/local/remote SHA observed the new runs but raced `spell-check` finishing naturally; `gh run cancel` returned `Cannot cancel a workflow run that is completed`.
3. API readback showed `spell-check` run `29471480739` completed successfully and runs `29471480705`, `29471480676`, `29471480704`, `29471480683`, and `29471480709` completed cancelled.
4. A separate verified-SHA helper invocation then settled after 3 polls. No unrelated SHA was touched.

A new complete from-scratch Self-Review round follows this comment.

samchon:
## Solo Self-Review round 3: clean

I re-reviewed the complete `origin/master...ba45e452be661cf194cfd3d04d1bc371ac207976` diff from scratch after the round-2 remediation. No further sound improvement or issue candidate was found.

The pass covered all 11 changed files and the full acceptance/consequence surface of #2033, #2034, and #1890:

- segment-aware, bidirectional parent/child de-duplication in both reporters, including declaration order, dotted/quoted keys, indexes, equal paths, and both ancestor directions
- generated-code runtime-stub parity
- Standard Schema total formatting for primitive, symbol, null-prototype, cyclic/BigInt-rich, hostile-coercion, throwing-proxy, and revoked-proxy values; root/property paths; validate/validateEquals; success identity
- OpenAPI array constraint ownership and referenced-array oneOf discrimination
- independent expected-side oneOf canonicalization with recursive remainder narrowing, while actual OpenAPI paths/counts remain raw and exact
- Standard Schema exact path reconstruction and success-result static oracle
- downstream OpenAPI validator consumers in LLM JSON and HTTP LLM composition, dynamic test discovery, package boundaries, and generated/ignored fixture behavior
- no dependency, lockfile, public API, package-version, documentation, or production native-Go changes required

Current-head verification evidence:

- full `pnpm --filter ./tests/test-utils-automated start`
- focused recursive/implicit/tuple/tagged union oracle cases
- full `pnpm --filter ./tests/test-typia-automated start -- --include standardSchema_createValidate`
- focused Standard Schema unknown diagnostics, typia reporter boundaries, and OpenAPI reporter boundaries
- plain source typechecks for `test-utils-automated`, `test-typia-schema`, and the round-1 `test-utils` surface
- round-1 native command Go test and the implementation commit's broader package/root test baseline
- explicit changed-file Prettier checks and `git diff --check origin/master...HEAD`

The worktree is clean; local, origin, and PR head are the same SHA. The PR remains draft and was not made ready or merged.

PR #2068, opened 2026-07-16
fix(typia): preserve type and notation contracts

### LINKED ISSUE #2048: DeepPartial loses parameterized functions and readonly arrays
## Problem

Exported `DeepPartial<T>` documents that functions pass through unchanged and arrays recurse, but parameterized functions and readonly arrays do not satisfy its guards.

## Evidence

The function branch is `T extends (...args: unknown[]) => unknown`. Under strict function variance, `(x: string) => number` does not extend a callable accepting arbitrary `unknown`, so it falls into the object mapped type and becomes `{}`-like. The current test covers only `() => void`. Likewise `readonly T[]` misses `T extends Array<infer U>` and is mapped as an object rather than preserving readonly array structure. Exact searches found no duplicate.

## Consequence surface

Required/optional/rest/overloaded/generic functions, constructors, mutable/readonly arrays and tuples, unions, nested object properties, and `IJsonParseResult.IFailure<T>.data` consumers.

## Approach

Use a variance-safe callable/constructor test and explicit readonly array/tuple handling while preserving mutability and tuple shape where the public contract requires it.

- Type-level assertions for parameterized, overloaded/generic/rest functions and constructors.
- Mutable/readonly arrays and tuples recurse without losing readonly/tuple information.
- Existing primitive/object behavior remains unchanged; run `test-interface` and repository type/build tests.

## Coordination

Interface-only and safe to implement independently.

### LINKED ISSUE #2053: Resolved and notation return types corrupt readonly containers
## Problem

`Resolved<T>` and the four public notation return types recognize only mutable `Array`, `Set`, and `Map`. Readonly arrays, tuples, sets, and maps fall into generic mapped-object branches, producing unusable types that do not match the runtime containers returned by the operations.

## Evidence

`Resolved`, `CamelCase`, `PascalCase`, `SnakeCase`, and `KebabCase` each test `T extends Array<infer U>`, then mutable `Set`/`Map`. A compiler-API reproduction for `CamelCase<readonly { user_id: string }[]>` reduces to an object that maps `length` and every array method rather than `readonly { userId: string }[]`; function-valued methods become `never`. Other notation families also rename method keys such as `toString`, making the mismatch more severe. `Resolved<ReadonlySet<T>>` and `Resolved<ReadonlyMap<K,V>>` likewise map collection methods instead of preserving a resolved readonly collection. Readonly tuples fail their tuple branch for the same mutable-array gate.

`Resolved<T>` is the result type for plain clone, random, Protobuf decode, HTTP decoders, and other public operations; notation functions return the parallel aliases. Runtime paths recognize/map the corresponding containers, while existing type tests cover mutable containers only. Exact open/closed issue and pull-request searches found no duplicate.

## Consequence surface

Plain clone, random, Protobuf/HTTP result surfaces using `Resolved<T>`, all four naming conventions and their assert/is/validate/factory variants, readonly/mutable arrays and tuples, optional/rest tuple elements, nested readonly containers, ReadonlySet/ReadonlyMap, preservation vs normalization of readonly modifiers, unions, and assignability of generated runtime results.

## Approach

Model readonly containers before the generic object branch and recursively resolve/transform their element/key/value types. Preserve readonlyness where the public input-output contract promises a structural transform, or document and consistently type an intentional mutable output. Share the container decision shape across `Resolved` and the four notation aliases.

- `Resolved` and every notation alias turn readonly arrays/tuples/sets/maps into usable transformed container types, not mapped method objects.
- Mutable container behavior remains unchanged; tuple optional/rest structure is preserved.
- Compile-time equality/assignability cases match generated runtime values across all four notation families.
- Run focused interface type tests, notation transform/runtime tests, build/typecheck, and the full suite.

## Coordination

Shared interface type-level container ownership; can share a PR with the companion runtime-key and `DeepPartial` issues if the compile-time matrix remains reviewable, but none should be hidden behind another.

### LINKED ISSUE #2054: Notation conversion loses colliding keys and skips dynamic keys
## Problem

Runtime notation conversion silently loses literal properties whose source keys normalize to the same destination and does not rename dynamic/index-signature keys at all. Both behaviors contradict the advertised recursive key rewrite and can make the runtime result disagree with its mapped return type.

## Evidence

`NotationJoiner.Object` emits every literal property assignment after `Rename` without detecting destination collisions. For `{ user_id, userId }`, camel conversion emits duplicate `userId` assignments and the latter silently overwrites the former. The mapped type combines colliding source members under one destination key, so runtime value loss is not represented.

For dynamic entries, the joiner iterates `Object.entries(input)` but assigns `output[key] = value` with the original key. It uses the selected rename operation only for sole literal metadata. A `Record<string, { inner_key: string }>` therefore keeps each outer runtime key unchanged even though the API says all object keys are converted. Exact open/closed issue and pull-request searches found no duplicate.

## Consequence surface

All four naming conventions and assert/is/validate/factory variants, literal collisions, Unicode/acronym/underscore/hyphen cases, index signatures and template-literal keys, mixed literal/dynamic objects, nested values, symbol/numeric keys within supported metadata, enumeration order, and type/runtime agreement.

## Approach

Apply the selected rename function to every supported runtime string key. Detect destination collisions deterministically before data is lost and define one public policy: reject at transform/validation time with source paths, or preserve a documented representation that the return type can express. Do not rely on object-literal overwrite order.

- Dynamic string keys are converted by the selected notation at every nesting level.
- Literal, dynamic, and mixed collisions follow one explicit non-lossy or rejecting policy across all variants.
- Type aliases and runtime output remain assignable for collision and index-signature cases.
- Add direct/factory/assert/is/validate matrices and run focused native/runtime plus full tests; native changes bump `typia`.

## Coordination

Runtime notation ownership; coordinate with the readonly return-type issue for shared type/runtime fixtures.

### LINKED ISSUE #2055: SpecialFields and ClassProperties retain excluded properties
## Problem

The public `SpecialFields`/`OmitNever` key-filter chain only inspects string keys, and `ClassProperties<T>` also retains optional methods because their property type includes `undefined`. Numeric and unique-symbol impossible/method properties therefore survive the utilities that promise to remove them.

## Evidence

`SpecialFields` builds a mapped type over `keyof Instance` but indexes the result with `[keyof Instance & string]`. Numeric and unique-symbol keys are excluded even when their values match `Target`, so `OmitNever<T>` cannot omit non-string `never` members.

`ClassProperties` maps each property with `T[K] extends Function ? never : T[K]` and then calls `OmitNever`. For an optional method, `T[K]` is `(() => void) | undefined`; the non-distributive indexed-access conditional is false, so the method survives. Required numeric/symbol methods become `never` but also survive because `SpecialFields` never returns their keys:

```ts
class Example {
  value = 1;
  reset?(): void;
}
type Actual = ClassProperties<Example>;
// { value: number; reset?: (() => void) | undefined }
```

This contradicts all three exported contracts and can expose impossible or function-valued members in derived DTO types. Existing focused tests use string keys and required methods only. Exact open/closed issue and pull-request searches found no duplicate.

## Consequence surface

String, numeric, and unique-symbol keys; required and optional methods; public `SpecialFields` targets; `OmitNever`; function-valued data properties; overloads; getters/setters; inherited members; exact optional property types on/off; unions; abstract classes; and downstream uses in validators/serializers/schemas.

## Approach

Make `SpecialFields` return every matching `PropertyKey`, allowing `OmitNever` to remove non-string impossible members. Classify optional method keys after excluding absence (`Exclude<T[K], undefined>`) or with an equivalent key-remapping predicate. Preserve the documented distinction, if any, between prototype methods and intentional function-valued data properties; add compile-time fixtures before selecting that boundary.

- `SpecialFields` and `OmitNever` honor matching string, numeric, and unique-symbol keys.
- Optional and required methods at every key kind are absent from `ClassProperties<T>`.
- Optional data properties remain optional and unchanged.
- The policy for function-valued fields, accessors, overloads, and inheritance is explicit and tested.
- Run focused interface type tests, dependent package typechecks/build, and the full suite.

## Coordination

Small interface type-utility issue; compatible with the `DeepPartial` and readonly-container type fixes but independently verifiable.

samchon:
## Intent

Restore the published type-utility and notation contracts across mapped types, readonly containers, dynamic key rewriting, and property exclusion.

## Scope

- preserve callable, constructor, readonly-array, and tuple semantics in `DeepPartial`;
- keep `Resolved` and all four notation aliases aligned with mutable and readonly runtime containers;
- rename dynamic notation keys and reject normalized destination collisions without silent data loss;
- make `SpecialFields`, `OmitNever`, and `ClassProperties` honor all property-key kinds and optional methods.

## Deferred

None currently. The tests-first consequence trace may narrow a claim only if the issue contract proves broader than TypeScript or runtime semantics.

## Verification

Pending tests-first implementation and package-scoped local verification. Campaign Actions are intentionally cancelled by exact pushed SHA after each push; repository Actions settings remain enabled and unchanged.

Closes #2048
Closes #2053
Closes #2054
Closes #2055

### DIFF
--- a/packages/interface/src/typings/CamelCase.ts
+++ b/packages/interface/src/typings/CamelCase.ts
@@ -53,9 +53,11 @@ type CamelizeObject<T extends object> =
                 : T extends NativeClass
                   ? T
                   : {
-                      [Key in keyof T as CamelizeString<
-                        Key & string
-                      >]: CamelizeMain<T[Key]>;
+                      [Key in keyof T as Key extends string
+                        ? CamelizeString<Key>
+                        : Key extends number
+                          ? Key
+                          : never]: CamelizeMain<T[Key]>;
                     };
 
 type CamelizeArray<T extends readonly unknown[]> = {
--- a/packages/interface/src/typings/CamelCase.ts
+++ b/packages/interface/src/typings/CamelCase.ts
@@ -1,4 +1,5 @@
 import { Equal } from "./internal/Equal";
+import { IsBroadString } from "./internal/IsBroadString";
 import { IsTupleLike } from "./internal/IsTupleLike";
 import { NativeClass } from "./internal/NativeClass";
 import { ValueOf } from "./internal/ValueOf";
@@ -64,15 +65,16 @@ type CamelizeArray<T extends readonly unknown[]> = {
   [P in keyof T]: CamelizeMain<T[P]>;
 };
 
-type CamelizeString<Key extends string> = string extends Key
-  ? string
-  : Key extends `_${infer R}`
-    ? `_${CamelizeString<R>}`
-    : Key extends `${infer _F}_${infer _R}`
-      ? CamelizeSnakeString<Key>
-      : Key extends Uppercase<Key>
-        ? Lowercase<Key>
-        : CamelizePascalString<Key>;
+type CamelizeString<Key extends string> =
+  IsBroadString<Key> extends true
+    ? string
+    : Key extends `_${infer R}`
+      ? `_${CamelizeString<R>}`
+      : Key extends `${infer _F}_${infer _R}`
+        ? CamelizeSnakeString<Key>
+        : Key extends Uppercase<Key>
+          ? Lowercase<Key>
+          : CamelizePascalString<Key>;
 type CamelizePascalString<Key extends string> =
   Key extends `${infer F}${infer R}` ? `${Lowercase<F>}${R}` : Key;
 type CamelizeSnakeString<Key extends string> = Key extends `_${infer R}`
--- a/packages/interface/src/typings/CamelCase.ts
+++ b/packages/interface/src/typings/CamelCase.ts
@@ -1,5 +1,5 @@
 import { Equal } from "./internal/Equal";
-import { IsTuple } from "./internal/IsTuple";
+import { IsTupleLike } from "./internal/IsTupleLike";
 import { NativeClass } from "./internal/NativeClass";
 import { ValueOf } from "./internal/ValueOf";
 
@@ -13,8 +13,13 @@ import { ValueOf } from "./internal/ValueOf";
  * @author Jeongho Nam - https://github.com/samchon
  * @template T Target type to transform
  */
-export type CamelCase<T> =
-  Equal<T, CamelizeMain<T>> extends true ? T : CamelizeMain<T>;
+export type CamelCase<T> = unknown extends T
+  ? T
+  : object extends T
+    ? T
+    : Equal<T, CamelizeMain<T>> extends true
+      ? T
+      : CamelizeMain<T>;
 
 type CamelizeMain<T> = T extends [never]
   ? never // special trick for (jsonable | null) type
@@ -28,34 +33,34 @@ type CamelizeMain<T> = T extends [never]
 
 type CamelizeObject<T extends object> =
   T extends Array<infer U>
-    ? IsTuple<T> extends true
-      ? CamelizeTuple<T>
-      : CamelizeMain<U>[]
-    : T extends Set<infer U>
-      ? Set<CamelizeMain<U>>
-      : T extends Map<infer K, infer V>
-        ? Map<CamelizeMain<K>, CamelizeMain<V>>
-        : T extends WeakSet<any> | WeakMap<any, any>
-          ? never
-          : T extends NativeClass
-            ? T
-            : {
-                [Key in keyof T as CamelizeString<Key & string>]: CamelizeMain<
-                  T[Key]
-                >;
-              };
+    ? IsTupleLike<T> extends true
+      ? CamelizeArray<T>
+      : Array<CamelizeMain<U>>
+    : T extends ReadonlyArray<infer U>
+      ? IsTupleLike<T> extends true
+        ? CamelizeArray<T>
+        : ReadonlyArray<CamelizeMain<U>>
+      : T extends Set<infer U>
+        ? Set<CamelizeMain<U>>
+        : T extends Map<infer K, infer V>
+          ? Map<CamelizeMain<K>, CamelizeMain<V>>
+          : T extends ReadonlyMap<infer K, infer V>
+            ? ReadonlyMap<CamelizeMain<K>, CamelizeMain<V>>
+            : T extends ReadonlySet<infer U>
+              ? ReadonlySet<CamelizeMain<U>>
+              : T extends WeakSet<any> | WeakMap<any, any>
+                ? never
+                : T extends NativeClass
+                  ? T
+                  : {
+                      [Key in keyof T as CamelizeString<
+                        Key & string
+                      >]: CamelizeMain<T[Key]>;
+                    };
 
-type CamelizeTuple<T extends readonly any[]> = T extends []
-  ? []
-  : T extends [infer F]
-    ? [CamelizeMain<F>]
-    : T extends [infer F, ...infer Rest extends readonly any[]]
-      ? [CamelizeMain<F>, ...CamelizeTuple<Rest>]
-      : T extends [(infer F)?]
-        ? [CamelizeMain<F>?]
-        : T extends [(infer F)?, ...infer Rest extends readonly any[]]
-          ? [CamelizeMain<F>?, ...CamelizeTuple<Rest>]
-          : [];
+type CamelizeArray<T extends readonly unknown[]> = {
+  [P in keyof T]: CamelizeMain<T[P]>;
+};
 
 type CamelizeString<Key extends string> = Key extends `_${infer R}`
   ? `_${CamelizeString<R>}`
--- a/packages/interface/src/typings/CamelCase.ts
+++ b/packages/interface/src/typings/CamelCase.ts
@@ -62,13 +62,15 @@ type CamelizeArray<T extends readonly unknown[]> = {
   [P in keyof T]: CamelizeMain<T[P]>;
 };
 
-type CamelizeString<Key extends string> = Key extends `_${infer R}`
-  ? `_${CamelizeString<R>}`
-  : Key extends `${infer _F}_${infer _R}`
-    ? CamelizeSnakeString<Key>
-    : Key extends Uppercase<Key>
-      ? Lowercase<Key>
-      : CamelizePascalString<Key>;
+type CamelizeString<Key extends string> = string extends Key
+  ? string
+  : Key extends `_${infer R}`
+    ? `_${CamelizeString<R>}`
+    : Key extends `${infer _F}_${infer _R}`
+      ? CamelizeSnakeString<Key>
+      : Key extends Uppercase<Key>
+        ? Lowercase<Key>
+        : CamelizePascalString<Key>;
 type CamelizePascalString<Key extends string> =
   Key extends `${infer F}${infer R}` ? `${Lowercase<F>}${R}` : Key;
 type CamelizeSnakeString<Key extends string> = Key extends `_${infer R}`
--- a/packages/interface/src/typings/CamelCase.ts
+++ b/packages/interface/src/typings/CamelCase.ts
@@ -18,37 +18,47 @@ export type CamelCase<T> = unknown extends T
   ? T
   : object extends T
     ? T
-    : Equal<T, CamelizeMain<T>> extends true
-      ? T
-      : CamelizeMain<T>;
+    : T extends readonly unknown[]
+      ? CamelizeMain<T> // avoid eagerly comparing recursive tuple rest aliases
+      : Equal<T, CamelizeMain<T>> extends true
+        ? T
+        : CamelizeMain<T>;
 
-type CamelizeMain<T> = T extends [never]
+// TupleStack closes recursive tuple rest cycles without limiting other nesting.
+type CamelizeMain<T, TupleStack = never> = T extends [never]
   ? never // special trick for (jsonable | null) type
   : T extends { valueOf(): boolean | bigint | number | string }
     ? ValueOf<T>
     : T extends Function
       ? never
       : T extends object
-        ? CamelizeObject<T>
+        ? CamelizeObject<T, TupleStack>
         : T;
 
-type CamelizeObject<T extends object> =
+type CamelizeObject<T extends object, TupleStack> =
   T extends Array<infer U>
     ? IsTupleLike<T> extends true
-      ? CamelizeArray<T>
-      : Array<CamelizeMain<U>>
+      ? T extends TupleStack
+        ? T
+        : CamelizeArray<T, TupleStack | T>
+      : Array<CamelizeMain<U, TupleStack>>
     : T extends ReadonlyArray<infer U>
       ? IsTupleLike<T> extends true
-        ? CamelizeArray<T>
-        : ReadonlyArray<CamelizeMain<U>>
+        ? T extends TupleStack
+          ? T
+          : CamelizeArray<T, TupleStack | T>
+        : ReadonlyArray<CamelizeMain<U, TupleStack>>
       : T extends Set<infer U>
-        ? Set<CamelizeMain<U>>
+        ? Set<CamelizeMain<U, TupleStack>>
         : T extends Map<infer K, infer V>
-          ? Map<CamelizeMain<K>, CamelizeMain<V>>
+          ? Map<CamelizeMain<K, TupleStack>, CamelizeMain<V, TupleStack>>
           : T extends ReadonlyMap<infer K, infer V>
-            ? ReadonlyMap<CamelizeMain<K>, CamelizeMain<V>>
+            ? ReadonlyMap<
+                CamelizeMain<K, TupleStack>,
+                CamelizeMain<V, TupleStack>
+              >
             : T extends ReadonlySet<infer U>
-              ? ReadonlySet<CamelizeMain<U>>
+              ? ReadonlySet<CamelizeMain<U, TupleStack>>
               : T extends WeakSet<any> | WeakMap<any, any>
                 ? never
                 : T extends NativeClass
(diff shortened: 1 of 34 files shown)


samchon:
## Pre-edit Self-Review: findings and tests-first plan

Complete declared surface: issues #2048, #2053, #2054, and #2055; public interface aliases and their consumers; all notation direct/assert/is/validate and factory transforms; native metadata/object joining; internal fallback notation helpers; public API JSDoc and website guide; interface, native-command, runtime, packaging, and version consequences.

### Findings

1. `DeepPartial<T>` uses a contravariant `unknown[]` callable predicate and a mutable-array-only branch. It therefore rewrites parameterized/generic/overloaded callables and constructors as objects, widens mutable tuples to arrays, and maps readonly arrays/tuples as generic objects. The invariant is to pass callable/construct signatures through unchanged and use a homomorphic tuple/array transform that preserves mutability, optional elements, labels, and variadic rest structure.
2. `Resolved<T>` and Camel/Pascal/Snake/Kebab aliases gate on mutable arrays, sets, and maps. Readonly containers fall into generic object key mapping, while recursive tuple helpers also lose readonly and variadic structure. The invariant is mutable-in/mutable-out, readonly-in/readonly-out, with homomorphic tuple recursion and recursive container members.
3. `NotationJoiner.Object` renames only statically known keys, assigns dynamic entries under the source key, and emits colliding literal destinations without a diagnostic. `_notationAny` uses `Object.fromEntries` and has the same silent overwrite. Literal collisions can be rejected during metadata analysis; runtime dynamic/mixed collisions require an own-property check before assignment. All four rename helpers must be available for dynamic keys, including a missing kebab helper.
4. `SpecialFields` discards numeric and symbol candidates at its final string-only index. `ClassProperties` applies a non-distributive function check to optional members, so `undefined` hides optional method/function types. Property-key selection must retain all `PropertyKey` kinds; class filtering must classify the present value while retaining optional non-function data. Type structure cannot distinguish a method from a function-valued field, so the existing documented function-typed-member policy remains the boundary.
5. Native source will change, so the `typia` package version must advance in the same PR. Concurrent native batches own the same line; the branch must refresh before merge and take the next valid version.

### Plan

1. Add one-case-per-file compile-time regression fixtures for callable/constructor and readonly/variadic arrays, readonly `Resolved`/notation containers, and string/numeric/unique-symbol property exclusion.
2. Add command-level transformations that exercise dynamic renaming and dynamic collisions across all four naming families and direct/assert/is/validate plus factory variants; add a separate literal-collision diagnostic matrix.
3. Run the new interface and native-command cases before implementation and retain the expected red evidence.
4. Implement shared semantic fixes without fixture-specific branches, update the native package version and collision documentation, then run focused interface/native/runtime tests and broader applicable gates.
5. Start a fresh whole-diff Self-Review after every remediation and stop only on a clean complete round.

samchon:
## Self-Review round 1 findings and remediation plan

I reread the complete PR diff across the interface aliases, native notation emitter, runtime helpers, generated-command tests, and documentation.

Findings:

1. `_notationAny` previously used `Object.fromEntries`, which defines data properties safely. The new shared assignment helper writes through `output[destination]`; a destination such as `__proto__` can invoke the legacy prototype setter instead of creating an enumerable own property. Static object-literal emission has the same special-key risk. This violates the dynamic-key preservation contract and can mutate the output prototype.
2. The dynamic collision matrix only requires that some exception is thrown. An unrelated runtime exception would satisfy it, so it does not prove the documented collision error names both source keys and the normalized destination. Its valid cases also do not exclude leftover source keys.
3. The Pascal dynamic-key type assertion uses two directional checks while the other naming families use one semantic-shape check, and the readonly-container test comment still describes the old mutable-only behavior.

Plan before editing:

- Define renamed keys with data-property semantics and emit static keys as computed properties, preserving `__proto__`-like names without prototype mutation.
- Extend the native runtime matrix with special-key preservation, exact output-key checks, and dynamic collision-message assertions.
- Make the Pascal index-signature test assert its broad key and indexed value directly, and correct the stale test prose.
- Re-run the interface suite/build, typia compile/build, and focused native notation tests before pushing the remediation SHA.

samchon:
## Self-Review round 2 findings and remediation plan

I reread the complete updated diff after round 1 and traced the remaining mapped-type and collision branches.

Findings:

1. `SpecialFields` preserves optional modifiers in its intermediate mapped type. Indexing that map can therefore add `undefined` to the selected-key union even when an optional property does not match `Target`. Expanding the final index to every property-key kind must also remove that optionality leak.
2. The runtime matrix now covers literal/literal and dynamic/dynamic destination collisions, but it does not directly pin the literal/dynamic boundary where the generated `sources` table is responsible for detecting a collision.

Plan before editing:

- Remove optional modifiers from the `SpecialFields` selection map and add an optional non-matching property to the property-key regression.
- Add a mixed explicit-property/index-signature case whose runtime key collides with the pre-emitted literal destination, and require the normal collision error details.
- Re-run the interface suite/build and focused native notation tests before pushing the next SHA.

samchon:
## Self-Review round 3 findings and remediation plan

I reread the complete updated diff again and checked it against issue #2054's explicit consequence surface.

Findings:

1. Runtime `Object.entries` preserves numeric property names as string entries, and notation conversion leaves digit-only names unchanged. The four mapped aliases still use `Key & string`, which removes numeric literal/index keys from the return type and creates a type/runtime disagreement.
2. Broad string index signatures exercise the dynamic branch, but the template-literal metadata pattern branch named by the issue is not directly pinned. A regression there could pass the current broad-record matrix.
3. `DeepPartial` now preserves mutable/readonly arrays and tuple positions homomorphically, while its public JSDoc still says only that an array element type is transformed.

Plan before editing:

- Preserve numeric keys in all four notation aliases while continuing to exclude symbol keys that `Object.entries` does not convert; add numeric literal/index-signature type cases.
- Add focused runtime cases for numeric index keys and a template-literal index signature with nested key conversion.
- Update the `DeepPartial` behavior note to state its array/tuple shape-preservation contract.
- Re-run the interface suite/build, typia compile/build, and focused native notation tests before the next push.

samchon:
## Self-Review round 4 findings and remediation plan

I reread the updated diff and compared the fixtures line-by-line with the acceptance and consequence surfaces of issues #2048, #2053, #2054, and #2055.

No new implementation defect was found, but four acceptance branches are only implied rather than directly locked:

1. The template-literal metadata runtime path is covered, but the mapped return type is not indexed at a representative converted template key.
2. `DeepPartial` tests callable types directly but do not pin nested callable properties or distributive callable/object unions.
3. `ClassProperties` covers required/optional function fields and methods at all key kinds, but not overloaded or inherited methods named by the issue surface.
4. The readonly notation fixture distributes array/tuple/set/map coverage across the four families. A family-specific branch regression could therefore escape; each notation alias should traverse one composite containing all four readonly container kinds.

Plan before editing:

- Add compile-time assertions for the converted template key, nested/union `DeepPartial`, and overloaded/inherited `ClassProperties` behavior.
- Replace the distributed readonly notation checks with a compact composite tuple asserted independently for camel, Pascal, snake, and kebab, while retaining mutable-container and optional-only tuple checks.
- Re-run the interface suite/build and dependent typia compile before the next push.

samchon:
## Self-Review round 5: clean

I reread the complete latest PR diff from the claim commit through `89ee29d0d85c500192947c5064ea1105fa8facff`, including every interface alias/helper, native emitter and runtime helper, generated-command fixture, public JSDoc, website paragraph, package version, and regression test.

I also rechecked the final behavior against all acceptance and consequence branches in #2048, #2053, #2054, and #2055: callable/constructor/nested/union `DeepPartial`; mutable and readonly array/tuple/set/map shapes in `Resolved` and all notation aliases; literal, dynamic, mixed, numeric, template, nested, special-key, and all-variant notation behavior; and string/number/symbol plus optional/overloaded/inherited property exclusion.

No sound improvement or remaining defect was found in this round. `git diff --check` is clean, and the focused type, package, and native runtime gates cited in the preceding rounds pass.

samchon:
## Broad verification finding after the clean review round

The root `pnpm build` passed. The serial `test-*` gate then exposed a real type regression in six existing random-test files (20 diagnostics): `ResolvedArray<string[] & MinItems<1>>` and similar recursive tagged arrays had their `length` property transformed to the element type.

Root cause:

- `IsTupleLike` currently treats any key outside `keyof any[]` as evidence of a tuple position.
- Typia tag intersections such as `Array<T> & MinItems<1>` add metadata keys, so ordinary branded arrays are misclassified as tuples and homomorphically map every array property.

Plan before editing:

- Recognize tuples only through an empty tuple or explicit numeric-string tuple positions (`"0"`, `"1"`, and so on), excluding unrelated brand/tag keys.
- Add a focused branded-array assertion to the interface fixture.
- Update the reserved native package version to the campaign's current serialized slot, `13.1.8`.
- Re-run the focused interface/build gates, the previously failing automated suite, then the full `test-*` and Go toolchain gates.

The initial parallel root test attempt also left sibling test processes holding generated directories; those orphan processes were terminated and the actionable diagnostics above came from a clean serial rerun, not from the earlier `EBUSY` infrastructure failure.

samchon:
Broad verification follow-up: the serial test-* gate exposed TS2589 in test_random_tuple_rest because homomorphic tuple mapping revisited a recursive rest alias indefinitely. The fix bypasses eager Equal comparison for top-level arrays and carries a tuple-cycle stack through Resolved and all four notation aliases, closing only repeated tuple nodes while leaving ordinary nesting unlimited. Added compile-time coverage for recursive tuple heads across all five helpers. Focused test-interface and test-typia-schema now pass; the full gate will be rerun after this change.

samchon:
Native toolchain verification finding: TestNotationKebabCaseTransform still searched for the old non-computed emitted fragment xmlparser:, while this PR intentionally emits every static renamed key as a computed string property to make special keys such as __proto__ safe. The converter, compile-time fixture, runtime runner, and naming-convention tests all agree that the key value is xmlparser. I will update only this emitted-shape assertion to expect the computed property fragment, then rerun the focused native test and full toolchain gate.

samchon:
## Broad local verification on inherited head\n\nAfter inheriting the batch at \267903786fddd3eb04649a699951ab0f6035cf12\, I reran the complete applicable local gates from the isolated worktree:\n\n- \pnpm build\ — passed (all six published packages)\n- \pnpm test:packages\ — passed (all 12 \	est-*\ workspaces, including the full automated matrix and generation/CLI suites)\n- \pnpm test:toolchain\ — passed (public Go contracts and every native Go package; the notation command tests passed)\n- \git diff --check\ — clean\n\nI also completed a fresh solo whole-diff inspection over all 33 changed files after those gates. No additional sound improvement was found on the current head. A final from-scratch review and full integrated gate will be repeated after the required semantic rebase on the serialized native landing chain (#2071, #2075, #2074, then #2077).

samchon:
Final tagged-native verification found one actionable integration failure after rebasing the serialized campaign: TestJoinerBranchCoverage still constructs NotationJoiner_ObjectProps without the new dynamic rename/assign callbacks, so the tagged-only test panics while building the dynamic-entry AST. I will update that internal coverage fixture to provide the required callback nodes, then rerun the focused tagged package, the full tagged-native suite, build, root test, and a fresh whole-diff Self-Review before delivery.

samchon:
## Final rebased Self-Review finding and remediation plan

I reopened the entire 34-file current diff after rebasing onto master `05f6331ec1d2046ecf0cb0ccbd6d4479f9637da5` and reproduced one remaining tuple boundary defect.

`IsTupleLike<T>` recognizes tuples only through fixed numeric-string keys. A suffix-only variadic tuple such as `[...Item[], Tail]` has a broad numeric index and a required trailing position but no fixed `"0"` key, so `Resolved` and every notation alias currently classify it as a homogeneous array and discard the required tail. This violates #2053's requirement to preserve optional/rest tuple structure.

Remediation:

- recognize a required tuple suffix separately from fixed numeric keys while continuing to exclude ordinary and tagged homogeneous arrays;
- extend the existing variadic-boundary fixture with mutable and readonly suffix-only cases across `Resolved` and all four notation families;
- rerun the focused interface gate, build, full root/package/toolchain gates, tagged-native suite, website package/build gate, and then restart the whole-diff Self-Review from scratch.

samchon:
## Final rebased Self-Review addendum

The tuple suffix remediation is behaviorally correct and its focused compile gate passes, but the new internal helper comment now understates its own classifier: it says variadic tuples are recognized through explicit index keys, while suffix-only variadics intentionally have no such key and are recognized through their required trailing element.

I will align that comment with the implemented invariant, rerun the focused compile/diff check, push the documentation-only remediation, pass the exact-SHA cancellation gate, and then restart the complete review round.

samchon:
## Final rebased Self-Review: clean

I reread every changed file and hunk on `58b8320a79b07d8809c6c1195af08d935640b393` after the suffix-only variadic and classifier-documentation remediations. The review traced the public type projections, runtime and generated notation paths, native transformer behavior, collision handling, documentation, package version, and downstream website/compiler boundary. No further sound improvement remains in the declared #2048/#2053/#2054/#2055 scope.

Final verification:

- `pnpm --filter ./tests/test-interface start`
- `go -C packages/typia/test test ../native/core/programmers/helpers/... ../native/core/programmers/notations/... ../native/cmd/ttsc-typia/...`
- `go -C packages/typia/test test -tags typia_native_internal ../native/...`
- root `pnpm test` (all package workspaces, every generate mode, public Go, and native Go)
- root `pnpm install --frozen-lockfile`
- root `pnpm build`
- root `pnpm package:tgz` (`typia@13.1.8`)
- website `pnpm run test:compiler` (14/14 assertions)
- `go -C website/compiler test ./cmd/playground`
- website `pnpm build` (Wasm, examples, 74 static pages, Pagefind, and sitemap)
- `git diff --check`

The local sibling `ttsc` checkout was on unreleased `0.19.1` master and initially exposed a host-plugin API mismatch. The repository's website and release workflows explicitly pin the sibling checkout to the installed `@ttsc/wasm` version, so the compiler tests and production build were rerun against an isolated exact `ttsc v0.18.4` checkout (`b5ead8f8629a5f0442a04bb5f6c2c6ad6c2177b1`) matching the lockfile and CI contract. No out-of-scope compatibility patch was added.

The worktree is tracked-clean, `origin/master` remains `05f6331ec1d2046ecf0cb0ccbd6d4479f9637da5`, and that base is an ancestor of the PR head.

Issue #2035, opened 2026-07-16
Automated tests skip direct matrices and mask asynchronous failures


samchon:
## Problem

The generated automation harness skips the direct-operation matrix, can exit successfully after uncaught asynchronous failures, and uses equivalence helpers that deliberately approve whole fixture classes without comparing their results.

## Evidence

`tests/test-typia-automated/src/build/TestAutomationController.ts` calls `generateFeatureSet(..., false)` for every non-create-only template, but the function immediately returns when `create === false`. Twenty active direct public-operation families therefore lack the declared structure matrix while their `createX` variants run it.

Separately, `tests/test-typia-automated/src/index.ts` and `tests/test-utils-automated/src/index.ts` install `uncaughtException` and `unhandledRejection` handlers that only log. Installing those handlers suppresses Node's default nonzero termination; a minimal subprocess using the exact handler shape printed `Success` and exited 0 after each event.

Both behaviors were introduced in the monorepo migration `a7cbc4f1ae`. Closed unmerged PR #1864 recorded the direct-matrix gate but did not resolve it. No issue or pull request covers the asynchronous exit defect.

These are repository-integrity defects: control flow deliberately bypasses a requested test branch, and global failure handlers convert real test failures into green processes.

The active `resolved_equal_to` and `protobuf_equal_to` helpers add a third bypass: they return `true` whenever the fixture name contains `Class`; `resolved_equal_to` also approves any comparison involving a function. Six current class fixtures therefore execute clone/Protobuf code without an equivalence oracle. Both helpers compare only `Object.entries(x)`, so unexpected output properties are invisible. Native comparisons are also shallow: `DataView` checks only backing-buffer length and Blob/File only constructors, hiding byte-range, content, and metadata corruption. The class-name bypass was deliberately introduced in `6e314ed625b` and is tied to fixture naming rather than a documented semantic exception.

## Consequence surface

- direct `is`, assert/guard/validate and equals variants, random, protobuf, JSON, HTTP, and plain operations;
- worker/servant/timer/cleanup failures before, during, or after normal completion;
- generation counts, file naming collisions, run time, and product defects exposed when direct coverage returns.
- class fixtures, unexpected output properties, function-bearing shapes, DataView ranges/content, Blob/File bytes/metadata, and every operation that reuses the two equivalence helpers.

## Approach

[...]

PR #2070, opened 2026-07-16
fix(plain): clone supported native values safely

### LINKED ISSUE #2047: plain clone and classify mishandle supported binary natives
## Problem

`plain.clone` and `plain.classify` mishandle several supported native values:

- `DataView` is reconstructed from `input.buffer` only, so the output shares storage and expands a sliced view to the whole backing buffer;
- `Blob` and `File` fall through the native decoder's default branch, which emits a zero-argument function call such as `Blob()`/`File()` instead of constructing a copy and throws on standard runtimes.
- dynamic/`any` `RegExp` values have no `_plainCloneAny` branch and become `{}`, while dynamic `File` cloning drops `lastModified`.

## Evidence

Native `PlainCloneProgrammer.go` and `PlainClassifyProgrammer.go` emit `new DataView(input.buffer)`. The any fallback `_plainCloneAny` copies the full buffer but likewise discards `byteOffset`/`byteLength`. Mutating a cloned view can therefore mutate the source.

`NativeSimple` constructs only a zero-byte, full-buffer view and its oracle compares buffer length, so both `DataView` defects are hidden. Metadata explicitly recognizes `Blob`, `File`, and `RegExp` as natives. Both plain programmers' native switches omit Blob/File and fall through to a zero-argument `NewCallExpression(Type)`. The generic `_plainCloneAny` helper demonstrates Blob/File construction but omits `File.lastModified`, and because it lacks a RegExp branch a dynamic regex reaches `Object.entries` and becomes `{}`. Automated clone generation excludes the non-JSONable native fixture. Closed PR #1864 touched only part of the any fallback and no issue/open fix exists.

## Consequence surface

Typed/native and `any` paths, classify vs clone, offset/length views, ArrayBuffer and SharedArrayBuffer, `Blob`/`File` content/type/name/lastModified, RegExp source/flags/lastIndex, mutation independence, zero-length values, nested objects/collections, browser/Node constructors, and generated direct/factory operations.

## Approach

Clone exactly the viewed `DataView` bytes into independent storage and construct an equivalent view. Add explicit `Blob`/`File`/`RegExp` branches that preserve observable metadata and content. Share one helper between clone/classify/any paths where practical and define SharedArrayBuffer handling explicitly.

- A nonzero-offset, short `DataView` preserves visible bytes/length, exposes no outside bytes, and shares no storage.
- `Blob`/`File` clone and classify return valid independent instances with identical bytes and preserved `type`, plus `name`/`lastModified` for `File`.
- Dynamic/native RegExp values preserve source, flags, and the documented `lastIndex` policy instead of becoming plain objects.
- Mutations in either source or result do not affect the other where the native is mutable.
- Cover zero/full/sliced views, ArrayBuffer/SharedArrayBuffer, empty/non-empty Blob/File, any/native/nested/union paths.
- Strengthen `NativeSimple` and generated operation tests; run focused native/plain and full suites.

## Coordination

Native source change requires typia version bump; compatible with other plain-programmer work but independent of UniqueItems' DataView equality.

samchon:
## Intent

Make `plain.clone` and `plain.classify` preserve the observable value, range, and independence of every supported binary/native value instead of widening views, sharing storage, emitting invalid calls, or collapsing dynamic natives into plain objects.

## Scope

- clone `DataView` values from exactly their visible byte range into independent storage, including zero-length, full-range, sliced, ArrayBuffer-backed, and SharedArrayBuffer-backed inputs where supported;
- construct equivalent `Blob`, `File`, and `RegExp` values in typed and dynamic/`any` paths while preserving content and observable metadata;
- trace direct and factory clone/classify operations, nested/union paths, cross-realm and prototype semantics, and all helper callers;
- add content, range, mutation-independence, metadata, boundary, and negative-twin regressions before changing the implementation;
- bump the `typia` package version because shipped native Go source changes.

## Deferred

- No unrelated native metadata, equality, recovery, serializer, or validator change.
- No automation-harness repair owned by #2035 beyond focused regression coverage needed to prove #2047.
- No repository-wide formatting; the campaign Post-Campaign Cleanup phase owns that result.

## Verification

- tests-first regression failed on `Blob` construction before implementation and passed after the fix;
- all five focused plain-native cases pass across DataView/buffers, Blob/File, RegExp, collections, intrinsic brands, cross-realm values, unions, and negative twins;
- `pnpm build` passes for all packages;
- `pnpm test` passes all 12 test workspaces plus public and native Go toolchain suites;
- `NativeSimple` automated focus and the full schema suite pass;
- three solo Self-Review rounds were recorded on the PR, with both findings rounds remediated and the final round clean;
- after rebasing the benchmark-only `master` at `1cda8afc4e7795abc92e285212d128e942049410`, the focused group and `git diff --check` pass again.

Repository Actions remain enabled. Campaign runs were cancelled only for this branch's exact pushed SHA after every push or pull-request event, with two consecutive quiet polls required at each gate.

Closes #2047


samchon:
## Self-Review round 1

Base reviewed: `origin/master` at `fe05f28384094d030780b330e05781b03f8a3bbf`

Head reviewed: `0238616c85c007055dd3083f01f5110e1aaa0142`

### Findings

1. `tests/test-typia-schema/src/features/plain/test_plain_native_clone.ts` combines DataView/buffer, Blob/File, RegExp, cross-realm, dynamic, union, and negative-twin cases in one 539-line exported test. New and changed tests must keep one test case per file and name the file after the assertion surface.
2. `_plainCloneAny.cloneNative` routes on `Object.prototype.toString.call(value)` before applying intrinsic brand checks. A real supported native with an own `Symbol.toStringTag` can therefore fall through to plain-object cloning, and a plain object's `Symbol.toStringTag` getter is unnecessarily executed during clone. The implementation should derive brands only from intrinsic methods/getters while retaining cross-realm support and prototype-impostor rejection.

### Remediation plan

- Replace generic `Object.prototype.toString` routing with intrinsic brand probes. Use the intrinsic typed-array `@@toStringTag` getter for the concrete typed-array constructor and intrinsic getters/methods for Date, ArrayBuffer, SharedArrayBuffer, DataView, File, Blob, and RegExp.
- Add regressions proving an own `Symbol.toStringTag` does not demote a valid native and a throwing spoof getter is not invoked.
- Split the mixed regression into independently named one-case-per-file tests with a non-test shared helper module, preserving the existing direct/factory, clone/classify, dynamic, union, boundary, cross-realm, mutation-independence, and negative coverage.
- Run the focused plain-native group, affected package/native gates, then restart Self-Review from the full base-to-head diff.

samchon:
## Self-Review round 2

Base reviewed: `origin/master` at `fe05f28384094d030780b330e05781b03f8a3bbf`

Rebased head reviewed: `ddda7b70b4e76c5bf58e555efafb4b5db670b9e9`

### Finding

The issue consequence surface explicitly includes native values nested in collections. The split regression covers nested ordinary objects, typed/native unions, and root/dynamic values, but it no longer proves `Set` and `Map` recursion for the corrected native branches.

### Remediation plan

- Add one independently named collection test file.
- Exercise direct/factory clone and classify with a typed `Set<DataView>` and `Map<string, File>`, plus a dynamic map containing Blob and RegExp values.
- Reuse the shared content/metadata/range/independence assertions and keep production code unchanged.
- Run the focused group, push through the exact-SHA cancellation gate, and restart the complete Self-Review on the new head.

samchon:
## Self-Review round 3: clean

Base reviewed: `origin/master` at `e883425adfe080e9734ec4f5ff102635028c2218`

Head reviewed: `1b117463c964b20e00cca9c7561e3c4dfc53819f`

No actionable findings remain.

The round restarted from the complete base-to-head diff after the final rebase and collection remediation. It rechecked both Go emitters, their shared internal-import path, the dynamic clone helper's intrinsic branding and runtime availability guards, exact DataView range/backing semantics, Blob/File metadata and content, RegExp state policy, typed/dynamic/union/object/collection recursion, cross-realm behavior, prototype and toStringTag negatives, NativeSimple consequence coverage, the native package version bump, public deep-copy documentation, package boundaries, and all changed tests against the one-case-per-file rule.

The branch is based directly on `e883425adfe080e9734ec4f5ff102635028c2218`, `git diff --check` is clean, the PR reports mergeable, and the only worktree residue is the disposable ignored Go cache.

PR #2071, opened 2026-07-16
fix(openapi): enforce object safety invariants

### LINKED ISSUE #2032: UniqueItems semantics diverge across validators and random generation
## Problem

The documented `UniqueItems` contract—strict equality for primitives and deep structural equality for objects—is inconsistent across runtime validation, OpenAPI validation, JSDoc tags, and random generation. Some distinct values are rejected, some structural duplicates pass, and an unsatisfiable random request can loop forever.

## Evidence

Current helpers in typia and utils:

```text
_isUniqueItems([0, {x:1}, {x:1}])       => true   (duplicate passes)
_isUniqueItems([[1], [1,2]])             => false  (distinct arrays rejected)
_isUniqueItems([{x:undefined},{y:undefined}]) => false
```

Root causes include first-element-only shallow dispatch, array comparison without a length check, object comparison without own-key membership, and offset-blind `DataView` comparison. JSDoc `@uniqueItems` emits a shallow `Set` check instead of the deep helper. `_randomArray` also uses reference identity: it can return objects that validation rejects as duplicates and never terminate when requested cardinality exceeds a finite domain.

Issues #1062/#1138 establish the deep-validation and random-generation contracts. Closed unmerged PR #1864 fixed only a subset (array/DataView and an attempt cap); no open claim covers the full class.

## Consequence surface

- `tags.UniqueItems`, JSDoc `@uniqueItems`, and `OpenApiValidator`;
- primitives, mixed unions, plain/nested objects, arrays in both orders, maps/sets, buffers/views, files/blobs, and cycles;
- `typia.random` with structural domains and finite primitive domains;
- LLM/HTTP consumers that rely on OpenAPI `uniqueItems`.

## Approach

Define one deterministic structural-equality invariant and keep typia/utils copies behaviorally identical. Generated JSDoc validation must call the same owner. Random generation must apply the same uniqueness relation and terminate with a defined diagnostic/outcome when the requested domain is impossible.

- Add positive/negative/order-invariant cases for every affected container class and mixed unions.
- Prove JSDoc/type-tag/OpenAPI parity.
- Prove generated objects cannot contain structural duplicates and a deterministic unsatisfiable generator terminates.
- Preserve cycle safety and compare views by their visible bytes.
- Run focused typia/utils/random suites, native tests if generation changes, and `pnpm test`.

## Coordination

This touches shared runtime semantics and random generation; it should not be split into separate fixes that leave the documented contract divergent.

### LINKED ISSUE #2042: Lenient JSON and schema maps are prototype-sensitive
## Problem

Lenient JSON parsing, LLM coercion, schema conversion, and reference resolution use ordinary `{}` dictionaries plus prototype-sensitive reads/writes. Reserved keys can mutate a parsed value's prototype, become inherited tool arguments, or make missing schema references look resolved.

## Evidence

Public current-source reproduction:

```text
LlmJson.parse('{"__proto__":{"admin":true},}')
  success=true, own __proto__=false, prototype.admin=true
LlmJson.coerce(parsed, schema requiring admin)
  own admin=true; validation succeeds
OpenApiValidator($ref '#/components/schemas/toString', components.schemas={})
  success=true for arbitrary input
```

`parseLenientJson.parseObject()` assigns `result[key] = value` into `{}`. `coerceObject()` uses `key in value` and copies inherited values into a fresh ordinary object. Object validation reads `value[key]` without own-property membership. Separately, `LlmSchemaConverter` initializes `$defs = {}` and uses `props.$defs[key] !== undefined`; component names `constructor`, `toString`, and `__proto__` collide with inherited members, leaving dangling `$ref`s. Multiple resolvers similarly read `components.schemas?.[key]` without `Object.hasOwn`, and `OpenApiStationValidator` treats a genuinely missing reference as valid.

Exact issue/PR searches found no duplicate.

## Consequence surface

Malformed/lenient and native JSON paths, `__proto__`/`constructor`/`prototype`/`toString`, nested objects, additional properties, required LLM tool arguments, `$defs` conversion/inversion, OpenAPI components, missing/cyclic references, HTTP/LLM/MCP dispatch, and null-prototype user inputs.

## Approach

Treat parsed JSON objects and schema maps as data dictionaries: define own-property membership and safe assignment centrally, use null-prototype containers or equivalent creation where appropriate, and make unresolved references fail deterministically rather than validate as unknown. Do not blacklist only the witnessed key.

- Reserved keys round-trip as own JSON data without mutating prototypes.
- Inherited values never satisfy/copy a declared tool parameter.
- Reserved component names convert and resolve normally; missing references report a validation/conversion failure.
- Cover native JSON success, lenient fallback, nested arrays/objects, coercion, validator, converter/inverter, and cyclic references.
- Run focused utils/LLM/MCP tests and the full suite.

## Coordination

Security-sensitive correctness issue; keep the parser/coercer/schema-dictionary fixes in one batch so no bypass chain remains half-fixed.

### LINKED ISSUE #2043: OpenApiValidator and 3.1 tuple conversion omit core invariants
## Problem

`OpenApiValidator` and the OpenAPI 3.1 upgrader omit core type/tuple invariants:

- object schemas accept arrays and other non-plain JSON objects;
- unbounded number/integer schemas accept `NaN`/`±Infinity`;
- tuple validation ignores `minItems`, `maxItems`, and `uniqueItems`;
- tuple child validation inherits an optional parent's `required: false`, accepting explicit `undefined` elements;
- fixed-prefix logic treats every prefix item as mandatory instead of respecting `minItems`;
- OpenAPI 3.1 `prefixItems` conversion drops the standard `items` rest schema and consults deprecated `additionalItems` instead.

## Evidence

A public `OpenApiValidator` probe at current master returned success for `[]` against an object schema, `NaN` against number, `Infinity` against integer, two rest strings against `maxItems: 1`, and `[undefined]` against a string prefix when the root was optional. `OpenApiTupleValidator.ts` never reads the three array constraints and omits `required: true` on element calls. `OpenApiV3_1Upgrader.ts` sets `items: undefined` whenever `prefixItems` exists and maps only `schema.additionalItems`.

Closed PR #1864 observed the non-finite numeric gap but remained unmerged; no issue/open implementation covers this full class.

## Consequence surface

Root/property requiredness, object-vs-array/native values, JSON numeric domain, fixed/optional/variadic tuples, unique items, OpenAPI 3.1 conversion, LLM parameters, HTTP migration, and strict/non-strict (`equals`) validation. The separate `equals` option intentionally owns rejection of extra object properties and is not changed here.

## Approach

Make each schema kind enforce its intrinsic JSON type and constraints. Tuple requiredness is determined per index/range, not inherited from the containing property's optionality. Preserve 3.1 `items` as the internal tuple rest schema.

- Object schemas reject arrays, Date, Map, and other non-JSON-object instances while accepting null-prototype records.
- Number/integer reject non-finite values and preserve valid bound behavior.
- Tuple min/max/unique, optional prefix, required elements, false/schema rest, and empty cases match before/after 3.1 conversion.
- Add public validator, converter round-trip, LLM/HTTP consumer tests and run utils/full suites.

## Coordination

UniqueItems' deep equality fix is a dependency for tuple `uniqueItems` parity; batch after or with that issue.

samchon:
## Overview

Reserve one cohesive implementation batch for OpenAPI object safety:

- make `UniqueItems` validation and random generation use one deterministic deep structural equality relation;
- treat parsed JSON objects and schema/component maps as prototype-safe data dictionaries;
- enforce intrinsic object, numeric, tuple, reference, and OpenAPI 3.1 conversion invariants across OpenAPI and LLM consumers.

Closes #2032
Closes #2042
Closes #2043

## Scope

The implementation will remain within the shared typia/runtime and `@typia/utils` OpenAPI/LLM owners, with focused regression coverage for positive, negative, boundary, and oracle-derived cases. Fixture-specific workarounds, unrelated scalar semantics, and repository-wide formatting are excluded.

## Coordination

PR #2060 currently has an adjacent OpenAPI 3.1 upgrader change and PR #2062 has adjacent Swagger conversion work. This branch will refresh from current `master` before merge and resolve any converter overlap semantically, retaining both contracts.

## Verification

Pending. This draft is an implementation-free campaign claim; tests-first focused suites, affected package builds, broader applicable local verification, and repeated solo Self-Review rounds will be recorded before merge.

Repository Actions are deliberately cancelled by exact campaign commit SHA after every push and pull-request event. No workflow or repository Actions setting is disabled; local verification is the campaign gate.

### DIFF
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -237,7 +237,7 @@ var metadataCommentTagFactory_PARSER = map[string]metadataCommentTagFactory_pars
     Report func(msg string) any
     Value  string
   }) metadataCommentTagFactory_TagRecord {
-    return metadataCommentTagFactory_TagRecord{"array": {{Name: "UniqueItems", Target: "array", Kind: "uniqueItems", Value: true, Validate: "$input.length <= 1 || (new Set($input).size === $input.length)", Exclusive: true, Schema: map[string]any{"uniqueItems": true}}}}
+    return metadataCommentTagFactory_TagRecord{"array": {{Name: "UniqueItems", Target: "array", Kind: "uniqueItems", Value: true, Validate: "$importInternal(\"isUniqueItems\")($input)", Exclusive: true, Schema: map[string]any{"uniqueItems": true}}}}
   },
   "type": metadataCommentTagFactory_parse_type,
   "minimum": func(props struct {
--- a/packages/typia/package.json
+++ b/packages/typia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "13.1.3",
+  "version": "13.1.4",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "exports": {
--- a/packages/typia/src/internal/_isUniqueItems.ts
+++ b/packages/typia/src/internal/_isUniqueItems.ts
@@ -1,12 +1,4 @@
 export const _isUniqueItems = (elements: any[]): boolean => {
-  // EMPTY OR ONLY ONE
-  if (elements.length < 2) return true;
-
-  // SHALLOW COMPARISON
-  if (["boolean", "bigint", "number", "string"].includes(typeof elements[0]))
-    return new Set(elements).size === elements.length;
-
-  // DEEP COMPARISON
   for (let i = 0; i < elements.length; i++)
     for (let j = i + 1; j < elements.length; j++)
       if (equals(new WeakMap())(elements[i], elements[j])) return false;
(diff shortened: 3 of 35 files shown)


samchon:
Implementation wave pushed at `0c8aa153e3ccee7b6b1712c9197f4cd51d6547e7`.

Implemented:

- one byte-identical structural `UniqueItems` relation for typia and utils, including mixed values, nested containers, binary views, files/blobs, and cycles;
- JSDoc/type-tag/OpenAPI parity plus bounded random generation for impossible uniqueness domains;
- safe own-property dictionary reads/writes across lenient JSON, LLM coercion/schema conversion, OpenAPI converters, validators, and HTTP migration;
- deterministic missing/cyclic reference failures;
- intrinsic JSON object/finite numeric/tuple constraints and OpenAPI 3.1 `prefixItems` + `items` rest preservation.

Verification passed:

```text
pnpm --filter ./packages/utils build
pnpm --filter ./packages/typia build
pnpm --filter ./tests/test-utils start
pnpm --filter ./tests/test-typia-schema start
pnpm test:go:native
```

Focused RED→GREEN evidence and the public-prototype compatibility correction are retained in the campaign evidence. Repository-wide `pnpm build` / `pnpm test` and the required repeated solo Self-Review remain before merge.

samchon:
## Self-Review round 1

Head reviewed: `0c8aa153e3ccee7b6b1712c9197f4cd51d6547e7`

### Findings

1. `LlmSchemaConverter.parameters()` and `invert()` now expose newly created null-prototype `$defs` / `components.schemas` objects. Safe assignment is sufficient for reserved keys, and the full-suite parser/coercer failure already proved that changing public object prototypes is an unnecessary compatibility break.
2. `LlmTypeChecker.escapeReference()` still dereferences with a non-null assertion and unbounded recursion. Missing, inherited-only, or direct cyclic `$defs` aliases can therefore throw or overflow even though the other LLM/OpenAPI reference owners now fail deterministically.
3. `ObjectDictionary.has/get` do not accept runtime `null`; malformed external schema/component maps that optional chaining previously treated as absent now throw in the centralized helper.
4. Tuple validation delegates present elements to `OpenApiStationValidator`, whose unknown-schema fast path accepts `undefined`. A `{}` prefix or `additionalItems: true` can therefore still admit an explicit `undefined` or sparse hole despite tuple elements being required JSON values.
5. `_randomArray` calls `_isUniqueItems` over the complete accumulated array for every candidate, repeatedly rechecking all established pairs and turning generation into cubic work. Candidate-vs-existing comparisons preserve the invariant with quadratic total work.
6. The structural object fallback uses `Object.keys`, so enumerable symbol-keyed state is ignored and distinct objects can be misclassified as duplicates outside the OpenAPI-only string-key surface.

### Remediation plan

- Preserve ordinary public dictionary prototypes while retaining centralized own-property reads and `defineProperty` writes; make read helpers null-tolerant.
- Replace LLM reference recursion with bounded iterative own-definition traversal and add missing/inherited/cyclic coverage tests.
- Reject `undefined`/holes before tuple schema dispatch, including unknown-prefix and open-rest cases.
- Compare only each random candidate against established elements and include enumerable symbol keys in structural object equality, keeping both helper copies byte-identical.
- Run focused regressions, affected package/full workspace gates, push through the exact-SHA cancellation gate, and restart the complete Self-Review from the new base-to-head diff.

samchon:
## Self-Review round 1 resolution

Resolved at `649eefa53f60d98c344a0a29c569f924789d689a`.

- Public LLM `$defs` and newly created `components.schemas` retain ordinary object prototypes while reserved-key writes remain `defineProperty`-safe.
- `LlmTypeChecker` now resolves only own definitions with iterative missing/cycle termination; centralized reads tolerate runtime-null maps.
- Tuple validation rejects explicit `undefined` and sparse entries before unknown/open schema dispatch.
- Random uniqueness compares each candidate only with established elements.
- Structural object equality now includes enumerable symbol keys, with typia/utils helper copies still byte-identical.

Focused OpenAPI/LLM/HTTP and typia UniqueItems/random regressions passed. The next step is semantic rebase onto current `master`, including PR #2062's Swagger converter contracts, followed by both regression groups and a fresh full-diff Self-Review.

samchon:
## Semantic rebase over converter campaigns

Rebased onto `5c1c81b920d8a18619f88474c51a6f9f8f310a0e` at head `3fa2d945b82d31e39f15c0f6f157b2cf36f195d7`.

- PR #2062's Swagger v2 form encoding, media selection, server conversion, nullable/reference, unsupported-schema, and enum contracts remain unchanged; the base-to-head Swagger diff adds only centralized safe own-property component reads/writes.
- The sole manual conflict was `HttpMigrateRouteComposer.ts`. The current-master querystring/cookie/serialization/canonical-collision/optional-object behavior was selected in full. Its newly introduced `resolveSchema()` component lookup and the existing recursive sanitizer/emplacement paths were then routed through `ObjectDictionary`.
- OpenAPI 3.1 querystring behavior from the preceding HTTP campaign remains intact beside this branch's tuple `items` rest-schema conversion.
- PR #2070 already owns typia `13.1.2`, and #2072 is landing with `13.1.3`; this native-generator batch is serialized at `13.1.4`.

Both focused groups passed after rebase:

```text
#2071: prototype dictionaries, tuple/object/number invariants, V3.1 tuple conversion, structural UniqueItems/random
#2062: strict numeric defaults, unrepresentable Swagger, form arrays/data, server/media, urlencoded file, enum downgrade
```

The branch will refresh once more after #2072 lands before final verification and merge.

samchon:
Final semantic rebase evidence

- Rebasing onto `ed1b565cedaa772927bafbb8a4dea781b94fe62d` replayed the implementation and Self-Review remediation without content conflicts.
- The sole conflict was `packages/typia/package.json`: merged `master` already owns `13.1.3`, so this PR retains its required native-release successor `13.1.4`.
- The resulting merge base exactly matches `origin/master` and the PR head is `0aab2a475a36ffa6ba37a43861037cc375e36b67`.
- `git diff --check origin/master...HEAD` passes and no conflict marker remains in the changed files.
- The prior #2062 Swagger/OpenAPI contracts and #2072 shared native runtime contracts remain in the rebased base; the PR diff over that base contains only its own object-safety, structural uniqueness, tuple/reference validation, tests, and version changes.

samchon:
Self-Review round 2 finding and remediation plan

Broad verification found one compatibility defect in the rebased implementation:

- `OpenApiObjectValidator.isPlainObject` rejects every object with a user-defined prototype. This correctly rejects `Date`, `Map`, and `Set`, but it also rejects ordinary serializable class instances whose enumerable own properties are the JSON object value. The established `test-openapiautomated` class fixtures consequently fail at `$input` before their declared properties can be validated.

Remediation plan:

1. Classify JSON-object-shaped values by the intrinsic object tag, accepting ordinary, null-prototype, cross-realm, and user-class objects that present as `[object Object]` while still rejecting arrays and native container/value classes.
2. Update the focused invariant test so a serializable class instance remains accepted, while Date/Map/Set and arrays remain rejected and inherited properties remain absent.
3. Re-run the focused object invariant, the previously failing automated class cases, all affected/full suites, package/root builds, Go native tests, and root `pnpm test`.
4. Commit and push the minimal remediation, run the exact-SHA cancellation gate, then restart the full-diff Self-Review from the new head.

samchon:
Self-Review round 2 verification addendum

The complete OpenAPI automation suite exposed two additional consequence-surface mismatches after the class fix:

- Two spoiler cases replace a referenced object with an array. The strengthened object validator correctly reports the object-owning paths (`...p2` and `...target`), while the independent spoiler oracle still follows the old array's ad-hoc child properties and expects deeper paths. The production result is correct; the oracle must canonicalize an intrinsically non-object value to the object owner before descending.
- Typia-generated internal tuple schemas omit `minItems` while historically requiring every fixed prefix item. Treating absent `minItems` as zero made a required nested tuple accept `[]`. Conversely, an external OpenAPI 3.1 `prefixItems` schema follows JSON Schema 2020-12 and defaults `minItems` to zero.

Remediation plan extension:

1. Preserve the internal tuple contract by defaulting validation's minimum to `prefixItems.length` when `minItems` is absent.
2. During OpenAPI 3.1 `prefixItems` upgrade, materialize the standard default as `minItems: 0`; add a conversion/validation regression for an omitted prefix.
3. Teach the independent automated spoiler oracle to stop at an object owner when the current value is not JSON-object-shaped, matching the strengthened intrinsic type check without weakening production validation.
4. Remove the temporary diagnostic source, run all three formerly failing cases and the entire OpenAPI automation suite, then repeat the full repository acceptance gate.

samchon:
## Final-base integration finding and remediation plan

The bounded root test against newly merged PR #2091 exposed one stale adapter oracle, not a production regression:

- The new MCP output-validation matrix expects an array returned for an object result schema to fail at `$input.value`. That path reflects the previous validator behavior, which descended into a non-object array and only then noticed the missing property.
- This PR intentionally enforces the intrinsic JSON-object boundary at the object owner, so the same array correctly fails at `$input` before property traversal.
- The Vercel matrix carries the identical `$input.value` expectation and would fail for the same reason once reached.

Remediation plan:

1. Update only the MCP and Vercel array-case path oracles to `$input`; retain missing-property failures at `$input.value`.
2. Re-run the focused and full MCP/Vercel suites plus the existing OpenAPI/object-focused gates.
3. Re-run root `pnpm test` with bounded file-backed output and explicit process-tree cleanup, then restart the complete solo base-to-head Self-Review.

samchon:
## Self-Review round 3 — clean on final base

Fresh complete review covered every base-to-working-tree hunk across all 35 changed files after rebasing onto `98f552144c`, including native versioning/JSDoc generation, both byte-identical structural uniqueness helpers, bounded random generation, prototype-safe schema dictionaries, LLM/OpenAPI reference termination, converter and HTTP consumers, intrinsic object/number/tuple validation, generated-automation oracles, and the newly merged MCP/Vercel output-validation consequence.

No further correctness, boundary, compatibility, security, data-loss, cache/recovery, Windows/POSIX, packaging, documentation, or test-quality improvement survived verification.

Final direct verification:

- affected OpenAPI/LLM/HTTP and typia structural-uniqueness focused suites;
- complete `test-utils`, `test-typia-schema`, and `test-utils-automated` suites;
- MCP and Vercel output-validation focused cases after the final-base oracle repair;
- `pnpm run test:packages` — exit 0 (384 s);
- `pnpm run test:go:public` — exit 0 (16 s);
- `pnpm run test:go:native` — exit 0 (493.9 s);
- root `pnpm build` — exit 0 on the immediately preceding base; final master delta was adapter tests only and the affected integration was revalidated above;
- `git diff --check origin/master`, no conflict markers, typia/utils `_isUniqueItems` sources byte-identical, and typia version `13.1.4`.

The earlier monolithic `pnpm test` wrapper result is deliberately not counted: its child process ended while the wrapper pipe remained hung. The direct component gates above replace it with explicit exit-code evidence.

PR #2072, opened 2026-07-16
fix(typia): restore native mechanics contracts

### LINKED ISSUE #1264: [BUG] Returning type of createAssertGuard is invalid
## Bug Report

Returning types of `createAssertGuard` and `createAssertGuardEquals` are invalid.

**Document**

image

**Actual types**

https://github.com/samchon/typia/blob/cd6ed78a9815134617405b34d5ee891da56f539a/src/module.ts#L722-L724

https://github.com/samchon/typia/blob/cd6ed78a9815134617405b34d5ee891da56f539a/src/module.ts#L896-L898

### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoF4AOxgFMozUAxtzgBJALLcQAI17YWcOMAAmALjgEYUTnjgAyXIQB0AMWghUMADw0ArjeU0AfKwUdUIbmo1aOeF3HxPOA4baVldeQUDAkMAFWRua3suAGYAJic9SIUYfBixTlBQyzSABkcsqOjDAFEADwEAGxsCYAA3bjFUOqKQSwBGUvLWYhYBCA4NODwbVCglCTCoOABeBGQ0QwEobgtuAEECAl4YAHFZ+ctxSRkoRwAKAEpWccn4CCkAK1WsUZYZuYLG68e4fT6PNivAgQRrcQyNCB4e4AA2UagAJFgwYZlMRkY8gA

### LINKED ISSUE #1889: Native semantic helpers blanket-recover unexpected panics
## Problem

Three native paths catch every Go panic and silently substitute ordinary metadata/classification results. Expected upstream limitations and unexpected programmer/checker defects therefore become indistinguishable, allowing corrupt or incomplete transform output instead of a visible failure.

## Current sites

- `metadata_get_type_arguments` in `MetadataHelper.go` returns `nil` type arguments after any panic from `checker.GetTypeArguments`.
- Protobuf metadata handling uses a deferred `_ = recover()` around a capability-sensitive checker path.
- Plain tuple classification uses a blanket recovery while computing its tuple seed, even though its callers establish tuple ownership.

The scoped transformer recovery path is the counterexample: it handles the documented transform error and re-panics unexpected values.

## Evidence and provenance

The metadata helper remains unchanged from this issue's original report and has no merged implementation PR. A 2026-07 repository-wide campaign traced the same failure class into the Protobuf and plain-classification paths. These are not requests to remove all recovery; the defect is silently converting arbitrary panics into valid-looking semantic results.

## Approach

For each site, identify the precise expected upstream limitation. Prefer an explicit capability/type precondition that removes panic-based control flow. Where recovery is unavoidable, recognize only the documented expected value and re-panic everything else. Keep behavior unchanged for supported inputs and add a focused regression that proves unexpected failures cannot be swallowed.

- No native semantic helper blanket-swallows arbitrary panics at the three listed ownership points.
- Expected unsupported checker inputs retain their documented fallback, with tests and a comment naming the upstream condition.
- Unexpected panics remain observable and cannot yield incomplete metadata, Protobuf schema, or tuple classification.
- Run focused Go/native transform tests, repository build/typecheck/static checks, and the full test suite.
- Native changes bump the `typia` package version.
- During the issue campaign implementation, do not run repository-wide `pnpm format`; formatting is deferred to the campaign cleanup PR.

## Coordination

Reopened and expanded instead of creating a duplicate. The implementation may split commits by native owner, but one PR should keep the shared recovery contract explicit.

### LINKED ISSUE #1950: compare.less violates its NaN-last total-order contract
## Residual defect

The merged `compare.less` implementation does not satisfy this issue's explicit NaN-last total-order decision. Its primitive comparator is equivalent to:

```ts
x < y ? -1 : x > y ? 1 : 0
```

For `NaN`, both comparisons are false. Consequently `NaN` compares equal to every finite number, `less(NaN, 1)` and `less(1, NaN)` are both false, and the documented sort-comparator composition collapses distinct values. Invalid `Date` values have the same problem because they compare through `getTime()`.

## Provenance

This issue originally specified “NaN number leaves sort last among numbers” precisely to avoid this trap. PR #1954 merged the feature but its validation list did not cover NaN, and current tests contain no NaN/invalid-Date ordering matrix. Reopening this exact contract is preferable to creating a duplicate.

## Approach

Define the total scalar order explicitly, preserving the existing order for ordinary finite/infinite numbers. Handle NaN symmetrically and reuse the rule for numeric native projections such as `Date#getTime`. Ensure mixed primitive unions keep their type-rank behavior before applying the within-number rule.

- `NaN` sorts after every non-NaN number and compares equivalent only to NaN.
- `-Infinity`, finite values, signed zero, and `Infinity` retain a deterministic order.
- Invalid dates follow a documented total rule and do not compare equal to every valid date.
- Direct/factory/nested/array/tuple/mixed-union comparators satisfy antisymmetry and sort-composition cases.
- Run focused Go transform/runtime tests and the full suite; native changes bump `typia`.
- During the issue campaign implementation, do not run repository-wide `pnpm format`; formatting is deferred to the campaign cleanup PR.

### LINKED ISSUE #2052: typia.functional wrappers discard the call-site receiver
## Problem

Every `typia.functional.*` wrapper invokes the captured function as a bare call from an arrow function. A function with an explicit TypeScript `this` parameter retains that receiver in the advertised wrapped type, but the generated runtime call discards it.

## Evidence

The assert, is/equals, and validate families emit arrow wrappers and a `NewCallExpression` whose callee is the captured function identifier and whose arguments are only the declared parameters. No generated path forwards the wrapper's call-site receiver with `call`/`apply`/`Reflect.apply`.

```ts
function add(this: { base: number }, value: number): number {
  return this.base + value;
}
const checked = typia.functional.assertFunction(add);
checked.call({ base: 40 }, 2); // typed as 42, generated inner call has undefined `this`
```

Assert-family descriptors return `T` and document the same signature; the other families likewise derive parameters/output from `T extends (...args: any[]) => any`. Arrow wrappers cannot observe a dynamic `this`, so changing only the inner call is insufficient. Exact open/closed issue and pull-request searches found no duplicate.

## Consequence surface

All assert/assertEquals/is/equals/validate function/parameter/return variants, factories and direct calls, sync/async functions, explicit `this` parameters, methods extracted with binding, recursive/error-factory paths, thrown errors, and validation of receiver-dependent return values.

## Approach

Generate a receiver-preserving ordinary function wrapper and invoke the captured target with that `this` (or define and enforce an explicit receiver-free type contract). Preserve async behavior, parameter names/types, error factories, and output transformations. Because current public types retain `this`, the implementation-first default is receiver forwarding rather than silently narrowing the API.

- Direct and factory variants preserve a dynamic receiver for assert/is/equals/validate families.
- Parameter-only, return-only, and combined wrappers behave identically for receiver-dependent sync and async functions.
- Existing receiver-free arrow/function cases and error paths remain unchanged.
- Transform snapshots prove the wrapper is receiver-capable and runtime tests cover `.call`, bound methods, and extracted methods.
- Run focused functional Go/runtime tests and the full suite; native changes bump `typia`.

## Coordination

One native functional-programmer batch should cover the shared emission helper rather than patching individual variants independently.

samchon:
## Intent

Restore four shared native contracts together: assertion-guard factory typing, semantic panic ownership, total comparator ordering, and functional receiver forwarding.

## Changes

- Make `createAssertGuard` and `createAssertGuardEquals` return `AssertionGuard<T>` directly.
- Replace blanket panic recovery with explicit ownership checks in metadata, Protobuf, and plain tuple paths.
- Give `compare.less` a NaN-last order, including invalid `Date` projections, across direct and factory comparators.
- Preserve dynamic `this` through all `typia.functional.*` wrappers with `Reflect.apply`, while excluding TypeScript's erased `this` parameter from runtime arguments and validation.
- Preserve receiver types in the public functional helpers and document receiver/NaN behavior with runnable examples.
- Ship the native changes as `typia` 13.1.3 after rebasing onto the 13.1.2 predecessor.

## Verification

- Focused guard, recovery, comparison, functional, and plain-native regression suites: pass.
- `go test ./... -count=1` in `packages/typia/native`: pass.
- `go vet ./...` in `packages/typia/native`: pass.
- `go test ./... -count=1` in `packages/typia/test`: pass.
- `pnpm --filter typia build`: pass as 13.1.3.
- `pnpm --filter @typia/examples build`: pass.
- `git diff --check`: pass.
- Four solo Self-Review rounds completed; rounds 1-3 were remediated and round 4 was clean.

The repository-wide `pnpm run test:packages` produced no failure output but exceeded the 15-minute local limit while `test-typia-automated` was still running. The narrower native, transform/runtime, package build, and example build gates above completed successfully.

Campaign Actions were cancelled only for the exact pushed SHA after every push. On final HEAD `ece7583c6a8bdd7cede48e08509a2ff399441d03`, five workflow runs were cancelled and spell-check completed successfully before cancellation.

Closes #1264
Closes #1889
Closes #1950
Closes #2052

### DIFF
--- a/examples/src/functional/assertFunction.ts
+++ b/examples/src/functional/assertFunction.ts
@@ -1,7 +1,13 @@
 import typia from "typia";
 
-const func = typia.functional.assertFunction<(x: number, y: number) => number>(
-  (x: number, y: number): number => x + y,
-);
-func(3, 4);
-func(4, 5);
+interface Receiver {
+  offset: number;
+}
+
+function add(this: Receiver, x: number, y: number): number {
+  return this.offset + x + y;
+}
+
+const func = typia.functional.assertFunction(add);
+func.call({ offset: 10 }, 3, 4);
+func.call({ offset: 20 }, 4, 5);
--- a/examples/src/functional/isFunction.ts
+++ b/examples/src/functional/isFunction.ts
@@ -1,7 +1,13 @@
 import typia from "typia";
 
-const func = typia.functional.isFunction<(x: number, y: number) => number>(
-  (x: number, y: number): number => x + y,
-);
-func(3, 4);
-func(4, 5);
+interface Receiver {
+  offset: number;
+}
+
+function add(this: Receiver, x: number, y: number): number {
+  return this.offset + x + y;
+}
+
+const func = typia.functional.isFunction(add);
+func.call({ offset: 10 }, 3, 4);
+func.call({ offset: 20 }, 4, 5);
--- a/examples/src/functional/validateFunction.ts
+++ b/examples/src/functional/validateFunction.ts
@@ -1,7 +1,13 @@
 import typia from "typia";
 
-const func = typia.functional.validateFunction<
-  (x: number, y: number) => number
->((x: number, y: number): number => x + y);
-func(3, 4);
-func(4, 5);
+interface Receiver {
+  offset: number;
+}
+
+function add(this: Receiver, x: number, y: number): number {
+  return this.offset + x + y;
+}
+
+const func = typia.functional.validateFunction(add);
+func.call({ offset: 10 }, 3, 4);
+func.call({ offset: 20 }, 4, 5);
--- a/packages/typia/native/cmd/ttsc-typia/assert_guard_factory_contract_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/assert_guard_factory_contract_test.go
@@ -1,0 +1,152 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+
+  shimast "github.com/microsoft/typescript-go/shim/ast"
+  "github.com/samchon/ttsc/packages/ttsc/driver"
+)
+
+// TestAssertGuardFactoryContract verifies factory typing and runtime shape.
+//
+// An assertion-guard factory returns the guard itself. Its successful invocation
+// returns void while narrowing the input; it never returns another callable
+// assertion guard.
+//
+//  1. Compile normal and equals factories with explicit and inferred types.
+//  2. Prove successful invocations narrow and have a void return type.
+//  3. Exercise custom errors and the equals-only surplus-property boundary.
+func TestAssertGuardFactoryContract(t *testing.T) {
+  project := compareEqualCoverProject(t, "assert-guard-factory-", assertGuardFactorySource)
+  ttscTypiaTestAssertGuardFactoryType(t, project)
+  ttscTypiaTestTypecheck(t, project)
+  js := compareEqualCoverTransform(t, project)
+  if strings.Contains(js, "createAssertGuard<") {
+    t.Fatalf("assert guard factory call was not transformed:\n%s", js)
+  }
+
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(ttscTypiaTestRewriteCommonJS(t, js)), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
+  if err := os.WriteFile(runner, []byte(assertGuardFactoryRuntime), 0o644); err != nil {
+    t.Fatalf("write runtime runner: %v", err)
+  }
+  cmd := exec.Command(node, runner)
+  cmd.Dir = runtimeDir
+  if output, err := cmd.CombinedOutput(); err != nil {
+    t.Fatalf("assert guard factory runtime cases failed: %v\n%s", err, output)
+  }
+}
+
+func ttscTypiaTestAssertGuardFactoryType(t *testing.T, project string) {
+  t.Helper()
+  program, diagnostics, err := driver.LoadProgram(project, "tsconfig.json", driver.LoadProgramOptions{})
+  if err != nil {
+    t.Fatalf("load assertion guard fixture: %v", err)
+  }
+  if len(diagnostics) != 0 {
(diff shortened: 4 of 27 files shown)


samchon:
## Self-Review round 1

Reviewed the full implementation diff against current origin/master after the implementation push.

Findings:

1. The functional receiver regression exercises every receiver-dependent wrapper, but it does not include the required receiver-free arrow/function control. Add a normally invoked arrow wrapper so changing the generated wrapper shape cannot regress existing no-this calls.
2. Moving Protobuf emplacement after successful metadata analysis correctly removes blanket recovery, but the focused ownership test now covers only the escaping invariant panic. Add an unsupported-metadata control to the same test so the documented semantic diagnostic remains distinct from an unexpected emplacer failure.
3. The compare.less guide says NaN equals itself for ordering purposes, which can be confused with the separate compare.equals API. Reword this as ordering equivalence without implying structural equality.

Plan: add the two focused controls, adjust the guide wording, rerun the focused tests and format/diff checks, then perform another full solo review round.

samchon:
## Self-Review round 2

Re-read the full diff plus round-1 remediation and found two remaining precision issues:

1. The Protobuf unsupported-input control checks only that some diagnostic exists. Because the top-level shape check also emits a diagnostic, the test could pass even if the specific any-type fallback disappeared. Assert the exact does-not-support-any-type message.
2. The compare guide still says less and equals compose into the shown sort comparator, but the sample and the NaN contract use two directional less checks; structural equals is neither called nor appropriate for NaN. Describe the actual bidirectional-less composition.

Plan: tighten the assertion and wording, rerun the focused package tests, then perform a third complete solo review pass.

samchon:
## Self-Review round 3

The full diff is functionally sound after round 2, but one coverage gap remains around the exact consequence of erasing the TypeScript-only this parameter: invalid-parameter tests prove failure but not that error paths still begin at parameters[0]. A shifted index would regress the public error contract while all current assertions passed. The issue also names error-factory paths.

Plan: add one custom assert-parameters wrapper and assert both its custom error and parameters[0] path, assert the validation-family path as well, rerun the focused transform/runtime test, and then run another complete review pass.

samchon:
## Self-Review round 4 — clean

Completed another full solo pass over source, public declarations, generated-call mechanics, focused regressions, examples, documentation, and versioning after all remediation.

No sound correctness, compatibility, test-coverage, or documentation improvements remain in the declared #1264/#1889/#1950/#2052 scope. Focused recovery and functional tests pass, including receiver-free invocation, exact parameters[0] error paths, custom error factories, supported Protobuf diagnostics, and escaping invariant panics.

PR #2084, opened 2026-07-16
fix(typia): reject truncated Protobuf payloads

### LINKED ISSUE #2083: Protobuf decoders accept truncated length-delimited payloads
## Problem

Typia's Protobuf reader accepts truncated length-delimited `string` and `bytes` payloads. It returns the available prefix, advances its pointer beyond the buffer, and lets generated decoders report success for corrupted wire data.

## Evidence

`packages/typia/src/internal/_ProtobufReader.ts` reads a length varint in `bytes()`, records the current pointer, increments `ptr` by the declared length, and returns `Uint8Array.subarray(from, from + length)` without checking that the payload exists. Typed-array `subarray` clamps its end to the backing view instead of throwing.

After building current `master` at `1cda8afc4e7795abc92e285212d128e942049410`, an in-memory probe against the emitted `_ProtobufReader` used `[5, 97, 98]`: a declared five-byte payload with only two bytes available. `bytes()` returned `[97, 98]`, set `index()` to 6, and retained `size()` 3, with no exception. `string()` inherits the same behavior.

Generated decoders call `reader.bytes()`/`reader.string()` directly for known fields. The outer decode loop stops after the over-advanced pointer exceeds size, while assert/is/validate decode variants validate the already-truncated value; an unconstrained string or byte array therefore passes. The reader's `skip(length)` already performs the missing bounds check, showing that overflow is intended to be an error elsewhere.

Tests-first implementation auditing also reproduced bounded-subrecord overread after a top-level reader check: truncated map entry `[0a 02 0a 01 61]`, nested message `[0a 02 0a 01 61]`, and packed field `[0a 01 80 01]` can consume bytes past their declared local end and still succeed when trailing outer-buffer bytes exist. `ProtobufDecodeProgrammer` computes local end markers for these generated loops but does not require the final reader index to stay within and finish at the declared boundary. The full failure class therefore spans the TypeScript reader and the native generated-decoder boundary logic.

The Protobuf wire specification defines a length-delimited record as a varint size followed by exactly that payload and uses the same encoding for strings and bytes: https://protobuf.dev/programming-guides/encoding/ . History traces the unchecked reader method to the monorepo migration. #1073 concerned view offsets and #2051/#2066 fixed 64-bit varints; open and closed searches found no truncated-payload issue or PR.

## Consequence surface

- Direct and factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` operations.
- Singular, optional, repeated, map, nested-message, packed, string, and bytes length-delimited fields.
- Corrupted network/storage payloads can be accepted as shorter valid values, hiding transport truncation and integrity failures.
- Pointer overrun can suppress parsing of subsequent required or unknown fields.
- UTF-8 decoding can turn a truncated multibyte tail into replacement text unless strict payload bounds fail first.

## Approach

Centralize reader bounds validation for every fixed- and length-sized read. Validate the declared payload length before mutating the pointer or constructing a view, and make generated map, nested, and packed subrecords enforce their declared local end before returning to the outer record. Keep error behavior consistent with `skip`, and audit fixed32/fixed64 plus every bounded subreader for the same failure class rather than patching only the first demonstrated branch.

- Truncated string and bytes payloads throw a stable buffer-overflow error in direct and factory decode variants.
- assert/is/validate decode variants cannot convert malformed wire data into a successful product value; tests pin their documented error surface.
- Singular, repeated, map key/value, nested message, packed, unknown-field skip, zero-length, exact-end, and sliced-buffer cases have valid and one-byte-short controls.
- A declared length larger than the remaining buffer never advances the reader pointer or returns a partial view.
- Fixed32/fixed64 reads and length-delimited subreaders are audited and receive equivalent boundary coverage where applicable.
- Focused reader tests, generated Protobuf suites, full test-typia-schema, typia build, and public/native Go tests pass.
- Because the complete repair changes native Go generator source, the typia package version advances in the implementation PR.

## Coordination

This does not reopen the merged 64-bit varint issue #2051: it owns TypeScript reader bounds and the native generator's enforcement of length-delimited subrecord boundaries. Do not change unrelated Protobuf wire semantics or publish compatibility claims beyond rejecting malformed truncated data.

samchon:
## Intent

Reject truncated length-delimited Protobuf payloads before a decoder can return a partial string, byte array, nested message, packed value, or map entry as a successful product value.

## Scope

- Make `_ProtobufReader` validate fixed- and length-sized reads against one active boundary before mutating its pointer or constructing a view.
- Make `ProtobufDecodeProgrammer` fork and close that boundary for generated nested-message, map-entry, and packed-array subrecords, including cases where trailing outer bytes still exist.
- Cover direct and factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` paths across string, bytes, optional, repeated, map, nested, packed, unknown-field, fixed-width, zero-length, exact-end, and sliced-buffer cases.
- Ship the native generator correction as the reserved `typia` 13.1.10 after the 13.1.9 predecessor lands.

## Deferred

- No changes to the already-landed 64-bit varint semantics.
- No unrelated Protobuf wire-semantic or compatibility changes beyond rejecting malformed truncated records.
- No repository-wide formatting; Post-Campaign Cleanup owns that phase.

## Verification

- Tests-first reproductions failed on partial top-level string/bytes and on bounded map/nested/packed overreads with trailing bytes outside the subrecord.
- Full `test-typia-schema`: pass with the new reader and generated-decoder matrix.
- Focused Protobuf programmer and full `cmd/ttsc-typia` tests: pass.
- Public Go suite: pass.
- Native Go suite: pass.
- Tagged native-internal tests, typia build, final semantic rebase after #2089/13.1.9, and findings-first solo Self-Review through a clean round remain pending.

Campaign Actions remain enabled and are cancelled only for this branch's exact pushed SHA after every push or pull-request mutation.

Closes #2083

### DIFF
--- a/packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go
+++ b/packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go
@@ -174,34 +174,25 @@ func protobufDecodeProgrammer_write_object_function(props struct {
   Object  *schemametadata.MetadataObjectType
 }) *shimast.Node {
   f := nativecontext.EmitFactoryOf(protobufDecodeProgrammer_factory, props.Context.Emit)
-  lengthAssign := f.NewExpressionStatement(
-    protobufDecodeProgrammer_assignment(
-      f.NewIdentifier("length"),
-      shimast.KindEqualsToken,
-      nativefactories.ExpressionFactory.Conditional(
-        protobufDecodeProgrammer_lessThan(f.NewIdentifier("length"), nativefactories.ExpressionFactory.Number(0, props.Context.Emit), props.Context.Emit),
-        protobufDecodeProgrammer_callReader("size", nil, props.Context.Emit),
-        protobufDecodeProgrammer_add(protobufDecodeProgrammer_callReader("index", nil, props.Context.Emit), f.NewIdentifier("length"), props.Context.Emit),
-        props.Context.Emit,
-      ),
-      props.Context.Emit,
-    ),
-  )
-  body := []*shimast.Node{lengthAssign}
-  body = append(body, protobufDecodeProgrammer_write_object_function_body(protobufDecodeProgrammer_objectBodyProps{
+  body := protobufDecodeProgrammer_write_object_function_body(protobufDecodeProgrammer_objectBodyProps{
     Context:   props.Context,
-    Condition: protobufDecodeProgrammer_lessThan(protobufDecodeProgrammer_callReader("index", nil, props.Context.Emit), f.NewIdentifier("length"), props.Context.Emit),
+    Condition: protobufDecodeProgrammer_lessThan(protobufDecodeProgrammer_callReader("index", nil, props.Context.Emit), protobufDecodeProgrammer_callReader("size", nil, props.Context.Emit), props.Context.Emit),
     Tag:       "tag",
     Output:    "output",
     Object:    props.Object,
-  })...)
+  })
+  body = append(body, f.NewIfStatement(
+    protobufDecodeProgrammer_lessThan(nativefactories.ExpressionFactory.Number(-1, props.Context.Emit), f.NewIdentifier("previous"), props.Context.Emit),
+    f.NewExpressionStatement(protobufDecodeProgrammer_callReader("close", []*shimast.Node{f.NewIdentifier("previous")}, props.Context.Emit)),
+    nil,
+  ))
   body = append(body, f.NewReturnStatement(f.NewIdentifier("output")))
   return f.NewArrowFunction(
     nil,
     nil,
     f.NewNodeList([]*shimast.Node{
       nativefactories.IdentifierFactory.Parameter("reader", nil, nil, props.Context.Emit),
-      nativefactories.IdentifierFactory.Parameter("length", nativefactories.TypeFactory.Keyword("number", props.Context.Emit), nativefactories.ExpressionFactory.Number(-1, props.Context.Emit), props.Context.Emit),
+      nativefactories.IdentifierFactory.Parameter("previous", nativefactories.TypeFactory.Keyword("number", props.Context.Emit), nativefactories.ExpressionFactory.Number(-1, props.Context.Emit), props.Context.Emit),
     }),
     nativefactories.TypeFactory.Keyword("any", props.Context.Emit),
     nil,
@@ -460,12 +451,13 @@ func protobufDecodeProgrammer_decode_array(props protobufDecodeProgrammer_decode
       f.NewBlock(f.NewNodeList([]*shimast.Node{
         nativefactories.StatementFactory.Constant(nativefactories.StatementFactory_ConstantProps{
           Name:  "piece",
-          Value: protobufDecodeProgrammer_add(protobufDecodeProgrammer_callReader("uint32", nil, emit), protobufDecodeProgrammer_callReader("index", nil, emit), emit),
+          Value: protobufDecodeProgrammer_callReader("fork", nil, emit),
         }, emit),
         f.NewWhileStatement(
-          protobufDecodeProgrammer_lessThan(protobufDecodeProgrammer_callReader("index", nil, emit), f.NewIdentifier("piece"), emit),
+          protobufDecodeProgrammer_lessThan(protobufDecodeProgrammer_callReader("index", nil, emit), protobufDecodeProgrammer_callReader("size", nil, emit), emit),
           f.NewExpressionStatement(protobufDecodeProgrammer_arrayPush(props.Accessor, decoder, emit)),
         ),
+        f.NewExpressionStatement(protobufDecodeProgrammer_callReader("close", []*shimast.Node{f.NewIdentifier("piece")}, emit)),
       }), true),
       f.NewExpressionStatement(protobufDecodeProgrammer_arrayPush(props.Accessor, decoder, emit)),
     ))
@@ -486,7 +478,7 @@ func protobufDecodeProgrammer_decode_regular_object(props struct {
   }
   args := []*shimast.Node{f.NewIdentifier("reader")}
   if props.Top == false {
-    args = append(args, protobufDecodeProgrammer_callReader("uint32", nil, emit))
+    args = append(args, protobufDecodeProgrammer_callReader("fork", nil, emit))
   }
   return f.NewCallExpression(
     f.NewIdentifier(fmt.Sprintf("%so%d", protobufDecodeProgrammer_PREFIX, index)),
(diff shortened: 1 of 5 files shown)


samchon:
## Self-Review round 1 — clean

**Findings:** No correctness, boundary, compatibility, test-isolation, packaging, or documentation improvement survived a fresh review of the full base-to-head surface at `c8e00fa18063567acc1d43c1b76118aaee22537a`.

The review re-traced top-level sized reads and generated nested-message, map-entry, packed-array, unknown-field, fixed-width, sliced-buffer, direct/factory, assert/is/validate, zero-length, exact-end, and pointer rollback behavior. The reader owns the active boundary; the native generator forks and closes it only for length-delimited subrecords; the package version is reserved as 13.1.10 behind #2089/13.1.9.

**Verification at this round:**
- `pnpm --filter ./tests/test-typia-schema start --include protobuf.decode` — pass
- `go -C packages/typia/test test ../native/core/programmers/protobuf ../native/cmd/ttsc-typia` — pass
- `pnpm test:go:public` — pass
- `pnpm test:go:native` — pass
- `go -C packages/typia/test test -tags typia_native_internal ../native/...` — pass
- `pnpm --filter typia build` — pass
- `pnpm --filter ./tests/test-typia-schema start` — pass
- `git diff --check origin/master...HEAD` — pass

Final semantic rebase, repeat verification, and another clean Self-Review round remain intentionally pending until #2089 lands.

samchon:
## Solo Self-Review — round 1 (findings and plan)

Surface: `origin/master...c8e00fa180` — `_ProtobufReader.ts`, `ProtobufDecodeProgrammer.go`, `package.json`, and the two new `test-typia-schema` cases. Fresh full read of every hunk plus the consequence surface (generated emit, consumers, docs, examples, packaging, conventions).

### Verified sound

- **Reader boundary.** `take()` validates before mutating `ptr`, so no read advances past the active `end` or builds a clamped `subarray`; `u8`/`float`/`double`/`bytes`/`skip` all route through it. `atomic()` restores `ptr`/`end` when a multi-step read (`uint32()` then `take()`) fails midway, which is what makes the one-byte-short cases report a stable error instead of a half-consumed reader.
- **Subrecord stack.** `fork()`/`close(previous)` compose a real stack: `fork` returns the enclosing `end`, generated nested-message / map-entry / packed-array loops iterate `while (index < size)` against the forked `end`, and `close` restores it. The `-1` sentinel is safe because `end` is never negative, so a top-level `_poN(reader)` correctly skips `close` while every nested call closes exactly once. Recursive and repeated-message shapes nest correctly via the JS call stack.
- **No false rejection.** Every generated loop can only exit at `index === end`, because reads are capped at `end`. `close`'s `ptr !== end` throw is therefore defensive rather than reachable on well-formed data — valid payloads keep their existing semantics.
- **Conventions.** Both new tests mirror the peer `test_protobuf_reader_64bit_varints.ts`: one exported `test_*` per matching filename, the three-part doc comment, and a local assert helper.
- **Docs/artifacts.** `website/src/content/docs/protobuf/decode.mdx` makes no claim this change falsifies (its "on failure" column describes *validation* failure, and it already tells readers the decoder — not the validators — owns structural decoding). `examples/bin/**` and `tests/test-typia-generate/src/output/**` are not committed, so there is no stale generated artifact to refresh.

### Finding 1 — verification gap (closed, no code change)

`test-typia-automated` was not in the recorded verification even though this PR changes `ProtobufDecodeProgrammer`, and that suite is the strongest regression gate for exactly this change. Ran it: it generates and executes **78 `protobuf.createDecode` round-trips** across the `@typia/template` matrix (recursive, hierarchical-pointer, nullable, and map structures) plus the 6 hand-maintained composite `ObjectSimple` protobuf cases. **Success, exit 0.** Gap closed with no defect found.

### Finding 2 — pre-existing, out-of-scope defect (reporting, not fixing here)

While auditing this issue's "unknown-field skip, zero-length ... valid and one-byte-short controls" acceptance line, I found a real defect that this PR does **not** own.

`skip(0)` is overloaded as the *skip-a-varint* sentinel for `ProtobufWire.VARIANT`, but `skipType` forwards `this.skip(this.uint32())` for `LEN`. An unknown length-delimited field whose declared length is `0` therefore takes the varint-skip branch and eats a byte belonging to the next field.

Reproduced on this branch:

| Probe | Expected | Actual |
| --- | --- | --- |
| `new _ProtobufReader(Uint8Array.of(0, 0x0a)).skipType(ProtobufWire.LEN)` | `index() === 1` | `index() === 2` |
| `typia.protobuf.decode<{ value?: string }>(Uint8Array.of(0x12, 0, 0x0a, 1, 0x61))` — a **valid** message (unknown field 2 = empty `LEN`, known field 1 = `"a"`) | `{ value: "a" }` | throws `Error on typia.protobuf.decode(): buffer overflow.` |

Disposition and evidence for the classification:

- **Pre-existing.** `git show origin/master:packages/typia/src/internal/_ProtobufReader.ts` carries the same `skip(this.uint32())` / `skip(0)` collision, and master reaches the same wrong outcome by the same path. **This PR neither introduces nor changes the behavior.**
- **Different root cause.** It is a sentinel collision that *rejects/misparses valid data*, not the truncation-accept class #2083 owns. Fixing it means changing wire-skip semantics for well-formed input, which this PR's Deferred section and #2083's Coordination section both explicitly exclude.
- **Consequence surface.** Any foreign producer emitting an empty length-delimited field the local type does not declare — routine under proto3 explicit presence and schema evolution (a removed or newer field).

Because it is out of scope and independently reachable, I am **not** widening this PR. Recommending a separate campaign issue; flagging for lead adjudication rather than deciding campaign state myself.

### Plan

Round 1 applied no code edits, so there is nothing to re-verify and no new commit. Round 1 closes clean.

### Local verification on `c8e00fa180`

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/core/programmers/protobuf/...` | pass |
| `go -C packages/typia/test test ../native/...` | pass (12 packages, 0 failures) |
| `go -C packages/typia/test test ./...` | pass (17 packages, 0 failures) |
| `pnpm --filter test-typia-schema start` | Success — includes both new cases |
| `pnpm --filter test-typia-automated start` | Success — 78 generated protobuf decode round-trips + 6 composite cases |

Campaign Actions remain unchanged; all six runs for `c8e00fa180` are terminal.

samchon:
Final semantic rebase onto current `master` and clean Self-Review round (whole `origin/master...HEAD` diff).

## Rebase

Rebased from the previous base `ed1b565ced` (#2072) onto current `master` `22f00400b5` (#2089). The only conflict was the `packages/typia/package.json` version line, which is the known single-line ownership conflict this campaign serializes rather than avoids. Resolved to **`13.1.10`**, the correct next patch on master's `13.1.9`. `git diff origin/master...HEAD -- packages/typia/package.json` is now exactly one line, and `git diff --check` is clean — no line-ending or whitespace damage from the resolution.

No other file conflicted: nothing merged since `ed1b565ced` touches `ProtobufDecodeProgrammer.go` or `_ProtobufReader.ts`.

## Self-Review round 2 (fresh, post-rebase, whole diff): clean

No findings, so no remediation commit followed. Round 1's reasoning was re-checked against the new base and still holds: `take()` validates before mutating `ptr`; `fork()`/`close()` compose a correct stack through the returned enclosing `end`; the `-1` sentinel is safe because `end` is never negative; and every generated loop can only exit at `index === end`, so `close()`'s throw is defensive rather than a path that could reject a valid payload.

## Local verification on the rebased head (campaign CI is deliberately cancelled)

| Gate | Command | Result |
| --- | --- | --- |
| Focused programmer | `go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/core/programmers/protobuf/...` | pass |
| Native Go tree | `go -C packages/typia/test test ../native/...` | pass |
| Public Go contract tree | `go -C packages/typia/test test ./...` | pass |
| Affected TS workspace | `pnpm --filter test-typia-schema start` | pass, both new cases |
| Generated decode matrix | `pnpm --filter test-typia-automated start` | pass, 78 generated `protobuf.createDecode` round-trips plus 6 composite |

`test-typia-automated` was added to this batch's evidence during review: the PR changes `ProtobufDecodeProgrammer`, so the generated decode matrix over recursive, hierarchical-pointer, nullable, and map structures is its strongest regression gate, and it was absent from the recorded evidence.

## Deferred, with a published issue

Round 1 surfaced a **pre-existing** defect in the same file and deliberately did not widen this batch: `skip(0)` is the varint-skip sentinel, but `skipType` forwards `skip(this.uint32())` for `LEN`, so a legal zero-length unknown field consumes the next field's tag. It is now tracked as **#2116**.

It is excluded from this PR on purpose: it reproduces identically on `master`, is behaviourally unchanged by this diff, and is the opposite failure direction — #2083 is about *accepting truncated* payloads, while #2116 is about *rejecting valid* ones. #2116 lands after this PR and rebases onto it, so the two contracts cannot relabel each other's errors.

The pre-rebase head `c8e00fa180` settled after three polls with all six runs terminal (five `cancelled`; spell-check `29484109155` completed `success` before the cancel landed — the benign race already recorded in this campaign's ledger).

The rebased head `bbad991a91` settled after three polls with **all six runs `cancelled`** and none racing to terminal: `29523382825` (build), `29523382920` (website), `29523382924` (test), `29523382884` (spell-check), `29523382950` (nestia), `29523382817` (experiments). Record in `.wiki/typia-repository-2026-07/pr-2084-rebase-push-gate-bbad991a91.json`. No run for any other SHA was touched.

PR #2066, opened 2026-07-16
fix(protobuf): preserve 64-bit varint semantics

samchon:
## Intent

Restore Protocol Buffer 64-bit varint interpretation so unsigned values retain their full range and signed values are converted only at their semantic boundary.

## Scope

- Decode raw 64-bit varints as unsigned bit patterns.
- Apply two's-complement interpretation only for `int64` and ZigZag interpretation only for `sint64`.
- Cover authoritative high-bit and boundary vectors for scalar and packed fields, plus encode/decode consequences.
- Bump the `typia` package version because the shipped native Go source changes.

## Deferred

- No unrelated Protobuf malformed-input policy, native recovery, or transformer change.
- No repository-wide formatting; the campaign cleanup phase owns that result.

## Verification

Pending. This draft is an implementation-free campaign claim. Focused native/public Protobuf regressions, the affected feature suite, broader build/test verification, and solo Self-Review will be recorded before merge.

Repository Actions are deliberately cancelled only for this campaign branch's exact pushed SHA after every push and pull-request event. Local verification is the implementation gate.

Closes #2051

PR #2090, opened 2026-07-16
fix(typia): reject invalid Protobuf UTF-8 strings

### LINKED ISSUE #2088: Protobuf proto3 decoders accept invalid UTF-8 string payloads
## Problem

Typia's proto3 decoders accept invalid UTF-8 in `string` fields. The reader silently replaces malformed byte sequences with U+FFFD, and every decode variant can report a successful product value even though proto3 requires string payloads to be valid UTF-8.

## Evidence

`packages/typia/src/internal/_ProtobufReader.ts` constructs a default `TextDecoder("utf-8")`. Its non-fatal mode replaces malformed input instead of throwing, and `string()` applies it directly to every length-delimited string payload.

On current master `ed1b565cedaa772927bafbb8a4dea781b94fe62d`, an actual generated `createDecode<IFile>` decoder accepted `[10, 1, 255, 24, 0, 34, 0]`, where field 1 declares and contains exactly one payload byte `0xff`. `createDecode` and `createAssertDecode` returned U+FFFD, `createIsDecode` returned a non-null value, and `createValidateDecode` returned `success: true`. The length and reader pointer are both exact, so this reproduction contains no truncation or bounds overflow. The same payload throws when decoded with `new TextDecoder("utf-8", { fatal: true })`.

Typia's Protobuf schema generator emits `syntax = "proto3"`. The official proto3 guide requires a `string` to contain UTF-8 or 7-bit ASCII, and distinguishes `bytes` as the arbitrary-octet type: https://protobuf.dev/programming-guides/proto3/ . The official validation guidance states that proto3 parsers must validate UTF-8: https://protobuf.dev/reference/java/java-utf8-validation/ .

#1306 fixed length calculation for valid emoji strings, #912/#1073 covered sliced-buffer boundaries, #2051/#2066 covered 64-bit varints, and #2083 owns truncated length-delimited payloads. Open and closed issue/PR searches found no invalid-UTF-8 or replacement-character repair.

## Consequence surface

- Direct and factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` operations.
- Singular, optional, repeated, nested, and map key/value string fields.
- Invalid wire bytes are irreversibly changed to U+FFFD, so decode/re-encode does not preserve or reject the original payload.
- Schema and runtime disagree: the emitted `.proto` promises proto3 strings while the runtime applies a lossy non-validating policy.
- `bytes` fields must remain able to carry the same arbitrary octets and serve as the compatibility control.

## Approach

Decode Protobuf string payloads with strict UTF-8 validation and preserve the existing arbitrary-byte behavior for `bytes`. Centralize the strict decoder/error path so every generated string position receives identical behavior without changing JSON or other TextDecoder consumers.

- Valid empty, ASCII, multi-byte, emoji, and boundary strings decode unchanged across every direct and factory decode family.
- Lone continuation bytes, overlong encodings, truncated multi-byte sequences, encoded surrogate code points, and out-of-range code points are rejected rather than replaced.
- Singular, optional, repeated, nested, and map key/value string paths have malformed and valid controls.
- The same malformed octets remain valid in a Protobuf `bytes` field.
- assert/is/validate variants expose their documented failure shape and never return a successful value containing replacement text for invalid wire input.
- Focused reader tests, generated Protobuf suites, full `test-typia-schema`, typia build, and public/native Go tests pass.

## Coordination

This is independent from #2083/PR #2084: that batch checks declared length against remaining bytes, while this issue uses an exact-length payload whose contents are invalid UTF-8. Keep the claims and regressions separate and do not broaden the fix to non-Protobuf text decoding.

samchon:
## Intent

Reject invalid UTF-8 in proto3 `string` payloads before generated decoders can return lossy U+FFFD replacement text as a successful product value.

## Scope

- Apply strict UTF-8 decoding only to Protobuf `string` reads.
- Preserve arbitrary octets in Protobuf `bytes` fields as the negative compatibility control.
- Cover direct and factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` across singular, repeated, nested, and map key/value string positions.
- Pin valid ASCII, multibyte, and emoji controls plus lone-continuation, overlong, truncated-multibyte, encoded-surrogate, and out-of-range failures.
- Keep this TypeScript-runtime-only repair version-neutral.

## Deferred

- #2083 / PR #2084 owns truncated length-delimited bounds; this pull request does not duplicate that repair.
- Final verification and merge wait for PR #2084, followed by a semantic rebase that preserves its bounds regressions.
- No non-Protobuf text-decoder policy changes and no repository-wide formatting; Post-Campaign Cleanup owns formatting.

## Verification

Pending tests-first reader and generated-decoder regressions, focused suites, full `test-typia-schema`, `typia` build, public/native Go gates, and findings-first solo Self-Review through a clean round.

Campaign Actions remain enabled and are cancelled only for this branch's exact pushed SHA after every push or pull-request mutation. Local verification is the implementation gate.

Closes #2088

### DIFF
--- a/packages/typia/src/internal/_ProtobufReader.ts
+++ b/packages/typia/src/internal/_ProtobufReader.ts
@@ -77,7 +77,15 @@ export class _ProtobufReader {
   }
 
   public string(): string {
-    return utf8.get().decode(this.bytes());
+    const bytes: Uint8Array = this.bytes();
+    const decoder: TextDecoder = utf8.get();
+    try {
+      return decoder.decode(bytes);
+    } catch {
+      throw new Error(
+        "Error on typia.protobuf.decode(): invalid UTF-8 string.",
+      );
+    }
   }
 
   public fork(): number {
@@ -231,4 +239,6 @@ export class _ProtobufReader {
   }
 }
 
-const utf8 = new Singleton(() => new TextDecoder("utf-8"));
+const utf8 = new Singleton(
+  () => new TextDecoder("utf-8", { fatal: true }),
+);
--- a/packages/typia/src/internal/_ProtobufReader.ts
+++ b/packages/typia/src/internal/_ProtobufReader.ts
@@ -240,5 +240,5 @@ export class _ProtobufReader {
 }
 
 const utf8 = new Singleton(
-  () => new TextDecoder("utf-8", { fatal: true }),
+  () => new TextDecoder("utf-8", { fatal: true, ignoreBOM: true }),
 );
--- a/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
+++ b/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
@@ -1,13 +1,23 @@
 import typia from "typia";
 
-/** Verifies every generated proto3 string position rejects malformed UTF-8. */
+/**
+ * Verifies every generated proto3 string position rejects malformed UTF-8.
+ *
+ * Generated decode wrappers share the reader but add distinct assertion,
+ * predicate, and validation layers. This matrix prevents any wrapper or
+ * structural position from turning replacement text into a successful value.
+ *
+ * 1. Decode valid strings and arbitrary bytes through all eight public paths.
+ * 2. Move each malformed UTF-8 class through every generated string position.
+ * 3. Require one wire error while the identical bytes field remains valid.
+ */
 export const test_protobuf_decode_invalid_utf8 = (): void => {
-  for (const [, text] of VALID_TEXTS) {
+  for (const [label, text] of VALID_TEXTS) {
     const payload: Uint8Array = ENCODER.encode(text);
     const bytes: Uint8Array = Uint8Array.of(0xff, 0x80, 0xc0, 0xaf);
     const wire: Uint8Array = message({ bytes, defaultString: payload });
     for (const [decoder, decode] of DECODERS)
-      assertDecoded(decoder, decode(wire), text, bytes);
+      assertDecoded(`${decoder}: valid ${label}`, decode(wire), text, bytes);
   }
 
   for (const [label, payload] of INVALID_UTF8) {
@@ -16,7 +26,12 @@ export const test_protobuf_decode_invalid_utf8 = (): void => {
       defaultString: ASCII,
     });
     for (const [decoder, decode] of DECODERS)
-      assertDecoded(decoder, decode(bytesOnly), ASCII_TEXT, payload);
+      assertDecoded(
+        `${decoder}: bytes negative twin: ${label}`,
+        decode(bytesOnly),
+        ASCII_TEXT,
+        payload,
+      );
 
     for (const position of STRING_POSITIONS) {
       const wire: Uint8Array = message({
--- a/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
+++ b/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
@@ -5,9 +5,11 @@ import typia from "typia";
  *
  * Generated decode wrappers share the reader but add distinct assertion,
  * predicate, and validation layers. This matrix prevents any wrapper or
- * structural position from turning replacement text into a successful value.
+ * structural position from turning replacement text into a successful value, or
+ * from silently dropping a leading U+FEFF that the writer encoded verbatim.
  *
- * 1. Decode valid strings and arbitrary bytes through all eight public paths.
+ * 1. Decode valid strings, byte order marks, and arbitrary bytes through all eight
+ *    public paths.
  * 2. Move each malformed UTF-8 class through every generated string position.
  * 3. Require one wire error while the identical bytes field remains valid.
  */
@@ -65,9 +67,7 @@ const directIs = (input: Uint8Array): typia.Resolved<IUtf8Surface> => {
   if (value === null) throw new Error("direct isDecode returned null");
   return value;
 };
-const directValidate = (
-  input: Uint8Array,
-): typia.Resolved<IUtf8Surface> =>
+const directValidate = (input: Uint8Array): typia.Resolved<IUtf8Surface> =>
   unwrap(
     "direct validateDecode",
     typia.protobuf.validateDecode<IUtf8Surface>(input),
@@ -165,11 +165,7 @@ const message = (props: {
 };
 
 const field = (sequence: number, payload: Uint8Array): Uint8Array =>
-  Uint8Array.from([
-    (sequence << 3) | 2,
-    ...varint(payload.length),
-    ...payload,
-  ]);
+  Uint8Array.from([(sequence << 3) | 2, ...varint(payload.length), ...payload]);
 
 const concat = (...arrays: readonly Uint8Array[]): Uint8Array =>
   Uint8Array.from(arrays.flatMap((array) => [...array]));
@@ -203,9 +199,15 @@ const VALID_TEXTS = [
   ["empty", ""],
   ["ASCII", "Protocol Buffers"],
   ["literal replacement character", "\ufffd"],
+  ["leading byte order mark", "\ufefftext"],
+  ["byte order mark only", "\ufeff"],
+  ["interior byte order mark", "a\ufeffb"],
   ["multibyte", "é水값"],
   ["emoji", "🫢"],
-  ["UTF-8 boundaries", "\u0000\u007f\u0080\u07ff\u0800\uffff\u{10000}\u{10ffff}"],
+  [
+    "UTF-8 boundaries",
+    "\u0000\u007f\u0080\u07ff\u0800\uffff\u{10000}\u{10ffff}",
+  ],
 ] as const;
 
 const INVALID_UTF8 = [
--- a/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
+++ b/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_invalid_utf8.ts
@@ -1,0 +1,202 @@
+import typia from "typia";
+
+/** Verifies every generated proto3 string position rejects malformed UTF-8. */
+export const test_protobuf_decode_invalid_utf8 = (): void => {
+  for (const [, text] of VALID_TEXTS) {
+    const payload: Uint8Array = ENCODER.encode(text);
+    const bytes: Uint8Array = Uint8Array.of(0xff, 0x80, 0xc0, 0xaf);
+    const wire: Uint8Array = message({ bytes, defaultString: payload });
+    for (const [decoder, decode] of DECODERS)
+      assertDecoded(decoder, decode(wire), text, bytes);
+  }
+
+  for (const [label, payload] of INVALID_UTF8) {
+    const bytesOnly: Uint8Array = message({
+      bytes: payload,
+      defaultString: ASCII,
+    });
+    for (const [decoder, decode] of DECODERS)
+      assertDecoded(decoder, decode(bytesOnly), ASCII_TEXT, payload);
+
+    for (const position of STRING_POSITIONS) {
+      const wire: Uint8Array = message({
+        bytes: payload,
+        defaultString: ASCII,
+        malformed: { payload, position },
+      });
+      for (const [decoder, decode] of DECODERS)
+        assertInvalid(`${decoder}: ${position}: ${label}`, () => decode(wire));
+    }
+  }
+};
+
+interface IUtf8Surface {
+  singular: string;
+  optional?: string;
+  repeated: string[];
+  nested: IUtf8Nested;
+  dictionary: Map<string, string>;
+  bytes: Uint8Array;
+}
+
+interface IUtf8Nested {
+  value: string;
+}
+
+type StringPosition = (typeof STRING_POSITIONS)[number];
+
+const directIs = (input: Uint8Array): typia.Resolved<IUtf8Surface> => {
+  const value = typia.protobuf.isDecode<IUtf8Surface>(input);
+  if (value === null) throw new Error("direct isDecode returned null");
+  return value;
+};
+const directValidate = (
+  input: Uint8Array,
+): typia.Resolved<IUtf8Surface> =>
+  unwrap(
+    "direct validateDecode",
+    typia.protobuf.validateDecode<IUtf8Surface>(input),
+  );
+
+const factoryDecode = typia.protobuf.createDecode<IUtf8Surface>();
+const factoryAssert = typia.protobuf.createAssertDecode<IUtf8Surface>();
+const factoryIs = typia.protobuf.createIsDecode<IUtf8Surface>();
+const factoryValidate = typia.protobuf.createValidateDecode<IUtf8Surface>();
+
+const DECODERS: readonly (readonly [
+  string,
+  (input: Uint8Array) => typia.Resolved<IUtf8Surface>,
+])[] = [
+  ["direct decode", (input) => typia.protobuf.decode<IUtf8Surface>(input)],
+  [
+    "direct assertDecode",
+    (input) => typia.protobuf.assertDecode<IUtf8Surface>(input),
+  ],
+  ["direct isDecode", directIs],
+  ["direct validateDecode", directValidate],
+  ["factory decode", factoryDecode],
+  ["factory assertDecode", factoryAssert],
+  [
+    "factory isDecode",
+    (input) => {
+      const value = factoryIs(input);
+      if (value === null) throw new Error("factory isDecode returned null");
+      return value;
+    },
+  ],
+  [
+    "factory validateDecode",
+    (input) => unwrap("factory validateDecode", factoryValidate(input)),
+  ],
+];
+
+const unwrap = <T>(label: string, validation: typia.IValidation<T>): T => {
+  if (validation.success === false)
+    throw new Error(`${label} rejected a structurally valid message.`);
+  return validation.data;
+};
+
+const assertDecoded = (
+  decoder: string,
+  value: typia.Resolved<IUtf8Surface>,
+  text: string,
+  bytes: Uint8Array,
+): void => {
+  const mapValue: string | undefined = value.dictionary.get(text);
+  if (
+    value.singular !== text ||
+    value.optional !== text ||
+    value.repeated.length !== 1 ||
+    value.repeated[0] !== text ||
+    value.nested.value !== text ||
+    value.dictionary.size !== 1 ||
+    mapValue !== text ||
+    value.bytes.length !== bytes.length ||
+    value.bytes.some((byte, index) => byte !== bytes[index])
+  )
+    throw new Error(`${decoder} changed a valid string or bytes payload.`);
+};
+
+const assertInvalid = (label: string, task: () => unknown): void => {
+  try {
+    task();
+  } catch (error) {
+    if (error instanceof Error && error.message === INVALID_UTF8_ERROR) return;
+    throw new Error(`${label} raised an unstable error.`, { cause: error });
+  }
+  throw new Error(`${label} accepted invalid UTF-8.`);
+};
+
+const message = (props: {
+  bytes: Uint8Array;
+  defaultString: Uint8Array;
+  malformed?: { payload: Uint8Array; position: StringPosition };
+}): Uint8Array => {
+  const string = (position: StringPosition): Uint8Array =>
+    props.malformed?.position === position
+      ? props.malformed.payload
+      : props.defaultString;
+  return concat(
+    field(1, string("singular")),
+    field(2, string("optional")),
+    field(3, string("repeated")),
+    field(4, field(1, string("nested"))),
+    field(
+      5,
+      concat(field(1, string("map key")), field(2, string("map value"))),
+    ),
+    field(6, props.bytes),
+  );
+};
+
+const field = (sequence: number, payload: Uint8Array): Uint8Array =>
+  Uint8Array.from([
+    (sequence << 3) | 2,
+    ...varint(payload.length),
+    ...payload,
+  ]);
+
+const concat = (...arrays: readonly Uint8Array[]): Uint8Array =>
+  Uint8Array.from(arrays.flatMap((array) => [...array]));
+
+const varint = (value: number): number[] => {
+  const output: number[] = [];
+  do {
+    const byte: number = value & 0x7f;
+    value >>>= 7;
+    output.push(value === 0 ? byte : byte | 0x80);
+  } while (value !== 0);
+  return output;
+};
+
+const INVALID_UTF8_ERROR =
+  "Error on typia.protobuf.decode(): invalid UTF-8 string.";
+const ENCODER = new TextEncoder();
+const ASCII_TEXT = "valid";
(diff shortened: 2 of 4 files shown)


samchon:
## Self-Review round 1 — finding before remediation

A fresh full pass over the current three-file diff found one regression gap: the valid controls do not include a literal U+FFFD replacement character. The implementation correctly uses fatal decoding of the original bytes, but the existing malformed-input matrix would also pass under an incorrect future repair that decodes lossily and then rejects every decoded string containing U+FFFD. I will add U+FFFD as a valid string control in both the reader and generated-decoder suites, run the focused workspace tests, and then repeat the complete review after the mandatory post-#2084 rebase. No production-code change is required for this finding.

samchon:
Round 1 remediation landed at `9001a0088268afaa0a2feac56c9dc77209cf5257`. Both the shared reader and all generated decoder families now accept a correctly encoded literal U+FFFD while continuing to reject the six malformed byte classes. `pnpm --filter ./tests/test-typia-schema start -- --include invalid_utf8` passed both focused cases; `git diff --check` passed. The exact-SHA gate settled after two quiet polls with no Actions runs. Final broad gates and a new complete clean Self-Review remain intentionally deferred until #2084 lands and this branch is semantically rebased onto the integrated bounds repair.

samchon:
## Solo Self-Review round 2 — findings before edits

A fresh complete round over the whole `origin/master...HEAD` diff (3 files: `_ProtobufReader.ts`, `test_protobuf_decode_invalid_utf8.ts`, `test_protobuf_reader_invalid_utf8.ts`) found **one** sound defect. It is published here with its remediation plan before any fix edit.

### Finding 1 — the strict decoder still silently corrupts a *valid* proto3 string (leading U+FEFF)

`_ProtobufReader.string()` now constructs `new TextDecoder("utf-8", { fatal: true })`. `fatal` seals the malformed-byte class this pull request owns, but the option that governs the **other** lossy branch of the same decoder is still at its default: `ignoreBOM` defaults to `false`, which means the decoder **removes** a leading U+FEFF instead of returning it.

U+FEFF is a perfectly legal Unicode code point and a legal proto3 `string` payload. The write and read halves therefore disagree:

- `_ProtobufWriter` encodes through a plain `new TextEncoder()` (`_ProtobufWriter.ts:135`), which has no BOM policy and emits U+FEFF verbatim as `EF BB BF`.
- `_ProtobufReader.string()` decodes that exact byte triple back to *nothing*.

Verified decoder behavior (Node 24):

| Input | `{ fatal: true }` (this PR) | `{ fatal: true, ignoreBOM: true }` |
| --- | --- | --- |
| `EF BB BF 41 42` | `"AB"` — BOM dropped | `"﻿AB"` |
| `encode("﻿hello")` | `"hello"` — **lossy roundtrip** | `"﻿hello"` |
| `encode("a﻿b")` (interior) | `"a﻿b"` | `"a﻿b"` |

Only the *leading* position is stripped, so `typia.protobuf.encode<T>({ v: "﻿hello" })` followed by `decode` returns `"hello"` — silent data loss on input that was never malformed, and no error on any of the eight decode paths.

**Why it belongs to this pull request, not a follow-up.** This is the same defect class #2088 names — "invalid wire bytes are irreversibly changed ... decode/re-encode does not preserve or reject the original payload" — reached through the sibling option on the single line this diff already rewrites. #2088's acceptance criteria explicitly require that "valid empty, ASCII, multi-byte, emoji, and **boundary** strings decode unchanged across every direct and factory decode family"; a leading U+FEFF is exactly such a boundary and today does *not* decode unchanged. Leaving it would be whack-a-mole under the development skill's Forbidden section: same owner, same line, same lossy-string class, one byte sequence away from the case that was fixed.

**Why the current tests cannot see it.** `VALID_TEXTS` in both new test files covers empty, ASCII, a literal U+FFFD, multibyte, emoji, and the UTF-8 code-point boundaries — but no *leading* U+FEFF. The matrix is blind to this branch, so a decoder that silently truncates the first code point of a legal payload passes every assertion.

**Remediation plan (tests-first, in this order):**

1. Add a leading-U+FEFF case (and a BOM-only payload) to `VALID_TEXTS` in both new test files, and confirm they fail red against the current head `9001a00882`.
2. Set `ignoreBOM: true` on the singleton decoder so `string()` returns exactly the code points the writer encoded. `ignoreBOM: true` means "do not treat the BOM specially" — it is the correct policy for a Protobuf payload, which is a length-delimited byte string with no document-encoding preamble to sniff. `fatal` is retained unchanged, so every malformed class this PR already rejects stays rejected.
3. Re-run the focused reader and generated-decoder regressions, then the affected suites.

### Verified sound — no change proposed

These were inspected this round and are correct as written; recording them so the round is auditable rather than silent:

- **`bytes()` is outside the `try`.** `string()` calls `this.bytes()` before entering the `try`, so #2083 / PR #2084's forthcoming length-bound errors propagate untouched and cannot be relabelled as a UTF-8 failure. This is the property the rebase after #2084 must preserve.
- **The shared singleton is not poisoned by a throw.** The decoder is module-level and shared by every decode in the process, so a leaked `fatal` error state would corrupt unrelated later reads. Empirically confirmed it does not: after a throwing `decode`, subsequent valid decodes return correct text, and `[[BOMSeen]]` does not leak between calls. This holds because a non-`stream` `decode()` re-initializes decoder state at the start of every call.
- **`string()` is the single choke point.** `TextDecoder` appears nowhere else in `packages/typia/src`, and both native string emit sites in `ProtobufDecodeProgrammer.go` (the scalar path and the map/atomic path) route through `protobufDecodeProgrammer_callReader("string", ...)`. Every singular, optional, repeated, nested, and map key/value position therefore inherits one policy, and no JSON or other `TextDecoder` consumer is affected.
- **`bytes` compatibility is intact.** `bytes()` never touches the decoder, so arbitrary octets remain readable; both test files pin the same malformed octets passing through a `bytes` field as the negative twin.
- **The error message matches the file's convention** (`Error on typia.protobuf.decode(): ...`, as in `buffer overflow`) and is stable enough for the tests to assert on by equality rather than a substring, which avoids the self-catching false-pass hazard.
- **Native-free.** The diff touches no Go source, so the `packages/typia/package.json` version rule does not apply — the same disposition as the TypeScript-runtime-only Protobuf varint repair in PR #2066.

samchon:
## Solo Self-Review round 2 — remediation resolved

Round 2's single finding is fixed at `798540c600bea30f1aeb8f45ce4982851ab07e40`, tests-first as planned.

**Red before, green after.** The new `VALID_TEXTS` cases failed against the previous head `9001a00882` through the real product path, not just the reader unit:

```
Error: leading byte order mark decoded as "text".
Error: direct decode: valid leading byte order mark changed a valid string or bytes payload.
```

After setting `ignoreBOM: true` on the singleton decoder, both cases pass.

**Both options are independently load-bearing.** To prove the matrix is not vacuous, each half of the option pair was reverted in isolation and the suite re-run:

| Mutant | Decoder | Result |
| --- | --- | --- |
| A | `{ fatal: true }` (drop `ignoreBOM`) | **Fails** — `leading byte order mark decoded as "text"` |
| B | `{ ignoreBOM: true }` (drop `fatal`) | **Fails** — `invalid leading byte was accepted as a Protobuf string` |
| Fix | `{ fatal: true, ignoreBOM: true }` | **Passes** |

So the malformed-rejection matrix and the BOM-preservation matrix each pin their own option, and neither can regress silently.

**Cases added** to `VALID_TEXTS` in both test files: leading BOM (`"﻿text"`), BOM-only (`"﻿"`), and interior BOM (`"a﻿b"`) as the position control proving only the leading position was ever affected. Both doc comments were updated to state the byte-exactness requirement and its rationale.

**Verification at this head**

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-schema start -- --include invalid_utf8` | pass |
| `go -C packages/typia/test test ./utils/protobuf/... ./helpers/protobuf/...` | pass (2 packages ok) |
| `pnpm --filter ./packages/typia build` | pass (rolldown 32.35s) |
| `pnpm --filter ./tests/test-typia-schema start` (full workspace) | pass |
| `git diff --check` | clean |
| `prettier --check` on the three changed files | clean |

Targeted Prettier was applied to the three branch-owned files only; the adjacent pre-existing `test_protobuf_*_64bit_varints.ts` files are also unformatted on master but were deliberately left untouched, since repository-wide formatting belongs to Post-Campaign Cleanup.

The diff remains three files and touches no native Go, so no `packages/typia/package.json` version bump applies — the same disposition as the TypeScript-runtime-only PR #2066. The stale version line on this branch is intentionally left for the lead to own at rebase time.

Exact-SHA gate for `798540c600`: runs `29520405789`, `29520405996`, `29520406041`, `29520405921`, `29520406074`, and `29520405203` all ended `cancelled`; readback shows zero active runs.

Round 3 now reviews the complete diff from scratch.

samchon:
## Solo Self-Review round 3 — findings before edits

Fresh complete round over the whole `origin/master...HEAD` diff at `798540c600`. The runtime and test surface came back clean this round; **one** finding remains, on the documentation half of the change surface.

### Finding 1 — the new public failure mode is undocumented, and one documented sentence is now partly wrong

This pull request changes observable public behavior: `typia.protobuf.decode` and every sibling now **throw** on wire bytes that were previously accepted and returned as U+FFFD text. The repository development skill requires the matching page under `website/src/content/docs/**` to be updated in the same change when public behavior changes. `website/src/content/docs/protobuf/decode.mdx` was not touched.

Two specific gaps on that page:

1. **The Variants table understates the failure surface.** It documents `decode` as "output is whatever fell out of the binary reader", `isDecode` as "returns `null`", and `validateDecode` as "returns `IValidation<...>`". Those describe *validation* outcomes. A malformed-UTF-8 payload now raises a plain `Error` out of the reader **before** any validator runs, so all four variants — including the two documented as non-throwing — propagate it. This is pre-existing reader behavior (`buffer overflow` already throws the same way), but nothing on the page tells the reader that a wire-level error escapes the documented per-variant contract.

2. **The IMPORTANT note's last claim no longer holds for this class.** It currently states the reader "already did that decoding work" and that these validators catch "the producer encoded a UUID that's actually not a UUID" but not "someone fed me random bytes." Random bytes that are invalid UTF-8 in a `string` field are now rejected — by the reader, not the validator. The note's overall point stays correct (the validators check tag constraints, not wire structure), but that sentence needs to stop implying malformed text passes through.

**Remediation plan:**

1. Add a short subsection to `decode.mdx` stating the proto3 contract this PR enforces: `string` fields must carry valid UTF-8 and a malformed payload throws, while `bytes` fields still carry arbitrary octets. Name the rejected classes concretely rather than adjectivally.
2. Amend the IMPORTANT note's final sentence so the "random bytes" claim is scoped to structural validation and no longer implies malformed text is returned.
3. Keep the Variants table intact and add the wire-error caveat next to it rather than rewriting the per-variant contract, which is unchanged for validation failures.

Following the documentation skill: plain technical voice, one source line per paragraph, no marketing adjectives, no wrap-up restatement. Scope stays `decode.mdx`; `encode.mdx` and `message.mdx` describe behavior this PR does not change.

### Verified sound this round — no change proposed

- **`bytes()` remains outside the `try`**, so #2083 / PR #2084's forthcoming bounds errors keep their own identity and cannot be relabelled `invalid UTF-8`. Re-confirmed as the invariant the post-#2084 rebase must preserve.
- **`ignoreBOM: true` matches the reference oracle, not just our tests.** A proto3 payload is a length-delimited byte string with no encoding preamble; U+FEFF is an ordinary code point (ZWNBSP), and reference implementations do not strip it. The expectation is taken from the proto3 contract rather than from typia's own output.
- **Both `assertInvalid` helpers cannot self-catch.** The "was accepted" / "accepted invalid UTF-8" throws sit outside their own `try`, and the `catch` compares the message by equality and rethrows anything else with `cause`. This avoids the false-pass hazard found on PR #2062 round 4.
- **The negative twin holds in both files**, with the identical malformed octets passing through a `bytes` field, plus an interior-BOM control proving only the leading position was ever affected.
- **The diff is still native-free** (three files, no Go), so no `packages/typia/package.json` bump applies, matching PR #2066's disposition. The stale version line stays the lead's to own at rebase.

samchon:
## Solo Self-Review round 4 — findings before edits

Fresh complete round over the whole `origin/master...HEAD` diff at `8a7328c243`, now four files including the new documentation. The runtime and test surface came back clean again. **One** finding, in the documentation added by round 3.

### Finding 1 — the new doc snippet annotates a return type the signature does not promise

The `## Wire errors` snippet added to `decode.mdx` reads:

```typescript
const user: User = typia.protobuf.assertDecode<User>(buffer);
```

Two problems, both self-inflicted by round 3:

1. **The annotation is not what the function returns.** The page's own signature block documents `assertDecode<T>(buffer: Uint8Array): Resolved<T>`, not `T`. `Resolved<User>` collapses to `User` for this particular flat interface, so the snippet happens to compile, but as written it teaches the wrong return contract for any type where `Resolved` actually rewrites — the `Date`, class-instance, and `toJSON` cases the page links `Resolved.ts` for. Documentation should not depend on an incidental structural coincidence.
2. **It contradicts the page's own house style one screen above.** The `## First example` writes `const user = typia.protobuf.assertDecode<User>(bytes);` with no annotation, deliberately letting inference carry the type. The new snippet also renamed the buffer variable to `buffer`, so a reader scanning both sees two spellings of the same call for no reason.

Neither is a runtime defect, but the documentation is part of this change surface and the rest of the diff's correctness claims are only as good as the review that produced them, so it gets fixed on the same standard.

**Remediation plan:** drop the annotation and reuse the page's existing `bytes` variable so the snippet mirrors `## First example` exactly, leaving `try`/`catch` as the only thing the snippet is actually teaching.

### Verified sound this round — no change proposed

- **The doc's factual claims are oracle-backed, not test-backed circularity.** `Uint8Array` maps to proto3 `bytes` was confirmed in the native `ProtobufFactory.go` (`IProtobufSchema_IByte{Type: "bytes"}`) rather than inferred from typia's own output; the "throws out of `isDecode`/`validateDecode`" claim is pinned by `assertInvalid` running across all eight decoders.
- **`decode.mdx` is the only doc surface that needed changing.** `encode.mdx` and `message.mdx` make no UTF-8, replacement, or lossiness claims, and `misc.mdx`'s only Protobuf mention is an unrelated `reflect` metadata reference. Scope stays correct.
- **The `## Wire errors` section sits after the signature `<Tabs>` block**, so it does not orphan the Variants type tabs under the wrong heading — the placement was moved for exactly this reason before commit.
- **Runtime and tests unchanged and still green** at this head; the reader diff, both option semantics, the `bytes()`-outside-`try` invariant, and the negative twins were re-inspected from scratch and stand.
- **Diff remains native-free** (no Go, no `package.json`), so PR #2066's no-bump disposition still applies and the stale version line stays the lead's to own at rebase.

samchon:
## Solo Self-Review round 5 — clean round, implementation complete

Fresh complete round over the whole `origin/master...HEAD` diff at `b45f302ff2` (4 files). **No finding survived verification, so Self-Review terminates here** under the review skill's stop rule.

### What the batch ended up being

Rounds 2-4 changed the shape of the fix twice, so the summary is worth stating plainly. Issue #2088's own repair (`fatal: true`) was already in place at handoff. This session's rounds found and fixed:

| Round | Finding | Outcome |
| --- | --- | --- |
| 2 | Strict decoding still silently corrupted a **valid** proto3 string: `ignoreBOM` defaults to `false`, so a leading U+FEFF was dropped while the writer encoded it verbatim | `ignoreBOM: true`; three BOM cases added to both matrices |
| 3 | Public failure mode undocumented; one documented sentence implied malformed text is returned | `## Wire errors` added to `decode.mdx`; IMPORTANT note scoped |
| 4 | The new snippet annotated `User` where the signature returns `Resolved<T>`, and diverged from the page's own style | annotation dropped, snippet mirrors `## First example` |
| 5 | — | clean |

The round-2 finding is the substantive one: without it this PR would have shipped a decoder that rejects malformed bytes correctly while still silently losing a code point from legal input — fixing half the lossy-string class named in #2088 and leaving the other half in the same function.

### Acceptance criteria from #2088

| Requirement | Status |
| --- | --- |
| Valid empty, ASCII, multi-byte, emoji, boundary strings decode unchanged, all 8 families | pass — plus literal U+FFFD and the BOM cases round 2 added |
| Lone continuation, overlong, truncated multi-byte, encoded surrogate, out-of-range rejected | pass |
| Singular, optional, repeated, nested, map key/value have malformed + valid controls | pass |
| Same malformed octets remain valid in a `bytes` field | pass — negative twin in both files |
| assert/is/validate never return a successful value containing replacement text | pass |

### Verification at `b45f302ff2`

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-schema start -- --include invalid_utf8` | pass |
| `go -C packages/typia/test test ./utils/protobuf/... ./helpers/protobuf/...` | pass (2 ok) |
| `pnpm --filter ./packages/typia build` | pass |
| `pnpm --filter ./tests/test-typia-schema start` (full workspace) | pass |
| `prettier --check` (3 branch-owned code files) | clean |
| `git diff --check` | clean |

The full workspace run was taken at `798540c600`; every commit after it is documentation-only and cannot affect it. Mutation evidence for both decoder options is in the round-2 resolution comment.

**Not run:** the production website build. Its documented prerequisites (sibling `ttsc` checkout, root `pnpm build`, `pnpm package:tgz`) are campaign-excluded, and the website dependencies are not installed in this worktree. The MDX was instead verified structurally: no raw `<`/`{` outside code fences, fences balanced, and `## Wire errors` deliberately placed after the signature `<Tabs>` block so it does not orphan the Variants tabs. This matches the deferral already recorded for the HTTP contracts batch.

### Merge readiness

Ready on the merits; **not** ready to merge yet, by ledger order:

- Merge waits for **PR #2084** (#2083, adjacent reader bounds), then a semantic rebase. The rebase must preserve one invariant this PR established deliberately: `string()` calls `this.bytes()` **outside** the `try`, so #2084's length-bound errors keep their own identity and are never relabelled `invalid UTF-8`.
- The diff touches **no native Go and no `package.json`**, so no version bump applies — same disposition as the TypeScript-runtime-only PR #2066. The branch's stale `13.1.3` line is intentionally untouched and remains the lead's to own at rebase.

Exact-SHA gate for `b45f302ff2`: runs `29520907113`, `29520907146`, `29520906459`, `29520906317`, `29520906208`, `29520906212` all `cancelled`; readback shows zero active. Every push this session gated clean with no spell-check race.

samchon:
Lead Self-Review round 6 (fresh, post-rebase onto merged #2084): one finding, with the remediation plan, posted before any edit.

## Rebase

Rebased from `ed1b565ced` onto current `master` `ce9a8d13ab`, which now contains **#2084** (`72d9ce646f`, typia 13.1.10). All eight commits replayed with **no conflict** even though both branches modify `_ProtobufReader.ts`: #2084 owns `bytes()`, `take()`, `atomic()`, `fork()`, `close()`, and the previously unchecked `float()`/`double()`/`u8()` paths, while this branch owns `string()` and the decoder singleton. Git merged them by region.

The integration is semantically correct: `bytes()` retains #2084's bounds-checked `atomic`/`take` form, and `string()` sits on top of it. No `packages/typia/package.json` bump, correctly — the final diff contains no native Go, matching the #2066 precedent.

## Finding: the rebase invariant this batch depends on is not pinned by any test

The handoff correctly identifies the invariant — `string()` must call `this.bytes()` **outside** the `try`, so #2084's length-bound errors are not relabelled as an encoding fault:

```ts
public string(): string {
  const bytes: Uint8Array = this.bytes();   // outside: #2084's "buffer overflow" propagates as-is
  const decoder: TextDecoder = utf8.get();
  try {
    return decoder.decode(bytes);           // inside: only a decode failure becomes "invalid UTF-8"
  } catch {
    throw new Error("Error on typia.protobuf.decode(): invalid UTF-8 string.");
  }
}
```

But that invariant is enforced **only by the statement's position**. Nothing fails if a later change moves `this.bytes()` one line down into the `try`. In that state a *truncated* length-delimited string field — a wire-framing fault that #2084 exists to report precisely — would be reported as `invalid UTF-8 string`, blaming the payload's encoding for a length error and sending a user to debug the wrong layer. Both of this batch's new tests frame their payloads with a correct length, so neither can observe the difference.

This became a real, reachable interaction only when #2084 merged a few minutes ago; before that, `bytes()` could not throw for a truncated frame at all. The two error contracts now share one method and must be pinned against each other.

## Remediation plan

Add one regression to `test_protobuf_reader_invalid_utf8.ts`: frame a payload with a declared length larger than the bytes that follow and require `string()` to throw #2084's `buffer overflow` message — explicitly **not** `invalid UTF-8 string`. The existing `assertInvalid` helper already rejects an "unstable error", so the new case will use a symmetric assertion against the other message. I will observe it fail against a deliberately moved `bytes()` call before keeping the fix, so the test is proven non-vacuous rather than merely green.

Then rerun the focused reader regressions and the full `test-typia-schema` workspace, and restart a complete fresh Self-Review round.

samchon:
Round 6 remediation is pushed as `622d8cc584232180b8c32fb74f75a41cdee804f5`, and the follow-up round over the new diff is clean.

## Remediation

`test_protobuf_reader_invalid_utf8` now frames three payloads whose declared length overruns the buffer and requires #2084's `buffer overflow` message, explicitly rejecting `invalid UTF-8 string`. Every one of those payloads is itself valid UTF-8, so the frame is the only thing that can fail — that isolates the boundary contract from the encoding contract instead of entangling them.

**Proven non-vacuous by mutation, not merely observed green:**

| Reader state | Result |
| --- | --- |
| `bytes()` resolved before the `try` (as shipped) | `test_protobuf_reader_invalid_utf8: 4 ms` — pass |
| `bytes()` moved inside the `try` | fails with `declared length past the buffer was relabelled as an encoding fault.` |

The mutation was reverted and the invariant restored. The new assertion distinguishes the two failure messages rather than accepting any throw, so a relabelled error is reported as exactly that rather than as a generic instability.

## Verification on `622d8cc584`

| Gate | Command | Result |
| --- | --- | --- |
| Focused reader/decode regressions | `pnpm --filter test-typia-schema start -- --include protobuf_reader_invalid_utf8` | pass |
| Affected TS workspace | `pnpm --filter test-typia-schema start` | Success, 11,169 ms; `test_protobuf_decode_invalid_utf8` 30 ms and `test_protobuf_reader_invalid_utf8` 4 ms |
| Public Go protobuf contract | `go -C packages/typia/test test ./utils/protobuf/... ./helpers/protobuf/...` | pass (2 packages) |

No `packages/typia/package.json` bump: the final diff contains no native Go, matching the #2066 precedent for a TypeScript-runtime-only Protobuf fix.

Note for anyone re-running these: `typia/lib/internal/*` resolves through the package's export map to `src/internal/*.ts`, so these tests read the source and no `packages/typia` build is required. The `lib/` directory on a working checkout can be a stale artifact and is not what this suite exercises.

## Self-Review round 7 (fresh, whole diff): clean

No further findings. The four-file diff is the decoder singleton (`fatal` plus `ignoreBOM`), `string()`'s boundary-preserving shape, two regression matrices, and the `decode.mdx` wire-error documentation that the round-3/4 work added for the newly public failure mode.

`622d8cc584` settled with all six runs terminal: `29524738458` (experiments), `29524738386` (build), `29524738517` (test), `29524738505` (website), `29524738285` (nestia) `cancelled`, and `29524738542` (spell-check) `success` — it reached terminal before the cancel landed, the benign race this campaign has recorded on several batches and which the gate script now records instead of treating as fatal. No run for any other SHA was touched.

PR #2091, opened 2026-07-16
fix(adapters): enforce reflected output schemas

### LINKED ISSUE #2086: MCP and Vercel tools do not enforce their declared output schemas
## Problem

`@typia/mcp` and `@typia/vercel` advertise output schemas derived from controller return types but do not validate runtime execution values against those schemas. A controller implementation that returns `undefined`, `null`, an array, or a wrong-shaped object can therefore produce a nominally successful adapter result that violates the adapter's own wire contract.

## Evidence

`packages/mcp/src/internal/McpControllerRegistrar.ts` includes `outputSchema` whenever `entry.function.output` exists. After execution it returns the text `Success` for `undefined`, emits text only for null/array/primitive values, and casts any non-null object to `structuredContent` without schema validation. An actual SDK `Client` over `InMemoryTransport` rejects these declared-output calls: missing structured content becomes MCP error `-32600`, while a wrong object becomes `-32602` for output-schema mismatch. The server's direct private-handler tests cover valid objects and void functions but bypass the SDK transport validation that exposes the protocol failure.

`packages/vercel/src/internal/VercelToolsRegistrar.ts` similarly advertises `VercelParameterConverter.convertToolOutput(func.output)`, whose success branch requires typed `data`. It converts only `undefined` into the failure branch; null, arrays, and wrong objects are returned as `{ success: true, data: result }` without checking `func.output`. Because the converter supplies a JSON Schema description without a runtime validator, the installed AI SDK does not repair or reject this mismatch on behalf of the adapter.

The MCP specification requires structured tool results to conform to an advertised output schema: https://modelcontextprotocol.io/specification/2025-06-18/server/tools . #1988/#1990/#1999 and #2020/#2021 introduced argument validation, adapters, structured output, and duplicate-payload policy, but searches found no runtime output-validation issue or repair. LangChain does not advertise the same output-schema contract and is outside this issue.

## Consequence surface

- Class and HTTP controllers registered through `createMcpServer` or `toVercelTools`.
- MCP clients that correctly validate declared outputs and turn a server's apparent success into a protocol error.
- Vercel tool consumers that trust `outputSchema` for typed tool-result handling but receive invalid success data.
- `undefined`, null, arrays, primitives, missing required properties, wrong property types, and extraneous properties where exact schemas apply.
- Controller exceptions and argument-validation failures, whose existing error behavior must remain distinct from invalid return values.

## Approach

Validate every declared runtime output with the reflected function output schema before constructing a successful adapter result. Convert validation failure into each adapter's existing tool-error channel with actionable path/expected/value detail, while preserving the no-output behavior of genuinely `void` functions and the current optional MCP text fallback.

- A valid declared object result remains one MCP `structuredContent` value and one Vercel `{ success: true, data }` result.
- For a declared output, `undefined`, null, array, primitive, missing/wrong required property, and forbidden extra-property controls cannot escape as successful results.
- MCP invalid outputs travel through the public SDK client/transport and return a stable tool error rather than `-32600`/`-32602` protocol rejection caused by the adapter's malformed success response.
- Vercel invalid outputs use the advertised failure branch and include validation paths; the result itself conforms to `outputSchema`.
- A true `void` controller remains successful without a fabricated structured result, and `textFallback` still duplicates only valid structured output when enabled.
- Both class and HTTP controllers are covered, including thrown exceptions and declared-object return violations.
- MCP/Vercel package builds, public client integration tests, full `test-mcp` and `test-vercel` suites, and relevant utils/interface tests pass.

## Coordination

The two adapters share the same reflected-output trust boundary and should be one implementation batch with separate public integration matrices. Do not add LangChain output schemas or change controller return types as part of this repair.

samchon:
## Intent

Enforce the reflected runtime output schemas that `@typia/mcp` and `@typia/vercel` already advertise so invalid controller return values become adapter-level tool errors instead of malformed successes.

## Scope

- Validate declared class and HTTP controller outputs before producing successful MCP or Vercel results.
- Preserve true `void` success, controller exceptions, argument-validation behavior, and MCP `textFallback` semantics.
- Cover the public MCP SDK client/transport boundary and both adapter integration matrices.

## Deferred

- LangChain remains unchanged because it does not advertise an output schema.
- Controller return types, dependency versions, and adapter dependencies remain unchanged.

## Verification

Pending tests-first implementation, package builds, public integration, full `test-mcp`/`test-vercel`, relevant utils/interface suites, and solo Self-Review. Campaign CI is cancelled per exact-SHA policy; local verification is authoritative.

Closes #2086

### DIFF
--- a/packages/mcp/src/internal/McpControllerRegistrar.ts
+++ b/packages/mcp/src/internal/McpControllerRegistrar.ts
@@ -35,6 +35,10 @@ export namespace McpControllerRegistrar {
       registry.set(func.name, {
         function: func,
         execute: composeExecute(controller, func),
+        validateOutput:
+          func.output === undefined
+            ? undefined
+            : LlmJson.validate(func.output, true),
       });
     }
 
@@ -126,29 +130,38 @@ export namespace McpControllerRegistrar {
 
     try {
       const result: unknown = await entry.execute(validation.data);
-      if (result === undefined) {
-        return { content: [{ type: "text" as const, text: "Success" }] };
-      }
+      if (entry.validateOutput === undefined)
+        return {
+          content: [
+            {
+              type: "text" as const,
+              text: result === undefined ? "Success" : JSON.stringify(result),
+            },
+          ],
+        };
+      const output: IValidation<unknown> = entry.validateOutput(result);
+      if (!output.success)
+        return {
+          isError: true,
+          content: [
+            {
+              type: "text" as const,
+              text:
+                `Type errors in "${entry.function.name}" output:\n\n` +
+                LlmJson.stringify(output),
+            },
+          ],
+        };
       // When the reflected return type schema exists, ship the result as
       // structured output, once. The spec's duplicate text fallback for
       // clients that ignore `outputSchema` doubles the payload, so it is
       // opt-in through `textFallback: true`.
       //
-      // A result that cannot ship as `structuredContent` keeps its text
-      // block regardless — dropping it would leave the call empty.
-      const structured: boolean =
-        entry.function.output !== undefined &&
-        typeof result === "object" &&
-        result !== null &&
-        !Array.isArray(result);
       return {
-        content:
-          structured && !textFallback
-            ? []
-            : [{ type: "text" as const, text: JSON.stringify(result) }],
-        ...(structured
-          ? { structuredContent: result as Record<string, unknown> }
-          : {}),
+        content: textFallback
+          ? [{ type: "text" as const, text: JSON.stringify(result) }]
+          : [],
+        structuredContent: output.data as Record<string, unknown>,
       };
     } catch (error) {
       return {
@@ -181,4 +194,5 @@ export namespace McpControllerRegistrar {
 interface IToolEntry {
   function: ILlmFunction;
   execute: (args: unknown) => Promise<unknown>;
+  validateOutput?: ((output: unknown) => IValidation<unknown>) | undefined;
 }
--- a/packages/vercel/src/internal/VercelToolsRegistrar.ts
+++ b/packages/vercel/src/internal/VercelToolsRegistrar.ts
@@ -172,7 +172,7 @@ export namespace VercelToolsRegistrar {
                 success: false,
                 error:
                   `Type errors in "${name}" output:\n\n` +
-                  `\`\`\`json\n${LlmJson.stringify(output)}\n\`\`\``,
+                  LlmJson.stringify(output),
               } satisfies ITryResult);
         } catch (error) {
           return {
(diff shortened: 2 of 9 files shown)


samchon:
Self-Review Round 1 findings and remediation plan

Reviewed the complete 9-file base-to-head surface from scratch, including MCP protocol success/error shapes, Vercel wrapper conformance, class and HTTP execution, exact additional-property behavior, text fallback, true void results, exception preservation, public documentation, dependency boundaries, and test isolation.

Findings: no correctness, compatibility, security, data-loss, concurrency, packaging, or test-oracle defect survived verification. The implementation preserves schema-free behavior and validates only functions that advertise reflected outputs.

Remediation plan: no code remediation is required from this round. The remaining verification gap is the production website build for the two guide edits; I will run it before the final clean review round.

samchon:
Self-Review Round 2 findings and remediation plan

Fresh review of the complete base-to-head surface found one output-feedback defect: the new Vercel output-validation branch wraps LlmJson.stringify(output) in an additional Markdown code fence even though LlmJson.stringify already returns a fenced annotated payload. That would produce nested fences in model-visible error feedback.

Remediation plan: remove only the redundant outer fence from the new output-validation branch, preserve the established argument-error behavior, rerun formatting checks plus the MCP/Vercel and shared validation suites, push the focused repair, then restart the complete solo review.

samchon:
Self-Review Round 3 — clean

Repeated the complete solo review from scratch after the Round 2 repair. Reviewed every changed hunk and the full consequence surface: declared vs schema-free outputs, exact object validation, class and HTTP execution, MCP Client/InMemoryTransport protocol behavior, Vercel wrapper conformance, valid/undefined/null/array/primitive/missing/extra/wrong-nested cases, text fallback, exceptions, dependencies, documentation, test isolation, and packaging.

No further correctness, compatibility, security, data-loss, concurrency, packaging, documentation, or test-quality improvement survived verification.

Final local verification:
- pnpm --filter ./packages/mcp build
- pnpm --filter ./packages/vercel build (repeated after the review repair)
- pnpm --filter ./packages/utils build
- pnpm --filter ./packages/interface build
- full 	est-mcp, 	est-vercel, 	est-utils, and 	est-interface workspaces (all repeated after the repair)
- focused MCP SDK Client/InMemoryTransport and Vercel output matrices
- changed-file Prettier check and git diff --check
- direct production 
ext build: compiled, type-checked, generated all 73 static pages, and exported successfully

Environment note: the full website prepare wrapper could not install fresh compiler dependencies before the build because campaign tarballs were not staged; its later wasm step also hardcodes the user's sibling 	tsc checkout, which is v0.19.0-3-gd72f46d62 rather than locked v0.18.4. That user checkout was not modified. The guide MDX itself passed the direct production Next.js build.

PR #2093, opened 2026-07-16
fix(generate): preserve filesystem identity and traversal bounds

### LINKED ISSUE #2081: typia generate path identity uses platform heuristics instead of filesystem semantics
## Problem

`typia generate` derives path identity from the operating-system name instead of the actual filesystem. It assumes every macOS filesystem is case-insensitive and every Linux filesystem is case-sensitive; case-sensitive APFS, case-insensitive Linux mounts, and per-directory case settings can therefore make its keys and containment checks disagree with physical identity.

## Evidence

`packages/typia/src/executable/TypiaGenerateWizard.ts` defines `isCaseSensitive()` as false for both Windows and `darwin`. `projectFileKey()` and `filesystemKey()` therefore lowercase all macOS paths, and `indexTransformedOutputs()` stores transformed sources in a `Map` under those folded keys. Case-distinct entries overwrite one another before `getTransformedOutput()` maps them back to targets.

Apple documents both `APFS (Case-sensitive)` and `APFS (Case-sensitive, Encrypted)` formats, where names differing only by case are distinct: https://support.apple.com/en-gb/guide/disk-utility/dsku19ed921c/mac . The Linux kernel documents per-directory case-insensitive lookup for ext4: https://www.kernel.org/doc/html/v5.9/admin-guide/ext4.html . Microsoft likewise documents per-directory case-sensitivity flags on NTFS: https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity . Platform identity therefore cannot determine the path semantics of the source and output locations.

The path-key helper was introduced by #1894's generate hardening work, but tests and CI run on ordinary Windows/Linux volumes and do not exercise case-sensitive macOS. Open and closed issue/PR searches found no case-sensitive APFS fix.

## Consequence surface

- Directory and explicit-file `typia generate` modes wherever filesystem case behavior differs from the platform default.
- Transformed-output lookup, duplicate target detection, output ancestor checks, declaration/source extension matching, and symlink containment comparisons that share `isCaseSensitive()`.
- Projects that intentionally contain case-distinct filenames or directories, including repositories checked out onto case-sensitive APFS, case-insensitive Linux mounts, or Windows directories with case sensitivity enabled.
- Wrong-file generation can be silent because both physical inputs exist while only the normalized map key collides.

## Approach

Determine case behavior from the relevant filesystem/path rather than from `process.platform`, and use a consistent path-identity policy for source, project, target, and realpath comparisons. When reliable probing is impossible, fail safely on ambiguous case-distinct identities instead of selecting one transformed output silently.

- On a case-sensitive filesystem, `Foo.ts` and `foo.ts` retain distinct transformed-output and target identities.
- On a case-insensitive filesystem, aliases differing only by case are still rejected or coalesced according to the existing safety contract.
- The same modeled filesystem behavior produces the same decision regardless of the mocked host platform.
- Tests model both filesystem behaviors independently of the CI host and cover directory input, explicit file input, target collisions, declaration filtering, and realpath lookup.
- No output is written when case identity is ambiguous or inconsistent during a run.
- The test-typia-generate package, typia build, CLI smoke tests, and path-safety regression matrix pass.

## Coordination

The input-symlink traversal defect is a separate cause but shares `TypiaGenerateWizard` path ownership and can be implemented in the same generate-filesystem batch after both issues are published and independently locked by tests.

### LINKED ISSUE #2082: typia generate follows directory symlinks without physical traversal guards
## Problem

`typia generate --input <directory>` follows directory symlinks recursively without tracking physical identity or enforcing input-root containment. A symlink cycle can recurse until path/stack failure, and a symlink to an external directory silently imports sources outside the requested tree.

## Evidence

`packages/typia/src/executable/TypiaGenerateWizard.ts` implements directory gathering with `readdir`, `fs.promises.stat(next)`, and unconditional recursion when `stat.isDirectory()`. `stat` follows symlinks. The function has no `lstat` decision, realpath containment check, device/inode visited set, or physical-directory cycle guard, even though the same file already contains realpath/identity helpers for output safety.

An input entry `input/loop -> input` repeatedly re-enters the same physical directory through ever-longer lexical paths. An entry `input/vendor -> ../outside` traverses and queues supported source files outside the selected input root. Existing generate symlink tests cover output paths and ancestors, not input directory traversal.

The gather loop dates to #1816, while later #1894 hardening focused explicit inputs and output symlink escapes without adding directory-input identity tracking. Open and closed issue/PR searches found no input-cycle or external-input traversal fix.

## Consequence surface

- Directory-mode generation on Windows junctions and symbolic links, macOS aliases/symlinks exposed as filesystem links, and Unix symlinks.
- Infinite/expanding recursion, path-length or resource-exhaustion failures, duplicate physical source transformation, and nondeterministic overwrite order.
- Accidental reading and generated copying of TypeScript sources outside the requested input boundary.
- Output exclusion based only on lexical path can be bypassed when an input symlink reaches the physical output tree.

## Approach

Define and enforce directory-input symlink semantics explicitly. Track visited physical directory identities, prevent cycles and duplicate traversal, and require every followed realpath to remain within the physical input root unless external traversal is deliberately unsupported and rejected. Apply the same rule to the output directory reached through aliases.

- A self-cycle, parent-cycle, two-link cycle, and Windows junction equivalent terminate deterministically without recursion growth or partial output.
- A link to an external directory is rejected with a path-specific diagnostic and no external source is transformed or written.
- A link to the physical output directory is excluded even when its lexical path is inside input.
- Repeated aliases to one physical in-root directory cannot transform or write the same source twice.
- Ordinary nested directories and supported source symlinks retain the documented behavior.
- Tests use bounded subprocess timeouts and assert output-tree contents, not only exit status.
- The test-typia-generate package, typia build, and CLI path-safety matrix pass.

## Coordination

The case-sensitive macOS path-identity defect is separate but shares the same generator owner. Batch the implementation only if each issue keeps its own focused regressions and acceptance proof.

samchon:
## Intent

Make `typia generate` use observed filesystem case behavior instead of host-platform heuristics, and bound directory input traversal by physical containment and identity.

## Implementation

- Observe one consistent case policy across the project, input, and output locations; fail closed on inconsistent or ambiguous filesystems.
- Preserve case-distinct transformed outputs on sensitive filesystems and reject folded target collisions on insensitive filesystems.
- Track physical directories/files with inode plus realpath fallback, reject cyclic or repeated directory links, reject external links, and exclude physical output aliases.
- Build a temporary ttsc config from only the validated inputs so the compiler cannot independently re-enter rejected directory links.
- Preserve final-component source symlinks while canonicalizing aliased project parents.
- Preflight linked output ancestors before transformation or writes.

## Verification

- `pnpm --filter typia build` (ttsc + rolldown)
- `pnpm --dir tests/test-typia-generate run check:files:errors`
- Modeled case-sensitive and case-insensitive identity cases
- Filesystem matrix: 9 passed, 1 local Windows privilege skip for file-symlink creation; the case runs on Linux CI
- Directory, file, rerun, literal, project-directory, implicit-project, and jsconfig generation/build paths
- Four findings-first solo Self-Review rounds; final round clean

Campaign GitHub Actions were cancelled per exact-SHA policy; local verification is authoritative.

Closes #2081
Closes #2082

### DIFF
--- a/packages/typia/src/executable/FileSystemIdentity.ts
+++ b/packages/typia/src/executable/FileSystemIdentity.ts
@@ -1,0 +1,171 @@
+import fs from "fs";
+import path from "path";
+
+export namespace FileSystemIdentity {
+  export interface IIdentity {
+    readonly caseSensitive: boolean;
+    contains(file: string, directory: string): boolean;
+    filesystemKey(file: string): string;
+    isDeclarationFile(file: string): boolean;
+    isSamePath(x: string, y: string): boolean;
+    isSupportedExtension(file: string): boolean;
+    projectFileKey(file: string): string;
+  }
+
+  export function create(
+    caseSensitive: boolean,
+    pathApi: typeof path.posix = path,
+  ): IIdentity {
+    const fold = (value: string): string =>
+      caseSensitive ? value : value.toLowerCase();
+    const filesystemKey = (file: string): string =>
+      fold(pathApi.normalize(file));
+    return {
+      caseSensitive,
+      contains: (file, directory) => {
+        const fileKey: string = filesystemKey(pathApi.resolve(file));
+        const directoryKey: string = filesystemKey(pathApi.resolve(directory));
+        const prefix: string = directoryKey.endsWith(pathApi.sep)
+          ? directoryKey
+          : `${directoryKey}${pathApi.sep}`;
+        return fileKey === directoryKey || fileKey.startsWith(prefix);
+      },
+      filesystemKey,
+      isDeclarationFile: (file) =>
+        DTS_PATTERN.test(fold(pathApi.basename(file))),
+      isSamePath: (x, y) => filesystemKey(x) === filesystemKey(y),
+      isSupportedExtension: (file) => {
+        const filename: string = fold(pathApi.basename(file));
+        return (
+          TS_PATTERN.test(filename) && DTS_PATTERN.test(filename) === false
+        );
+      },
+      projectFileKey: fold,
+    };
+  }
+
+  export class Policy {
+    private caseSensitive_: boolean | undefined;
+    private location_: string | undefined;
+
+    public observe(caseSensitive: boolean | undefined, location: string): void {
+      if (caseSensitive === undefined) return;
+      if (this.caseSensitive_ === undefined) {
+        this.caseSensitive_ = caseSensitive;
+        this.location_ = location;
+      } else if (this.caseSensitive_ !== caseSensitive) {
+        throw new URIError(
+          `Error on TypiaGenerateWizard.generate(): inconsistent filesystem case behavior between ${this.location_} and ${location}.`,
+        );
+      }
+    }
+
+    public get(): IIdentity {
+      if (this.caseSensitive_ === undefined) {
+        throw new URIError(
+          "Error on TypiaGenerateWizard.generate(): unable to determine filesystem case behavior for this run.",
+        );
+      }
+      return create(this.caseSensitive_);
+    }
+  }
+
+  export async function inspectDirectory(
+    directory: string,
+  ): Promise<boolean | undefined> {
+    const names: string[] = await fs.promises.readdir(directory);
+    const entries: Set<string> = new Set(names);
+    for (const name of names) {
+      const alternate: string | undefined = alternateCase(name);
+      if (alternate === undefined) continue;
+      if (entries.has(alternate)) return true;
+      try {
+        await fs.promises.lstat(path.join(directory, alternate));
+        return false;
+      } catch (error) {
+        if (isMissingFileError(error)) return true;
+        throw error;
+      }
+    }
+    return undefined;
+  }
+
+  export async function probeDirectory(directory: string): Promise<boolean> {
+    const inspected: boolean | undefined = await inspectDirectory(directory);
+    if (inspected !== undefined) return inspected;
+
+    for (let attempt = 0; attempt < 10; ++attempt) {
+      const name: string =
+        `.typia-case-probe-${process.pid}-${Date.now()}-${Math.random()
+          .toString(36)
+          .slice(2)}`.toLowerCase();
+      const original: string = path.join(directory, name);
+      const alternate: string = path.join(directory, name.toUpperCase());
+      let handle: fs.promises.FileHandle;
+      try {
+        handle = await fs.promises.open(original, "wx");
+      } catch (error) {
+        if (isAlreadyExistsError(error)) continue;
+        throw new URIError(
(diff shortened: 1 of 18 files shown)


samchon:
## Self-Review Round 1 — findings before remediation

1. **Physical identity fallback:** ileIdentityKey() currently trusts stat.ino unconditionally. Filesystems that report zero/unusable inode values would collapse unrelated directories and files into one visited identity. Use canonical realpath as the fallback identity when the inode is zero.
2. **Case-probe cleanup:** probeDirectory() suppresses unlink failures in inally. A failed cleanup could leave the hidden probe behind despite a no-partial-output safety failure. Surface non-ENOENT cleanup failures with a path-specific diagnostic.
3. **Dynamic glob ambiguity:** when the static glob base has no ASCII-cased entry, inspectDirectory() returns undefined and the glob silently inherits the already-observed output policy. Fail before glob expansion when source case behavior cannot be observed reliably.

Remediation will add focused assertions where practical, rerun typia ttsc plus identity/filesystem matrices, then push through the exact-SHA gate before the next full review.

samchon:
## Self-Review Round 2 — finding before remediation

1. **Final-component source symlinks:** the bounded temporary config now records every input as its full realpath. That is correct for a project reached through an aliased parent directory, but it changes the established positive case for a supported .ts symlink whose physical target has a different filename or extension. Preserve the final lexical link component while canonicalizing parent directories, and try that same compiler key during transformed-output lookup.

Remediation will strengthen the positive source-link case to use a non-TypeScript target name, rerun ttsc and the complete focused matrices, then push through the exact-SHA gate before Round 3.

samchon:
## Self-Review Round 3 — finding before remediation

1. **Repeated-alias test contract wording:** the case intentionally fails closed before transformation, but its headline and test name still say sources are emitted/deduplicated. Align the test description with the asserted rejection so future maintainers do not infer successful alias traversal is part of the contract.

No implementation defect was found in this pass. I will correct the test wording, run the focused matrix, push/gate the documentation-only test correction, then perform a final clean review pass.

samchon:
## Self-Review Round 4 — clean

Reviewed the complete diff again across case observation/folding, glob expansion, directory and file physical identity, symlink/junction containment, output preflight, bounded ttsc project construction, transformed-output lookup, cleanup paths, and every new regression fixture.

No sound implementation improvement or meaningful missing issue candidate remains in the declared scope.

Final evidence at HEAD 8aa5f12b64:
- pnpm --filter typia build — pass (ttsc + rolldown)
- modeled identity cases — pass
- filesystem matrix — 9 pass, 1 local Windows privilege skip (file symlink; runnable on Linux CI)
- legacy check-file-errors path-safety matrix — pass
- directory/file/rerun/literal/project-directory/implicit-project/jsconfig generation and builds — pass
- exact-SHA campaign gate — all observed workflows cancelled

PR #2089, opened 2026-07-16
fix(typia): classify promised return types semantically

samchon:
## Intent

Replace name-based Promise recognition in the native transformer with compiler-semantic promised-type classification so generated functional wrappers and reflected metadata agree with TypeScript's public Promise contract.

Closes #2080.

## Scope

- classify built-in Promise returns, derived classes and interfaces, and branded Promise intersections by semantic awaited-type resolution;
- reject shadowed non-thenable types that merely carry the name `Promise`;
- use one shared classification path for functional wrapper generation and reflected function metadata;
- preserve receiver behavior merged in #2072;
- cover direct and factory functional families, fulfilled-value validation, invalid-parameter Promise shape, rejection propagation, reflection, LLM/JSON/controller consumers, and generated output;
- reserve `typia` version 13.1.9 after the preceding native release sequence through 13.1.8 is integrated.

## Deferred coordination

This branch is initially based on `ed1b565cedaa772927bafbb8a4dea781b94fe62d`. It must not become ready or merge until PRs #2071, #2075, #2074, #2077, and #2068 are merged sequentially and this branch is semantically rebased onto that resulting master.

## Verification

Pending tests-first implementation. Planned verification covers focused native/public Go tests, functional runtime matrices, reflected metadata, JSON/LLM/controller consumers, package build, and the broader repository suite required after the final semantic rebase.

Campaign CI is deliberately cancelled per exact pushed SHA. Local verification and solo Self-Review are the merge gates.

PR #2096, opened 2026-07-16
fix(typia): make native checks and failed builds atomic

### LINKED ISSUE #2079: Native check mode skips typia diagnostics and failed builds publish artifacts
## Problem

The native `ttsc-typia` host does not run typia's transformer when `check` or `--noEmit` is requested, so source files with typia transform errors can pass analysis with exit code 0. In emitting builds, the same transform diagnostics are evaluated only after JavaScript and the optional manifest have already been written, leaving failed-build artifacts that can be packaged or executed.

## Evidence

On current `master` at `1cda8afc4e7795abc92e285212d128e942049410`, `packages/typia/native/cmd/ttsc-typia/main.go` maps `check` to `runBuild(... --noEmit)`. `build.go` computes `shouldEmit` from `NoEmit` and invokes `prog.EmitWithPluginTransformers(...)` only inside `if shouldEmit`. The typia transformer and its `AddDiagnostic` callback therefore never execute for no-emit analysis, `transformDiags` remains empty, and `runBuild` returns 0 after TypeScript-only diagnostics pass.

The emitting path calls `driver.DefaultWriteFile` for each output as the transformer runs, writes the manifest after emit, and checks `len(transformDiags)` only afterward. `packages/typia/native/transform/FileTransformer.go` catches an expected `TransformerError`, records a diagnostic, and returns no replacement node; traversal then retains the original source node. A failed transform can therefore leave JavaScript containing an untransformed typia runtime stub, and the manifest advertises that file before the command exits 3.

History traces the no-emit host to #1816/#1831 and transform diagnostic collection to the later native integration. Existing diagnostic tests assert stderr/status for emit mode and TypeScript semantic diagnostics, but do not require typia diagnostics during `check` or prove that a failed build leaves output and manifest paths untouched. Open and closed issue/PR searches found #1972/#1975 for preserving native diagnostic detail and #2059/#2064 for a separate playground consumer, but no issue covering this host-level no-emit/transactionality defect.

## Consequence surface

- `ttsc-typia check`, `--noEmit`, and tsconfig-driven no-emit validation can report success without validating any typia call.
- Editors, CI jobs, and package scripts that use check-only analysis receive false-green results.
- Emitting builds can leave stale or newly written JavaScript from a command that reports transform failure.
- Optional manifests can describe outputs from a failed transform and feed downstream packaging or cleanup logic.
- Multi-file projects can mix valid transformed files with invalid untransformed stubs before the final nonzero exit.

## Approach

Give analysis and emit a shared typia-diagnostic phase. No-emit mode must traverse every relevant source through the transformer without writing files. Emitting mode must not publish outputs or a manifest until typia diagnostics are known clean; buffer writes or use another transactional strategy that preserves existing files on failure.

- A project with a TypeScript-clean but typia-invalid call makes `check`, explicit `--noEmit`, and tsconfig `noEmit` exit nonzero with the same source location/code/message as emit mode.
- No-emit validation writes no JavaScript, declarations, source maps, build info, or manifest.
- An emit project containing one valid and one typia-invalid source leaves every pre-existing output byte-identical and creates no new output or manifest entries.
- A clean multi-file project emits and manifests the same files as before.
- Subprocess tests cover direct `runBuild`, the public command dispatch, quiet/verbose output, and both explicit and config-owned no-emit.
- Native Go tests, public Go tests, typia build, and the relevant ttsc integration suites pass.
- Because native Go source changes, the typia package version advances in the implementation PR.

## Coordination

This is distinct from #1972/#1975, which preserve native diagnostic detail, and #2059/#2064, which fixed the website playground's treatment of a transform subprocess. This issue owns the native host's validation and artifact-commit semantics for every caller.

samchon:
## Summary

- run typia transform diagnostics in native check and explicit/configured no-emit modes
- make emitted JavaScript, declarations, maps, build info, and manifests transactional with respect to transform failures
- preserve pre-existing outputs byte-for-byte when a multi-file transform fails

This draft currently contains only the campaign claim commit. Tests-first implementation, full native/public verification, and solo Self-Review follow after the serialized native landing prerequisites.

Closes #2079

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/build.go
+++ b/packages/typia/native/cmd/ttsc-typia/build.go
@@ -55,26 +55,50 @@ func runBuild(args []string) int {
       return 2
     }
   }
-  prog, diags, err := driver.LoadProgram(cwd, *tsconfigPath, driver.LoadProgramOptions{
+  loadProgram := func(options driver.LoadProgramOptions) (*driver.Program, int) {
+    prog, diags, err := driver.LoadProgram(cwd, *tsconfigPath, options)
+    if err != nil {
+      fmt.Fprintf(stderr, "ttsc-typia build: %v\n", err)
+      return nil, 2
+    }
+    if len(diags) > 0 {
+      if prog != nil {
+        prog.Close()
+      }
+      driver.WritePrettyDiagnostics(stderr, diags, cwd)
+      return nil, 2
+    }
+    if diags := prog.Diagnostics(); len(diags) > 0 {
+      prog.Close()
+      driver.WritePrettyDiagnostics(stderr, diags, cwd)
+      return nil, 2
+    }
+    return prog, 0
+  }
+  prog, code := loadProgram(driver.LoadProgramOptions{
     ForceEmit:   *emit,
     ForceNoEmit: *noEmit,
     OutDir:      *outDir,
   })
-  if err != nil {
-    fmt.Fprintf(stderr, "ttsc-typia build: %v\n", err)
-    return 2
+  if code != 0 {
+    return code
   }
-  if len(diags) > 0 {
-    driver.WritePrettyDiagnostics(stderr, diags, cwd)
-    return 2
+  shouldEmit := !prog.ParsedConfig.ParsedConfig.CompilerOptions.NoEmit.IsTrue()
+  if !shouldEmit {
+    // The plugin transformer is part of tsgo's emit pipeline. Reload with emit
+    // enabled so check/noEmit traverses the exact same source set, then discard
+    // every generated output below.
+    prog.Close()
+    prog, code = loadProgram(driver.LoadProgramOptions{
+      ForceEmit: true,
+      OutDir:    *outDir,
+    })
+    if code != 0 {
+      return code
+    }
   }
   defer prog.Close()
-  if diags := prog.Diagnostics(); len(diags) > 0 {
-    driver.WritePrettyDiagnostics(stderr, diags, cwd)
-    return 2
-  }
 
-  shouldEmit := !prog.ParsedConfig.ParsedConfig.CompilerOptions.NoEmit.IsTrue()
   transformDiags := []typiaTransformDiagnostic{}
   pluginOptions := readTypiaPluginOptions(cwd, *tsconfigPath)
   transformOptions := pluginOptions.TransformOptions()
@@ -96,19 +120,33 @@ func runBuild(args []string) int {
   if !*quiet {
     fmt.Fprintf(stdout, "// ttsc-typia build: tsconfig=%s cwd=%s emit=%v rewrite=%s\n", *tsconfigPath, cwd, shouldEmit, *rewriteMode)
   }
-  if shouldEmit {
-    emitted := []string{}
-    writeFile := shimcompiler.WriteFile(func(fileName, text string, data *shimcompiler.WriteFileData) error {
+  emitted := []string{}
+  pending := []buildPendingOutput{}
+  writeFile := shimcompiler.WriteFile(func(fileName, text string, data *shimcompiler.WriteFileData) error {
+    if shouldEmit {
       emitted = append(emitted, fileName)
-      return driver.DefaultWriteFile(fileName, text)
-    })
-    eDiags, err := prog.EmitWithPluginTransformers([]driver.PluginTransform{typiaTransform}, writeFile)
-    if err != nil {
-      fmt.Fprintf(stderr, "ttsc-typia build: emit failed: %v\n", err)
-      return 3
+      pending = append(pending, buildPendingOutput{FileName: fileName, Text: text})
     }
-    for _, d := range eDiags {
-      fmt.Fprintln(stderr, "  -", d.String())
+    return nil
+  })
+  eDiags, err := prog.EmitWithPluginTransformers([]driver.PluginTransform{typiaTransform}, writeFile)
+  if err != nil {
+    fmt.Fprintf(stderr, "ttsc-typia build: emit failed: %v\n", err)
+    return 3
+  }
+  for _, d := range eDiags {
+    fmt.Fprintln(stderr, "  -", d.String())
+  }
+  if len(transformDiags) > 0 {
+    writeTypiaTransformDiagnostics(stderr, transformDiags, cwd)
+    return 3
+  }
+  if shouldEmit {
+    for _, output := range pending {
+      if err := driver.DefaultWriteFile(output.FileName, output.Text); err != nil {
+        fmt.Fprintf(stderr, "ttsc-typia build: emit write failed: %v\n", err)
+        return 3
+      }
     }
     if !*quiet {
       fmt.Fprintf(stdout, "// ttsc-typia build: emitted=%d files\n", len(emitted))
@@ -136,11 +174,12 @@ func runBuild(args []string) int {
       }
     }
   }
-  if len(transformDiags) > 0 {
-    writeTypiaTransformDiagnostics(stderr, transformDiags, cwd)
-    return 3
-  }
   return 0
+}
+
+type buildPendingOutput struct {
+  FileName string
+  Text     string
 }
 
 type typiaTransformDiagnostic struct {
--- a/packages/typia/native/cmd/ttsc-typia/build.go
+++ b/packages/typia/native/cmd/ttsc-typia/build.go
@@ -55,7 +55,7 @@ func runBuild(args []string) int {
       return 2
     }
   }
-  loadProgram := func(options driver.LoadProgramOptions) (*driver.Program, int) {
+  loadProgram := func(options driver.LoadProgramOptions, diagnose bool) (*driver.Program, int) {
     prog, diags, err := driver.LoadProgram(cwd, *tsconfigPath, options)
     if err != nil {
       fmt.Fprintf(stderr, "ttsc-typia build: %v\n", err)
@@ -68,10 +68,12 @@ func runBuild(args []string) int {
       driver.WritePrettyDiagnostics(stderr, diags, cwd)
       return nil, 2
     }
-    if diags := prog.Diagnostics(); len(diags) > 0 {
-      prog.Close()
-      driver.WritePrettyDiagnostics(stderr, diags, cwd)
-      return nil, 2
+    if diagnose {
+      if diags := prog.Diagnostics(); len(diags) > 0 {
+        prog.Close()
+        driver.WritePrettyDiagnostics(stderr, diags, cwd)
+        return nil, 2
+      }
     }
     return prog, 0
   }
@@ -79,7 +81,7 @@ func runBuild(args []string) int {
     ForceEmit:   *emit,
     ForceNoEmit: *noEmit,
     OutDir:      *outDir,
-  })
+  }, true)
   if code != 0 {
     return code
   }
@@ -87,12 +89,14 @@ func runBuild(args []string) int {
   if !shouldEmit {
     // The plugin transformer is part of tsgo's emit pipeline. Reload with emit
     // enabled so check/noEmit traverses the exact same source set, then discard
-    // every generated output below.
+    // every generated output below. The original program already proved the
+    // project's diagnostics; diagnosing the private reload would reject valid
+    // analysis-only options such as allowImportingTsExtensions.
     prog.Close()
     prog, code = loadProgram(driver.LoadProgramOptions{
       ForceEmit: true,
       OutDir:    *outDir,
-    })
+    }, false)
     if code != 0 {
       return code
     }
--- a/packages/typia/native/cmd/ttsc-typia/build_no_emit_preserves_analysis_options_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/build_no_emit_preserves_analysis_options_test.go
@@ -53,6 +53,9 @@ export const isValue = (input: unknown): input is { value: number } =>
   if code != 0 {
     t.Fatalf("valid analysis-only project failed with code %d\nstdout=%s\nstderr=%s", code, out, errText)
   }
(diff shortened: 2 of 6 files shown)


samchon:
## Self-Review round 1 ? finding

**Medium ? the clean multi-file commit-path test does not prove that every source output survives buffering.**

`TestBuildTransactionalEmitPublishesCleanProject` currently requires only that the manifest contains at least one `.js`, `.js.map`, `.d.ts`, and `.d.ts.map`. It would still pass if one of the two source files were silently dropped, so it does not fully pin the issue acceptance criterion that a clean multi-file project emits and manifests the same files as before.

Remediation: compare the manifest, relative to the project root, against the exact eight expected outputs for both source files; continue requiring every advertised path to exist and the JavaScript transform stubs to be removed. Then rerun the focused regression tests and the command package before restarting the full review.

samchon:
## Self-Review round 2 ? clean

Restarted the review from the full four-file PR diff after the round-1 test remediation.

No further sound correctness, compatibility, diagnostic, artifact-publication, or test-adequacy finding remains in the current draft:

- check, explicit `--noEmit`, and config-owned `noEmit` all force an authoritative transform traversal while discarding every compiler output;
- emitting builds retain all compiler outputs in memory until typia diagnostics are known clean;
- transform failures preserve seeded output and manifest trees byte-for-byte;
- the clean two-source path now pins the exact eight JavaScript/declaration/map outputs, verifies every manifest path exists, and verifies emitted JavaScript contains no runtime typia stub;
- manifest publication remains after output commit and is skipped for no-emit/transform-failure paths;
- the standalone native host has no matching public documentation surface that requires an update for this corrective behavior.

Verification on the implementation head:

- focused no-emit/failure/success regressions: PASS;
- full `ttsc-typia` command package: PASS;
- full native Go tree: PASS;
- full public Go tree: PASS;
- `pnpm --filter typia build`: PASS;
- targeted Go formatting and `git diff --check`: PASS;
- exact-SHA cancellation gate: all six late-arriving Actions runs cancelled and read back completed.

The PR intentionally remains draft. Final readiness still requires rebasing onto the serialized native landing chain, assigning the reserved typia package version, rerunning the full gates, and performing a fresh post-rebase Self-Review.

samchon:
Pre-landing consequence review found and remediated one no-emit boundary before the serialized version rebase. A project may legally combine \
oEmit\ with \llowImportingTsExtensions\. The private emit-enabled reload needed to reach typia's transformer was being diagnosed as though the user requested emit, producing artificial TS5096 after the original no-emit program had already passed.\n\nRemediation at \afaf77b9f8b7149b2694f7d0af3433ab513ff61\ diagnoses the user-configured program once and treats the emit-enabled reload only as a discarded transform traversal. The new independent regression failed before the fix with TS5096 and now passes.\n\nVerified locally:\n- \go -C packages/typia/test test ../native/cmd/ttsc-typia -run 'TestBuild(NoEmitReportsTransformDiagnostics|NoEmitPreservesAnalysisOnlyOptions|TransactionalEmitPublishesCleanProject|TransformFailurePreservesArtifacts)$' -count=1\\n- exact-SHA campaign gate: six runs observed and cancelled, then two quiet polls\n\nThe final fresh whole-diff Self-Review and full gates remain pending until #2068 -> #2089 -> #2084 -> #2090 lands and this branch receives its actual final typia version.

samchon:
## Self-Review round 3 — clean on the declared surface, with one out-of-scope defect referred to the lead

Round 2 declared clean, but the analysis-only compiler-option remediation (`bafaf77b9f` → `1c00aa5064`) landed afterwards, so the Review Law required a fresh complete round over the current head. This is that round: a from-scratch pass over the whole five-file `origin/master...HEAD` diff at `1c00aa5064` (merge-base `98f552144c`; `build.go` has not moved on master since, so the diff is semantically current).

### Verdict on the #2079 surface

No finding survives verification. Three candidates were raised and killed or accepted on evidence, and the round applied no edit — the branch is byte-identical to `1c00aa5064`.

### Candidate 1 — build-info leak on a failing *incremental* emit — DISPROVEN

The issue names build info explicitly, and `buildAtomicityProject` sets no `incremental`, so the emit-failure tests cannot detect a `.tsbuildinfo` leak. I built the branch binary and ran a failing emit with `"incremental": true` + `"tsBuildInfoFile": "cache.tsbuildinfo"`:

```
exit=3, files remaining: tsconfig.json only
tsbuildinfo present? NO
```

The clean incremental build then emitted **exactly the same 8 outputs and still no build info**. This host never produces build info through `writeFile` at all, so adding `incremental` to the fixture would assert a condition the compiler cannot violate — fake coverage. Candidate dropped; **no test change made**.

### Candidate 2 — check now type-checks twice — VERIFIED TRADEOFF, no change

The no-emit path deliberately reloads the program with emit enabled. Measured with both binaries against a real 181-file project (`tests/template`):

| binary | `check` elapsed | result |
| --- | --- | --- |
| master `22f00400b5` | 750–830 ms | exit 0 — **false green, no typia work done** |
| branch `1c00aa5064` | 1130–1375 ms | exit 0 — full typia traversal, wrote nothing |

~1.5x is the cost of performing the validation the command previously skipped. The cheaper alternative — flipping `NoEmit` on the already-checked program's shared `CompilerOptions` — mutates compiler state after checking and is not worth the risk. Keeping the explicit reload.

### Candidate 3 — partial publication if a write fails mid-commit — ACCEPTED RESIDUAL

If `driver.DefaultWriteFile` fails on output *N*, outputs `1..N-1` are already on disk. This is an I/O failure, not the transform-failure class #2079 owns, and the manifest is still withheld (`return 3` precedes manifest publication), so nothing advertises the partial tree. A true fix needs temp-file + rename with rollback — a separate design. Not expanding scope here.

### Regression tests are load-bearing (mutation-verified)

I did not take the green run on trust. Reverting only the `diagnose` flag on the private reload (`}, false)` → `}, true)`) makes `TestBuildNoEmitPreservesAnalysisOnlyOptions` fail with exactly the artificial diagnostic the fix exists to prevent:

```
build_no_emit_preserves_analysis_options_test.go:54: valid analysis-only project failed with code 2
  tsconfig.json:19:35 - error TS5096: Option 'allowImportingTsExtensions' can only be used
  when one of 'noEmit', 'emitDeclarationOnly', or 'rewriteRelativeImportExtensions' is set.
```

Mutation reverted; worktree clean.

### End-to-end atomicity proof (not a unit assertion)

Both halves of #2079 were reproduced out-of-process with real binaries built from master and from this branch, against a live project seeded with prior artifacts.

**Failed emit build** (`--emit --manifest`, one valid + one typia-invalid source), both exit 3:

| | master `22f00400b5` | branch `1c00aa5064` |
| --- | --- | --- |
| pre-existing `dist/valid.js` | **overwritten** (prior bytes destroyed) | byte-identical |
| new artifacts | **7 created** (`.js`/`.js.map`/`.d.ts`/`.d.ts.map` × 2) | none |
| `dist/invalid.js` | **leaked, containing `typia_1.default.is(input)`** — the untransformed runtime stub the issue predicted | absent |
| manifest | **rewritten to advertise all 8 outputs of the failed build** | still `["previous.js"]` |

**Check mode**, same project:

| | master | branch |
| --- | --- | --- |
| `ttsc-typia check` | **exit 0 — false green** | exit 3 |
| diagnostic | none | `src/invalid.ts:3:9 - error TS(typia.is): non-specified generic argument.` |

The branch leaves `dist/` and `artifacts/` byte-for-byte identical after a failed build. The invariant holds against the real filesystem, not only in-process.

### Out-of-scope defect found — same root cause, referred to the lead

The sibling single-file host in `transform.go` has the identical "publish before validating" defect and **this PR does not fix it**. Verified with the branch binary:

```
$ ttsc-typia transform --file src/invalid.ts --output ts --out out.ts
exit=0        stderr: (empty)
out.ts:  export function generic<T>(input: T): boolean {
           return typia.is<T>(input);   // untransformed stub, silently published
         }
```

`runTransformSingleJS` and the `--output ts` path call `writeSingleOutput` without ever consulting `transformDiags`, while `runTransformProject` correctly returns 3 with the diagnostic in its envelope. `defaultTransformOutput()` returns `"ts"`, so this is the **default** single-file behavior, reachable through `packages/typia/src/executable/generate/ttsc.ts` and mapped into the wasm host (`main_wasm.go` → `jsTransform`).

I am **not** fixing it in #2096, deliberately:

- #2079's acceptance criteria cover `check`/`--noEmit`/emit atomicity; `transform --file` is neither, even though the issue's "owns … validation and artifact-commit semantics for every caller" line arguably reaches it;
- flipping that path's exit code 0 → 3 changes a public CLI contract whose consumers (the ttsc transform stage, the playground wasm host) sit outside this batch's verified surface and outside the serialized native chain.

Recommending a separate campaign issue. The lead owns the batching call.

### Verification on `1c00aa5064`

| command | result |
| --- | --- |
| focused no-emit/failure/success/analysis-options matrix | PASS |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia` (full command package) | PASS (41s) |
| `go -C packages/typia/test test ../native/...` (full native tree) | PASS |
| `go -C packages/typia/test test ./...` (full public tree) | PASS |
| `go vet` on the changed package | PASS |
| `git diff --check` | PASS |
| out-of-process before/after atomicity + check repro | PASS (table above) |
| mutation check on the analysis-options regression | PASS (fails without the fix) |

One native-tree run first died with `fatal error: runtime: cannot allocate memory` while **linking** a test binary under concurrent-agent contention; it passed cleanly on retry with `-p 2`. That was machine memory exhaustion, not a code failure.

The PR stays draft. Still outstanding and owned by the lead: rebase onto the completed native serial chain and assignment of the real typia package version (the branch still carries a stale `13.1.3`, and `packages/typia/package.json` is intentionally not in this diff).

samchon:
Final semantic rebase onto current `master`, version assignment, and clean Self-Review round.

## Rebase and version

Rebased from `98f552144c` onto `14ea77e03c` (#2090). All five commits replayed with **no conflict**: nothing merged since the old base touches `build.go`, confirming the handoff's read that the implementation was still semantically current. `transform.go` did move on master via #2100/#2109, but this branch does not touch it.

Version assigned as **`13.1.11`**, the next patch on master's `13.1.10` (#2084). The bump is mandatory here because the diff changes native Go, per the development skill's Change Integrity rule; it was deliberately withheld from the branch until this rebase so the serial chain could not strand a stale number. `git diff origin/master...HEAD -- packages/typia/package.json` is exactly one line and `git diff --check` is clean.

## Lead Self-Review (fresh, post-rebase, whole diff): clean

No findings, so no remediation commit followed. The two halves of #2079 are each fixed at their real cause rather than papered over:

1. **Check mode no longer skips typia diagnostics.** The root cause is structural: the plugin transformer runs inside tsgo's *emit* pipeline, so under `noEmit` it never executed and its diagnostics could not exist. The fix reloads the program with emit forced, then discards every generated output through the `writeFile` callback rather than suppressing the transform. Passing `diagnose: false` on that private reload is correct and load-bearing: the original program already proved the project's diagnostics, and diagnosing the reload would reject valid analysis-only options such as `allowImportingTsExtensions`. `TestBuildNoEmitPreservesAnalysisOnlyOptions` pins exactly that and fails with TS5096 if the flag is reverted.
2. **A failed build no longer publishes partial output.** `writeFile` now collects into `pending` instead of writing, and the `len(transformDiags) > 0` check moved from the end of the function to immediately after emit — so the failure path returns `3` before a single file or the manifest is written. Previously that check sat *after* both the writes and the manifest update, which is precisely why a failed build could leave seven artifacts and a manifest advertising them.

Program lifetime is sound across the new control flow: the reload closes the first program explicitly before replacing it, `loadProgram` closes on each of its own error paths, and the `defer prog.Close()` is registered only once the final program is settled, so there is no double close and no leak.

## Local verification on `fdd19f197a` (campaign CI is deliberately cancelled)

| Gate | Command | Result |
| --- | --- | --- |
| Focused atomicity matrix | `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` | ok, 86.0s |
| Vet, changed package | `go -C packages/typia/native vet ./cmd/ttsc-typia/` | clean |
| Public Go contract tree | `go -C packages/typia/test test ./...` | pass, 0 failures |

The out-of-process proof recorded earlier on this thread stands and was not invalidated by the rebase, since `build.go` is unchanged on master: built from `master` and from this branch against a live project seeded with prior artifacts, both exit 3, but master overwrote a pre-existing output, created seven new artifacts, leaked an untransformed stub, rewrote the manifest to advertise all eight outputs of the failed build, and then reported `check` as exit 0 — a false green. This branch preserves every prior byte, writes nothing, withholds the manifest, and reports the diagnostic.

## Accepted residual

If an I/O error strikes midway through draining `pending`, some outputs are already on disk. True all-or-nothing would need a temp-and-rename commit. The residual is accepted and bounded: the manifest is still withheld, so no consumer is told the failed build's outputs exist. Check-mode cost roughly doubles (measured 750ms → 1130ms on a real 181-file project), which is the price of running the validation it previously skipped.

## Deferred, with a published issue

Review surfaced that `transform.go`'s single-file host carries the **identical** root cause and is unfixed: `ttsc-typia transform --file src/invalid.ts --output ts` exits `0`, emits no diagnostic, and publishes the untransformed stub, while `runTransformProject` correctly returns `3`. It is now tracked as **#2117** and deliberately excluded here, because it falls outside #2079's acceptance criteria and flips a public CLI exit-code contract whose consumers — the ttsc transform stage and the playground wasm host — are outside this batch's verified surface. #2117 lands after this PR so both hosts tell one story.

`fdd19f197a` settled after three polls with all six runs `cancelled` and none racing to terminal: `29526363760`, `29526363831`, `29526363736`, `29526363733`, `29526363793`, `29526363724`. Record in `.wiki/typia-repository-2026-07/pr-2096-rebase-push-gate-fdd19f197a.json`. No run for any other SHA was touched.

PR #2094, opened 2026-07-16
docs(setup): restore the TypeScript 6 legacy guide

samchon:
## Summary

- restore a maintained `/docs/setup/legacy` route for typia 12 and TypeScript versions below 7
- preserve `/docs/setup` as the authoritative typia 13 / TypeScript 7 setup
- document the verified v12 setup wizard, manual `ts-patch`, and `@typia/unplugin@12` paths
- record the Next.js type-only cache limitation, its rebuild workaround, and the product repair tracked by #2095

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm run build`
- `pnpm run package:tgz`
- `pnpm --dir website run test:compiler` (14 tests)
- `go -C website/compiler test ./cmd/playground`
- `pnpm --dir website run build`
- verified generated `/docs/setup` and `/docs/setup/legacy` artifacts and sitemap entries
- findings-first solo Self-Review completed

The full build used an isolated CI-like checkout with the lockfile-matched `ttsc v0.18.4`; the user-owned sibling ttsc checkout was not changed.

Closes #2092

Issue #2092, opened 2026-07-16
Typescript 6 documentation


nene:
When I added Typia to my project a few weeks ago, the documentation definitely contained information about TypeScript 6. But today, when I came to look for help on solving problems with my Typia + TypeScript 6 setup, I can no more find this documentation.

It might be that Typia doesn't support TypeScript 6 any more because in the mean time TypeScript 7 has been released.

Is there any place from where I could find the old documentation?

samchon:
You can find its README in https://github.com/samchon/typia/tree/v12.1.1

However, if you need legacy setup document, I can revive them as https://typia.io/docs/setup/legacy. If you need it, please tell me.

nene:
Thanks. In the mean time I managed to find the [setup documentation](https://github.com/samchon/typia/blob/v12.1.1/website/src/content/docs/setup.mdx) inside the Github repo.

I can manage with it, but I imagine I'm not the only one who's still running TypeScript 6. If you could revive the old docs, it would be awesome.

nene:
BTW, would it be OK to ask questions about the TypeScript 6 version of Typia, or is it completely unsupported now?

Specifically, I'm running my app with NextJS on Vercel, and it turns out that the compiled typescript code gets quite heavily cached. Which means that when I change `MyType` in one file and there's no change in a file that imports this type for use with `typia.assert<MyType>()`, that file does not get recompiled and so the assertion code remains unchanged. For now I've found a workaround by completely deleting `.next/cache` before the build.

I wonder if that's a known issue in Typia, and whether the TypeScript 7 version fixes it.

samchon:
Yes, questions about the TypeScript 6 line are welcome. `typia@12.1.1` is the matching legacy release (`typescript >=4.8 <7`); current development targets the TypeScript 7 toolchain, so fixes will normally land there unless they are specifically backported.

I audited the cache behavior you described. It is a real dependency-invalidation gap, not an application mistake: the v12 `@typia/unplugin` path cannot tell Next.js that the generated caller depends on a type-only imported file. I also checked the current TypeScript 7 path. `@ttsc/unplugin` has the required watch-dependency protocol, but typia's native transform does not yet populate that dependency list, so I cannot honestly promise that migrating alone fixes this today.

I opened #2095 for the product repair. PR #2094 will restore `/docs/setup/legacy` and document this boundary. Until #2095 lands, clearing `.next/cache` (or otherwise forcing every typia call-site module to rebuild after type-only changes) is the safe workaround.

PR #2077, opened 2026-07-16
fix(typia): backtrack tagged union and template branches

samchon:
Closes #2040.
Closes #2041.

## Campaign scope

This batch fixes two tagged branch-matching failures discovered by the July 2026 issue campaign.

- preserve wrapper tag predicates when an any/unknown-element array participates in an array or tuple union
- find a valid tagged interpolation split when a template-literal boundary is ambiguous instead of dropping placeholder tags
- cover positive, negative, boundary, union-order, validation-diagnostic, and generated-value consequences
- document runtime behavior and the intentionally lossy schema boundary

## Verification

- focused command and internal native regressions pass
- full public and `typia_native_internal` Go suites pass
- typia/package/root builds, package tarball, package matrix, and website production build pass
- five complete Self-Review rounds ended clean
- the sole later package-matrix failure was reproduced unchanged on exact base `e30b0ccf6c`, proving it is an existing master baseline issue
- exact-SHA workflow cancellation gate is quiet for head `8c50c0589f`

Repository-wide formatting, Actions permissions, and workflow activation state are unchanged.

PR #2097, opened 2026-07-16
fix(typia): invalidate bundlers on type-only changes

### LINKED ISSUE #2095: Typia transform omits type dependencies and leaves bundler validators stale
## Problem

Typia's TypeScript 7 transform host does not report the source files consulted while generating each validator. Bundlers therefore cannot invalidate a module when only an imported type changes, so incremental Next.js, Vite, Rollup, Webpack, and related builds can serve stale validators until their cache is cleared or the validator module itself changes.

## Evidence

Issue #2092 contains an external Next.js/Vercel reproduction: a file calls `typia.assert<MyType>()`, `MyType` changes in another file, the caller is not recompiled, and the old assertion remains in `.next/cache`. Deleting `.next/cache` is the reporter's current workaround.

The TypeScript 6 `@typia/unplugin` implementation confirms the failure mechanism. Its transform cache key is derived only from the current module id and source text, and its `sourceCache` retains imported source files independently. More importantly, a bundler that does not rerun the unchanged caller never reaches either cache. Type-only imports are absent from the JavaScript module graph, so Next.js has no dependency edge to invalidate.

The TypeScript 7 `@ttsc/unplugin` already implements the required consumer channel. samchon/ttsc#214 and PR samchon/ttsc#220 added a per-file `dependencies` transform-envelope field and forward it to bundler `addWatchFile`/loader dependency hooks; the regression suite explicitly covers type-only validator inputs and cache-served transforms. This has shipped since `@ttsc/unplugin` v0.15.3.

Typia does not populate that channel. On current master `98f552144c56ccf6f009f789b00fca467b2a776b`, `packages/typia/native/cmd/ttsc-typia/transform.go` encodes `transformProjectOutput` with only `diagnostics` and `typescript`. There is no `dependencies` field or consulted-source collection anywhere in the native transform host, so `@ttsc/unplugin` receives no watch files for typia-generated code. Open and closed typia issue/PR searches found no existing repair.

## Consequence surface

- Type-only changes used by `typia.is`, `assert`, `validate`, serializers, schemas, Protobuf, LLM, and every other generated operation can leave previously transformed callers stale.
- Next.js production and development caches, Vite/Rollup watch mode, Webpack/Rspack loaders, Turbopack, and other `@ttsc/unplugin` adapters all depend on the same missing producer data.
- The generated JavaScript can disagree with the current TypeScript type graph while the build still succeeds.
- Clearing the bundler cache or touching every call-site file masks the problem but does not establish a correct dependency graph.
- TypeScript 6 remains a legacy limitation, but the current TypeScript 7 toolchain exposes a dependency protocol that typia should satisfy.

## Approach

Collect the source files actually consulted while transforming each project source and emit them through the existing ttsc `dependencies` envelope. Use project-relative or absolute paths accepted by the protocol, exclude the transformed file itself, and keep the result deterministic and deduplicated. If exact checker dependency tracking is not available at the transform boundary, prefer a conservative correct project-source set over an unsound incomplete set, then narrow it only with proof.

- A project has `model.ts` exporting a type and an unchanged `validator.ts` importing it only as a type for a typia operation.
- The first transform generates code for the original model; changing only `model.ts` causes the bundler/watch transform to invalidate and regenerate `validator.ts` without clearing caches or touching it.
- The native transform JSON contains a deterministic `dependencies[validator]` entry that includes `model.ts` and excludes `validator.ts` itself.
- Cache-served `@ttsc/unplugin` transforms replay the same dependency registrations.
- Unrelated files do not disappear from required dependency coverage through aliases, re-exports, interfaces, type aliases, generic instantiations, or declaration files.
- Next.js or its shared Webpack/Turbopack loader boundary has an end-to-end regression, plus focused native transform-envelope and `@ttsc/unplugin` integration tests.
- Typia build, public/native Go suites, relevant ttsc/unplugin compatibility tests, and an incremental two-build reproduction pass.

## Coordination

The ttsc consumer protocol is already released; this issue owns typia's producer side only and must not mutate the sibling ttsc repository. #2092/PR #2094 restores the legacy TypeScript 6 documentation and may document the cache-clearing workaround, but this product repair remains a separate native implementation batch. Coordinate its typia version bump after the existing native landing sequence and #2079.

samchon:
## Intent

Emit the type-source dependency graph that Typia's native transformer consults so `@ttsc/unplugin` can invalidate unchanged validator call sites after type-only changes.

## Scope

- Produce the released ttsc `dependencies` transform-envelope field from Typia's native host.
- Keep dependency paths deterministic, deduplicated, and self-excluding.
- Cover type-only imports, re-exports, aliases, generic instantiations, and declaration inputs conservatively.
- Verify cache replay and incremental invalidation through the existing released `@ttsc/unplugin` contract without mutating the sibling ttsc repository.

## Deferred

- TypeScript 6 legacy unplugin behavior remains owned by #2092 / PR #2094.
- Final native version `13.1.12`, semantic rebase, and merge wait for #2096.
- No repository-wide formatting; Post-Campaign Cleanup owns formatting.

## Verification

Pending tests-first native transform-envelope coverage, deterministic/deduplicated/self-exclusion controls, released `@ttsc/unplugin` compatibility, incremental two-build reproduction, Typia build, public/native Go suites, and findings-first solo Self-Review through a clean round.

Campaign Actions remain enabled and are cancelled only for this branch's exact pushed SHA after every push or pull-request mutation. Local verification is the implementation gate.

Closes #2095

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/transform.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform.go
@@ -7,6 +7,7 @@ import (
   "fmt"
   "os"
   "path/filepath"
+  "sort"
   "strings"
 
   shimast "github.com/microsoft/typescript-go/shim/ast"
@@ -118,9 +119,11 @@ func runTransformProject(
   transformDiags *[]typiaTransformDiagnostic,
 ) int {
   out := transformProjectOutput{
-    Diagnostics: []transformCompilerDiagnostic{},
-    TypeScript:  map[string]string{},
+    Dependencies: map[string][]string{},
+    Diagnostics:  []transformCompilerDiagnostic{},
+    TypeScript:   map[string]string{},
   }
+  dependencySources := transformProjectDependencySources(prog, cwd)
   for _, sf := range prog.SourceFiles() {
     if sf.IsDeclarationFile {
       continue
@@ -129,7 +132,11 @@ func runTransformProject(
     if filepath.IsAbs(key) || key == ".." || strings.HasPrefix(key, "../") {
       continue
     }
-    out.TypeScript[key] = transformFileToTypeScript(prog, typiaTransform, sf)
+    transformed := transformFileToTypeScriptResult(prog, typiaTransform, sf)
+    out.TypeScript[key] = transformed.TypeScript
+    if transformed.Changed {
+      out.Dependencies[key] = excludeTransformDependency(dependencySources, key)
+    }
   }
   for _, diag := range *transformDiags {
     out.Diagnostics = append(out.Diagnostics, transformDiagnosticToCompilerDiagnostic(diag))
@@ -152,6 +159,19 @@ func transformFileToTypeScript(
   typiaTransform driver.PluginTransform,
   sf *shimast.SourceFile,
 ) string {
+  return transformFileToTypeScriptResult(prog, typiaTransform, sf).TypeScript
+}
+
+type transformFileOutput struct {
+  Changed    bool
+  TypeScript string
+}
+
+func transformFileToTypeScriptResult(
+  prog *driver.Program,
+  typiaTransform driver.PluginTransform,
+  sf *shimast.SourceFile,
+) transformFileOutput {
   options := prog.TSProgram.Options()
   ec := shimprinter.NewEmitContext()
   result := sf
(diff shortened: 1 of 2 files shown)


samchon:
## Tests-first implementation evidence

The claim base failed the new focused regression exactly because the project transform emitted no `dependencies` entry for `src/validator.ts`. The implementation now passes the same regression and its consequence matrix:

- a type-only import crossing a tsconfig path alias, re-export, type alias, generic instantiation, ambient declaration, and outside-root declaration;
- deterministic sorted values, deduplication, validator self-exclusion, and no entry for untouched files;
- relative project paths plus an absolute outside-root path, matching the released ttsc protocol;
- two identical transforms with byte-stable dependency maps;
- a second build changing only `model.ts`, which regenerates the unchanged validator while preserving its dependency graph;
- `--rewrite-mode none` as the no-dependency negative control.

The checker exposes no per-query provenance, so the producer uses the issue's conservative sound fallback: every watchable source in the TypeScript program is reported for files whose Typia transform actually changes the AST. `bundled:///` TypeScript library URIs are excluded because they have no filesystem watch target. Untouched outputs receive no dependency entry.

Focused transform/command/diagnostic tests pass. The repository-pinned `ttsc` and `@ttsc/unplugin` 0.18.4 sources were inspected read-only and confirm parsing of this optional envelope, project-relative/absolute normalization, self/dedup filtering, and cache-hit `addWatchFile` replay. The sibling ttsc checkout was not mutated.

The full command package currently reaches unrelated tagged-type and assertion-guard baseline failures owned by the pending #2077/#2068/#2089 landing chain. Final package-wide gates, Typia 13.1.12, and clean solo Self-Review remain pending the required post-#2096 semantic rebase.

samchon:
Closing unmerged: both halves of this draft's approach are now owned elsewhere, so no scope remains.

- The **precise half** landed in PR #2100 (closed #2098/#2092/#2095): per-file consulted-declaration collection at the metadata-analysis funnel, emitted through the `dependencies` envelope. It strictly dominates a coarse whole-program list in precision while remaining sound for the files typia actually consults.
- The **conservative-superset half** (whole-program `dependencies`) is the degenerate "treat the whole project as input" form: every file change invalidates every module, which is sound but carries zero precision — and its rightful owner is the compiler host, not a plugin. That host-owned baseline (per-file reference graph incl. type-only edges, ambient files, config chain — `tsc --incremental` semantics) is being implemented plugin-agnostically in samchon/ttsc#716.

The serialized landing chain this draft was deferred behind (#2096 → #2079) is unrelated to dependency reporting and is unaffected by this closure.

PR #2074, opened 2026-07-16
fix(openapi): align scalar validation semantics

samchon:
## Overview

Align OpenAPI scalar semantics across generated validators, shared schema validation/containment, random generation, and documentation:

- enforce exact finite-decimal `multipleOf` divisibility for type/JSDoc tags and OpenAPI validators;
- generate only representable in-range multiples, including one-sided and exclusive ranges, and reject empty discrete ranges;
- make schema coverage prove constant, bound, divisibility, string, array, tuple, and uniqueness containment;
- make OneOf array/object validation independent of discriminator branch order while retaining useful failure paths;
- count JSON Schema string lengths in Unicode code points in generated and shared validators.

Closes #2044
Closes #2045
Closes #2046

## Integration

This branch is based on current `master` after #2071 and #2075. Their prototype-safe object access, reference-cycle handling, tuple minimum semantics, float32 bounds, and format validation contracts are retained. Package version advances from `13.1.5` to `13.1.6`.

## Verification

Exact HEAD: `0bc53eb0c53158dc223728b83b4f180186a2cc35`

- `pnpm --filter @typia/interface --filter @typia/utils --filter typia -r build`
- full `@typia/test-utils` suite
- full `@typia/test-typia-schema` suite
- `pnpm test:go` (public and untagged native suites)
- `go -C packages/typia/test test -tags typia_native_internal ../native/...`
- focused affected native transform/runtime fixtures
- numeric containment property matrix: 698 schemas, 201 values, 175,712 claimed-cover pairs, zero counterexamples
- string containment property matrix: 94 schemas, 15 values, 2,252 covered pairs, zero counterexamples
- array/tuple containment property matrix: 450 schemas, 40 values, 48,870 covered pairs, zero counterexamples
- deterministic random matrix: 48 valid checks
- CI-equivalent exact-head root build
- exact-head website production build, including compiler WASM, examples, Next typecheck, 74 static pages, Pagefind, and sitemap
- final fresh solo Self-Review: no findings
- `git diff --check`; clean worktree; local and remote HEAD match

Repository Actions are deliberately cancelled by exact campaign commit SHA after every push and pull-request event. No workflow or repository Actions setting is disabled; the local verification above is the campaign gate.

PR #2100, opened 2026-07-16
feat(native): emit per-file consulted-declaration dependencies in the transform envelope

### LINKED ISSUE #2092: Typescript 6 documentation

### LINKED ISSUE #2098: native transform: emit per-file consulted-declaration dependencies (standalone fix for #2092)
## Problem

typia's native transform host emits a stdout envelope containing only `diagnostics` + `typescript` — `transformProjectOutput` in `packages/typia/native/cmd/ttsc-typia/transform.go` has no `dependencies` field — so the dependency channel that ttsc shipped in v0.15.3 (samchon/ttsc#214 → samchon/ttsc#220) carries nothing. Bundler caches therefore never learn that a module's generated validator depends on the files where its validated types are declared, and both watch mode and persistent caches serve stale validators. Field report: #2092 (Next.js `.next/cache` keeps a stale `typia.assert<MyType>()` when only `MyType`'s file changes; workaround is deleting the cache before every build). #1722 requested exactly this and was closed when the protocol half landed in ttsc; the producer half was never implemented.

## Evidence

- Reproduced on npm artifacts (typia 13.1.1 + ttsc/@ttsc/unplugin 0.19.1), webpack 5 filesystem cache: add a required property to `MyType` in its own file, rebuild with the cache kept → bundle still contains the old validator and accepts the now-invalid input; deleting the cache produces the correct rejection.
- The consumer half works end-to-end today: a one-line fixture loader registering the type's source file via `this.addDependency()` for the consumer module makes the cached rebuild pick up the type change. Production is the only gap.
- `grep -ri dependenc packages/typia/native` — zero hits: no collection, no emission.

## Approach

Ownership context (see samchon/ttsc#716 for the full design): the plugin-agnostic reference-graph baseline is the ttsc driver's job. typia owns two things — its extra-programmatic inputs (none today) and the **precise** consulted-file knowledge that only its own analysis has.

1. **Collect per transformed file the set of source files owning the declarations the transform actually consulted.** The natural funnel is metadata analysis: `MetadataFactory.Analyze` (`native/core/factories/MetadataFactory.go`) resolves every named type to checker symbols, and each symbol's declarations carry their `SourceFile().FileName()`. Attribute the set to the file being transformed (the call-expression site), covering the full resolution the analysis performs — nested properties, generic arguments, union/intersection members, inherited members, and ambient declarations. Skip nothing the analysis read; include `node_modules` declaration files (registering them is sound; consumers decide how to snapshot them). Default TS `lib.*.d.ts` files may be excluded — they are toolchain-versioned, not project inputs.
2. **Emit the collection as `dependencies` in the transform envelope** from `runTransformProject`, keyed like `typescript` (project-relative paths), values project-relative or absolute per the protocol.
3. This is **standalone-sufficient and union-safe today**: the shipped `@ttsc/unplugin` (≥ 0.15.3) already registers reported entries, so this change alone closes #2092 without waiting for samchon/ttsc#716. Once #716 lands, the host graph baseline unions with this list; narrowing below the baseline is deferred until ttsc specifies the completeness contract.
4. **Follow-up after samchon/ttsc#716 releases** (separate change): route envelope encoding through the driver's SDK encoder so typia also carries the host `graph` section.

## Consequence surface

- All bundler adapters benefit identically (webpack/Next.js persistent cache, watch/HMR, Turbopack) — the channel is adapter-shared.
- Single-file transform modes (`--file`) and `build` emit no envelope; only the project envelope path changes.
- Envelope size: one string list per transformed file, bounded by the type graph actually consulted — no measurable protocol concern.

- Go tests: for a fixture where `a.ts` calls `typia.validate<T>()` with `T` declared in `b.ts` (and a case where `T` references `c.ts` transitively through a property type), the envelope's `dependencies["a.ts"]` contains `b.ts` (and `c.ts`); a file with no typia call reports no entry; the transformed file itself may be omitted (the consumer drops it).
- End-to-end: the webpack filesystem-cache reproduction above passes **without deleting the cache** — changing the type's file rebuilds the consumer and the new validator rejects the stale input.
- Existing native transform tests and the TypeScript-side test suites keep passing.

## Coordination

- samchon/ttsc#716 — host-owned reference graph (plugin-agnostic baseline); this issue is independent of it for delivery, and its follow-up item 4 depends on its release.
- History: #1722 (original request), samchon/ttsc#214 / samchon/ttsc#220 (protocol + consumer, shipped).
- Field report: #2092.

samchon:
## Intent

Implements #2098: the native transform host collects, per transformed file, the set of source files owning the declarations its metadata analysis actually consulted, and emits that collection as `dependencies` in the project transform envelope.

This is the producer half of the dependency channel that ttsc shipped in v0.15.3 (samchon/ttsc#214 / samchon/ttsc#220). The shipped `@ttsc/unplugin` consumer already forwards reported entries to the bundler (`addWatchFile` / `addDependency`), so this change alone closes #2092: webpack/Next.js persistent caches and watch mode learn that a module's generated validator depends on the files where its validated types are declared.

## Scope

- Collect consulted declaration files during metadata analysis (`packages/typia/native/core/factories/...`), attributed to the file being transformed, covering nested property types, generic arguments, union/intersection members, inherited members, type tags, and ambient declarations. `node_modules` declaration files are included; default `lib.*.d.ts` files are excluded (toolchain-versioned, not project inputs).
- Emit `Dependencies map[string][]string` (`"dependencies,omitempty"`) from `runTransformProject` in `packages/typia/native/cmd/ttsc-typia/transform.go`, keyed like the `typescript` map (project-relative, forward slashes).
- Go tests for direct, transitive, and no-typia-call cases; end-to-end verification against a webpack 5 filesystem-cache project using the locally built typia.

## Deferred

- Issue #2098 approach item 4 (routing envelope encoding through the ttsc driver SDK encoder so typia also carries the host `graph` section) is a follow-up after samchon/ttsc#716 releases and is intentionally NOT part of this pull request.

## Verification (to be recorded on this thread as work lands)

- `pnpm test:go:native` and the `typia_native_internal`-tagged transform tests.
- `pnpm test` (packages + toolchain).
- Webpack 5 + `cache: { type: "filesystem" }` + `@ttsc/unplugin/webpack` + esbuild-loader reproduction: build, mutate the type's own file, rebuild with cache kept, assert the bundle carries the new check and rejects the previously-valid input.

Campaign note: this branch follows the issue-campaign CI discipline — runs for campaign SHAs are cancelled immediately after each push and local verification is the gate; no workflow or Actions setting is touched.

Closes #2092.
Closes #2098.

https://claude.ai/code/session_01U3kDYEkHutvSMcHPBXhd7P


samchon:
## Implementation status and local verification (through 731e064)

### What landed

1. **Collection funnel** (`packages/typia/native/core/schemas/metadata/MetadataDependency.go`): a per-checker dependency listener registry. Metadata analysis reports every consulted declaration file through three touch surfaces:
   - `Iterate_metadata` entry — every checker type the analysis reads (top types, union/intersection members, alias targets, array/tuple elements, generic arguments, property types) reports its structural symbol's and type-name symbol's declaration files.
   - `MetadataCollection.ApparentProperties` — every enumerated property symbol reports its declaring file, which covers inherited members (their symbols stay declared in the base file).
   - `Emplace_metadata_object` — the `extends`/`implements` chain is walked explicitly so a base that currently declares no property (an empty interface) still registers; gaining its first property is exactly the change that must invalidate the consumer.
2. **Written-reference walker** (same file + `CallExpressionTransformer`): an alias of an intrinsic (`type Id = string`) leaves no symbol on the resolved checker type — the checker interns the intrinsic — so the type-graph touches alone missed its declaring file (verified by probe before the fix: `type Outer = { value: Inner }` with `type Inner = number` in its own file reported only the outer file). The walker resolves every reference written in call-site type arguments, heritage elements, and type-annotated declarations (following import aliases) and recurses through type-alias declarations. Probe after: both the intermediate and the collapsed alias file are reported.
3. **Envelope producer** (`cmd/ttsc-typia/transform.go`): `runTransformProject` registers a collector, attributes touches to the file being transformed, and emits `Dependencies map[string][]string` (`"dependencies,omitempty"`), keyed like `typescript` (project-relative, forward slashes). Default `lib.*.d.ts` files and the transformed file itself are dropped; `node_modules` declarations are included. Values under the project are project-relative; others fall back per `sourceFileKey` (the consumer absolutizes against the project root).

### Local verification

- `pnpm test:go` — all green (public + native trees), including the two new command-level tests:
  - `TestProjectDependenciesEnvelopeTransform` — `a.ts` → `typia.validate<Bee>` with `Bee` in `b.ts` referencing `Cee` in `c.ts`: `dependencies["src/a.ts"]` contains both; no entry for `b.ts`/`c.ts`/`d.ts` (no typia call); no self entry; no `lib.*.d.ts` leak.
  - `TestProjectDependenciesEmptyHeritageTransform` — empty `extends` target registers.
- `pnpm test` (full TS + Go suite) — every suite passes except `tests/test-typia-generate`, see below.
- **End-to-end #2092 reproduction (acceptance)**: webpack 5 + `cache: { type: "filesystem" }` + `@ttsc/unplugin/webpack` + esbuild-loader, using the LOCALLY BUILT typia tarball (`pnpm package:tgz`):
  - Build 1: `typia.validate<MyType>` accepts `{"id":"x"}`, rejects `{}` — correct.
  - Add required `age: number` to `mytype.ts` only; rebuild WITH the cache kept (webpack log confirms `[cached]` modules in play): the bundle now contains the `age` check, **rejects** `{"id":"x"}` (`$input.age` error) and accepts `{"id":"x","age":3}`.
  - **Control** with npm `typia@13.1.1` (no producer), same harness: the cached rebuild keeps the stale validator — `{"id":"x"}` still accepted, zero `age` occurrences in the bundle. The producer half is what closes #2092.

### Pre-existing failure, not from this change

`tests/test-typia-generate` `check:files:filesystem` fails 3 of 10 node:test cases on this Windows machine (`repeated-directory-aliases`: "input directory link revisits a physical directory"). Verified on a clean `master` worktree (bc06291926) in the same environment: identical 3 failures. Environmental/pre-existing; unrelated to this PR (the generate wizard's directory-alias traversal is TypeScript-side and this PR only touches the native transform envelope path).

### Scope note (owner's standing order)

The development skill asks for a `typia` version bump alongside native Go changes; per the owner's standing rule that versioning is his domain and PRs carry only functional changes, no `package.json` version field is touched here.

samchon:
## Self-Review Round 1 — findings and remediation plan

Full fresh pass over every changed hunk plus the consequence surface (producer, collection funnels, walker, tests, packaging, CI, sibling PRs). Findings, to be applied after this comment:

1. **Walk precision: declaration walks include bodies and initializers.** `metadataDependency_touchVisited` walks whole method/property declarations, so a reference appearing only in a method BODY or property initializer registers as a dependency although the generated validator never consults it — noise that causes unnecessary invalidations. Remediation: restrict the walk to the written type surface — the whole node for `type` aliases (type parameters + aliased type only), the type annotation for properties and parameters, and parameter types + return type for methods, accessors, and index signatures.
2. **Bundled URI sources must never become dependency values.** tsgo serves default libraries from `bundled:///` URIs; the `lib.*.d.ts` basename filter already drops them, but the collector gets an explicit `://` scheme guard so no URI can ever reach the consumer's `path.resolve`.
3. **Collector cost.** `Touch` recomputes `filepath.Rel` for every reported file; memoize the fileName → envelope-value mapping.
4. **Verification gap found in my own evidence.** The earlier "full `pnpm test`" stopped at the first failing workspace (`test-typia-generate`, pre-existing Windows filesystem failures — verified identical on a clean master worktree), so suites ordered after it never ran locally. `test-typia-schema`, `test-utils`, `test-vercel`, `test-utils-automated` are being run explicitly now; a full re-run will be recorded before merge.
5. **CI status is master-inherited, not from this change.** The `test` / `nestia` workflow failures on this branch are the `OpenApiValidator` union regressions that PR #2101 documents as failing on exact clean master (`e30b0ccf`, regressed by #2074). Plan: land after the fix merges (rebase onto master) and require green runs on the head SHA before merging, per the ordinary check loop.
6. **Overlap disclosure.** Draft PR #2097 also emits the `dependencies` envelope but as a conservative whole-program source universe (every changed file depends on all program sources), targets #2095, and defers its merge pending #2096. This PR implements the precise consulted-declaration set that #2098 specifies and is standalone. Flagging so the owner can reconcile #2097 after this lands.

Applying 1–3 now; 4–5 gate the merge.

samchon:
## Self-Review Round 2 — findings and remediation plan

Fresh full pass over the current head (d2b63d2), including the shim API surface behind the new walk (`Node.Type()` covers every kind fed to it with a nil default; `Parameters()` is guarded to function-like kinds), the registry lifecycle, envelope determinism, packaging (`files` includes `native`, tarball carries the new source), and the wasm target (compiles).

Finding:

1. **The written-reference walker has no committed regression test.** The collapsed-alias behavior (`type Outer = { value: Inner }` / `type Inner = number` in its own file) and the body-reference exclusion (a type referenced only inside a method body must NOT register) were verified only by ad-hoc probes. Remediation: add a command-level Go test pinning both the positive closure (intermediate and collapsed alias files reported) and the negative twin (body-only reference absent).

Applying now. Local verification running on the final tree: full `pnpm test` and the `typia_native_internal` tagged suite; results will be recorded before merge together with the CI plan from round 1 (rebase after the master-side #2074 regression fix lands via #2101).

samchon:
## Self-Review Round 3 — clean round; final verification

Third complete fresh pass over the whole diff (head `9d67933`, master merged after #2101): registry lifecycle and nil-safety, walk boundaries (`Node.Type()` kind coverage, function-like `Parameters()` guards), heritage recursion and cycle guards, collector determinism, attribution across the sequential project loop, single-file/build/wasm modes (no listener, no envelope change), packaging (`files` ships `native`), and all three committed tests. No further sound improvement found — Self-Review concludes after rounds 1 and 2 produced applied findings and round 3 is clean.

### Final verification record

**CI (head `9d67933`, all workflows executed normally): all green** — build, test, nestia, experiments, website, spell-check. The earlier `test`/`nestia` failures were master's #2074 union regression; they turned green after merging master with #2101's fix.

**Local (Windows, final tree):**
- `pnpm run test:go` — 29 packages ok (public + native), including the three new envelope tests (direct + transitive attribution, no-call exclusion, self/lib exclusion; empty heritage; collapsed alias closure + body-reference boundary).
- `go test -tags typia_native_internal ../native/...` — 27 packages ok (includes the project-catalog transform coverage over the repository's test projects with the collector active).
- `pnpm test` — every workspace passes except `tests/test-typia-generate`, whose `check:files:filesystem` directory-junction cases fail identically on a clean master worktree in this environment (Windows-only, pre-existing; the same suite is green in the Ubuntu CI `test` workflow above).

**End-to-end acceptance (#2092), locally built typia tarball, webpack 5 + filesystem cache + `@ttsc/unplugin/webpack` + esbuild-loader:**
- Build 1: validator accepts `{"id":"x"}`, rejects `{}`.
- Add required `age: number` to the type's own file, rebuild with the cache kept (2–7 s cached rebuild): bundle carries the new check, rejects `{"id":"x"}` with `$input.age`, accepts `{"id":"x","age":9}`.
- Control with npm `typia@13.1.1` (no producer): identical harness keeps serving the stale validator from cache.

Marking ready for review and merging under the standing autonomous mandate.

Issue #2103, opened 2026-07-16
Generated OpenAPI component names can violate the Components key grammar and local-reference syntax


samchon:
## Problem

`typia.json.schema()`, `typia.json.schemas()`, and `typia.json.application()` expose OpenAPI 3.0 and 3.1 compatible schema output, but generated `components.schemas` names are not constrained to the OpenAPI Components Object key grammar. Legal TypeScript identifiers and generic literal arguments can therefore produce component keys containing `/`, `~`, `%`, `#`, or non-ASCII characters.

The generator then forms `$ref` and discriminator mapping values by concatenating the component name onto `#/components/schemas/` without JSON Pointer token escaping or URI-fragment encoding. Some generated documents are consequently invalid OpenAPI documents, and some references do not identify the component that the generator stored.

No invalid TypeScript or manual schema mutation is required.

## Public reproduction

With the normal typia transformer configured:

```typescript
import typia from "typia";

type Recursive<T extends string> = {
  value: T;
  children: Recursive<T>[];
};

interface Café {
  value: string;
  child?: Café;
}

interface IArguments {
  plain: Recursive<"Plain">;
  dotted: Recursive<"A.B">;
  dashed: Recursive<"A-B">;
  underscored: Recursive<"A_B">;
  slash: Recursive<"A/B">;
  tilde: Recursive<"T~N">;
  percent: Recursive<"C%D">;
  hash: Recursive<"A#B">;
  space: Recursive<"A B">;
  colon: Recursive<"A:B">;
  unicode: Café;
}

const v30 = typia.json.schema<IArguments, "3.0">();
const v31 = typia.json.schema<IArguments, "3.1">();

console.log(JSON.stringify({ v30, v31 }, null, 2));
```

Both versions emit the same component names and matching raw references:

| Type input | Generated component key | Generated `$ref` | OpenAPI key |
| --- | --- | --- | --- |
| `Recursive<"Plain">` | `RecursivePlain` | `#/components/schemas/RecursivePlain` | valid control |
| `Recursive<"A.B">` | `RecursiveA.B` | `#/components/schemas/RecursiveA.B` | valid control |
| `Recursive<"A-B">` | `RecursiveA-B` | `#/components/schemas/RecursiveA-B` | valid control |
| `Recursive<"A_B">` | `RecursiveA_B` | `#/components/schemas/RecursiveA_B` | valid control |
[...]

PR #2104, opened 2026-07-16
Fix LLM local-reference integrity

### LINKED ISSUE #2102: LLM $defs references violate JSON Pointer semantics and can fail open
## Problem

`ILlmSchema.IParameters.$defs` accepts arbitrary string keys and `ILlmSchema.IReference.$ref` is documented as a JSON Pointer, but typia's LLM reference pipeline treats the suffix after `#/$defs/` as a raw object key. Producers also interpolate raw keys into `$ref` values.

Valid definition keys containing `/`, `~`, spaces, `%`, or other URI-fragment characters therefore cannot be addressed with their RFC 6901 and URI-fragment encodings. Malformed raw spellings may be followed as if they were valid, while valid escaped spellings, percent-encoded spellings, and missing targets can be replaced with `{}` during LLM-to-OpenAPI inversion. The empty schema is then treated as unknown and accepts arbitrary values.

This makes traversal, coercion, validation, conversion, and provider adapters disagree about the same schema graph. It is reachable from both public schema-accepting APIs and generated recursive typia schemas.

## Public evidence

The following schema uses only the exported `ILlmSchema.IParameters` and `LlmJson.validate()` contracts:

```typescript
import { ILlmSchema } from "@typia/interface";
import { LlmJson } from "@typia/utils";

const parameters: ILlmSchema.IParameters = {
  type: "object",
  properties: {
    value: {
      $ref: "#/$defs/A~1B",
    },
  },
  required: ["value"],
  additionalProperties: false,
  $defs: {
    "A/B": {
      type: "number",
    },
  },
};

const validate = LlmJson.validate(parameters);
console.log(validate({ value: 42 }));
console.log(validate({ value: "wrong" }));
```

`#/$defs/A~1B` is the RFC 6901 spelling for the `A/B` definition. The number passes, but the string also returns `success: true`. The valid reference misses the raw-key lookup, inversion installs an empty component schema, and OpenAPI validation treats that schema as accept-all.

The same public generator path emits non-canonical references without any manually constructed schema:

```typescript
import typia from "typia";

type Recursive<T extends string> = {
  value: T;
  children: Recursive<T>[];
};

interface IArguments {
  slash: Recursive<"A/B">;
  tilde: Recursive<"T~N">;
  percent: Recursive<"C%D">;
  combined: Recursive<"A~/B">;
}

console.log(JSON.stringify(typia.llm.parameters<IArguments>(), null, 2));
```

The generated `$defs` keys are correct JSON object keys, but their references are emitted raw:

```text
"RecursiveA/B"  -> "#/$defs/RecursiveA/B"
"RecursiveT~N"  -> "#/$defs/RecursiveT~N"
"RecursiveC%D"  -> "#/$defs/RecursiveC%D"
"RecursiveA~/B" -> "#/$defs/RecursiveA~/B"
```

Canonical fragment spellings include `RecursiveA~1B`, `RecursiveT~0N`, `RecursiveC%25D`, and `RecursiveA~0~1B`.

## Standards oracle and case matrix

[RFC 6901](https://www.rfc-editor.org/rfc/rfc6901.html) requires `~` to be encoded as `~0`, `/` as `~1`, decoding in the order `~1` then `~0`, and unresolved pointer evaluation to produce an error condition. Its URI-fragment representation additionally requires percent encoding where appropriate. [JSON Schema 2020-12](https://json-schema.org/draft/2020-12/json-schema-core#name-direct-references-with-ref) defines `$ref` as a URI-reference and `$defs` as an object whose schemas are addressed by references.

| `$defs` key | `$ref` | Standards result | Current result |
| --- | --- | --- | --- |
| `Plain` | `#/$defs/Plain` | resolves | control behaves correctly |
| empty string | `#/$defs/` | resolves the empty token | control behaves correctly |
| `A/B` | `#/$defs/A~1B` | resolves | target is missed; validation can fail open |
| `A/B` | `#/$defs/A/B` | extra pointer segment / unresolved | raw key may be followed by traversal and coercion, then split differently during inversion |
| `T~N` | `#/$defs/T~0N` | resolves | target is missed; validation can fail open |
| `T~N` | `#/$defs/T~N` | malformed token | malformed raw spelling is accepted as the key |
| `~1` | `#/$defs/~01` | resolves after `~1`, then `~0` decoding | decode-order boundary is missed |
| `A B` | `#/$defs/A%20B` | resolves | percent-encoded fragment is missed |
| `C%D` | `#/$defs/C%25D` | resolves | percent-encoded fragment is missed |
| `C%D` | `#/$defs/C%D` | malformed percent encoding | malformed raw spelling is accepted as the key |
| no target | `#/$defs/Missing` | resolution error | inversion can synthesize `{}` and validation accepts arbitrary values |

AJV 8.18.0 independently resolves and enforces the valid escaped and percent-encoded cases. RFC 6901 remains the syntax oracle for malformed raw-tilde tokens because some JSON Schema implementations are permissive there.

## Verified root cause

There is no shared owner for local LLM references. Each producer and consumer slices or interpolates the string independently:

- `LlmTypeChecker` uses raw `split` and `replace` operations for traversal, flattening, and coverage.
- `coerceLlmArguments` uses a raw prefix replacement for public coercion and schema-assisted parsing.
- `LlmSchemaConverter` extracts and emits raw keys in compose, invert, and discriminator paths; inversion falls back to `{}` when a definition lookup misses.
- the native `LlmSchemaProgrammer` extracts OpenAPI keys and constructs `#/$defs/` values without token or fragment encoding.

The fail-open transition is:

```text
valid escaped or encoded local reference
  -> raw lookup misses $defs
  -> inversion installs component {}
  -> OpenApiValidator treats {} as unknown
  -> an incompatible runtime value validates successfully
```

Malformed raw references follow a different path: traversal and coercion may find the raw key, but inversion or validation splits the same text differently. Fixing one call site would leave the graph inconsistent.

## Consequence surface

- `LlmTypeChecker.visit`, `escape`, and `covers` omit, misresolve, or compare the wrong referenced schema.
- `LlmJson.coerce` and schema-assisted `LlmJson.parse` do not coerce provider strings through valid encoded references, while malformed raw references may be followed.
- `LlmJson.validate` and schema-first `LlmJson.structuredOutput(...).validate` can accept incompatible values for valid escaped, encoded, or missing references.
- generated `typia.llm.parameters`, `typia.llm.schema`, applications, and structured-output schemas can expose raw recursive references to JSON Schema and provider tooling.
- `LlmSchemaConverter` compose/invert and discriminator mappings do not preserve definition identity.
- MCP, LangChain, and Vercel adapter input/output boundaries depend on these coercion and validation paths; they can reject recoverable inputs or accept invalid outputs depending on direction.
- ordinary keys, empty keys, reserved own-property names, chained references, and valid recursion must remain supported.

## Approach

Establish one shared local-reference codec and resolver for LLM `$defs` and use it at every producer and consumer boundary.

The invariant is:

1. Emit canonical local URI-fragment references by JSON Pointer token encoding first and URI-fragment encoding second.
2. Parse only the supported local-reference shape, percent-decode the fragment, split the JSON Pointer, then decode tokens in RFC order.
3. Distinguish malformed syntax, unsupported reference shapes, missing targets, and valid recursive graphs through each public path's existing error channel.
4. Never substitute an empty schema for a malformed or missing reference in validation conversion.

Do not implement another collection of call-site replacements, and do not special-case only `/` and `~`; the fragment, decode order, empty token, malformed encodings, and graph states are one correctness boundary.

- Generated references for `/`, `~`, `A~/B`, `~1`, spaces, `%`, Unicode, and ordinary keys are canonical and resolve to the intended `$defs` entry.
- The empty key `#/$defs/` and reserved own-property names continue to work.
- Decoding proves the `~1`-before-`~0` boundary.
- Raw slash, malformed tilde, raw or malformed percent escapes, extra pointer segments, wrong prefixes, unsupported non-local references, and missing targets fail deterministically.
- Chained references and valid recursion terminate with consistent traversal, coercion, validation, conversion, and coverage behavior.
- `LlmTypeChecker`, `LlmSchemaConverter`, `LlmJson.parse`, `LlmJson.coerce`, `LlmJson.validate`, and both structured-output paths share the invariant.
- MCP, LangChain, and Vercel tests exercise their actual argument and output validation boundaries.
- Independent JSON Schema validation accepts generated canonical references and enforces the referenced constraints.
- Run focused `@typia/utils` and adapter suites, native transform tests for generated references, affected package builds, and the broader applicable repository suite.

## Coordination

Closed issue #2042 and merged PR #2071 partially overlap only the missing-reference edge: they established prototype-safe schema dictionaries and required missing references to fail, but `LlmSchemaConverter` still converts a missing `$defs` target to `{}`. They did not implement JSON Pointer token or URI-fragment handling.

The linked OpenAPI component-name defect is separate. OpenAPI restricts component keys and therefore needs deterministic legal name allocation with collision handling. LLM `$defs` keys are unrestricted JSON object keys and need correct reference encoding and decoding. Pointer encoding alone cannot legalize an OpenAPI component name, and OpenAPI name allocation cannot fix arbitrary LLM `$defs` references.

Issue #1593 is related history about generated reference length and possible collisions, not JSON Pointer semantics or fail-open validation. Closed PRs #1778 and #1864 touch LLM references generally but do not implement this invariant. Exact open and closed issue and pull-request searches found no duplicate.

samchon:
## Overview

This draft claims one coordinated repair for #2102 across LLM schema generation and every internal consumer of local `$defs` references.

The implementation will establish a single RFC 6901 token plus URI-fragment codec, emit canonical references from native generation and OpenAPI-to-LLM conversion, decode them consistently in checking/coercion/inversion, and fail deterministically for malformed or missing references through `LlmJson` and the LangChain, MCP, and Vercel adapter boundaries.

OpenAPI component-key allocation is intentionally excluded and tracked separately by #2103: OpenAPI names have a restrictive grammar, while LLM `$defs` keys remain unrestricted and require lossless reference encoding.

## Delivery state

- [x] Lead-vetted issue and evidence packet published
- [x] Isolated worktree and claim-only Lore commit
- [ ] Regression tests demonstrating current failures
- [ ] Implementation
- [ ] Package-scoped verification
- [ ] Solo Self-Review rounds
- [ ] Fresh-master semantic rebase and current native package version assignment

The native version is deliberately not speculated on this claim commit. This branch will take the next valid version only after the active native landing chain has completed and the branch has been refreshed from `master`.

Closes #2102

### DIFF
--- a/packages/interface/src/schema/ILlmSchema.ts
+++ b/packages/interface/src/schema/ILlmSchema.ts
@@ -65,8 +65,9 @@ export namespace ILlmSchema {
      * Named schema definitions for reference.
      *
      * Contains type definitions that can be referenced throughout the schema
-     * using `$ref: "#/$defs/TypeName"`. Essential for recursive types and
-     * shared structures.
+     * using `$ref: "#/$defs/TypeName"`. Definition names are encoded as one RFC
+     * 6901 token in URI-fragment form. Essential for recursive types and shared
+     * structures.
      */
     $defs: Record<string, ILlmSchema>;
 
@@ -394,14 +395,15 @@ export namespace ILlmSchema {
    *
    * Points to a schema defined in the `$defs` map using a JSON pointer. Used to
    * avoid schema duplication and enable recursive type definitions. The
-   * reference path format is `#/$defs/TypeName`.
+   * reference path format is `#/$defs/TypeName`, where `TypeName` is one RFC
+   * 6901 token encoded as a URI fragment.
    */
   export interface IReference extends IJsonSchemaAttribute {
     /**
      * JSON pointer to the referenced schema.
      *
-     * Must be in the format `#/$defs/TypeName` where TypeName is a key in the
-     * parent schema's `$defs` map.
+     * Must be in the format `#/$defs/TypeName`, where `TypeName` decodes to a
+     * key in the parent schema's `$defs` map.
      */
     $ref: string;
   }
--- a/packages/typia/native/core/programmers/llm/LlmSchemaProgrammer.go
+++ b/packages/typia/native/core/programmers/llm/LlmSchemaProgrammer.go
@@ -318,7 +318,7 @@ func llmSchemaProgrammer_convert_schema_config(schema nativeiterate.JsonSchema, 
           defs[key] = llmSchemaProgrammer_convert_schema_config(target, components, defs, config)
         }
       }
-      union = append(union, map[string]any{"$ref": "#/$defs/" + key})
+      union = append(union, map[string]any{"$ref": llmSchemaProgrammer_encode_reference(key)})
       return
     }
     if input["type"] == "object" {
@@ -643,6 +643,23 @@ func llmSchemaProgrammer_ref_key(ref string) string {
   return ref[index+1:]
 }
 
+func llmSchemaProgrammer_encode_reference(key string) string {
+  token := strings.NewReplacer("~", "~0", "/", "~1").Replace(key)
+  var output strings.Builder
+  output.WriteString("#/$defs/")
+  for _, value := range []byte(token) {
+    if (value >= 'A' && value <= 'Z') ||
+      (value >= 'a' && value <= 'z') ||
+      (value >= '0' && value <= '9') ||
+      strings.ContainsRune("-_.!~*'()", rune(value)) {
+      output.WriteByte(value)
+    } else {
+      fmt.Fprintf(&output, "%%%02X", value)
+    }
+  }
+  return output.String()
+}
+
 func llmSchemaProgrammer_component(components *nativeiterate.OpenApi_IComponents, key string) (nativeiterate.JsonSchema, bool) {
   if components == nil || components.Schemas == nil {
     return nil, false
@@ -677,7 +694,7 @@ func llmSchemaProgrammer_discriminator(schema nativeiterate.JsonSchema, union []
   if rawMapping, ok := llmSchemaProgrammer_schema_map(discriminator["mapping"]); ok {
     mapping := map[string]any{}
     for key, value := range rawMapping {
-      mapping[key] = "#/$defs/" + llmSchemaProgrammer_ref_key(fmt.Sprint(value))
+      mapping[key] = llmSchemaProgrammer_encode_reference(llmSchemaProgrammer_ref_key(fmt.Sprint(value)))
     }
     output["mapping"] = mapping
   }
--- a/packages/typia/native/core/programmers/llm/llm_reference_codec_test.go
+++ b/packages/typia/native/core/programmers/llm/llm_reference_codec_test.go
@@ -1,0 +1,24 @@
+package llm
+
+import "testing"
+
+// TestLlmReferenceCodec locks JSON Pointer token escaping before URI-fragment encoding.
+func TestLlmReferenceCodec(t *testing.T) {
+  t.Parallel()
+  cases := map[string]string{
+    "Plain": "#/$defs/Plain",
+    "":      "#/$defs/",
+    "A/B":   "#/$defs/A~1B",
+    "T~N":   "#/$defs/T~0N",
+    "A~/B":  "#/$defs/A~0~1B",
+    "~1":    "#/$defs/~01",
+    "A B":   "#/$defs/A%20B",
+    "C%D":   "#/$defs/C%25D",
+    "Café":  "#/$defs/Caf%C3%A9",
+  }
+  for key, expected := range cases {
+    if actual := llmSchemaProgrammer_encode_reference(key); actual != expected {
+      t.Fatalf("encode %q: expected %q, got %q", key, expected, actual)
+    }
+  }
+}
--- a/packages/typia/package.json
+++ b/packages/typia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typia",
-  "version": "13.1.11",
+  "version": "13.1.12",
   "description": "Superfast runtime validators with only one line",
   "main": "src/index.ts",
   "exports": {
--- a/packages/utils/src/converters/LlmSchemaConverter.ts
+++ b/packages/utils/src/converters/LlmSchemaConverter.ts
@@ -7,6 +7,7 @@ import {
 } from "@typia/interface";
 
 import { JsonDescriptor } from "../utils/internal/JsonDescriptor";
+import { LlmReference } from "../utils/internal/LlmReference";
 import { ObjectDictionary } from "../utils/internal/ObjectDictionary";
 import { OpenApiSchemaSanitizer } from "../utils/internal/OpenApiSchemaSanitizer";
 import { LlmTypeChecker } from "../validators/LlmTypeChecker";
@@ -245,7 +246,7 @@ export namespace LlmSchemaConverter {
           const out = () => {
             union.push({
               ...input,
-              $ref: `#/$defs/${key}`,
+              $ref: LlmReference.write(key),
             });
           };
           if (ObjectDictionary.has(props.$defs, key)) return out();
@@ -433,7 +434,10 @@ export namespace LlmSchemaConverter {
                         Object.entries(props.schema.discriminator.mapping).map(
                           ([key, value]) => [
                             key,
-                            `#/$defs/${value.split("/").at(-1)}`,
+                            LlmReference.write(
+                              value.split("#/components/schemas/")[1] ??
+                                value.split("/").at(-1)!,
+                            ),
                           ],
                         ),
                       )
@@ -511,20 +515,28 @@ export namespace LlmSchemaConverter {
         );
       else if (LlmTypeChecker.isAnyOf(schema)) schema.anyOf.forEach(visit);
       else if (LlmTypeChecker.isReference(schema)) {
-        const key: string =
-          schema.$ref.split("#/$defs/")[1] ?? schema.$ref.split("/").at(-1)!;
-        if (ObjectDictionary.get(props.components.schemas, key) === undefined) {
+        const resolved: LlmReference.IResolved | undefined =
+          LlmReference.resolve(props.$defs, schema.$ref);
+        if (resolved === undefined) {
+          union.push({ oneOf: [] });
+          return;
+        }
+        const componentKey: string = LlmReference.componentKey(resolved.key);
+        if (
+          ObjectDictionary.get(props.components.schemas, componentKey) ===
+          undefined
+        ) {
           props.components.schemas ??= {};
-          ObjectDictionary.set(props.components.schemas, key, {});
+          ObjectDictionary.set(props.components.schemas, componentKey, {});
           ObjectDictionary.set(
             props.components.schemas,
-            key,
-            next(ObjectDictionary.get(props.$defs, key) ?? {}),
+            componentKey,
+            next(resolved.schema),
           );
         }
         union.push({
           ...schema,
-          $ref: `#/components/schemas/${key}`,
+          $ref: `#/components/schemas/${componentKey}`,
         });
       } else if (LlmTypeChecker.isBoolean(schema))
         if (!!schema.enum?.length)
@@ -586,13 +598,9 @@ export namespace LlmSchemaConverter {
                         props.schema["x-discriminator"].propertyName,
                       mapping:
                         props.schema["x-discriminator"].mapping !== undefined
-                          ? Object.fromEntries(
-                              Object.entries(
-                                props.schema["x-discriminator"].mapping,
-                              ).map(([key, value]) => [
-                                key,
-                                `#/components/schemas/${value.split("/").at(-1)}`,
-                              ]),
+                          ? invertDiscriminatorMapping(
+                              props.$defs,
+                              props.schema["x-discriminator"].mapping,
                             )
                           : undefined,
                     }
@@ -601,6 +609,23 @@ export namespace LlmSchemaConverter {
     } satisfies OpenApi.IJsonSchema;
   };
 }
+
+const invertDiscriminatorMapping = (
+  $defs: Record<string, ILlmSchema>,
+  mapping: Record<string, string>,
+): Record<string, string> | undefined => {
+  const entries: [string, string][] = [];
+  for (const [discriminator, reference] of Object.entries(mapping)) {
+    const resolved: LlmReference.IResolved | undefined =
+      LlmReference.resolve($defs, reference);
+    if (resolved === undefined) return undefined;
+    entries.push([
+      discriminator,
+      `#/components/schemas/${LlmReference.componentKey(resolved.key)}`,
+    ]);
+  }
+  return Object.fromEntries(entries);
+};
 
 const validateStrict = (
   schema: OpenApi.IJsonSchema,
--- a/packages/utils/src/converters/LlmSchemaConverter.ts
+++ b/packages/utils/src/converters/LlmSchemaConverter.ts
@@ -102,6 +102,7 @@ export namespace LlmSchemaConverter {
                 description: result.value.description,
               },
               escape: true,
+              key: LlmReference.readOpenApi(props.schema.$ref)!,
             })
           : result.value.description,
       } satisfies ILlmSchema.IParameters,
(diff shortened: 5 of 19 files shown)


samchon:
## Solo Self-Review round 1 findings

I reviewed the complete 13-file base-to-head surface from scratch, including the shared codec/resolver, native producer, converter/inverter, traversal/coverage/coercion consumers, generated output, public docs, and real adapter boundaries.

Two test-integrity improvements survived verification:

1. The LangChain, MCP, and Vercel fixtures assert canonical emitted references, but their valid calls already contain correctly typed values. They therefore do not prove that each adapter's schema-assisted input coercion actually follows the encoded reference. I will add a numeric field beneath the recursive referenced type, submit it as a string at each adapter boundary, and assert the executed result contains a number.
2. The native change updates both ordinary `$ref` emission and `x-discriminator.mapping`, but the generated-schema regression currently covers only ordinary recursive references. I will add a reserved-character discriminated union and assert the native mapping values use the same canonical codec.

Remediation remains inside #2102. OpenAPI restricted component-name allocation stays excluded in #2103/#2105. After the tests and implementation evidence are updated, I will rerun the affected focused gates and begin a fresh complete Self-Review round.

samchon:
## Solo Self-Review round 1 additional finding

The converter's first implementation encoded an LLM `$defs` key once, then used that encoded text both as the OpenAPI component object key and as the reference suffix. That is internally accepted only because the current OpenAPI validator slices references as raw strings.

It is not a valid JSON Pointer round-trip. For example, storing a component under `A~1B` and emitting `#/components/schemas/A~1B` makes an RFC resolver look for the raw key `A/B`; storing `A%20B` and emitting the same suffix makes it look for `A B`. External schema tooling therefore misses the component even though typia's internal raw lookup finds it.

I will separate component identity from reference encoding and add an independent resolver assertion. This remains coordinated with #2103/#2105: #2104 owns lossless pointer identity for arbitrary LLM definitions, while #2105 owns allocating OpenAPI-legal component names. I will not introduce a competing legal-name allocator here.

samchon:
Self-Review round 1 found one additional reference-identity defect before remediation.

`LlmSchemaConverter.transform` still extracts OpenAPI component names with string splitting in its validation pass, reference visit, and discriminator mapping. A standards-compliant reference such as `#/components/schemas/A~1B` must resolve the raw component key `A/B`, but the converter instead looks for `A~1B`. If both raw keys exist it can silently select the wrong schema, and discriminator conversion double-escapes the already encoded token to `#/$defs/A~01B`. The current converter regression constructs references as `#/components/schemas/${key}`, so slash, percent, whitespace, and Unicode cases are malformed URI-fragment JSON Pointers and mask this path.

Remediation will remain in #2104's reference-codec scope:

1. make the converter validation, visit, and discriminator paths decode only canonical local component references through the shared helper;
2. reject malformed, missing, or unsupported external references instead of recovering a suffix with `split("/").at(-1)`;
3. change the regression inputs to canonical OpenAPI references and add colliding raw keys such as `A/B` and `A~1B`, proving the referenced raw identity is selected and emitted once as a canonical `$defs` reference;
4. preserve raw unrestricted component identities in this intermediate conversion. #2105 remains responsible for allocating OpenAPI-legal names when LLM definitions are inverted outward.

samchon:
Self-Review round 1 found one more forward-conversion entry point before remediation.

`LlmSchemaConverter.parameters` delegates a root reference to `LlmParametersFinder`, which currently calls the generic `OpenApiTypeChecker.unreference` string-splitting path. A canonical root reference such as `#/components/schemas/A~1B` therefore searches for the encoded spelling instead of the raw component key `A/B`; if both raw keys exist it can select the wrong schema. After conversion, `JsonDescriptor.cascade` repeats the raw suffix extraction, so even a separately resolved root would lose or misattribute its component description.

Remediation will make the LLM parameters entry point dereference canonical local component references through `LlmReference.readOpenApi`, with deterministic malformed, missing, and cyclic failure, then pass the decoded raw root identity into description cascading. A root-reference collision regression will prove that `A~1B` resolves `A/B` rather than the literal `A~1B` key. This is converter-local reference handling; it does not change the general OpenAPI component allocator owned by #2105.

samchon:
Self-Review round 1 found an additional validation consumer before remediation.

`OpenApiOneOfValidator.getFlattened` still takes `schema.$ref.split("/").pop()` while inspecting inverted LLM unions. #2104 now correctly stores the raw component key `A/B` and emits `#/components/schemas/A~1B`; the discriminator probe therefore misses the component, represents it as unknown `{}`, and selects that first unresolved-looking branch as the entire remainder set. A value valid only for a later encoded-reference branch can then be rejected without that branch being tried. Missing, malformed, or cyclic aliases likewise become an accept-all discriminator sentinel during branch selection even though the station validator would reject them later.

Remediation will decode component references with `LlmReference.readOpenApi` in the flattening probe and represent malformed, missing, or cyclic flattening as an impossible schema rather than unknown. The LLM JSON Pointer matrix will validate values belonging to both encoded discriminator branches through the public `LlmJson.validate` path. This is the validation consumer of #2104's raw-key/reference invariant; OpenAPI name allocation remains #2105's responsibility.

samchon:
## Self-Review round 1 remediation: two finding premises were wrong

Round 1's five findings-first comments are now implemented. While running the verification gate on that remediation, two of the round-1 test plans turned out to rest on incorrect premises. Correcting them changed the tests, not the implementation.

### 1. The native discriminator regression asserted a discriminator that is never emitted

Finding 2 planned "a reserved-character discriminated union" to lock the native `x-discriminator.mapping` codec. It was written with a generic **type alias** (`type Variant<T> = { kind: T; value: number }`), and the assertion failed. The cause is not the codec: `json_schema_discriminator` emits nothing for that shape at all, on the JSON-schema layer, for reserved **and** plain names alike. My diff never touches that gating, so this is pre-existing native behavior and out of #2102's scope.

A generic **interface** does produce the discriminator. Empirical matrix over `typia.llm.parameters`:

| Shape | `x-discriminator` |
| --- | --- |
| `ICat \| IDog` (named interfaces, inline union) | emitted |
| `IVariant<"Alpha"> \| IVariant<"Beta">` (generic interface) | emitted |
| `IVariant<"A/B"> \| IVariant<"T~N">` (generic interface, reserved) | emitted, `#/$defs/IVariantA~1B` + `#/$defs/IVariantT~0N` |
| `Variant<"Alpha"> \| Variant<"Beta">` (generic type alias) | **none** |

The test now uses `IVariant<T>`, so it actually exercises the codec. Red-proved: reverting the native mapping line to `"#/$defs/" + ref_key(...)` fails the suite on `encodes discriminator mapping references`; restoring it passes.

I am **not** fixing the type-alias gating here — it is a separate pre-existing question, unrelated to JSON Pointer semantics. Flagging it for the lead as a possible candidate.

### 2. LangChain cannot demonstrate typia's input coercion — it pre-validates

Finding 1 planned to submit a numeric string at each adapter boundary and assert typia coerces it through the encoded reference. That is unreachable for LangChain: `toLangChainTools` passes `entry.function.parameters` as the tool `schema`, and LangChain validates input with `@cfworker/json-schema` **before** the adapter closure calls `LlmJson.validateArguments`. A numeric string never reaches typia's coercion. MCP and Vercel do not pre-validate, so their coercion assertions stand and pass.

That gate is more valuable than the original plan: it is an independent JSON Schema implementation acting as an oracle. With **raw** refs, cfworker registers anchors under the canonical spelling and cannot resolve `#/$defs/RecursiveA/B`, throwing on the **valid** tree:

```
- https://github.com/cfworker#/$defs/RecursiveA~1B/properties/children
  at validate (@cfworker/json-schema/dist/commonjs/validate.js:77:19)
  at DynamicStructuredTool.call (@langchain/core/dist/tools/index.cjs:117:54)
```

This independently reproduces #2102 outside typia and confirms `~1` is the resolvable spelling. The LangChain test now asserts that a valid recursive value executes and that two violations of **only** the referenced definition (`count: "42"`, `value: "wrong"`) are rejected at LangChain's own gate — which it can only do by resolving the encoded reference. A fail-open reference would admit both.

### Verification on this remediation

| Command | Result |
| --- | --- |
| `go -C packages/typia/native test ./core/programmers/llm/...` | pass |
| `go -C packages/typia/native test ./...` | pass (12 packages) |
| `pnpm --filter @typia/utils build` | pass |
| `pnpm --filter test-utils start` | pass |
| `pnpm --filter test-typia-schema start` | pass |
| `pnpm --filter test-langchain start` | pass |
| `pnpm --filter test-mcp start` | pass |
| `pnpm --filter test-vercel start` | pass |

A fresh complete Self-Review round follows this push.

samchon:
## Solo Self-Review round 2 findings

Fresh complete round over the whole `origin/master...HEAD` surface (17 files) at `d705d5f151`: shared codec, native producer, converter forward/invert, parameters entry point, description cascade, type-checker traversal and coverage, coercion, the two OpenAPI validators, generated output, docs, and the five test suites. I re-derived the traversal rewrite against `OpenApiTypeCheckerBase.visit` and confirmed it preserves node coverage, accessor formats, and dedup semantics while adding discriminator-mapping validation.

Two improvements survive verification.

### 1. The extra-pointer-segment guard has no test

`LlmReference.decode` rejects a decoded token containing `/`:

```ts
if (token.includes("/")) return undefined;
```

That guard implements the issue's acceptance criterion "reject extra pointer segments where only one `$defs` key is supported". It is the only thing standing between `#/$defs/A%2FB` and silently resolving the `A/B` definition. `%2F` is a **well-formed** percent triplet, so it passes the fragment regex and `decodeURIComponent`; per RFC 6901 §6 it percent-decodes to `/$defs/A/B`, a three-token pointer, not a one-token `$defs` key. No case in either malformed list covers it — every existing negative (`A/B`, `T~N`, `C%D`, `A B`, `A%`, `%C3%28`, `A#B`, `#/wrong/A`, absolute URI) fails earlier, at the regex or the tilde scan. Deleting the guard keeps the suites green, which makes it unpinned behavior and admits two spellings for one key.

I will add `#/$defs/A%2FB` and `#/components/schemas/A%2FB` to the malformed matrices.

For the record, these are correctly **not** rejected and must keep resolving, since percent-decoding precedes token-decoding: `#/$defs/A%7E1B` and `#/$defs/A%7E0B` are alternate spellings of `A/B` and `A~B`.

### 2. `llm/json.mdx` is the schema-first boundary and was not updated

`llm/schema.mdx` documents the canonical reference form, but `website/src/content/docs/llm/json.mdx` is the page that owns the consumer side this PR changes. It documents `LlmJson.coerce`'s pipeline with "**Resolve `$ref`** — follow schema references" as step 1, and it recommends `LlmJson.validate` exactly where hand-built and third-party references arrive:

| Scenario | Use |
| --- | --- |
| Schema comes from a remote registry / database | `LlmJson.validate` |
| You're building a generic library that accepts any LLM schema | `LlmJson.validate` |

That is a public behavior change for those readers: a malformed or missing reference previously degraded to an accept-all `{}` and returned `success: true`, and now fails validation. The development skill requires the matching page to change with the behavior. I will document the canonical form and the fail-closed contract on that page.

### Not changed, reported to the lead

- **Codec naming.** `LlmReference` is an LLM-named internal that now also serves the general `OpenApiStationValidator` and `OpenApiOneOfValidator`. It is functionally correct and domain-consistent (both are consumers of inverted LLM schemas), but the name under-describes it. I am not renaming it here: #2105 owns that same OpenAPI surface and would absorb pure churn.
- **Native version bump.** Go source under `packages/typia/native` changed, which the development skill normally pairs with a `typia` version bump in the same PR. Deliberately omitted — the lead serializes the native version after the chain rebase.
- **Pre-existing discriminator gating** for generic type aliases, reported in the previous comment, remains untouched and out of scope.

samchon:
## Solo Self-Review round 3: clean

Fresh complete round over the whole `origin/master...HEAD` surface (18 files) at `88b4246618`. Round 2 changed only a test file and a guide page — source and the codec are byte-identical to `d705d5f151`, which I verified rather than assumed — so the round-1 suite results still hold for this head.

I re-derived the surface from scratch rather than rechecking the last fix:

- **Codec.** Encode is `~`→`~0`, `/`→`~1`, then `encodeURIComponent`. Decode is percent-decode, reject an embedded separator, then a single left-to-right `~` scan, which is equivalent to the RFC's `~1`-before-`~0` order. Empty-key boundary round-trips (`""` ↔ `#/$defs/`), and the `""` key is never confused with `undefined` because every caller tests `=== undefined`. `read`/`readOpenApi` reject each other's prefix.
- **Native/TS parity.** Go's allowed set `-_.!~*'()` is exactly `encodeURIComponent`'s unreserved set, and both percent-encode UTF-8 bytes; the Go and TypeScript matrices agree on `A%20B`, `C%25D`, `Caf%C3%A9`, `A~0~1B`, `~01`.
- **Traversal rewrite.** The hand-rolled `validate` matches `OpenApiTypeCheckerBase.visit` node for node (reference, oneOf, object + `additionalProperties`, array, tuple + `prefixItems`/`additionalItems`), shares its accessor formats and its `props.accessor ?? "$input.schema"` / `$input.components.schemas` defaults, and keeps the same visited-dedup placement after the per-node check. It adds discriminator-mapping validation, which is what makes `convertDiscriminatorMapping`'s `undefined` return unreachable for malformed input.
- **Fail-closed sentinels.** `isUnknown({})` is `true`, so the old `escaped: {}` made one unresolvable branch collapse the whole `oneOf` into `remainders` and hijack selection. `{ oneOf: [] }` is not unknown, matches no branch builder, and leaves the station validator to report. `getFlattened`'s `visited` is fresh per call site, so it trips only on pure-ref cycles — ordinary recursion descends through `properties` and is unaffected. `OpenApiStationValidator`'s `references` likewise resets when descending, preserving legitimate recursion.
- **Ownership.** No LLM path still resolves references by string slicing; `LlmTypeChecker` uses the base only for scalar cover helpers. The remaining `split("/").pop()` sites are OpenAPI document upgraders/downgraders and the prefix-parameterized base checker, which are #2103/#2105's surface, not this one.
- **Accessor change.** `visit` now emits `$defs["Foo"]`; that is not a divergence but alignment with `OpenApiTypeCheckerBase`'s existing `[${JSON.stringify(key)}]` convention, and no test or consumer depends on the old unquoted form.

Nothing further survived verification. No new finding; no change made in this round.

### Final verification at `88b4246618`

| Command | Result |
| --- | --- |
| `go -C packages/typia/native test ./core/programmers/llm/...` | pass |
| `go -C packages/typia/native test ./...` | pass (12 packages) |
| `pnpm --filter @typia/utils build` | pass |
| `pnpm --filter test-utils start` | pass |
| `pnpm --filter test-typia-schema start` | pass |
| `pnpm --filter test-langchain start` | pass |
| `pnpm --filter test-mcp start` | pass |
| `pnpm --filter test-vercel start` | pass |

Every regression here was red-proved, not merely observed green: reverting the native mapping encoder fails the discriminator assertion, emitting raw `$ref`s makes LangChain's `@cfworker/json-schema` fail to resolve the **valid** tree, and deleting the `token.includes("/")` guard fails the new extra-segment assertion.

Not run: the root `pnpm test` and the website build, which needs a sibling ttsc checkout plus staged tarballs. The guide edit is prose-only.

Two items remain for the lead, restated from the round-2 comment: the `typia` native version bump is deliberately **not** in this PR, and #2105 must integrate unrestricted `$defs` inversion with OpenAPI-legal component allocation on top of this head.

samchon:
Final semantic rebase onto current `master`, version assignment, and lead verification.

## Rebase and version

Rebased from `f1895bc2a7` onto `40cfb302f3` (#2096). All five commits replayed with **no conflict** — nothing merged in the chain (#2089, #2084, #2115, #2090, #2096) touches this batch's 18 files.

Version assigned as **`13.1.12`**, the next patch on master's `13.1.11` (#2096). The bump is mandatory because the diff changes native Go (`LlmSchemaProgrammer.go`), per the development skill's Change Integrity rule; it was withheld from the branch until this rebase so the serial chain could not strand a stale number. The handoff explicitly flagged this as a lead action item and it is now discharged. `git diff origin/master...HEAD -- packages/typia/package.json` is exactly one line and `git diff --check` is clean.

## Lead Self-Review of the codec: clean

The RFC 6901 implementation in `LlmReference` is correct where it is easiest to get wrong:

- `encode` applies `~` -> `~0` **before** `/` -> `~1`. The order is load-bearing: reversed, the `~` pass would double-escape the `~1` it just produced. Because of it, a literal key `a~1b` and an escaped key `a/b` remain distinguishable across a round trip.
- `decode`'s fragment class admits neither `/` nor `"`. Excluding `/` is what keeps a deeper pointer path from being read as a single `$defs` key; excluding `"` agrees with the RFC 3986 correction being made separately in #2110.
- The `token.includes("/")` guard runs **after** percent-decoding, which is the only place a `%2F` can surface as a raw slash. Since `encode` never emits a raw `/`, a decoded slash proves malformed input rather than a legal key. Round 2 correctly identified that this guard had no covering test and that the obvious test shape would not have reached it.
- The `~` unescape loop rejects an invalid escape (`~2`) and a trailing `~` instead of passing either through, so a malformed pointer cannot silently decode to a plausible key.
- `resolve` reads through `ObjectDictionary.get`, keeping `__proto__`-style keys as data, and `dereference` carries a `visited` set that rejects a cyclic alias chain rather than looping.

## Local verification on `7573f3b6e8` (campaign CI is deliberately cancelled)

| Gate | Command | Result |
| --- | --- | --- |
| Focused native LLM programmer | `go -C packages/typia/native test -count=1 ./core/programmers/llm/...` | ok, 0.23s |
| Interface build | `pnpm --filter @typia/interface build` | pass |
| Utils build | `pnpm --filter @typia/utils build` | pass |
| Affected workspaces | `pnpm --filter test-utils start` | Success |
| | `pnpm --filter test-typia-schema start` | Success |
| | `pnpm --filter test-langchain start` | Success |
| | `pnpm --filter test-mcp start` | Success |
| | `pnpm --filter test-vercel start` | Success |

## What this thread's review actually corrected

Worth recording, because the campaign ledger was wrong about this batch: the round-1 remediation existed only as ~430 lines of **uncommitted working-tree edits** that had never been committed, pushed, or compiled. Trusting the worktree over the ledger prose is what surfaced that.

Two of round 1's own findings then proved to have **wrong premises**, and were resolved by fixing the tests rather than the implementation: the discriminator case used a generic type *alias*, which emits no discriminator at all under pre-existing `json_schema_discriminator` gating untouched by this diff (a generic *interface* does emit it, correctly encoded); and LangChain pre-validates with `@cfworker/json-schema` before typia coerces, so a numeric string can never reach coercion there. The second was repurposed into an independent oracle: with raw `$ref`s, cfworker registers anchors under the canonical spelling and cannot resolve the valid tree — reproducing #2102 outside typia entirely.

The generic-type-alias discriminator gating is recorded as an unvetted candidate for the campaign's next discovery round; it is pre-existing and out of scope here.

`7573f3b6e8` settled after three polls with all runs `cancelled` and none racing to terminal. Record in `.wiki/typia-repository-2026-07/pr-2104-rebase-push-gate-7573f3b6e8.json`. No run for any other SHA was touched.

## Coordination

PR #2105 (#2103, OpenAPI component-name grammar) integrates only on top of this final SHA. Its integration plan is already written against this batch's real final API — `LlmReference.componentKey` is gone, `writeOpenApi`/`readOpenApi` are the identity on a legal key and need no change there.

Issue #1593, opened 2025-06-18
logic or limits of '$ref' in typia.json.schemas


salisbury-espinosa:
```typescript
import { RequireOneOrNone } from 'type-fest';
import typia, { tags } from 'typia';
type ListParamsType<InterfaceType extends Record<string, any>> =
  RequireOneOrNone<
    {
      limit?: tags.Maximum<10000> &
        tags.Minimum<1> &
        tags.Type<'uint32'> &
        number;
      offset?: tags.Type<'uint32'> & number;
      orderBy?: {
        fieldName: keyof InterfaceType;
        direction: 'ASC' | 'DESC';
      };
      orderByList?: Array<{
        fieldName: keyof InterfaceType;
        direction: 'ASC' | 'DESC';
      }> &
        tags.MaxItems<10> &
        tags.MinItems<1>;
    },
    'orderBy' | 'orderByList'
  >;
console.log(
  JSON.stringify(
    typia.json.schemas<
      [
        ListParamsType<{
          one: number;
          two: string;
        }>,
      ],
      '3.1'
    >(),
  ),
);
```

=> 

```json
{
  "version": "3.1",
  "components": {
    "schemas": {
      "_RequireOneOrNone__typeorderByorderByList": {
        "oneOf": [
          {
            "$ref": "#/components/schemas/RequiredPicklimit(Maximum10000Minimum1Typeuint32number)undefinedoffset(Typeuint32number)undefinedorderBy...undefinedorderByList(......1more...MinItems...)undefinedorderByPartial...Omit..."
          },
          {
            "$ref": "#/components/schemas/RequiredPicklimit(Maximum10000Minimum1Typeuint32number)undefinedoffset(Typeuint32number)undefinedorderBy...undefinedorderByList(......1more...MinItems...)undefinedorderByListPartial...Omit..."
          },
          {
            "$ref": "#/components/schemas/PartialRecordorderByorderByListneverOmitlimit(Maximum10000Minimum1Typeuint32number)undefinedoffset(Type...number)undefinedorderBy...undefinedorderByList(......1more...MinItems...)undefinedorderByorderByList"
          }
        ]
      },
      "RequiredPicklimit(Maximum10000Minimum1Typeuint32number)undefinedoffset(Typeuint32number)undefinedorderBy...undefinedorderByList(......1more...MinItems...)undefinedorderByPartial...Omit...": {
        "type": "object",
        "properties": {
          "orderBy": {
            "type": "object",
            "properties": {
[...]

PR #2105, opened 2026-07-16
fix(typia): legalize OpenAPI component names

### LINKED ISSUE #2103: Generated OpenAPI component names can violate the Components key grammar and local-reference syntax

samchon:
## Intent

Reserve the complete implementation surface for #2103 before writing product or test code.

Closes #2103.

## Scope

- allocate deterministic, nonempty OpenAPI 3.0/3.1 component keys that satisfy `^[a-zA-Z0-9\.\-_]+$`;
- preserve legal alphanumeric, dotted, hyphenated, underscored, and qualified controls;
- keep component insertion, recursive `$ref`s, discriminator mappings, collision allocation, and downstream consumers synchronized;
- cover all three public JSON generators plus validator, type checker, converters, HTTP composition/migration, and generated LLM conversion surfaces;
- add tests first for illegal names, Unicode, empty/punctuation boundaries, legal controls, collisions, recursion, and independent reference resolution.

## Coordination

- Unrestricted LLM `$defs` reference parsing and encoding remains owned by #2102. This pull request legalizes generated OpenAPI names and does not claim the manual LLM dictionary contract.
- The implementation may proceed in parallel, but landing follows #2102 where shared converter/reference handling overlaps.
- Native source changes require a `typia` version bump. No version is changed on the claim or stale implementation base; the next version is selected only after rebasing onto the completed native landing chain.

## Verification

Pending tests-first implementation. The claim commit is intentionally empty and changes no product files.

Campaign Actions remain enabled. Runs for each exact campaign SHA are cancelled and locally scoped verification is the implementation gate.

### DIFF
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -1,6 +1,7 @@
 package metadata
 
 import (
+  "hash/fnv"
   "maps"
   "slices"
   "strconv"
@@ -557,15 +558,16 @@ func MetadataCollection_replaceOpenApi(str string) string {
   var escaped strings.Builder
   var quote rune
   quotedEscape := false
+  disambiguate := false
   for _, ch := range str {
     if quote != 0 {
       if quotedEscape {
-        metadataCollection_writeOpenApiNameRune(&escaped, ch)
+        disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
         quotedEscape = false
         continue
       }
       if ch == '\\' {
-        metadataCollection_writeOpenApiNameRune(&escaped, ch)
+        disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
         quotedEscape = true
         continue
       }
@@ -573,14 +575,18 @@ func MetadataCollection_replaceOpenApi(str string) string {
         quote = 0
         continue
       }
-      metadataCollection_writeOpenApiNameRune(&escaped, ch)
+      disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
       continue
     }
     if ch == '\'' || ch == '"' || ch == '`' {
       quote = ch
       continue
     }
-    escaped.WriteRune(ch)
+    if ch == '$' {
+      disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+    } else {
+      escaped.WriteRune(ch)
+    }
   }
   normalized := MetadataCollection_replace(escaped.String())
   if len(normalized) == 0 {
@@ -588,19 +594,31 @@ func MetadataCollection_replaceOpenApi(str string) string {
   }
   var builder strings.Builder
   for _, ch := range normalized {
-    metadataCollection_writeOpenApiNameRune(&builder, ch)
+    disambiguate = metadataCollection_writeOpenApiNameRune(&builder, ch) || disambiguate
+  }
+  if disambiguate {
+    builder.WriteString(".x")
+    builder.WriteString(metadataCollection_openApiNameHash(str))
   }
   return builder.String()
 }
 
-func metadataCollection_writeOpenApiNameRune(builder *strings.Builder, ch rune) {
+func metadataCollection_writeOpenApiNameRune(builder *strings.Builder, ch rune) bool {
   if metadataCollection_isOpenApiNameRune(ch) {
     builder.WriteRune(ch)
-    return
+    return false
   }
   builder.WriteString("_x")
   builder.WriteString(strings.ToUpper(strconv.FormatInt(int64(ch), 16)))
   builder.WriteByte('_')
+  return true
+}
+
+func metadataCollection_openApiNameHash(str string) string {
+  hasher := fnv.New64a()
+  _, _ = hasher.Write([]byte(str))
+  encoded := strings.ToUpper(strconv.FormatUint(hasher.Sum64(), 16))
+  return strings.Repeat("0", 16-len(encoded)) + encoded
 }
 
 func metadataCollection_isOpenApiNameRune(ch rune) bool {
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -603,11 +603,20 @@ func MetadataCollection_replaceOpenApi(str string) string {
     disambiguate = metadataCollection_writeOpenApiNameRune(&builder, ch) || disambiguate
   }
   if disambiguate {
-    builder.WriteString(".x")
+    builder.WriteString(metadataCollection_openApiNameSuffix)
     builder.WriteString(metadataCollection_openApiNameHash(str))
   }
   return builder.String()
 }
+
+// metadataCollection_openApiNameSuffix separates an escaped base name from its
+// disambiguating hash. It must not be ".": JsonDescriptor.cascade reads a dot
+// in a component key as a namespace boundary and inherits the parent
+// component's description, so a dotted suffix would present the escaped base as
+// a fake parent and pull an unrelated type's description into the escaped
+// schema. "-" is in the OpenAPI key alphabet, needs no JSON Pointer or URI
+// escaping, and carries no such meaning.
+const metadataCollection_openApiNameSuffix = "-x"
 
 func metadataCollection_writeOpenApiNameRune(builder *strings.Builder, ch rune) bool {
   if metadataCollection_isOpenApiNameRune(ch) {
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -552,34 +552,39 @@ func MetadataCollection_replace(str string) string {
 // replacement used by LLM `$defs`: OpenAPI restricts keys to an ASCII grammar,
 // while an LLM definition map can own arbitrary JSON object keys.
 func MetadataCollection_replaceOpenApi(str string) string {
-  if len(str) == 0 {
-    return "_"
-  }
   var escaped strings.Builder
   var quote rune
+  quotedContent := false
   quotedEscape := false
   disambiguate := false
   for _, ch := range str {
     if quote != 0 {
       if quotedEscape {
         disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+        quotedContent = true
         quotedEscape = false
         continue
       }
       if ch == '\\' {
         disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+        quotedContent = true
         quotedEscape = true
         continue
       }
       if ch == quote {
+        if quotedContent == false {
+          disambiguate = true
+        }
         quote = 0
         continue
       }
       disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+      quotedContent = true
       continue
     }
     if ch == '\'' || ch == '"' || ch == '`' {
       quote = ch
+      quotedContent = false
       continue
     }
     if ch == '$' {
@@ -590,7 +595,8 @@ func MetadataCollection_replaceOpenApi(str string) string {
   }
   normalized := MetadataCollection_replace(escaped.String())
   if len(normalized) == 0 {
-    return "_"
+    normalized = "_"
+    disambiguate = true
   }
   var builder strings.Builder
   for _, ch := range normalized {
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -3,6 +3,7 @@ package metadata
 import (
   "maps"
   "slices"
+  "strconv"
   "strings"
 
   nativeast "github.com/microsoft/typescript-go/shim/ast"
@@ -543,6 +544,72 @@ func MetadataCollection_replace(str string) string {
     str = strings.ReplaceAll(str, pair.Before, pair.After)
   }
   return str
+}
+
+// MetadataCollection_replaceOpenApi converts a metadata display name into an
+// OpenAPI Components Object key. Keep this separate from the general metadata
+// replacement used by LLM `$defs`: OpenAPI restricts keys to an ASCII grammar,
+// while an LLM definition map can own arbitrary JSON object keys.
+func MetadataCollection_replaceOpenApi(str string) string {
+  if len(str) == 0 {
+    return "_"
+  }
+  var escaped strings.Builder
+  var quote rune
+  quotedEscape := false
+  for _, ch := range str {
+    if quote != 0 {
+      if quotedEscape {
+        metadataCollection_writeOpenApiNameRune(&escaped, ch)
+        quotedEscape = false
+        continue
+      }
+      if ch == '\\' {
+        metadataCollection_writeOpenApiNameRune(&escaped, ch)
+        quotedEscape = true
+        continue
+      }
+      if ch == quote {
+        quote = 0
+        continue
+      }
+      metadataCollection_writeOpenApiNameRune(&escaped, ch)
+      continue
+    }
+    if ch == '\'' || ch == '"' || ch == '`' {
+      quote = ch
+      continue
+    }
+    escaped.WriteRune(ch)
+  }
+  normalized := MetadataCollection_replace(escaped.String())
+  if len(normalized) == 0 {
+    return "_"
+  }
+  var builder strings.Builder
+  for _, ch := range normalized {
+    metadataCollection_writeOpenApiNameRune(&builder, ch)
+  }
+  return builder.String()
+}
+
+func metadataCollection_writeOpenApiNameRune(builder *strings.Builder, ch rune) {
+  if metadataCollection_isOpenApiNameRune(ch) {
+    builder.WriteRune(ch)
+    return
+  }
+  builder.WriteString("_x")
+  builder.WriteString(strings.ToUpper(strconv.FormatInt(int64(ch), 16)))
+  builder.WriteByte('_')
+}
+
+func metadataCollection_isOpenApiNameRune(ch rune) bool {
+  return (ch >= 'a' && ch <= 'z') ||
+    (ch >= 'A' && ch <= 'Z') ||
+    (ch >= '0' && ch <= '9') ||
+    ch == '.' ||
+    ch == '-' ||
+    ch == '_'
 }
 
 func MetadataCollection_escape(str string) string {
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_cascade_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_cascade_test.go
@@ -1,0 +1,55 @@
+package metadata
+
+import (
+  "strings"
+  "testing"
+)
+
+// TestMetadataCollectionOpenApiNameCascade verifies the allocator never invents
+// a namespace parent for an escaped component name.
+//
+// JsonDescriptor.cascade reads every dotted prefix of a component key as a
+// parent namespace and inherits that parent's description. A disambiguating
+// suffix joined with "." would therefore present an escaped name's own base as
+// a fake parent, and an unrelated component owning that base would leak its
+// description into the escaped type's schema.
+//
+//  1. Allocate a legal name and the escaped name that shares its base.
+//  2. Collect every dotted prefix each allocation exposes.
+//  3. Require no allocation to expose another allocation as its parent.
+func TestMetadataCollectionOpenApiNameCascade(t *testing.T) {
+  inputs := []string{
+    "A_x2F_B",
+    "A/B",
+    "Recursive",
+    `Recursive<"">`,
+    `Recursive<"A/B">`,
+    "Qualified.Member",
+    "Café",
+    "$Plain",
+    "",
+  }
+  allocated := map[string]string{}
+  for _, input := range inputs {
+    output := MetadataCollection_replaceOpenApi(input)
+    if previous, ok := allocated[output]; ok {
+      t.Fatalf("allocation collided: %q and %q both produced %q", previous, input, output)
+    }
+    allocated[output] = input
+  }
+  for output, input := range allocated {
+    segments := strings.Split(output, ".")
+    for i := 1; i < len(segments); i++ {
+      parent := strings.Join(segments[:i], ".")
+      if owner, ok := allocated[parent]; ok {
+        t.Fatalf(
+          "allocation %q (from %q) claims unrelated allocation %q (from %q) as a namespace parent",
+          output,
+          input,
+          parent,
+          owner,
+        )
+      }
+    }
+  }
+}
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
@@ -21,15 +21,21 @@ func TestMetadataCollectionOpenApiName(t *testing.T) {
     "Qualified.Member": "Qualified.Member",
     "A-B":              "A-B",
     "A_B":              "A_B",
+    "A_x2F_B":          "A_x2F_B",
   }
+  outputs := map[string]string{}
   for input, expected := range legal {
-    if actual := MetadataCollection_replaceOpenApi(input); actual != expected {
+    actual := MetadataCollection_replaceOpenApi(input)
+    if actual != expected {
       t.Fatalf("legal OpenAPI component name changed: input=%q actual=%q expected=%q", input, actual, expected)
     }
+    outputs[actual] = input
   }
 
   inputs := []string{
     "",
+    "$Plain",
+    "A/B",
     `Recursive<"A/B">`,
     `Recursive<"T~N">`,
     `Recursive<"C%D">`,
@@ -42,7 +48,6 @@ func TestMetadataCollectionOpenApiName(t *testing.T) {
     "한글",
   }
   grammar := regexp.MustCompile(`^[a-zA-Z0-9.\-_]+$`)
-  outputs := map[string]string{}
   for _, input := range inputs {
     first := MetadataCollection_replaceOpenApi(input)
     second := MetadataCollection_replaceOpenApi(input)
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
@@ -1,8 +1,8 @@
 package metadata
 
 import (
-	"regexp"
-	"testing"
+  "regexp"
+  "testing"
 )
 
 // TestMetadataCollectionOpenApiName verifies the OpenAPI-specific metadata
@@ -16,45 +16,45 @@ import (
 //  2. Encode empty, punctuation, percent, fragment, and Unicode boundaries.
 //  3. Prove formerly colliding inputs remain legal, distinct, and repeatable.
 func TestMetadataCollectionOpenApiName(t *testing.T) {
-	legal := map[string]string{
-		"Plain":            "Plain",
-		"Qualified.Member": "Qualified.Member",
-		"A-B":              "A-B",
-		"A_B":              "A_B",
-	}
-	for input, expected := range legal {
-		if actual := MetadataCollection_replaceOpenApi(input); actual != expected {
-			t.Fatalf("legal OpenAPI component name changed: input=%q actual=%q expected=%q", input, actual, expected)
-		}
-	}
+  legal := map[string]string{
+    "Plain":            "Plain",
+    "Qualified.Member": "Qualified.Member",
+    "A-B":              "A-B",
+    "A_B":              "A_B",
+  }
+  for input, expected := range legal {
+    if actual := MetadataCollection_replaceOpenApi(input); actual != expected {
+      t.Fatalf("legal OpenAPI component name changed: input=%q actual=%q expected=%q", input, actual, expected)
+    }
+  }
 
-	inputs := []string{
-		"",
-		"A/B",
-		"T~N",
-		"C%D",
-		"A#B",
-		"A B",
-		"A:B",
-		"AB",
-		"///",
-		"Café",
-		"한글",
-	}
-	grammar := regexp.MustCompile(`^[a-zA-Z0-9.\-_]+$`)
-	outputs := map[string]string{}
-	for _, input := range inputs {
-		first := MetadataCollection_replaceOpenApi(input)
-		second := MetadataCollection_replaceOpenApi(input)
-		if first != second {
-			t.Fatalf("OpenAPI component allocation is not repeatable: input=%q first=%q second=%q", input, first, second)
-		}
-		if grammar.MatchString(first) == false {
-			t.Fatalf("illegal OpenAPI component name: input=%q output=%q", input, first)
-		}
-		if previous, ok := outputs[first]; ok && previous != input {
-			t.Fatalf("OpenAPI component allocation collided: first=%q second=%q output=%q", previous, input, first)
-		}
-		outputs[first] = input
-	}
+  inputs := []string{
+    "",
+    `Recursive<"A/B">`,
+    `Recursive<"T~N">`,
+    `Recursive<"C%D">`,
+    `Recursive<"A#B">`,
+    `Recursive<"A B">`,
+    `Recursive<"A:B">`,
+    `Recursive<"AB">`,
+    `Recursive<"///">`,
+    "Café",
+    "한글",
+  }
+  grammar := regexp.MustCompile(`^[a-zA-Z0-9.\-_]+$`)
+  outputs := map[string]string{}
+  for _, input := range inputs {
+    first := MetadataCollection_replaceOpenApi(input)
+    second := MetadataCollection_replaceOpenApi(input)
+    if first != second {
+      t.Fatalf("OpenAPI component allocation is not repeatable: input=%q first=%q second=%q", input, first, second)
+    }
+    if grammar.MatchString(first) == false {
+      t.Fatalf("illegal OpenAPI component name: input=%q output=%q", input, first)
+    }
+    if previous, ok := outputs[first]; ok && previous != input {
+      t.Fatalf("OpenAPI component allocation collided: first=%q second=%q output=%q", previous, input, first)
+    }
+    outputs[first] = input
+  }
 }
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
@@ -22,6 +22,8 @@ func TestMetadataCollectionOpenApiName(t *testing.T) {
     "A-B":              "A-B",
     "A_B":              "A_B",
     "A_x2F_B":          "A_x2F_B",
+    "_":                "_",
+    "Recursive":        "Recursive",
   }
   outputs := map[string]string{}
   for input, expected := range legal {
@@ -36,6 +38,7 @@ func TestMetadataCollectionOpenApiName(t *testing.T) {
     "",
     "$Plain",
     "A/B",
+    `Recursive<"">`,
     `Recursive<"A/B">`,
     `Recursive<"T~N">`,
     `Recursive<"C%D">`,
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_openapi_name_test.go
@@ -1,0 +1,60 @@
+package metadata
+
+import (
+	"regexp"
+	"testing"
+)
+
+// TestMetadataCollectionOpenApiName verifies the OpenAPI-specific metadata
+// name allocator preserves legal controls while encoding every other rune.
+//
+// The schema generators use the allocated name as both a Components Object
+// key and a local-reference token. Deleting punctuation is insufficient:
+// distinct names can collapse, and the surviving key can still violate OAS.
+//
+//  1. Preserve ordinary, dotted, hyphenated, and underscored legal controls.
+//  2. Encode empty, punctuation, percent, fragment, and Unicode boundaries.
+//  3. Prove formerly colliding inputs remain legal, distinct, and repeatable.
+func TestMetadataCollectionOpenApiName(t *testing.T) {
+	legal := map[string]string{
+		"Plain":            "Plain",
+		"Qualified.Member": "Qualified.Member",
+		"A-B":              "A-B",
+		"A_B":              "A_B",
+	}
+	for input, expected := range legal {
+		if actual := MetadataCollection_replaceOpenApi(input); actual != expected {
+			t.Fatalf("legal OpenAPI component name changed: input=%q actual=%q expected=%q", input, actual, expected)
+		}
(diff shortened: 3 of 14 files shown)


samchon:
Implementation increment `6dc41c1ecaf7e964404e958148068ca2642d1ee8` is now on the branch.

- added a dedicated OpenAPI component-key allocator while leaving unrestricted LLM `$defs` naming to #2102;
- routed `json.schema`, `json.schemas`, and `json.application` through the same allocator for both supported versions;
- made multi-root `json.schemas` analysis share collision ownership;
- covered legal controls, forbidden ASCII, Unicode identifiers/literals, empty/punctuation boundaries, reordered discovery, recursive object/array/alias/native producers, and discriminator mappings;
- verified the generated artifacts through an independent RFC 3986/RFC 6901 oracle, `OpenApiValidator`, `OpenApiTypeChecker`, `LlmSchemaConverter`, `OpenApiConverter`, `HttpMigration`, and `HttpLlm`.

Focused evidence so far:

- `go -C packages/typia/test test ../native/core/schemas/metadata -run TestMetadataCollectionOpenApiName -count=1`
- `go -C packages/typia/test test ../native/transform/features/json -count=1`
- focused `test-typia-schema` execution for `json_schema_openapi_component_names`

The exact-SHA campaign gate cancelled six Actions runs for this increment. External OpenAPI validation, broader suites, and solo Self-Review remain. The PR stays draft and intentionally has no native version bump until #2102 and the preceding native landing chain complete, after which it needs a fresh rebase and version selection.

samchon:
Self-Review round 1 found that the allocator still has deterministic collision gaps beyond the covered old normalization cases.

- A legal raw name such as `A_x2F_B` currently collides with the escaped output for `A/B`.
- A legal TypeScript identifier such as `$Foo` is passed through the legacy punctuation-deletion stage and collides with `Foo`.

I will first add regression cases that prove the allocated base names are distinct and repeatable, then change the OpenAPI-only encoding so forbidden characters are encoded before any lossy legacy normalization. After that I will rerun the whole PR surface as a fresh Self-Review round.

samchon:
Self-Review round 2 found one remaining allocator collision at the empty boundary.

- An empty raw name currently returns `_`, which aliases the already legal `_` control.
- A generic empty literal such as `Recursive<"">` loses its empty quoted token and aliases the legal `Recursive` candidate.

I will seed both legal controls into the allocator test, prove the two collisions first, then disambiguate empty raw input and empty quoted tokens without changing nonempty legal quoted controls.

Broad evidence before this follow-up: native Go tree, typia/utils package builds, the complete schema workspace, and external Redocly validation are green. Root `pnpm test` reached this PR's schema and downstream suites successfully but later hit an unrelated `test-typia-generate` child-process ETIMEDOUT under shared CPU contention.

samchon:
Self-Review round 3 found one sound defect in the committed allocator, plus a corrected diagnosis of the earlier "timeout".

## Finding: the `.x<hash>` disambiguator collides with the namespace-cascade separator

`JsonDescriptor.cascade` (`packages/utils/src/utils/internal/JsonDescriptor.ts:17`) is the one consumer that gives `.` a *meaning* inside a component key: it splits the key on `.` and treats every dotted prefix as a parent namespace, inheriting that parent's description. That is the intended `Qualified.Member` -> `Qualified` feature.

The allocator picks `.x<FNV>` as its disambiguator suffix, so every escaped name now presents its own escaped base as a fake parent namespace. When a real component owns that base, an unrelated type's description is injected into the escaped type's schema.

Reproduction on this branch (`71087011b8`):

```ts
/** UNRELATED INTERFACE that happens to own the escaped base name. */
interface RecursiveObjectA_x2F_B { unrelated: string; }

/** GENERIC RECURSIVE OBJECT whose "A/B" instantiation must be escaped. */
type RecursiveObject<T extends string> = { value: T; children: RecursiveObject<T>[] };

typia.json.schema<{ collision: RecursiveObjectA_x2F_B; slash: RecursiveObject<"A/B"> }, "3.1">();
```

allocates `RecursiveObjectA_x2F_B` and `RecursiveObjectA_x2F_B.xC8FBD4DA176083CC`, and `OpenApiTypeChecker.escape` on the second one emits:

```
Description of the current {@link RecursiveObjectA_x2F_B.xC8FBD4DA176083CC} type:
> GENERIC RECURSIVE OBJECT whose "A/B" instantiation must be escaped.
------------------------------
Description of the parent {@link RecursiveObjectA_x2F_B} type:
> UNRELATED INTERFACE that happens to own the escaped base name.
```

The two types are unrelated. `OpenApiTypeChecker.escape` applies `cascade` to every dereferenced `$ref` (`OpenApiTypeCheckerBase.ts:344,362`), so this reaches every escaped-name consumer, and the polluted text is exactly what an LLM reads out of a function-calling schema. This PR's own fixture pairs `RecursiveObject<"A_x2F_B">` with `RecursiveObject<"A/B">`, so it already builds the colliding shape and the current assertions do not look at descriptions.

Scope note: the pre-existing `.o<N>` collision suffix in `MetadataCollection.getName` has the same shape and is shared with the LLM and reflect paths. That is #1593 territory and I am not touching it here. But I am introducing a *new* suffix, so it must not add new instances of the defect.

## Remediation

- Change the disambiguator separator from `.x` to `-x`. `-` is in the OpenAPI key alphabet `^[a-zA-Z0-9.\-_]+$`, needs no JSON Pointer or URI escaping, and carries no cascade meaning. I verified `JsonDescriptor.ts:17` is the only place that splits a component key on `.`; `EndpointUtil.normalize` folds `.` and `-` identically and only ever sees URL path segments.
- Add a regression case that pins an escaped name against a real component owning its escaped base, asserting the escaped type's description does not inherit the unrelated one.
- Rerun the focused Go allocator test, the native tree, and the schema workspace, then start a fresh full Self-Review round.

## Correction: the earlier `test-typia-generate` failure was not CPU contention

The round-2 note recorded an "unrelated child-process ETIMEDOUT under shared CPU contention". That diagnosis was wrong. `test-typia-generate` fails here with:

```
compile: writing output: write $WORK\b236\_pkg_.a: There is not enough space on the disk.
```

`C:` is at 100% (3.3 GB free) and Go builds `$WORK` under `C:\Users\samch\AppData\Local\Temp`. It is an environment fault, not a code fault and not contention; re-running with `GOTMPDIR` on `D:` gets past the failure point. Recording it so the ledger stops attributing it to load.

## Verification on `71087011b8` before this round's edits

- `go -C packages/typia/test test ../native/core/schemas/metadata -run TestMetadataCollectionOpenApiName -count=1` -> ok
- `go -C packages/typia/native test ./...` -> ok (whole tree)
- `pnpm --filter test-typia-schema start` -> Success, `test_json_schema_openapi_component_names` included
- generated-document evidence for OAS 3.0 and 3.1: 21 component keys, 0 illegal; 20 `$ref`s and 2 discriminator mappings all resolve under an independent RFC 3986 + RFC 6901 oracle; `OpenApiValidator`, `OpenApiTypeChecker.escape`, `LlmSchemaConverter.schema` (legal `$defs` keys), `HttpMigration` (1 route, 0 errors), and `HttpLlm` (1 function, 0 errors) all green.

samchon:
Self-Review round 4 (fresh full round over the whole `origin/master...HEAD` diff, 8 files) found nothing further to fix. Round 3's remediation is on the branch at `4e10e534f9`.

## Round 3 remediation landed

- `1921a6fbc2` — tests first. `TestMetadataCollectionOpenApiNameCascade` and `test_json_schema_openapi_component_name_cascade` both fail on `71087011b8`:
  ```
  allocation "A_x2F_B.xFA76C319A0B010E1" (from "A/B") claims unrelated
  allocation "A_x2F_B" (from "A_x2F_B") as a namespace parent
  ```
  The Go fixture also catches the more plausible instance the reproduction missed: `Recursive<"">` allocates `Recursive.x<hash>`, whose fake parent `Recursive` is a real allocated component.
- `4e10e534f9` — the fix. The disambiguating suffix now joins with `-x` instead of `.x`.

Both tests pass after the fix; both are oracle-derived (grammar, RFC resolution, absence of an invented parent) rather than snapshots of the hash, and I confirmed by grep that **no committed baseline anywhere in `tests/` pins the suffix form**, so the separator change cannot silently relax an existing expectation.

## Verification on `4e10e534f9`

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test ../native/core/schemas/metadata -count=1` | ok (both allocator tests) |
| `go -C packages/typia/native test ./...` | ok, whole tree |
| `pnpm --filter test-typia-schema start` | Success — `..._component_names` and `..._component_name_cascade` both run |
| `pnpm --filter test-typia-generate start` | **exit 0** — see below |
| generated-document oracle, OAS 3.0 and 3.1 | 21 keys / 0 illegal / **0 invented parents**; 20 `$ref`s + 2 discriminator mappings all resolve; `OpenApiValidator`, `OpenApiTypeChecker.escape`, `LlmSchemaConverter` (legal `$defs`), `HttpMigration` (0 errors), `HttpLlm` (0 errors) all green |

`test-typia-automated` and `test-utils` are still running and are reported as inconclusive, not as a pass. Neither exercises the changed Go path (the change is native-only; template structure names are all legal ASCII), and the essential consumers above are green.

## `test-typia-generate` now passes — and the round-2 diagnosis was wrong

Round 2 recorded an "unrelated child-process ETIMEDOUT under shared CPU contention". That was not the cause. The suite fails with:

```
compile: writing output: write $WORK\b236\_pkg_.a: There is not enough space on the disk.
```

`C:` was at 100% (3.3 GB free, then 527 MB) and Go builds `$WORK` under `C:\Users\samch\AppData\Local\Temp`. After reclaiming 26 GB of stale `go-build*` temp directories left by killed builds, `pnpm --filter test-typia-generate start` completes **exit 0**. It was an environment fault throughout — not contention, and not this branch. The previously inconclusive suite is now green.

## Out of scope, verified, recorded for triage

The pre-existing `.o<N>` collision suffix (`MetadataCollection.go:312`) has the identical cascade defect — `Foo.o1` inherits `Foo`'s description, which is the issue body's own `RecursiveAB` / `RecursiveAB.o1` pair. I did not touch it: `.o<N>` is shared with the LLM `$defs` and `reflect.name` paths, so changing it is a public-contract change well outside #2103. Same for the order-dependent tie-break on unquoted punctuation (`Recursive<Foo>` vs `RecursiveFoo`) — still legal, still no overwrite, so #2103's invariant holds and the aliasing belongs to #1593.

Also worth a release note: this PR necessarily renames previously-legal-but-aliasing keys (`$Named` was `Named`, now `_x24_Named-x<hash>`; `Recursive<"A B">` was `RecursiveAB`). That rename is the point — it is what removes the aliasing the issue cites — but it is user-visible in generated documents.

## Still blocked

- `LlmSchemaConverter.invert` integration waits on #2104's final SHA `88b4246618`. Exact plan recorded at `.wiki/typia-repository-2026-07/pr-2105-2104-integration-plan.md`; the integration test is written and held uncommitted at `tests/test-utils/src/features/llm/invert/test_llm_invert_openapi_component_names.ts` (it needs #2104's reference encoding, so committing it now would leave the branch red).
- No version bump on this branch. Native Go changed, so one is mandatory — the lead assigns it at the serial-chain rebase; the stale `13.1.7` is left untouched deliberately.

samchon:
Integration status: **blocked on a rebase**, not on analysis. #2104 merged to master (`98f560253e`), but this branch is based on `f1895bc2a7` and does not contain it, so the integration cannot be written or proven here.

## Why the integration is not implementable on `4e10e534f9`

`packages/utils/src/utils/internal/LlmReference.ts` is **absent** from this branch, and `invert` still has its pre-#2104 shape:

```ts
// this branch, LlmSchemaConverter.ts:514
const key: string =
  schema.$ref.split("#/$defs/")[1] ?? schema.$ref.split("/").at(-1)!;
```

The three sites the plan targets — `const componentKey: string = resolved.key` (`:570`), `LlmReference.writeOpenApi(componentKey)` (`:585`), and `LlmReference.writeOpenApi(resolved.key)` (`:682`) — exist only on master. The delta this branch is missing is **633 insertions across 19 files** in `packages/utils` and `packages/interface`, including all 94 lines of `LlmReference.ts` and a 192-line reshaping of `LlmSchemaConverter.ts`. Reproducing that by hand to code against it would be merging #2104 manually, and would conflict at the real rebase.

## The red-then-green evidence is unobtainable here — the red is the wrong red

The held test does go red on this base, and the first assertion that trips is even the plausible-looking *"every component name satisfies the OpenAPI grammar"*. That is misleading. The cause is #2104's defect, not #2103's:

```
ref                 "#/$defs/~1"
key extracted       "~1"          <- no pointer decoding on this base
$defs lookup        undefined     <- real key is "/"
component stored    "~1" = {}     <- definition CONTENT is lost
```

The illegal key `~1` here comes from *failing to decode*, whereas on master the illegal key would be the raw `/` that #2104 deliberately stores. More decisively: **no allocator can turn this green.** Because the `$defs` lookup misses, the component is written as `{}`, so the test's `resolves to its own definition` assertion stays red regardless of how the key is allocated. Making it green requires #2104's `LlmReference.resolve`. Observing and "fixing" the red here would be fitting code to the wrong cause.

## Verification re-run on `4e10e534f9`

- `pnpm --filter test-utils start` -> **Success** with the held test set aside; the held test is the sole failure and fails only for the reason above. The previously inconclusive record is now green.
- `test-typia-automated` re-run in progress.
- Disk: `go env GOTMPDIR=D:\gotmp` is in effect, so Go scratch is off `C:` (which is back down to 3.4 GB free). No stale `go-build*` dirs remained to reclaim.

## Ask

Per the standing offer: please rebase this branch onto `98f560253e` and assign the version. Everything after that is ready to execute immediately — the plan's API assumptions were re-verified against merged master and hold byte-for-byte, and the held test's encoder is identical to #2104's `encode`.

samchon:
Integration finding, posted before the edit: **#2104's `inversion keeps raw component identity` assertion is a premise that #2103 exists to overturn.** I am correcting the test, not the implementation, and want the reasoning on the record first.

## The red appeared for the right reason

Before implementing, I verified the red on the rebased base was #2103's defect and not #2104's. The assertion message is identical to the one I saw on the old base, so I instrumented `invert` rather than trust it:

| | old base (wrong red) | rebased base (right red) |
| --- | --- | --- |
| key for `$defs["/"]` | `"~1"` — decode failure | `"/"` — raw key stored |
| definition content | **lost** (`{}`) | **preserved** (0 lost) |

On the rebased base: 7 illegal keys (`""`, `/`, `~`, `%`, `#`, ` `, `Café`) written verbatim, **0 components with lost content**, and the `$defs` `Taken` definition silently swallowed by the caller's pre-existing `Taken`. That is exactly #2103 and exactly the seeding data-loss the plan predicted. Implementation followed.

## The conflict

`tests/test-utils/src/features/llm/schema/test_llm_schema_json_pointer_references.ts:105` asserts `Object.hasOwn(components.schemas!, key)` — that the stored component key **is** the raw `$defs` key. That is mutually exclusive with #2103 for every forbidden key. Across #2104's own matrix, 9 of 11 cases conflict:

```
  "Plain"      -> "Plain"            | still holds
  ""           -> "_"                | BROKEN (renamed)
  "A/B"        -> "A_x2F_B"          | BROKEN (renamed)
  "A~1B"       -> "A_x7E_1B"         | BROKEN (renamed)
  "T~N"        -> "T_x7E_N"          | BROKEN (renamed)
  "~1"         -> "_x7E_1"           | BROKEN (renamed)
  "A B"        -> "A_x20_B"          | BROKEN (renamed)
  "C%D"        -> "C_x25_D"          | BROKEN (renamed)
  "Café"       -> "Caf_xE9_"         | BROKEN (renamed)
  "A~/B"       -> "A_x7E__x2F_B"     | BROKEN (renamed)
  "__proto__"  -> "__proto__"        | still holds
```

## Why the test yields, not the implementation

- #2103's acceptance is explicit: *"Assert that every emitted `components.schemas` key matches `^[a-zA-Z0-9.\-_]+$` and is nonempty."* An empty raw key cannot be both raw-identical and nonempty — the two contracts cannot both hold.
- The lead vetting note assigns this line to #2105 by name: *"`LlmSchemaConverter.invert` must allocate one stable legal component key for every resolved raw `$defs` key and reuse that mapping for component storage, `$ref`, recursion, and discriminator mappings."*
- `#/components/schemas/` is a **restricted** key space. #2104 owns the unrestricted `$defs` space and its lossless codec; storing a raw key on the OpenAPI side was never #2104's contract to keep.

## What I am changing, and what I am preserving

The assertion's *intent* — a definition survives inversion, is reachable, and is not mangled or lost — is still correct and my change keeps it. Only its *mechanism* (raw key == component key) is superseded. I will rewrite it to assert the definition survives under its **allocated legal** key and that the emitted `$ref` resolves to that same component by JSON Pointer, which is strictly stronger: it keeps #2104's resolution guarantee **and** adds the grammar guarantee. I am not deleting a case, weakening an oracle, or relaxing the `is not accept-all` checks.

Flagging under Change Integrity: this modifies a test committed by another batch. Evidence above; calling it out in the report.

samchon:
Self-Review round 5 (fresh full round over `origin/master...HEAD`, 12 files) found one sound defect **in my own integration commit** `1688b172ed`, plus one considered trade-off worth recording.

## Finding: allocation seeding does not match reference resolution

`LlmReference.resolve` finds a definition through `ObjectDictionary.get`, which keys off `hasOwnProperty` (own, **enumerable or not**). I seeded the allocation with `Object.keys`, which is own **and enumerable only**. A `$defs` key that is own-but-not-enumerable therefore resolves but is never allocated, and my `?? resolved.key` fallback then emits the raw key — reintroducing exactly the defect this issue exists to close.

Reproduced on `1688b172ed`:

```
Object.keys($defs)               : []
Object.getOwnPropertyNames($defs): ["A/B"]
hasOwnProperty("A/B")            : true

component keys written: ["A/B"]
illegal keys          : ["A/B"]
```

It is a narrow trigger — `ObjectDictionary.set`, object literals, and `JSON.parse` all produce enumerable keys, so only a deliberate `Object.defineProperty` reaches it — but the shape is wrong regardless: a fallback that silently emits a raw key sits in the one function whose entire purpose is to never emit one.

### Remediation

- Seed both `keys` and `reserved` with `Object.getOwnPropertyNames`, which matches `hasOwnProperty` exactly. The allocation then provably cannot miss a resolvable key.
- Change the unreachable fallback from `?? resolved.key` to an escaped legal name, so the grammar invariant holds even if the two ever diverge again.
- Add a regression case pinning the seeding-to-resolution symmetry, so reverting `getOwnPropertyNames` to `keys` fails loudly.

## Recorded trade-off: seeding vs. repeated inversion into shared components

Seeding makes a caller's pre-existing component an allocation *input*, so calling `invert` twice with the **same** `components` object and the same `$defs` now appends a second, differently named copy instead of reusing the first. That is inherent to the seeding rule the vetting note mandates (`seed pre-existing component keys before allocating forbidden names`) — from inside one call, a caller's foreign component and my own prior allocation are indistinguishable.

I verified this regresses nothing: `LlmJson.validate` (the only product caller of `invert`) builds a fresh `{ schemas: {} }` per call, and both `invert` calls in `test_llm_schema_parity_invert.ts` use separate `components` objects. Each call's returned schema stays internally consistent; only the shared map accumulates a duplicate. Recording it rather than adding speculative merge semantics for a pattern no caller uses.

## Verification of `1688b172ed` before this round's remediation

| Command | Result |
| --- | --- |
| `pnpm --filter test-utils start` | Success — held test green, corrected #2104 matrix green |
| `pnpm --filter @typia/utils build` | ok |
| `pnpm --filter test-mcp / test-vercel / test-langchain start` | PASS, PASS, PASS |

samchon:
Self-Review round 6 (fresh full round over `origin/master...HEAD`, 13 files) is **clean** — no sound improvement found. Round 5's remediation is on the branch at `adca55e33f`.

## Round 5 remediation landed, red-then-green proven

My first attempt at the regression test was **worthless and I caught it**: it asserted only "legal key + resolves", which my own escape fallback already satisfied, so it passed against the broken `Object.keys` seeding. A test that cannot fail against the defect it names is not coverage.

The rewritten case pins the collision that *only* correct seeding prevents — an own non-enumerable `A/B` whose escaped form is exactly a second, legal, enumerable `A_x2F_B`:

```
### RED (seeding reverted to Object.keys) ###
  - test_llm_invert_non_enumerable_definition: Error
  Error: Bug on the legal definition keeps its exact name: found different values - []

### GREEN (getOwnPropertyNames) ###
  - test_llm_invert_non_enumerable_definition: 2 ms
  - test_llm_invert_openapi_component_names:   5 ms
  - test_llm_schema_json_pointer_references:  11 ms
  Success
```

The escape fallback alone cannot fix this: it *produces* the very name the legal definition holds. Only seeding from the same own-property view that `resolve` uses closes it.

## Round 6 verification on `adca55e33f`

| Command | Result |
| --- | --- |
| `go -C packages/typia/native test ./...` | **ok**, whole tree — first run since the rebase pulled #2104's native changes |
| `pnpm --filter test-utils start` | Success |
| `pnpm --filter @typia/utils build` | ok |
| `test-mcp` / `test-vercel` / `test-langchain` | PASS / PASS / PASS |
| `test-typia-generate`, `test-typia-automated` | in flight |

Public contract confirmed intact: the emitted `invert` signature in `LlmSchemaConverter.d.ts` is byte-identical, and neither `invertInternal` nor `OpenApiComponentName` leaks to the public surface.

Re-verified after the rebase that the `-x` separator rationale still holds: #2104 added an optional `key` parameter to `JsonDescriptor.cascade` but left the `.split(".")` namespace-cascade logic unchanged, and both cascade regressions still pass.

## Considered and deliberately not done

- **A TS-side cascade test mirroring the native one.** The TS allocator cannot invent a dot by construction — the suffix is `-x<counter>` and `escape` only passes through dots the caller wrote — so such a test could only fail if someone edited the `SUFFIX` constant, whose doc comment carries the full rationale. The native test already pins the shared design rule. Adding a second one is a guard against editing a documented constant, not against a reachable defect.
- **Merge semantics for repeated `invert` into shared components.** Recorded in round 5; no caller does it, and inventing semantics for it would be speculative.

## Status

Native half and LLM inversion half are both complete and green. `packages/typia/package.json` is untouched at `13.1.12` — the version is yours to assign at the final rebase, per your note that #2113 takes `13.1.13`.

samchon:
_(Superseded by the corrected repost below — shell quoting emptied three code lines here.)_

samchon:
`test-typia-generate` re-run on `adca55e33f` exited 1. **Diagnosed: unrelated pre-existing Windows harness flakiness, not this branch and not disk.**

_(Reposted: shell quoting emptied three code lines in my previous comment — the substance is unchanged.)_

## Not disk this time

Checked that first, per the standing rule. `C:` had **6.52 GB** free and the log contains **zero** disk or memory errors (`not enough space` / `ENOSPC` / `out of memory` / `ETIMEDOUT` / `heap` all absent). The `GOCACHE`/`GOTMPDIR` relocation is holding — free space is now *rising* (3.4 -> 6.6 GB) across runs.

## The actual failure

```
> generate:files:implicit-jsconfig
> pnpm run clean && ttsx --cwd jsconfig-dir/src ... generate --output ../../src/output generate_jsconfig.ts
  ----------------------------------------
   Typia Generate Wizard          <- banner printed, CLI exited 0
  ----------------------------------------
> check:files:implicit-jsconfig
> node -e "require('fs').accessSync('src/output/generate_jsconfig.ts')"
  Error: ENOENT: no such file or directory
```

The CLI exited 0 — the `&&` chain reached `check`, so it believed it wrote the file — but the file was not there.

## Why it cannot be this branch

The jsconfig fixture is a plain validator with no component names, no JSON schema, and no LLM inversion anywhere in its path:

```ts
interface IJsConfigTemplate { id: string; value: number; }
export const isJsConfig = typia.createIs<IJsConfigTemplate>();
```

Nothing in this diff — native OpenAPI component allocation, the three JSON transformers, or `LlmSchemaConverter.invert` — is reachable from `typia.createIs`. Corroborating:

- Running that exact step standalone on the **same SHA** succeeds: `generate_jsconfig.ts` is written (372 bytes), exit 0, and the emitted validator is correct.
- The same suite passed **exit 0** earlier on `4e10e534f9` with the identical harness.

So it is nondeterministic, not a regression.

## Probable mechanism, for whoever owns the harness

```
clean                            :: node -e "require('fs').rmSync('src/output', { recursive: true, force: true })"
generate:files:implicit-jsconfig :: pnpm run clean && ttsx ... generate --output ../../src/output ...
check:files:implicit-jsconfig    :: node -e "require('fs').accessSync('src/output/generate_jsconfig.ts')"
```

Every mode does `rmSync` on `src/output` and then immediately writes back into it. On Windows a directory still holding an open handle (antivirus, an indexer, a concurrent agent) is only *marked* for deletion rather than removed; a file created inside it afterwards can succeed and then disappear with the pending delete. That matches the evidence exactly — the writer reports success, the reader finds nothing — and it only bites under the concurrent load this machine currently has.

This is a pre-existing harness robustness issue on Windows, outside #2103's scope. Recording it as a candidate rather than patching another batch's harness inside this PR.

## Status

Re-running the suite alone once `test-typia-automated` finishes, to confirm it goes green. Everything else on `adca55e33f` is green and the branch is unchanged.

samchon:
**Correcting my previous comment: my Windows-race diagnosis was wrong, and I over-committed to it.**

I attributed the `test-typia-generate` ENOENT to a Windows doomed-directory race between `rmSync` and the immediately following write. That was speculation dressed up as a mechanism. The reflog gives a far better-supported cause:

```
43dc90fd22 HEAD@{2026-07-17 05:29:40}: commit: chore(typia): bump version for OpenAPI component-name allocation
b7fffdcb2a HEAD@{2026-07-17 05:29:39}: rebase (finish): returning to refs/heads/fix/openapi-component-names
```

The lead's rebase landed at **05:29:40**. My `test-typia-generate` run ended at **05:37:35**. That suite chains seventeen steps, so it began before the rebase and finished after it — **the run spanned a mutation of its own working tree.**

It was not a benign mutation. Between `adca55e33f` and `43dc90fd22` the rebase replaced `packages/typia/src/internal/_ProtobufReader.ts` and `_isFormatUriReference.ts` (from #2119 and #2113) and bumped `packages/typia/package.json` from `13.1.12` to `13.1.14`. The suite drives `ttsx node_modules/typia/src/executable/typia.ts`, and ttsc keys its plugin cache by content — so the early steps ran against one typia source and version, and the later steps against another, with a plugin-cache invalidation in between.

**The run is void.** It is not evidence of a Windows race, and it is not evidence of a defect in this branch either. I cannot exclude the race mechanism, but I have no grounds to assert it, so I retract it. What still stands from that comment, because it rests on direct evidence rather than inference:

- it was **not** disk — `C:` had 6.52 GB free and the log contained zero disk or memory errors;
- it **cannot** be this branch — the jsconfig fixture is `typia.createIs<IJsConfigTemplate>()`, with no component names, no JSON schema, and no LLM inversion anywhere in its path;
- the failing step passes standalone on the rebased tree (file written, exit 0, correct emitted validator).

I have stopped the `test-typia-automated` run for the same reason — it was still executing when the tree changed under it, so its result would have been unreliable too. Both are now re-running cleanly against the settled head, and I will report only results measured on a tree that did not move underneath them.

Apologies for the noise; the corrected record matters more than the tidy story.

samchon:
Final rebase, version assignment, and lead certification.

## Rebase and version

Rebased onto `aefc24c86e` (#2119). All eleven commits replayed with no conflict; the 14-file diff is byte-identical across the rebase apart from the version line. Assigned **`13.1.14`**, the next patch on master's `13.1.13` (#2113) — mandatory because the diff changes native Go.

I reviewed the three commits master gained since this branch's last full round and agree with the batch's own read: #2114 governs property names and route accessors rather than component names, and #2113 and #2119 are format and Protobuf. Neither interacts with component allocation.

## Lead certification on `43dc90fd22`

| Gate | Result |
| --- | --- |
| `go -C packages/typia/test test -p 2 ../native/...` | pass, 0 failures |
| `pnpm --filter test-typia-generate start` | exit 0 |
| `pnpm --filter test-typia-automated start` | Success |
| `pnpm --filter test-utils start` | Success — `test_llm_invert_openapi_component_names`, `test_llm_invert_non_enumerable_definition`, `test_llm_schema_json_pointer_references` all run |
| `pnpm --filter test-typia-schema start` | Success — `test_json_schema_openapi_component_names`, `test_json_schema_openapi_component_name_cascade` all run |

**Two earlier failures on this head were my fault and are void, not findings.** I rebased this worktree while its suites were running, and later ran my own verification against orphaned suite processes that a `TaskStop` had failed to reap. A `test-typia-generate` run reported `Expected hard-linked output leaf to fail`, and two other runs died at exit 255. Every one of them passed cleanly once the tree was quiesced and measured sequentially. Recording this because a failure with a plausible-sounding message is exactly what gets mis-filed as a defect.

## Why this batch's evidence is trustworthy

The integration's red-then-green could not be taken at face value: on the pre-#2104 base the held test went red with an **identical message** for the wrong reason. Rather than trust it, the batch instrumented `invert` and separated the two causes:

| | pre-#2104 base (wrong red) | rebased base (right red) |
| --- | --- | --- |
| key for `$defs["/"]` | `"~1"` — a decode failure | `"/"` — a raw key stored |
| definition content | lost (`{}`) | preserved, 0 lost |

7 illegal keys written verbatim with 0 content lost is #2103's defect; a decode failure that loses the definition is #2104's, and no allocator could have turned that green. Fitting a fix to the first red would have been the *Evidence precedes correction* violation the development skill names.

## Change Integrity: this modifies a merged batch's test, and that is correct

The batch rewrote two assertions in #2104's `test_llm_schema_json_pointer_references`, and flagged it rather than slipping it through. I checked it independently and agree:

- `inversion keeps raw component identity` asserted `Object.hasOwn(components.schemas, "A/B")` — it pinned an **OpenAPI-invalid document**, since Components Object keys must match `^[a-zA-Z0-9.\-_]+$`. #2104's own scope line excludes component-name allocation, so that assertion pinned the status quo rather than a contract, and #2103 owns the line.
- The two contracts are mutually exclusive: an empty raw key cannot be both raw-identical and non-empty.
- The **forward** assertion — that `$defs` keys are unrestricted, which is #2104's real contract — is untouched. Only the inverse direction changed, and it was strengthened rather than relaxed: legal key *and* resolves.

## Self-Review round 5 caught a defect in its own plan

Round 5's regression test was worthless: it asserted "legal + resolves", which the escape fallback already satisfied, so it passed against the broken seeding it was written to catch. It was rewritten around the collision only correct seeding prevents and proved red-then-green.

The underlying defect was in the batch's own written plan: it seeded with `Object.keys` (enumerable only) while `resolve` uses `hasOwnProperty`, so an own non-enumerable definition resolved without being allocated and could seize an allocated name. Fixed with `getOwnPropertyNames`, and the plan document corrected. With campaign CI deliberately cancelled, a test that cannot fail is the only thing that can carry a defect through — this is the fourth such catch this session, and every one was self-reported.

## Deferred, recorded for the next discovery round

The pre-existing `.o<N>` disambiguator carries the identical cascade defect this batch fixed for `.x<hash>`, but it is shared with the LLM `$defs` and `reflect.name` paths, making it a public-contract change outside #2103. The discovery round has since confirmed it is **broader than recorded** — a real `namespace Foo { interface o1 }` collides with the `.o1` suffix and one type is silently dropped — so it is being adjudicated on its own evidence rather than bolted on here.

`43dc90fd22` settled after four polls; `29534452912` (spell-check) reached terminal `success` before the cancel landed and is recorded as `raced_terminal`; every other run `cancelled`.

PR #2109, opened 2026-07-16
fix(native): classify default libraries by compiler ownership and pin bundler-cache invalidation end to end

### LINKED ISSUE #2106: test: committed end-to-end bundler-cache invalidation regression
## Problem

PR #2100 landed the dependency producer (per-file consulted-declaration lists in the transform envelope) and its end-to-end acceptance — a webpack filesystem-cache rebuild picking up a type-only change — was verified manually, but no committed test exercises that path in CI. The invalidation pipeline (envelope `dependencies` → `@ttsc/unplugin` → loader `addDependency` → `fileDependencies` → persistent-cache snapshot) therefore has no regression guard in this repository. The original gap survived unnoticed for exactly this reason (#2092): every layer builds green while caches silently serve stale validators.

## Evidence

- Manual acceptance run for PR #2100: webpack 5 with `cache: { type: "filesystem" }`, `@ttsc/unplugin/webpack`, esbuild-loader, locally built typia. `src/mytype.ts` exports the validated interface; `src/index.ts` calls `typia.validate<MyType>(...)`. Build once; add a required property to `MyType`; rebuild **with the cache kept** → the second bundle rejects the now-invalid input (`$input.age`). Control with npm typia 13.1.1 (no producer) keeps serving the stale validator.
- Native Go tests added in #2100 cover envelope contents, not the bundler-cache round trip.

## Approach

Add a committed end-to-end regression that performs the two-build cache scenario against the locally built typia:

- A small fixture project (type file + consumer file + webpack config with a deterministic `cacheDirectory`) driven by a test that builds, mutates the type file, rebuilds without clearing the cache, and asserts the second bundle's runtime behavior (or generated source) reflects the new type.
- Keep it hermetic: local workspace/tarball linkage for `typia`, pinned `@ttsc/unplugin`/`webpack`/`esbuild-loader` devDependencies, isolated cache and output directories per run.
- Place it under the repository's test-suite conventions (a dedicated `tests/` project is the natural home; follow the existing suite layout and runner).

- The test fails when the producer is disabled (verifiable by temporarily reverting the `Dependencies` emission in `packages/typia/native/cmd/ttsc-typia/transform.go`) and passes on master.
- Runs in CI on the standard matrix; no reliance on network beyond the pinned installs the suites already perform.
- Existing suites unaffected.

## Coordination

- Follow-up to #2098 / PR #2100; carries forward the one unlanded acceptance item from #2095 (closed).
- Independent of samchon/ttsc#716 and of the SDK-encoder adoption follow-up; the test must keep passing unchanged when those land.

### LINKED ISSUE #2108: native transform: classify default-library files by compiler ownership, not basename pattern
## Problem

`transformDependencyDefaultLib` in `packages/typia/native/cmd/ttsc-typia/transform.go` (landed with PR #2100) drops a reported dependency file when its **basename** matches `lib.*.d.ts`. The intent is to exclude TypeScript's default libraries (toolchain-versioned, not project inputs), but basename is user-forgeable: a project's own ambient declaration file legitimately named `lib.custom.d.ts` — a common naming for hand-rolled ambient libs — is silently dropped from the `dependencies` envelope. Its changes then never invalidate consumers, reintroducing the exact stale-validator bug class this channel exists to prevent (#2092), for one filename shape.

## Evidence

- Lead audit of the merged PR #2100 diff (2026-07-16) identified the predicate:

  ```go
  func transformDependencyDefaultLib(fileName string) bool {
    base := filepath.Base(filepath.FromSlash(fileName))
    return strings.HasPrefix(base, "lib.") && strings.HasSuffix(base, ".d.ts")
  }
  ```

- Contrast with the orthodox classification in the sibling repo: ttsc's graph producer (samchon/ttsc#718, `packages/ttsc/driver/graph.go`) excludes compiler-owned libraries by their `bundled:///` virtual scheme — the compiler's own fact — never by name.
- The collector already drops virtual-URI sources (`strings.Contains(fileName, "://")` in `Touch`). tsgo embeds its default libraries as `bundled:///` virtual files, so on the standard path the basename branch may be **dead code whose only live effect is the false positive on user files**. Verify this with a probe before deleting; if any configuration surfaces default libs as on-disk paths (custom `--lib` resolution, `noLib` setups), classify those through the compiler, not the name.

## Consequence surface

- Any project containing a file matching `lib.*.d.ts` loses cache invalidation for types declared there — across every bundler adapter, watch mode, and persistent caches alike.
- Until typia#2107 lands (host graph union), the precise list is the only invalidation source, so the false positive is a correctness hole, not a precision loss.
- The `://` virtual-URI guard is not affected and stays (virtual sources cannot be watched regardless).

## Approach

Replace the basename predicate with compiler ownership: a file is excluded as a default library only if the compiler says it is one (tsgo program/driver classification; if the complete classification in tsgo is the `bundled:///` scheme already covered by the virtual-URI guard, delete the basename branch outright). Never classify project files by filename pattern.

- Go test: a fixture project with its own `lib.custom.d.ts` declaring a type consumed by a typia call — the envelope's `dependencies` entry for the consumer includes `lib.custom.d.ts`.
- Bundled default libraries remain excluded from every envelope entry (existing tests keep passing).
- The full native Go suite and the repository's standard test suites stay green.

## Coordination

- Follow-up from the lead audit of #2098 / PR #2100.
- Naturally batched with #2106 (same channel, same verification harness).
- Independent of the ttsc release sequence and of typia#2107.

samchon:
## Scope

One batch, two issues sharing the dependency-invalidation channel and its verification harness:

- Closes #2108 — replace the `lib.*.d.ts` basename heuristic in `transformDependencyDefaultLib` (`packages/typia/native/cmd/ttsc-typia/transform.go`) with compiler-owned default-library classification. The issue's probe runs first: verify whether tsgo's `bundled:///` embedding makes the basename branch dead code on the standard path, and whether any configuration surfaces default libs as on-disk paths.
- Closes #2106 — committed end-to-end bundler-cache invalidation regression test: webpack 5 filesystem cache, `@ttsc/unplugin/webpack`, two-build scenario against the locally built typia, placed per the repository test-suite conventions.

The #2106 fixture includes a project-owned `lib.custom.d.ts` type participating in invalidation so the end-to-end run also pins #2108.

## Plan

1. Probe tsgo's default-library surface (issue #2108 evidence step) and record findings on this thread.
2. Fix the classification; add the Go fixture test (`lib.custom.d.ts` in the envelope, bundled libs still excluded).
3. Add the committed e2e cache-invalidation suite; verify it fails with the producer reverted and passes on this branch.
4. Narrow-then-broad local verification (native Go tests, then standard suites), `pnpm format`, solo Self-Review (3+ rounds, findings commented before acting), merge on green checks.

Draft until implementation and Self-Review complete.

https://claude.ai/code/session_01U3kDYEkHutvSMcHPBXhd7P

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,10 @@ tests/test-typia-ttsc/fixtures/*/build/output/
 # transform test fixture scratch space (inside the npm-packed native tree)
 packages/typia/native/.tmp-ttsc-typia-tests/
 
+# bundler-cache e2e fixture scratch space (must stay inside the workspace so
+# the fixture resolves the suite's node_modules)
+tests/test-typia-bundler-cache/.tmp/
+
 package-lock.json
 !experiments/setup/package-lock.json
 
--- a/packages/typia/native/cmd/ttsc-typia/project_dependencies_custom_lib_name_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_dependencies_custom_lib_name_transform_test.go
@@ -1,0 +1,135 @@
+package main
+
+import (
+  "encoding/json"
+  "os"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestProjectDependenciesCustomLibNameTransform verifies a project-owned
+// ambient declaration file named `lib.custom.d.ts` appears in the transform
+// envelope's `dependencies`.
+//
+// Default-library exclusion must come from the compiler's own classification,
+// never from a filename pattern: a basename predicate (`lib.*.d.ts`) silently
+// drops a project's own ambient file with that common naming, so its changes
+// never invalidate consumers and persistent bundler caches serve stale
+// validators (samchon/typia#2108). Compiler-owned default libraries must still
+// stay excluded.
+//
+//  1. Build a project where `a.ts` calls `typia.validate<Bee>()` with `Bee`
+//     declared in `b.ts` consuming the ambient `CustomAmbient` interface
+//     declared in `src/lib.custom.d.ts`, plus a `Date` property so the
+//     analysis touches real default libraries.
+//  2. Run project transform mode and decode the JSON envelope.
+//  3. Assert `dependencies["src/a.ts"]` contains `src/lib.custom.d.ts`.
+//  4. Assert no compiler-owned default library (virtual `bundled:///` URI or
+//     any value outside the project) leaks into the entry.
+func TestProjectDependenciesCustomLibNameTransform(t *testing.T) {
+  project := projectDependenciesCustomLibNameProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--output", "ts",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("project transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  var envelope struct {
+    TypeScript   map[string]string   `json:"typescript"`
+    Dependencies map[string][]string `json:"dependencies"`
+  }
+  if err := json.Unmarshal([]byte(out), &envelope); err != nil {
+    t.Fatalf("decode envelope: %v\n%s", err, out)
+  }
+  entries := envelope.Dependencies["src/a.ts"]
+  if len(entries) == 0 {
+    t.Fatalf("dependencies must contain an entry for src/a.ts:\n%s", out)
+  }
+  found := map[string]bool{}
+  for _, entry := range entries {
+    found[entry] = true
+    if strings.Contains(entry, "://") {
+      t.Fatalf("dependencies must exclude virtual URI sources, got %q in %v", entry, entries)
+    }
+    if filepath.IsAbs(filepath.FromSlash(entry)) || strings.HasPrefix(entry, "../") {
+      t.Fatalf("dependencies must not report files outside the project, got %q in %v", entry, entries)
+    }
+  }
+  if !found["src/lib.custom.d.ts"] {
+    t.Fatalf("dependencies of src/a.ts must contain the project-owned src/lib.custom.d.ts: %v", entries)
+  }
+  if !found["src/b.ts"] {
+    t.Fatalf("dependencies of src/a.ts must contain the direct declaration file src/b.ts: %v", entries)
+  }
+}
+
+func projectDependenciesCustomLibNameProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "project-dependencies-custom-lib-name-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() { _ = os.RemoveAll(dir) })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(projectDependenciesCustomLibNameTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  for name, body := range map[string]string{
+    "a.ts":            projectDependenciesCustomLibNameSourceA,
+    "b.ts":            projectDependenciesCustomLibNameSourceB,
+    "lib.custom.d.ts": projectDependenciesCustomLibNameSourceLib,
+  } {
+    if err := os.WriteFile(filepath.Join(src, name), []byte(body), 0o644); err != nil {
+      t.Fatalf("write %s: %v", name, err)
+    }
+  }
+  return dir
+}
+
+const projectDependenciesCustomLibNameTSConfig = `{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "commonjs",
+    "moduleResolution": "bundler",
+    "ignoreDeprecations": "6.0",
+    "esModuleInterop": true,
+    "strict": true,
+    "skipLibCheck": true,
+    "rootDir": "src",
+    "outDir": "dist"
+  },
+  "include": ["src"]
+}
+`
+
+const projectDependenciesCustomLibNameSourceA = `import typia from "typia";
+
+import { Bee } from "./b";
+
+export const validateBee = (input: unknown) => typia.validate<Bee>(input);
+`
+
+const projectDependenciesCustomLibNameSourceB = `export interface Bee {
+  id: string;
+  when: Date;
+  ambient: CustomAmbient;
+}
+`
+
+const projectDependenciesCustomLibNameSourceLib = `declare interface CustomAmbient {
+  tag: string;
+}
+`
--- a/packages/typia/native/cmd/ttsc-typia/transform.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform.go
@@ -127,7 +127,10 @@ func runTransformProject(
     Diagnostics: []transformCompilerDiagnostic{},
     TypeScript:  map[string]string{},
   }
-  collector := newTransformDependencyCollector(cwd)
+  collector := newTransformDependencyCollector(cwd, func(fileName string) bool {
+    sf := prog.SourceFile(fileName)
+    return sf != nil && prog.TSProgram.IsLibFile(sf)
+  })
   schemametadata.MetadataDependency_listen(prog.Checker, collector.Touch)
   defer schemametadata.MetadataDependency_release(prog.Checker)
   for _, sf := range prog.SourceFiles() {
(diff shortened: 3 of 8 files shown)


samchon:
## #2108 probe findings (evidence step before the fix)

Probe: a temporary Go test loads a fixture project through `driver.LoadProgram` — a typia call consuming a type with `Date`, `Array<string>`, and a project-owned `src/lib.custom.d.ts` ambient interface — registers the raw `MetadataDependency` listener, and dumps every reported fileName plus the compiler's per-file classification.

Results:

- `bundled.Embedded = true`, `bundled.LibPath() = "bundled:///libs"` on the standard ttsc toolchain. Default libraries are touched by the listener as `bundled:///libs/lib.*.d.ts` virtual files (`lib.es5.d.ts`, `lib.es2020.date.d.ts`, ...), so the collector's existing `strings.Contains(fileName, "://")` guard already drops all of them. **The basename branch is dead code on the standard path; its only live effect is the false positive.**
- The fixture's own `src/lib.custom.d.ts` IS reported by the listener and `Program.IsLibFile` classifies it as *not* a library — but `transformDependencyDefaultLib` drops it by name. False positive confirmed exactly as the issue states.
- Configurations that surface default libs as on-disk paths do exist:
  - tsgo built with the `noembed` tag: `IsBundled` is always false and `LibPath()` is the executable's directory — on-disk libs would bypass the `://` guard. Not reachable through ttsc's standard plugin build (probe shows `Embedded = true`), but it is a real tsgo build mode.
  - `libReplacement: true` with installed `@typescript/lib-*` packages: `fileLoader.pathForLibFile` resolves the lib to an on-disk `node_modules` path, and that file is still registered in the program's lib-file map.
- The complete compiler-owned classification is `compiler.Program.IsLibFile(sourceFile)`, backed by `p.libFiles`, which is populated exclusively from lib-loading tasks — bundled and replaced alike. It is callable from the plugin through the shim alias (`shimcompiler.Program = compiler.Program`) on `prog.TSProgram`.

Fix shape (next commit): the dependency collector takes a compiler-backed `isLibraryFile` classifier (`prog.SourceFile(name)` + `prog.TSProgram.IsLibFile`); `transformDependencyDefaultLib` is deleted; the `://` virtual-URI guard stays. Since `IsLibFile` is the compiler's own fact, every configuration above classifies correctly and no project file can be dropped by its name.

samchon:
## #2106 implementation and acceptance evidence

New suite `tests/test-typia-bundler-cache` (picked up automatically by `pnpm test:packages`' `./tests/test-*` filter): a hermetic fixture project written into a workspace-internal scratch dir builds through webpack 5 with `cache: { type: "filesystem" }`, `@ttsc/unplugin/webpack` (catalog `0.18.4`), `esbuild-loader`, and the workspace-linked typia; then mutates only type files and rebuilds with the cache kept, asserting the bundle's runtime validation reflects each change:

1. Cold build: valid input accepted, input missing `id` rejected at `$input.id` (proves a live validator, not a stub).
2. Add required `MyType.age` in `mytype.ts`, rebuild with cache kept: previously valid input must fail at `$input.age`.
3. Add required `CustomLabel.flag` in the project-owned ambient `src/lib.custom.d.ts`, rebuild with cache kept: must fail at `$input.ambient.flag` — this step pins #2108 end to end.

Acceptance discriminations (both run locally, results quoted from the failing runs):

- **Producer disabled** (`out.Dependencies = collector.ToJSON()` emission commented out): suite fails at step 2 — `cached rebuild after adding MyType.age: expected validation failure at $input.age, got success` — the persistent cache served the stale validator, exactly the #2092 class.
- **Basename predicate reinstated** on top of the working producer: suite passes step 2 but fails at step 3 — `cached rebuild after adding CustomLabel.flag: expected validation failure at $input.ambient.flag, got success` — the `lib.custom.d.ts` false positive reproduced through the whole bundler pipeline.
- **This branch**: suite passes all steps.

Hermeticity: fixture and cache/output directories are per-run temp dirs inside `tests/test-typia-bundler-cache/.tmp` (gitignored; kept inside the workspace so the fixture resolves the suite's own pinned node_modules and the shared ttsc plugin cache); `webpack` is pinned to the `5.107.2` line already present in the lockfile so no other importer's resolution moved; the only lockfile additions are the suite's own packages.

## Local verification log

- `pnpm test:go:native`, `pnpm test:go:public`, and `go test -tags typia_native_internal ../native/cmd/ttsc-typia/`: pass.
- New Go test `TestProjectDependenciesCustomLibNameTransform`: fails on master's predicate (`dependencies of src/a.ts must contain the project-owned src/lib.custom.d.ts: [src/b.ts]`), passes with the fix.
- `pnpm build`: pass. `pnpm --filter ./tests/test-typia-bundler-cache start`: pass.
- Full `pnpm test`: every suite passes except two pre-existing `test-typia-generate` filesystem subtests (`parent directory links terminate`, `repeated directory aliases are rejected`) that fail identically on an unmodified `master` worktree on this Windows machine (symlink/junction environment; a third subtest self-skips with "file symbolic links require Windows developer mode"). CI's Ubuntu runs pass that suite, including on this branch's first commit.
- The first CI `test` run on 5979737 failed with `link: mapping output file failed: no space left on device` during the Go phase — runner disk exhaustion, unrelated to the diff (test:packages had already passed in that run; sibling branches were green at the same time). Fresh runs are triggered by this push.

samchon:
## Self-Review — Round 1 (findings before action)

Fresh full pass over the complete diff (`transform.go`, the new Go test, the new `tests/test-typia-bundler-cache` suite, `.gitignore`, `pnpm-lock.yaml`) plus the consequence surface (collector call sites, single-file transform mode, `MetadataDependency` listener contract, `@ttsc/unplugin` consumer, suite runner wiring, CI workflows, website docs).

Findings:

1. **Classifier correctness across configurations.** `prog.SourceFile` matches the exact `FileName()` strings the listener reports (both come from the same program), and `Program.IsLibFile` is keyed by the program's own lib map, covering bundled, `noembed` on-disk, and `libReplacement` resolutions. Lookups are memoized per fileName in the collector's existing `values` map, so the O(files) scan runs once per distinct reported file. No defect.
2. **Single-file transform mode** registers no listener (unchanged), so the new classifier only runs on the project path that owns the envelope. No defect.
3. **Go test scope.** The out-of-project assertion (`no absolute or ../ values`) is deliberate: in this fixture every consulted declaration is in-project, so any absolute leak would indicate an on-disk default library entering the envelope. Negative twin (bundled libs stay excluded) is covered by the `://` assertion plus the existing envelope test. Keep.
4. **Suite hermeticity on CI.** Lockfile committed; fixture resolves everything through the suite's own node_modules; `resolve.symlinks: false` keeps workspace-linked typia under its node_modules id so the unplugin transforms only the fixture project. esbuild ships its binary via optional platform packages, so pnpm's ignored build script does not affect it. No action.
5. **Website docs** carry no page describing the envelope's exclusion policy (only the unrelated `@typia/unplugin@12` legacy limitation), so no doc change belongs to this fix.
6. **Change-integrity call-out:** Go source under `packages/typia/native` changed; the development skill pairs that with a typia version bump, but versioning is the repository owner's domain and this PR deliberately carries no version change. Recording the deferral here instead of bumping.
7. **CI note:** the first `test` run failed with runner disk exhaustion during the Go link step (evidence in the previous comment); fresh runs for 6c5cd44 are in progress and gate the merge.

Remediation plan: no code change required by this round; items 6-7 are recorded as deferrals/observations. Round 2 follows as a fresh pass.

samchon:
## Self-Review — Round 2 (findings before action)

Fresh full pass over the current state: re-read `runTransformProject` and the collector in final form, the complete suite driver and fixture writers, the Go test, `.gitignore`, and the lockfile importer section.

Findings:

1. **Classifier evaluation order.** `value()` checks the virtual-URI guard before the compiler classification, so bundled libraries short-circuit and the O(program) `SourceFile` scan only runs for the (memoized) non-virtual file names. Verified the memo (`values`) covers the dropped case (`""`) as before. No defect.
2. **Envelope key discipline unchanged.** `Begin`/self-drop/`ToJSON` sorting untouched; the change is confined to how a reported file is classified. No defect.
3. **Suite discrimination scope.** The regression pins the staleness class: a cache that never restored anything would pass trivially, but that failure mode is a performance regression outside #2106's correctness contract, and the producer-disabled run (previous comment) already proved this exact fixture's cache actively restores and serves stale validators when the channel is absent — i.e. the discriminating power is real, not assumed. Recording the boundary rather than adding brittle webpack-stats assertions.
4. **Fixture mutation robustness.** `mutate` combines a pre-write delay with an explicit forward `utimes`, covering coarse-mtime filesystems in both directions; the scenario passed repeatedly locally, including through the full `pnpm test` chain.
5. **Suite failure modes.** Build failures, bundle execution failures, and missing JSON output each throw with the full child output; fixture cleanup runs in `finally`; the scratch root is gitignored and workspace-internal by design (module + plugin-cache resolution).

Remediation plan: no change required by this round. CI for 6c5cd44 is still in progress (`test`, `build`, and two experiment jobs pending; the rest green). Round 3 follows as a fresh pass; merge waits for green required checks.

samchon:
## Self-Review — Round 3 (findings before action)

Fresh full pass with a consumer, compatibility, and history lens over the final change list (`.gitignore`, the Go test, `transform.go`, `pnpm-lock.yaml`, the four suite files).

Findings:

1. **Consumer contract.** Re-read `@ttsc/unplugin`'s `selectFileDependencies`/`notifyFileDependencies`: envelope values may be project-relative or absolute, are absolutized against the project root, deduplicated, and self-dropped. The fix only adds legitimate project-relative values (e.g. `src/lib.custom.d.ts`) to that stream — fully within the existing contract; no consumer relied on the basename exclusion.
2. **Windows path safety of the `://` guard.** Drive-letter paths (`D:/...`) contain `:` followed by a single slash, never `://`; UNC paths carry no scheme. The guard cannot misfire on real filesystem paths, and the compiler classification behind it needs no path parsing at all.
3. **Concurrency and registration.** Listener registration/release is unchanged; the collector remains single-threaded per program exactly as before the change.
4. **Packaging.** The new colocated `*_test.go` follows the existing pattern of test files shipping inside the npm-packed `native/` tree; no `files`/pack change needed.
5. **No probe leftovers.** The temporary probe test and both acceptance-probe patches are absent from the branch; the final change list is exactly the eight intended files.
6. **History.** All four `TestProjectDependencies*` cases (PR #2100's three plus the new one) pass together; the new case fails on master's predicate, so the pair pins the regression in both directions.

Remediation plan: nothing to change. This is the third consecutive complete round with no sound improvement, satisfying the review stop rule. Remaining gate before ready-for-review + squash merge: green required checks for 6c5cd44 (currently `test`, `build`, and two experiment jobs still running).

Issue #2039, opened 2026-07-16
Runtime and generated JSON Schema format validators diverge


samchon:
## Problem

Four advertised JSON Schema format validators are incorrect or inconsistent across typia runtime, `@typia/utils`, and native generated checks:

| Format | Current defect |
| --- | --- |
| `byte` | multiline `m` makes one valid base64 line validate an otherwise invalid string |
| `ipv6` | native generated regex lost `\d` and `\.` escapes in IPv4-embedded branches |
| `iri-reference` | all three owners use the absolute-only `iri` pattern and reject relative IRI references |
| `date` / `date-time` | regex permits nonexistent calendar dates such as 2023-02-29 and 2024-02-30 |

Typia documents these as standard formats and emits the same names into JSON Schema, so runtime validation must not silently implement a different language.

Normative references: [JSON Schema Validation 2020-12](https://json-schema.org/draft/2020-12/json-schema-validation), [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html), [RFC 3987](https://www.rfc-editor.org/rfc/rfc3987.html), and [RFC 4291](https://www.rfc-editor.org/rfc/rfc4291.html).

## Evidence

- Runtime/utils `byte` and native `FormatCheatSheet` use the same anchored base64 expression with `gm`; invalid text containing a valid `YWJj` line passes.
- Runtime/utils IPv6 patterns correctly contain `\d` and `\.`, while `FormatCheatSheet.go` contains literal `d` and wildcard `.`. Generated checks reject valid `::ffff:192.168.0.1` and can accept malformed `::ffff:1ddX1ddX1ddX1dd`.
- `_isFormatIriReference` is byte-for-byte the scheme-required IRI pattern. JSON Schema defines `iri-reference` as either an IRI or a relative reference; `../경로`, `/경로`, and `#조각` are rejected.
- Public OpenAPI validation accepts `2023-02-29`, `2024-02-30`, and `2023-04-31`. RFC 3339 makes the allowed day depend on month/year.

Closed unmerged PR #1864 carried partial byte/IPv6/IRI observations but no issue or open replacement. Exact searches found no implementation claim.

## Consequence surface

Type tags, JSDoc tags, imported runtime helpers, `OpenApiValidator`, native generated checks, LLM/HTTP consumers, repeated calls, Unicode IRIs, IPv4-embedded IPv6 forms, leap years, and date-time offsets/fractions.

## Approach

[...]

Issue #2111, opened 2026-07-16
NamingConvention.variable misclassifies identifiers, leaking illegal bindings into generated HTTP accessors and parameter keys


samchon:
## Problem

`NamingConvention.variable()` is documented as *"Check if string is valid JavaScript variable name"* but returns `true` for eight strings that cannot be binding identifiers in an ES module. Because the public HTTP migration model derives `IHttpMigrateRoute.accessor` and path-parameter `key` from this predicate, a legal OpenAPI route or parameter name can be turned into an illegal identifier in the model consumed by RPC/SDK generation.

Affected users: `@typia/utils` consumers calling `NamingConvention.variable`/`reserved` directly, and every downstream SDK/RPC generator built on `HttpMigration.application()`.

## Evidence

`packages/utils/src/utils/NamingConvention.ts`:

```ts
export const variable = (str: string): boolean =>
  reserved(str) === false && /^[a-zA-Z_$][a-zA-Z_$0-9]*$/g.test(str);
export const reserved = (str: string): boolean => RESERVED.has(str);
```

`variable()` is documented as an identifier-legality predicate but implemented as "not in my word list, and looks like an identifier". Using Node's own ES-module parser as the oracle — generated SDK artifacts are ES modules, which are always strict:

```js
import vm from "node:vm";
const legal = (w) => { try { new vm.SourceTextModule(`export function f(${w}) { return ${w}; }`); return true; } catch { return false; } };
```

| word | `reserved()` | `variable()` | legal binding in a module | class |
| --- | --- | --- | --- | --- |
| `await` | false | **true** | false | ECMAScript `ReservedWord` |
| `yield` | false | **true** | false | `ReservedWord` / strict |
| `let` | false | **true** | false | strict future reserved |
| `static` | false | **true** | false | strict future reserved |
| `implements` | false | **true** | false | strict future reserved |
| `interface` | false | **true** | false | strict future reserved |
| `eval` | false | **true** | false | **strict BoundNames restriction** |
| `arguments` | false | **true** | false | **strict BoundNames restriction** |
| `package`, `private`, `protected`, `public` | true | false | false | already handled |
| `foo`, `async`, `of`, `get`, `undefined`, `NaN` | false | true | true | correctly accepted |

Two observations establish the root cause:

1. `package`, `private`, `protected`, and `public` are present while `implements`, `interface`, `let`, and `static` — the *same* strict-future-reserved class — are absent. The omission is an oversight, not a policy.
[...]

PR #2113, opened 2026-07-16
fix(typia): reject raw double quotes in uri-reference format validation

### LINKED ISSUE #2110: uri-reference format validation accepts a raw double quote in all three validator owners
## Problem

`tags.Format<"uri-reference">` accepts strings containing a raw `"` character, which RFC 3986 does not permit in any component of a `URI-reference`. Both TypeScript runtime helpers and the native code generator embed the same wrong character classes, so validation, the emitted JSON Schema `format` claim, and `@typia/utils` OpenAPI validation all disagree with the standard they name.

Affected users: anyone validating `uri-reference` values via typia type tags or JSDoc tags, and any `@typia/utils` consumer routing through `OpenApiValidator` (adapter, LLM, and HTTP validation paths).

## Evidence

RFC 3986 composes every component from `unreserved`, `sub-delims` (`!$&'()*+,;=`), and pct-encoded octets. A raw `"` belongs to none; `%22` is its only legal spelling.

Three owners carry a byte-identical expression with `"` inside 8 character classes (`&'"(` where it should read `&'(`):

- `packages/typia/src/internal/_isFormatUriReference.ts` — `_isFormatUriReference`
- `packages/utils/src/validators/functional/_isFormatUriReference.ts` — `_isFormatUriReference`
- `packages/typia/native/core/factories/FormatCheatSheet.go` — `FormatCheatSheet["uri-reference"]`

Reproduction (Node, no build required — reads the pattern from source):

```js
const fs = require("fs");
const src = fs.readFileSync("packages/typia/src/internal/_isFormatUriReference.ts", "utf8");
const re = eval("(" + src.match(/\/\^.*\$\/i/s)[0] + ")");
for (const s of ['"', 'a"b', '/a"b', '?q="', '#"', 'http://a"b/c'])
  console.log(JSON.stringify(s), re.test(s));   // observed: true   expected: false
console.log(JSON.stringify("%22"), re.test("%22"));   // true (correct — must stay true)
```

Observed: all six raw-quote inputs return `true`. Expected: `false`.

Two independent facts show the `"` is an accidental insertion rather than a deliberate relaxation:

1. **Internal asymmetry.** Within the *same* expression the `userinfo` and `IPvFuture` classes correctly omit the quote; only reg-name, path, query, and fragment carry it.
2. **The sibling disagrees.** `packages/typia/src/internal/_isFormatUri.ts` is the same expression with no quote anywhere and correctly rejects `http://a"b/c`.

**typia's own coverage contract is violated.** `packages/utils/src/utils/internal/OpenApiTypeCheckerBase.ts` declares `(x === "iri-reference" && y === "uri-reference")` — `iri-reference` covers `uri-reference`, so every valid `uri-reference` must be a valid `iri-reference`. Building both validators from current source:

| Input | `uri-reference` | `iri-reference` | Invariant |
| --- | --- | --- | --- |
| `"`, `a"b`, `/a"b`, `?q="`, `#"`, `http://a"b/c` | `true` | `false` | **violated** |
| `%22`, `a/b`, `a'b`, `http://h/p?q=1#f` | `true` | `true` | held |

Six violations, all raw-quote; every quote-free probe holds. The subset accepts what its declared superset rejects.

Provenance: blame traces the character class to `dd942d835ae` (2024-11-15). Closed #2039 / merged #2075 corrected adjacent `byte`, `ipv6`, `iri-reference`, `date`, and `date-time` behavior; #2039's body never mentions `uri-reference` and #2075 changed no `_isFormatUriReference` file.

## Consequence surface

- **`tags.Format<"uri-reference">`** (type tag and JSDoc tag): runtime validation accepts values the emitted JSON Schema `format` declares RFC 3986.
- **`@typia/utils`**: `_isStringFormat` routes `uri-reference` to the same helper, so `OpenApiValidator` and adapter/LLM/HTTP validation inherit the false positive.
- **Native generation**: generated checks embed the identical expression. Fixing only the two TypeScript helpers would create runtime/generated divergence and repeat the parity failure class fixed by #2039/#2075.
- **`OpenApiTypeChecker`**: the `iri-reference`/`uri-reference` coverage relation depends on precise format language.
- **Platforms**: TS runtime and native Go carry the same expression; no platform-specific divergence.
- **Compatibility**: strictly narrowing. Previously-accepted quote-bearing strings become correctly rejected; `%22` and all valid references are unaffected.

**Complete case matrix.** `"` is the only mishandled character. An exhaustive 128-codepoint sweep across reg-name, path, query, and fragment flags `"`, `#`, `/`, `:`, `?`, `@` — but the latter five are RFC delimiters that merely re-parse the string under different component boundaries (`http://a#b/p` is validly authority `a` + fragment `b/p`), not defects. A differential over a 300-string corpus (valid absolute/relative/IPv4/IPv6/urn/mailto references plus `"` injected at every offset) confirms: removing `"` newly rejects 218 strings (all quote-bearing) and newly accepts 0, with all 39 quote-free seeds keeping their verdict.

| Case | Expected |
| --- | --- |
| `"`, `a"b`, `/a"b`, `?q="`, `#"`, `http://a"b/c`, `http://x/p"q` | reject |
| `%22`, `%2F`, other valid pct-encodings | accept |
| `a'b` and every other sub-delim | accept |
| `""`, empty reference, absolute, relative, IPv4/IPv6 authority | unchanged |
| `%2`, malformed `%` escapes, controls, spaces, `<`, `>`, `\` | reject (already correct) |

## Approach

The invariant: **`uri-reference` accepts exactly RFC 3986 `URI-reference` = `URI / relative-ref`**, and `iri-reference` must remain a superset of it.

The architectural owner is the shared format table. `packages/typia/native/core/factories/FormatCheatSheet.go` is authoritative for generated checks; the two TypeScript helpers must carry the same language. Remove the raw `"` from every RFC 3986 component class while preserving apostrophes, `%22`, and every other sub-delim. Do not fix only the TypeScript side — the three owners must stay byte-identical, as they are today.

- **Positive:** absolute URIs, relative references, the empty reference, IPv4 and IPv6 authorities, `urn:`/`mailto:` forms, `%22` and other valid pct-encodings, and all sub-delims (notably `'`) remain accepted.
- **Negative:** raw `"` is rejected in reg-name, path, query, and fragment, and as a whole reference.
- **Boundary:** `%2` and other malformed escapes, controls, spaces, `<`, `>`, `\` stay rejected; repeated calls are stable (the pattern must not carry `/g`).
- **Regression:** the `iri-reference` ⊇ `uri-reference` coverage relation holds for the full probe set — this is the oracle that made the bug visible and should be pinned as a test.
- **Parity:** one authoritative matrix exercised across the runtime helper, `@typia/utils` `_isStringFormat`/`OpenApiValidator`, type tags, JSDoc tags, and native-generated checks. There is currently **no** `uri-reference` negative matrix anywhere in the repository — its absence is why three owners agreed on the same wrong language for ~20 months. Take expectations from RFC 3986, not from current output.
- **Commands:** `pnpm --filter ./tests/test-typia-schema start` and `pnpm --filter ./tests/test-utils start` (narrow); `pnpm test:go:native` for the generated side; `pnpm test` as the broader gate.

## Coordination

- Native Go changes require the coordinated `typia` package version bump in the same change (see the development skill's Change Integrity).
- Natural sibling of #2039 / merged PR #2075 (format parity), which deliberately deferred formats it did not own. Not covered by #2102 or #2103, which concern reference pointers and component names rather than URI lexis.
- Safe to parallelize with non-format work; serialize against any other branch bumping `packages/typia/package.json`.

samchon:
Fixes #2110.

## Intent

`tags.Format<"uri-reference">` accepts strings containing a raw `"`, which RFC 3986 permits in no component of a `URI-reference` (`%22` is its only legal spelling). Three owners carry a byte-identical expression with `"` wrongly present in 8 character classes (`&'"(` where RFC 3986's `sub-delims` yield `&'(`):

- `packages/typia/src/internal/_isFormatUriReference.ts`
- `packages/utils/src/validators/functional/_isFormatUriReference.ts`
- `packages/typia/native/core/factories/FormatCheatSheet.go` — `FormatCheatSheet["uri-reference"]`

The insertion is accidental, not policy: within the *same* expression the `userinfo` and `IPvFuture` classes correctly omit the quote, and the sibling `_isFormatUri.ts` carries no quote anywhere.

## Scope

- Remove the raw `"` from the 8 affected classes in all three owners, keeping them byte-identical. Fixing only the TypeScript side would create runtime/generated divergence — the parity failure class #2039/#2075 fixed.
- Add the `uri-reference` regression matrix that did not previously exist anywhere in the repository. Its absence is why three owners agreed on the same wrong language for ~20 months. Expectations are taken from RFC 3986, not from current output.
- Pin the `iri-reference` ⊇ `uri-reference` coverage relation declared by `OpenApiTypeCheckerBase` — the oracle that made the bug visible.

Strictly narrowing: previously-accepted quote-bearing strings become correctly rejected. `%22`, `a'b` and every other sub-delim, and all valid references are unaffected.

## Version

**Pending lead assignment.** This changes native Go (`FormatCheatSheet.go`), which per the development skill's Change Integrity requires a coordinated `packages/typia/package.json` version bump. The lead serializes the native version chain across several concurrent native pull requests and assigns the version at rebase time, so `packages/typia/package.json` is deliberately untouched here.

## Verification

Local verification only; campaign CI is deliberately suspended and this commit's runs are cancelled under the exact-SHA gate. Commands recorded in the thread as increments land.


samchon:
## Self-Review — round 1 (findings before remediation)

Complete round over `origin/master...f0a6442816` (3 owner files + 1 new test).

### Verified sound

- **Only the quote changed.** Each owner is exactly its previous content with the 8 `&'"(` → `&'(` substitutions applied and nothing else — asserted mechanically (`after === before.split("&'\"(").join("&'(")` → `true` for all three, 8 substitutions each). No collateral edit to the neighbouring `uri`, `iri`, or IPv6 language.
- **Three owners byte-identical again.** `typia src === utils src` and `typia src === native go` both `true`; raw-quote count `0 / 0 / 0`; the 2 classes that were always correct (`userinfo`, `IPvFuture`) plus the 8 repaired ones give 10 apostrophe classes per owner.
- **No fourth owner.** A tracked-file sweep for the expression family returns only the 3 changed files plus `_isFormatUri.ts` (typia + utils), which is the separate `uri` format and is correctly quote-free. `_isStringFormat.ts` routes `uri-reference` to the repaired helper.
- **No documentation change owed.** The only website mention of `uri-reference` (`website/src/content/docs/validators/tags.mdx:152`) lists the format name; it makes no claim about the character repertoire.
- **No fixture or random-generation fallout.** `_randomFormatUriReference` delegates to `_randomFormatUrl`, so generated values carry no quote; the `@typia/template` fixtures use `git://github.com/samchon/typia.git/../nestia.git` (valid) and spoil with `http://깃허브.com` (non-ASCII), neither affected.

### Finding 1 — the repeated-call stability boundary is not pinned (accept, remediate)

The issue's acceptance criteria list, under **Boundary**: *"repeated calls are stable (the pattern must not carry `/g`)"*. The matrix does not assert this, so the criterion is unmet.

This is not hypothetical for this file: a `/g` regex carries `lastIndex` across `.test()` calls and alternates verdicts, and the sibling `NamingConvention.variable` in this repository already ships a `/g` identifier regex. The peer `test_openapi_validator_format_parity.ts` guards it explicitly ("Repeat base64 checks in alternating order to expose regular-expression state").

**Remediation:** re-run an accept/reject pair alternately several times through all four surfaces, so a `/g` regression fails the matrix rather than passing silently.

### Non-findings (recorded, deliberately not acted on)

- **No `test-typia-schema` case added.** Type tags, JSDoc tags, the runtime helper, and the native-generated expression are all already exercised from the `test-utils` matrix (that suite runs the transform), so a second copy there would duplicate coverage rather than add a surface. `pnpm --filter ./tests/test-typia-schema start` is still run as a regression gate and passes.
- **Fix not widened beyond `"`.** The lead's exhaustive sweep and 300-string differential establish `"` as the only mishandled character (218 newly rejected, 0 newly accepted). Probes such as `1bad:relative` (RFC `path-noscheme` forbids `:` in a first segment) are out of scope for this issue and are not introduced here.
- **`packages/typia/package.json` deliberately untouched** — native Go changed, so a version bump is owed; the lead serializes the native version chain and assigns it at rebase time.

samchon:
## Self-Review — round 2 (fresh full round, clean)

Round 1's finding was remediated in `4267b2fe24`, so this is a complete fresh round over `origin/master...4267b2fe24`, not a recheck of the fix.

### Round 1 remediation was itself defective, and this round's evidence caught it

The first attempt at the stability pin alternated an accept with a reject:

```ts
for (let i = 0; i < 3; ++i) {
  validate("https://example.com/path?query=1#fragment", true);
  validate('http://a"b/c', false);
}
```

That pin has **no teeth**. `RegExp.prototype.test` only advances `lastIndex` on a match and resets it to `0` on a failure, so an interleaved reject silently repairs the state a global flag would corrupt. Simulated against the shipped pattern:

| sequence | `/i` (shipped) | `/g` (regression) |
| --- | --- | --- |
| accept, reject, accept, reject … | `true false true false` | `true false true false` — **identical, undetectable** |
| accept, accept, accept, accept | `true true true true` | `true false true false` — **detected** |

The pin now repeats one accepted reference consecutively. Verified end-to-end by injecting `/gi` into the shipped pattern, rebuilding, and observing the matrix fail (`Bug on uri-reference accepts "https://example.com/" through type tag`), then reverting.

### The three-owner parity claim is now proven, not assumed

The doc comment claims the matrix runs through every owner. Verified per owner by injecting a regression into one owner at a time:

| Surface | Owner it exercises | Proof |
| --- | --- | --- |
| `_isFormatUriReference` (direct) | `packages/typia/src/internal/_isFormatUriReference.ts` | `/gi` injection into typia src fails the matrix |
| `typia.is` type tag | same typia helper (the transform imports it) | same injection fails on `through type tag` |
| `@format` JSDoc tag | **`FormatCheatSheet.go`** — `metadataCommentTagFactory_FORMATS` builds the comment-tag table from it | re-injecting the raw quote into the Go entry **only** fails on `Bug on uri-reference rejects "\"" through comment tag` |
| `OpenApiValidator` | `packages/utils/src/validators/functional/_isFormatUriReference.ts` | routed via `_isStringFormat` |

So the native Go owner is genuinely covered by a dedicated surface, and the fix is not TypeScript-only in effect as well as in text. Both injections were reverted; `git status` is clean and the three owners are byte-identical (`typia src === utils src`, `typia src === native go`, raw-quote count `0 / 0 / 0`, apostrophe classes `10 / 10 / 10`).

### Also checked this round, nothing to change

- **Diff hygiene.** Each owner equals its previous content with only the 8 `&'"(` → `&'(` substitutions (asserted mechanically). Branch surface is exactly 4 files; no `lib/`, tarball, or `package.json` crept in; working tree clean after every suite, including the regenerating automated matrix.
- **Test identity.** `test_openapi_validator_format_uri_reference` is exported once, matches its filename, and is unique across all suites — the filename/export mismatch class recorded elsewhere in this campaign is not reintroduced.
- **Formatting.** `prettier --check` passes for the new file. The repository-wide formatter was not run, per the campaign's no-format rule. Stored line endings are LF, matching peers.
- **Consequence surface.** `test-typia-automated` passes: `_randomFormatUriReference` delegates to `_randomFormatUrl`, so generated values carry no quote, and the `TypeTagFormat`/`CommentTagFormat` fixtures use a quote-free valid value with a non-ASCII spoil. No documentation is owed — the sole website mention lists the format name without a repertoire claim.

**Round 2 is clean; Self-Review concludes.** Remaining non-finding: `packages/typia/package.json` is deliberately unbumped although native Go changed, pending the lead's serialized native version assignment at rebase time.

samchon:
## Local verification (campaign CI suspended)

Final pushed SHA `4267b2fe24`. Automatic CI for every campaign SHA on this branch was cancelled under the exact-SHA gate, so the following local runs are the gate.

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils start` | **Success** (full suite) |
| `pnpm --filter ./tests/test-typia-schema start` | **Success** (full suite) |
| `pnpm test:go:native` (`go -C packages/typia/test test ../native/...`) | **Success**, all packages `ok` |
| `go -C packages/typia/test test -tags typia_native_internal ../native/core/factories/` | **Success** (`ok … core/factories`) |
| `pnpm --filter ./tests/test-typia-automated start` | **Success** (consequence check: the `TypeTagFormat` / `CommentTagFormat` fixtures carry a `uriReference` field) |
| `npx prettier --check` on the new test file only | **Success** (repository-wide formatter deliberately not run) |

### Fail-then-pass evidence for the new matrix

The matrix was written first, from RFC 3986 rather than from current output, and observed failing on unfixed source:

```
- test_openapi_validator_format_uri_reference: Error
Error: Bug on uri-reference rejects "\"" directly: found different values
{ "x": false, "y": true }        # expected false, observed true
```

After removing the raw quote from all three owners it passes (`test_openapi_validator_format_uri_reference: 12 ms — Success`). The run that first exercised the fix rebuilt the ttsc plugin from the changed `FormatCheatSheet.go`, so the generated side was rebuilt rather than served from cache.

### Zero-collateral evidence

Each owner is exactly its previous content with the 8 `&'"(` → `&'(` substitutions and nothing else:

```
packages/typia/src/internal/_isFormatUriReference.ts               substitutions: 8 | only-that-substitution: true
packages/utils/src/validators/functional/_isFormatUriReference.ts  substitutions: 8 | only-that-substitution: true
packages/typia/native/core/factories/FormatCheatSheet.go           substitutions: 8 | only-that-substitution: true
```

and the three remain byte-identical (`typia src === utils src: true`, `typia src === native go: true`, raw-quote `0 / 0 / 0`, apostrophe classes `10 / 10 / 10` — the 8 repaired plus the `userinfo` and `IPvFuture` classes that were always correct).

### Environment note

`pnpm test:go:native` first failed with `There is not enough space on the disk` — the `C:` volume is full and Go's `$WORK` lives there. Re-run with `GOTMPDIR` on a volume with space; all packages then pass. This is an environment condition on this machine, not a property of the change, and no shared Go cache was cleared.

### Not done here (lead-owned)

`packages/typia/package.json` is **not** bumped although native Go changed and Change Integrity requires a coordinated bump. The lead serializes the native version chain across the concurrent native pull requests and assigns the version at rebase time. This branch is not rebased for the same reason.

samchon:
Final semantic rebase onto current `master`, version assignment, and lead verification.

## Rebase and version

Rebased from `22f00400b5` onto `98f560253e` (#2104). All four commits replayed with **no conflict** — nothing merged in the chain touches the three format owners.

Version assigned as **`13.1.13`**, the next patch on master's `13.1.12` (#2104). The bump is mandatory because `FormatCheatSheet.go` is native Go, per the development skill's Change Integrity rule. `git diff origin/master...HEAD -- packages/typia/package.json` is exactly one line and `git diff --check` is clean.

## Lead verification of the fix itself

The three owners are byte-identical again, checked mechanically rather than by eye — raw-quote character classes now count **0** in `packages/typia/src/internal/_isFormatUriReference.ts`, `packages/utils/src/validators/functional/_isFormatUriReference.ts`, and `FormatCheatSheet.go`.

Runtime proof on the rebased head, built from the shipped source:

| Input | Before | After |
| --- | --- | --- |
| `"`, `a"b`, `/a"b`, `?q="`, `#"`, `http://a"b/c`, `http://x/p"q` | all `true` | all **`false`** |
| `%22`, `a'b`, `http://h/p?q=1#f`, empty, `a/b`, `urn:x:y`, `mailto:a@b.c`, `//[::1]/p` | `true` | `true` — unchanged |

Every raw-quote case is now rejected and every valid reference is preserved, including the percent-encoded spelling `%22` and the apostrophe sub-delim that RFC 3986 does permit. No regression in the accept direction.

## Local verification on `dc9eef8f0c` (campaign CI is deliberately cancelled)

| Gate | Command | Result |
| --- | --- | --- |
| Owning workspace | `pnpm --filter test-utils start` | Success |
| Schema workspace | `pnpm --filter test-typia-schema start` | Success |

## What this thread's Self-Review caught, and why it matters

Round 2 found that **round 1's own remediation was defective**. The issue requires pinning that the pattern must not carry `/g`, but an accept/reject pair cannot distinguish `/i` from `/gi`: a failed `test()` resets `lastIndex`, so the pin had no teeth. It now repeats one accepted value consecutively, proven by injecting `/gi`, watching it fail, and reverting.

That same `lastIndex` hazard independently surfaced on the #2111 batch, where hoisting a `/g` regex out of a per-call literal would have introduced a real bug. Two batches, one trap — worth remembering.

Round 2 also **proved rather than assumed** the three-owner claim: `FormatCheatSheet.go` backs the JSDoc `@format` path through `metadataCommentTagFactory_FORMATS`, and re-injecting the quote into the Go entry alone fails `rejects "\"" through comment tag`. So the native owner has a genuine covering surface rather than being fixed on faith.

## Deferred, recorded for the next discovery round

Review raised that `1bad:relative` is accepted although RFC 3986 forbids `:` in the first segment of a scheme-less relative reference. It is real but **not one-sided**, and the lead verification changes its framing: `iri-reference` rejects `./1bad:relative` and `good/1bad:relative`, which RFC 3986 **permits** (only `segment-nz-nc`, the first segment, forbids the colon; every later `segment` is `*pchar`, and `pchar` includes `:`). So there are two defects in opposite directions across two owners — `uri-reference` under-restricts the first segment while `iri-reference` over-restricts later ones. It is correctly out of scope here, is not quote-related, and is carried into the campaign's next discovery round for proper adjudication rather than bolted onto this batch.

`dc9eef8f0c` settled after three polls with all six runs `cancelled` and none racing to terminal. Earlier on this branch, `871b9a82a6`'s spell-check ended `failure` — that was **not** the benign race but a real defect it caught (`excludeds`, plus literal U+0000/U+007F bytes making the file read as binary), fixed in `f0a6442816`. Worth recording that a spell-check failure is worth reading rather than dismissing as the race.

PR #2114, opened 2026-07-16
fix(utils): make NamingConvention.variable an identifier-legality predicate

### LINKED ISSUE #2111: NamingConvention.variable misclassifies identifiers, leaking illegal bindings into generated HTTP accessors and parameter keys

samchon:
## Intent

Resolves #2111.

`NamingConvention.variable()` is documented as *"Check if string is valid JavaScript variable name"* but was implemented as `reserved(str) === false && /^[a-zA-Z_$][a-zA-Z_$0-9]*$/g.test(str)` — "not in my word list, and looks like an identifier". Those are different predicates, and the gap returned `true` for eight strings that cannot be binding identifiers in an ES module.

**The abstraction was wrong, not just the data.** `eval` and `arguments` are not reserved words at all — they are illegal only as *BoundNames* in strict code — so no reserved-word set can ever classify them. Adding eight strings to `RESERVED` would have left the same category error in place.

This branch models the three restriction classes the grammar actually defines and derives each public predicate from the ones that apply to it:

| Class | Contents | Feeds |
| --- | --- | --- |
| `RESERVED_WORDS` | ECMAScript `ReservedWord` (38, now including `await`/`yield`) | `reserved()`, `variable()` |
| `STRICT_RESERVED_WORDS` | strict future reserved (`implements`, `interface`, `let`, `package`, `private`, `protected`, `public`, `static`) | `reserved()`, `variable()` |
| `POLICY_RESERVED_WORDS` | `module` — typia policy, not grammar | `reserved()`, `variable()` |
| `STRICT_RESTRICTED_BINDINGS` | `eval`, `arguments` — restricted BoundNames | `variable()` only |

- `reserved()` = `ReservedWord` ∪ strict future reserved ∪ policy. This completes a class that already shipped `package`/`private`/`protected`/`public` while omitting `implements`/`interface`/`let`/`static`.
- `variable()` = identifier shape ∧ not `reserved()` ∧ not a restricted binding. It is no longer derivable from a word list alone.

## Deliberate public-API decisions

These are behavior changes to exported surface, stated explicitly rather than slipped in:

1. **`reserved()` gains 6 words** (`await`, `yield`, `implements`, `interface`, `let`, `static`). It previously returned `false` for words that *are* reserved in the strict dialect typia generates.
2. **`reserved()` still excludes `eval`/`arguments`** — deliberately. They are not reserved words; saying otherwise to make `variable()` work would re-introduce the category error. `variable()` rejects them separately.
3. **`module` stays reserved.** It is a legal identifier (`function f(module) {}` parses), so this is typia policy, not a grammar claim: shadowing it would break `module.exports` in CommonJS output. It is now documented and asserted as policy instead of sitting unexplained in a word list.
4. **The identifier pattern loses its `g` flag** and is hoisted to a module constant. A global regex carries `lastIndex` across `test()` calls; hoisting it *with* `/g` would have made the predicate alternate between answers for the same input. A stability assertion pins this.
5. **ASCII-only identifier shape is preserved** and now documented as a deliberate conservative bias — it can only cause a legal Unicode name to be escaped, never let an illegal binding through. Unicode support remains a separate policy decision, per the issue.

## Consequence surface

- **`EndpointUtil.normalize`** keeps gating on `reserved()`, so the 6 added words now escape there (`let` becomes `_let`), exactly as `package` already did. `eval`/`arguments` are not reserved and are caught downstream by the `variable()` gate, so both paths converge on a single `_` escape with no double-escaping.
- **`EndpointUtil.pascal`** consequently derives `_Let` where it derived `Let`, matching the existing `_Package` treatment. This is a generated *type-name* change for paths containing those 6 words.
- **Diagnostic consumers** (`stringifyValidationFailure`, `OpenApiObjectValidator`, `JsonDescriptor`) are intentionally untouched. They ask "can I use dot notation?", where a keyword is legal (`x.let`), so they were never defective. They now emit bracket notation instead of dot notation for these words — more verbose, still correct. Their call sites were left alone rather than quoting every consumer, per the issue's consumer split.
- **Compatibility:** accessors and parameter keys for the eight words change (`let` becomes `_let`), which is a generated-name change for anyone depending on today's broken output.
- **Scope:** `@typia/utils` only. No native Go change, so **no `typia` version bump** (development skill, Change Integrity).

### A latent divergence this does *not* fix

`packages/typia/src/internal/_accessExpressionAsString.ts` carries its own private copy of this predicate with a *different* 36-word set (no `module`/`package`/`public`/`private`/`protected`). The two already disagreed before this change; it widens that gap from 5 words to 13. It is a diagnostic property-access path, so it is not a defect, and `packages/typia` is outside this issue's scope. `test-typia-automated` predicts typia's generated paths through `@typia/utils`'s `variable()`, so the two are coupled — but no `@typia/template` structure uses a restricted word as a key, and the suite passes. **Flagged for the lead as a possible follow-up issue**, not addressed here.

## Verification

Tests were written first and observed failing against the original implementation, then passing — the final test forms were re-confirmed against a stashed revert:

| Test | Fails before on |
| --- | --- |
| `test_naming_convention_variable` | `variable("await")` returns `true`, engine says `false` |
| `test_naming_convention_reserved` | `reserved("await")` returns `false`, expected `true` |
| `test_http_migrate_route_accessor_identifier` | accessor segment `"await"` of `/await/{await}` is not a legal binding |

Expectations come from Node's own parser, not from typia's output. Because this suite does not run with `--experimental-vm-modules`, `vm.SourceTextModule` is unavailable; the helper uses `vm.Script` with `"use strict"` plus an `async` context, which was verified to reproduce module-goal binding legality **exactly across 69 probed words** (strict mode supplies the future-reserved words and the `eval`/`arguments` restriction; the async context supplies the `await` reservation).

Commands run locally (campaign CI is suspended):

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils start` | **pass** (full suite) |
| `pnpm --filter ./packages/utils build` | **pass** |
| `pnpm --filter ./tests/test-utils-automated start` | **pass** |
| `pnpm --filter ./tests/test-typia-automated start` | **pass** (the `variable()`-coupled suite) |

### Real `HttpMigration.application()` probe

Before — 8 of 11 routes emit a declaration that does not parse:

| route | accessor | param key | declaration compiles |
| --- | --- | --- | --- |
| `/let/{let}` | `["let","getByLet"]` | `let` | **NO — SyntaxError** |
| `/static/{static}` | `["static","getByStatic"]` | `static` | **NO — SyntaxError** |
| `/implements/{implements}` | `["implements","getByImplements"]` | `implements` | **NO — SyntaxError** |
| `/interface/{interface}` | `["interface","getByInterface"]` | `interface` | **NO — SyntaxError** |
| `/yield/{yield}` | `["yield","getByYield"]` | `yield` | **NO — SyntaxError** |
| `/await/{await}` | `["await","getByAwait"]` | `await` | **NO — SyntaxError** |
| `/eval/{eval}` | `["eval","getByEval"]` | `eval` | **NO — SyntaxError** |
| `/arguments/{arguments}` | `["arguments","getByArguments"]` | `arguments` | **NO — SyntaxError** |
| `/package/{package}` | `["_package","getBy_package"]` | `_package` | yes |
| `/module/{module}` | `["_module","getBy_module"]` | `_module` | yes |
| `/foo/{foo}` | `["foo","getByFoo"]` | `foo` | yes |

After — 0 of 11:

| route | accessor | param key | declaration compiles |
| --- | --- | --- | --- |
| `/let/{let}` | `["_let","getBy_let"]` | `_let` | yes |
| `/static/{static}` | `["_static","getBy_static"]` | `_static` | yes |
| `/implements/{implements}` | `["_implements","getBy_implements"]` | `_implements` | yes |
| `/interface/{interface}` | `["_interface","getBy_interface"]` | `_interface` | yes |
| `/yield/{yield}` | `["_yield","getBy_yield"]` | `_yield` | yes |
| `/await/{await}` | `["_await","getBy_await"]` | `_await` | yes |
| `/eval/{eval}` | `["_eval","getBy_eval"]` | `_eval` | yes |
| `/arguments/{arguments}` | `["_arguments","getBy_arguments"]` | `_arguments` | yes |
| `/package/{package}` | `["_package","getBy_package"]` | `_package` | yes (unchanged) |
| `/module/{module}` | `["_module","getBy_module"]` | `_module` | yes (unchanged) |
| `/foo/{foo}` | `["foo","getByFoo"]` | `foo` | yes (unchanged) |

`getBy_let` rather than `getByLet` follows the existing `getBy_package` shape, since `getName` capitalizes the already-escaped key.

## Documentation

No `website/src/content/docs/**` page documents `NamingConvention`, `HttpMigration`, or route accessors (verified by grep), so the JSDoc is the public documentation surface for this API and was rewritten to state the contract, the `module` policy, the `eval`/`arguments` split, the property-access caveat, and the ASCII bias. `experiments/esm` exercises only `camel`/`snake` and is unaffected.

## Campaign CI

Automatic CI for this branch is deliberately suspended under the issue campaign's per-push exact-SHA cancellation gate; every pushed SHA had its runs enumerated and cancelled. Local verification above is the gate. See `.agents/skills/issue-campaign/development.md`.

`pnpm format` was deliberately **not** run, per the campaign's no-format rule.

### DIFF
--- a/packages/utils/src/utils/NamingConvention.ts
+++ b/packages/utils/src/utils/NamingConvention.ts
@@ -141,18 +141,18 @@ export namespace NamingConvention {
   /**
    * Check if string is valid JavaScript variable name.
    *
-   * Returns `true` only when `str` can be used as a *binding* identifier — a
-   * declaration, parameter, or import name — in the dialect typia generates:
-   * an ES module, which is always strict mode. A string is a valid binding
-   * when it has identifier shape and is neither {@link reserved} nor one of
-   * the strict-mode restricted binding names (`eval`, `arguments`).
+   * Returns `true` only when `str` can be used as a _binding_ identifier — a
+   * declaration, parameter, or import name — in the dialect typia generates: an
+   * ES module, which is always strict mode. A string is a valid binding when it
+   * has identifier shape and is neither {@link reserved} nor one of the
+   * strict-mode restricted binding names (`eval`, `arguments`).
    *
    * The restricted binding names are why this is not derived from
    * {@link reserved} alone. `eval` and `arguments` are not reserved words at
    * all; they are rejected only in binding position under strict mode, so no
    * reserved-word list can express this predicate.
    *
-   * Property *access* is a weaker requirement than binding, since dot notation
+   * Property _access_ is a weaker requirement than binding, since dot notation
    * accepts reserved words (`x.let` is legal). Callers escaping a property key
    * rather than declaring a name are therefore held to a stricter rule than
    * they need; that is safe, but they cannot rely on this predicate to mean
@@ -178,8 +178,8 @@ export namespace NamingConvention {
    * it is a legal identifier, but shadowing it would break `module.exports` in
    * CommonJS output.
    *
-   * `eval` and `arguments` are deliberately absent. They are not reserved
-   * words — they are merely illegal as binding names in strict code — so
+   * `eval` and `arguments` are deliberately absent. They are not reserved words
+   * — they are merely illegal as binding names in strict code — so
    * {@link variable} rejects them separately.
    *
    * @param str String to check
@@ -191,12 +191,12 @@ export namespace NamingConvention {
 /**
  * Identifier shape, ASCII subset.
  *
- * Declared without the `g` flag on purpose: a global regular expression
- * carries `lastIndex` across `test()` calls, which would make
- * {@link NamingConvention.variable} alternate between answers for the same
- * input once the pattern is hoisted out of the function body.
+ * Declared without the `g` flag on purpose: a global regular expression carries
+ * `lastIndex` across `test()` calls, which would make
+ * {@link NamingConvention.variable} alternate between answers for the same input
+ * once the pattern is hoisted out of the function body.
  */
-const IDENTIFIER = /^[a-zA-Z_$][a-zA-Z_$0-9]*$/;
+const IDENTIFIER: RegExp = /^[a-zA-Z_$][a-zA-Z_$0-9]*$/;
 
 /** ECMAScript `ReservedWord`. */
 const RESERVED_WORDS: string[] = [
--- a/packages/utils/src/utils/NamingConvention.ts
+++ b/packages/utils/src/utils/NamingConvention.ts
@@ -12,7 +12,8 @@
  * - {@link pascal}: Convert to PascalCase (`FooBar`)
  * - {@link snake}: Convert to snake_case (`foo_bar`)
  * - {@link kebab}: Convert to kebab-case (`foo-bar`)
- * - {@link variable}: Test if string is valid JavaScript variable name
+ * - {@link variable}: Test if string is a legal binding identifier
+ * - {@link reserved}: Test if string is a JavaScript reserved word
  *
  * @author Jeongho Nam - https://github.com/samchon
  */
@@ -140,14 +141,46 @@ export namespace NamingConvention {
   /**
    * Check if string is valid JavaScript variable name.
    *
+   * Returns `true` only when `str` can be used as a *binding* identifier — a
+   * declaration, parameter, or import name — in the dialect typia generates:
+   * an ES module, which is always strict mode. A string is a valid binding
+   * when it has identifier shape and is neither {@link reserved} nor one of
+   * the strict-mode restricted binding names (`eval`, `arguments`).
+   *
+   * The restricted binding names are why this is not derived from
+   * {@link reserved} alone. `eval` and `arguments` are not reserved words at
+   * all; they are rejected only in binding position under strict mode, so no
+   * reserved-word list can express this predicate.
+   *
+   * Property *access* is a weaker requirement than binding, since dot notation
+   * accepts reserved words (`x.let` is legal). Callers escaping a property key
+   * rather than declaring a name are therefore held to a stricter rule than
+   * they need; that is safe, but they cannot rely on this predicate to mean
+   * "needs bracket notation".
+   *
+   * Identifier shape is restricted to ASCII. JavaScript also admits Unicode
+   * identifiers, so this may report `false` for an otherwise legal name. The
+   * bias is deliberate: it can only cause a name to be escaped, never to leak
+   * an illegal binding.
+   *
    * @param str String to check
    * @returns True if valid variable name
    */
   export const variable = (str: string): boolean =>
-    reserved(str) === false && /^[a-zA-Z_$][a-zA-Z_$0-9]*$/g.test(str);
+    IDENTIFIER.test(str) && RESTRICTED.has(str) === false;
 
   /**
    * Check if string is JavaScript reserved word.
+   *
+   * Covers every ECMAScript `ReservedWord` plus the future reserved words that
+   * apply in strict mode, since typia generates ES modules, which are always
+   * strict. `module` is reserved by typia policy rather than by the language:
+   * it is a legal identifier, but shadowing it would break `module.exports` in
+   * CommonJS output.
+   *
+   * `eval` and `arguments` are deliberately absent. They are not reserved
+   * words — they are merely illegal as binding names in strict code — so
+   * {@link variable} rejects them separately.
    *
    * @param str String to check
    * @returns True if reserved word
@@ -155,7 +188,19 @@ export namespace NamingConvention {
   export const reserved = (str: string): boolean => RESERVED.has(str);
 }
 
-const RESERVED: Set<string> = new Set([
+/**
+ * Identifier shape, ASCII subset.
+ *
+ * Declared without the `g` flag on purpose: a global regular expression
+ * carries `lastIndex` across `test()` calls, which would make
+ * {@link NamingConvention.variable} alternate between answers for the same
+ * input once the pattern is hoisted out of the function body.
+ */
+const IDENTIFIER = /^[a-zA-Z_$][a-zA-Z_$0-9]*$/;
+
+/** ECMAScript `ReservedWord`. */
+const RESERVED_WORDS: string[] = [
+  "await",
   "break",
   "case",
   "catch",
@@ -178,13 +223,8 @@ const RESERVED: Set<string> = new Set([
   "import",
   "in",
   "instanceof",
-  "module",
   "new",
   "null",
-  "package",
-  "public",
-  "private",
-  "protected",
   "return",
   "super",
   "switch",
@@ -197,6 +237,53 @@ const RESERVED: Set<string> = new Set([
   "void",
   "while",
   "with",
+  "yield",
+];
+
+/**
+ * Future reserved words that apply only in strict mode.
+ *
+ * Generated ES modules are always strict, so these are reserved for every name
+ * typia emits.
+ */
+const STRICT_RESERVED_WORDS: string[] = [
+  "implements",
+  "interface",
+  "let",
+  "package",
+  "private",
+  "protected",
+  "public",
+  "static",
+];
+
+/**
+ * Reserved by typia policy rather than by the language.
+ *
+ * `module` is a legal JavaScript identifier, but a generated name that shadows
+ * it would break `module.exports` in CommonJS output.
+ */
+const POLICY_RESERVED_WORDS: string[] = ["module"];
+
+/**
+ * Names that are illegal in binding position under strict mode without being
+ * reserved words.
+ *
+ * These are the reason {@link NamingConvention.variable} cannot be a
+ * reserved-word lookup: `x.eval` and `const { eval: e } = x` are fine, while
+ * `function f(eval) {}` is a `SyntaxError` in strict code.
+ */
+const STRICT_RESTRICTED_BINDINGS: string[] = ["eval", "arguments"];
+
+const RESERVED: Set<string> = new Set([
+  ...RESERVED_WORDS,
+  ...STRICT_RESERVED_WORDS,
+  ...POLICY_RESERVED_WORDS,
+]);
+
+const RESTRICTED: Set<string> = new Set([
+  ...RESERVED,
+  ...STRICT_RESTRICTED_BINDINGS,
 ]);
 
 const unsnake =
--- a/tests/test-utils/src/features/migrate/test_http_migrate_route_accessor_identifier.ts
+++ b/tests/test-utils/src/features/migrate/test_http_migrate_route_accessor_identifier.ts
@@ -1,0 +1,121 @@
+import { TestValidator } from "@nestia/e2e";
+import { IHttpMigrateApplication } from "@typia/interface";
+import { HttpMigration } from "@typia/utils";
+import { OpenApi } from "typia";
+
+import {
+  _isLegalBinding,
+  _isLegalDeclaration,
+} from "../../internal/_isLegalBinding";
+
+/**
+ * Verifies HttpMigration.application derives only legal binding identifiers for
+ * accessors and parameter keys.
+ *
+ * `IHttpMigrateRoute.accessor` and each parameter `key` are documented as the
+ * generated RPC accessor and parameter variable name, and downstream SDK
+ * generators emit them directly into an ES module. Because both are derived
+ * from `NamingConvention.variable`, that predicate's false positives (#2111)
+ * turned legal OpenAPI paths such as `/let/{let}` into `getByLet(connection,
+ * let)` — a real `SyntaxError`, not a hypothetical. A set-only unit test
+ * cannot prove this contract, so each route's emitted declaration is compiled.
+ *
+ * 1. Build one route per restricted word, plus controls that must not change.
+ * 2. Assert every accessor segment and parameter key is a legal binding.
+ * 3. Compile the SDK-shaped declaration each route implies.
+ * 4. Assert the already-escaped and ordinary routes keep their exact names.
+ */
+export const test_http_migrate_route_accessor_identifier = (): void => {
+  const words: string[] = [
+    // the eight #2111 false positives
+    "await", "yield", "let", "static", "implements", "interface", "eval",
+    "arguments",
+    // words already escaped before #2111 -- must not regress
+    "package", "private", "protected", "public", "case", "class", "enum",
+    "module",
+    // ordinary names -- must stay untouched
+    "foo", "async", "of", "get",
+  ];
+  const document: OpenApi.IDocument = {
+    openapi: "3.2.0",
+    "x-typia-emended-v12": true,
+    paths: Object.fromEntries(
+      words.map((word) => [
+        `/${word}/{${word}}`,
+        {
+          get: {
+            parameters: [
+              {
+                name: word,
+                in: "path" as const,
+                required: true,
+                schema: { type: "string" as const },
+              },
+            ],
+          },
+        },
+      ]),
+    ),
+    components: {},
+  };
+  const app: IHttpMigrateApplication = HttpMigration.application(document);
+  TestValidator.equals("every route migrated", app.routes.length, words.length);
+
+  for (const route of app.routes) {
+    // 2. EVERY DERIVED NAME IS A LEGAL BINDING
+    for (const accessor of route.accessor)
+      TestValidator.equals(
+        `accessor segment ${JSON.stringify(accessor)} of ${route.path} is a legal binding`,
+        _isLegalBinding(accessor),
+        true,
+      );
+    for (const parameter of route.parameters)
+      TestValidator.equals(
+        `parameter key ${JSON.stringify(parameter.key)} of ${route.path} is a legal binding`,
+        _isLegalBinding(parameter.key),
+        true,
+      );
+
+    // 3. THE DECLARATION THE ROUTE IMPLIES MUST COMPILE
+    //
+    // This is the contract that actually breaks: the accessor becomes the
+    // function name and each parameter key becomes one of its parameters.
+    // Compiling the whole declaration also proves the keys do not collide.
+    TestValidator.equals(
+      `generated declaration for ${route.path} compiles`,
+      _isLegalDeclaration({
+        name: route.accessor.at(-1)!,
+        parameters: ["connection", ...route.parameters.map((p) => p.key)],
+      }),
+      true,
+    );
+  }
+
+  // 4. CONTROLS KEEP THEIR EXACT NAMES
+  const accessorOf = (path: string): string[] =>
+    app.routes.find((r) => r.path === path)!.accessor;
+  const keyOf = (path: string): string =>
+    app.routes.find((r) => r.path === path)!.parameters[0]!.key;
+
+  TestValidator.equals("ordinary route is untouched", accessorOf("/foo/{foo}"), [
+    "foo",
+    "getByFoo",
+  ]);
+  TestValidator.equals("ordinary parameter key is untouched", keyOf("/foo/{foo}"), "foo");
+  TestValidator.equals("contextual keyword is untouched", accessorOf("/async/{async}"), [
+    "async",
+    "getByAsync",
+  ]);
+  TestValidator.equals(
+    "already-escaped route is unchanged",
+    accessorOf("/package/{package}"),
+    ["_package", "getBy_package"],
+  );
+  TestValidator.equals("newly escaped route", accessorOf("/let/{let}"), [
+    "_let",
+    "getBy_let",
+  ]);
+  TestValidator.equals("newly escaped parameter key", keyOf("/let/{let}"), "_let");
+  TestValidator.equals("eval is escaped", keyOf("/eval/{eval}"), "_eval");
+  TestValidator.equals("arguments is escaped", keyOf("/arguments/{arguments}"), "_arguments");
+};
--- a/tests/test-utils/src/features/migrate/test_http_migrate_route_accessor_identifier.ts
+++ b/tests/test-utils/src/features/migrate/test_http_migrate_route_accessor_identifier.ts
@@ -3,10 +3,8 @@ import { IHttpMigrateApplication } from "@typia/interface";
 import { HttpMigration } from "@typia/utils";
 import { OpenApi } from "typia";
 
-import {
-  _isLegalBinding,
-  _isLegalDeclaration,
-} from "../../internal/_isLegalBinding";
+import { _isLegalBinding } from "../../internal/_isLegalBinding";
+import { _isLegalDeclaration } from "../../internal/_isLegalDeclaration";
 
 /**
  * Verifies HttpMigration.application derives only legal binding identifiers for
@@ -17,8 +15,8 @@ import {
  * generators emit them directly into an ES module. Because both are derived
  * from `NamingConvention.variable`, that predicate's false positives (#2111)
  * turned legal OpenAPI paths such as `/let/{let}` into `getByLet(connection,
- * let)` — a real `SyntaxError`, not a hypothetical. A set-only unit test
- * cannot prove this contract, so each route's emitted declaration is compiled.
+ * let)` — a real `SyntaxError`, not a hypothetical. A set-only unit test cannot
+ * prove this contract, so each route's emitted declaration is compiled.
  *
  * 1. Build one route per restricted word, plus controls that must not change.
  * 2. Assert every accessor segment and parameter key is a legal binding.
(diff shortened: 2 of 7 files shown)


samchon:
## Self-Review — Round 1 (findings before remediation)

Complete solo round over the entire `origin/master...HEAD` diff (5 files, +503/-8): source, tests, helper, docs, consumers, packaging, and platform behavior. Findings first, per the campaign rule; remediation follows in the next push.

### Findings to act on

**1. `IDENTIFIER` is the only untyped constant** — *style, `NamingConvention.ts`*

Every sibling declares its type (`RESERVED_WORDS: string[]`, `RESERVED: Set<string>`); `IDENTIFIER` does not. Add `: RegExp`.

**2. `_isLegalBinding.ts` exports two symbols** — *convention, `tests/test-utils/src/internal/`*

Every one of the 12 peer helpers under `tests/test-typia-automated/src/internal/` exports exactly one symbol named after its file. This file exports both `_isLegalBinding` and `_isLegalDeclaration`, so the filename names only half its surface. Split into `_isLegalBinding.ts` and `_isLegalDeclaration.ts`, with the shared `compiles()` private to each or hoisted honestly.

**3. Four of five new/changed files are not prettier-conformant** — *`pnpm exec prettier --check` on the five paths*

```
[warn] packages/utils/src/utils/NamingConvention.ts
[warn] tests/test-utils/src/features/naming/test_naming_convention_variable.ts
[warn] tests/test-utils/src/features/naming/test_naming_convention_reserved.ts
[warn] tests/test-utils/src/features/migrate/test_http_migrate_route_accessor_identifier.ts
```

The campaign forbids running the **repository-wide** formatter, and I will not run `pnpm format` (`prettier --write "**/*.ts"` + `format:go`). But the development skill's base rule — *"never commit unformatted output"* — is not suspended, and the campaign carve-out exists to keep repo-wide churn out of issue PRs. Running prettier against **only the five paths this branch touches** honors both: zero unrelated churn, and no unformatted output handed to Post-Campaign Cleanup. Flagging the reasoning explicitly so it can be objected to.

**4. Already-escaped parameter keys are swept but never pinned** — *coverage, `test_http_migrate_route_accessor_identifier`*

`case`, `class`, `enum`, `private`, `protected`, and `public` are in the probe set, but step 4 pins exact names only for `package`, `foo`, `async`, and `let`. The issue's acceptance criteria require that *currently-handled keywords stay unchanged*; a legality sweep cannot prove "unchanged", only "legal". Add an exact-name assertion for a pre-existing keyword's parameter key so a regression in the untouched path is caught by name.

### Verified as *not* defects (recorded so the next round need not re-derive them)

**`EndpointUtil.normalize` still gating on `reserved()` is correct.** I traced every binding position reachable from `HttpMigrateRouteAccessor` and `HttpMigrateRouteComposer` for a position gated *only* by `reserved()`, since `reserved()` deliberately excludes `eval`/`arguments`:

| Position | Gate | Safe? |
| --- | --- | --- |
| `HttpMigrateRouteAccessor:23-24` namespace segments | `normalize()` then `variable()` | yes — `variable()` is the final gate |
| `HttpMigrateRouteAccessor:33-34` `entry.alias` | `normalize()` only | yes — `getName()` derives it from an HTTP method (`delete` is already remapped to `erase`), so it always begins with a non-reserved verb |
| `HttpMigrateRouteComposer:427-435` parameter `key` | `normalize()` then a `variable()` loop | yes |

No binding position is gated by `reserved()` alone, so `eval`/`arguments` cannot leak. Switching `normalize()` to `variable()` would have been actively **wrong**: it double-escapes (`foo bar` → `_foo bar` at `normalize`, then `__foo bar` at the accessor gate) for names `_`-prefixing cannot rescue.

**`x-samchon-accessor` remains ungated** (`HttpMigrateRouteAccessor:54-56` assigns it raw). Pre-existing, user-supplied, and explicitly "keep working / unchanged" in the issue. Untouched here; noting it exists rather than silently expanding scope.

**`Set` rather than an object literal** means `variable("constructor")`, `variable("prototype")`, and `variable("__proto__")` cannot be corrupted by prototype keys. Swept against the engine and passing.

**`test_naming_convention_reserved` restating ECMA-262's word list is justified.** No runtime oracle exists for "is this a reserved word" — an engine can only report *binding* legality, which conflates the reserved-word and BoundNames classes that this change deliberately separates. The list is the spec, written independently, and every word in it is additionally cross-checked against the engine by `test_naming_convention_variable`.

samchon:
## Self-Review — Round 2 (findings before remediation)

Fresh complete round over the whole `origin/master...HEAD` diff, re-derived from the current state rather than rechecking Round 1's fixes. Two findings to act on, one of which is a defect in my own **verification** rather than in the code.

### 1. My post-commit fail-first verification silently no-opped — *verification integrity*

Round 1's claim table said each test "fails before". After the implementation was committed, I re-ran that check with `git stash push packages/utils/src/utils/NamingConvention.ts`. With the file already committed and clean, **`git stash push` has nothing to save and exits without reverting anything** — so the test ran against the *fixed* implementation and "passed", which I would have read as a green fail-first check. It surfaced because a newly added test reported `Success` where it had to fail.

The two *earlier* fail-first checks were genuine — the implementation was still uncommitted then, and the stash reported `Saved working directory` (the before/after `HttpMigration` probe tables in the PR body came from that working revert). But the method is unsound once committed, so every fail-first claim is re-established with a method proven to actually revert:

```
git checkout origin/master -- packages/utils/src/utils/NamingConvention.ts
# sanity: grep -c STRICT_RESTRICTED_BINDINGS => 0  (proves the revert landed)
```

Result against the `origin/master` implementation — all four fail:

| Test | Against origin/master |
| --- | --- |
| `test_naming_convention_variable` | **Failed** |
| `test_naming_convention_reserved` | **Failed** |
| `test_http_migrate_route_accessor_identifier` | **Failed** |
| `test_http_migrate_route_parameter_key_escape` | **Failed** |

No product impact; the fix and tests were already correct. Recording it because an unverified fail-first claim is worthless, and this one was.

### 2. The `variable()`/duplicate escaping interaction is untested — *coverage*

`HttpMigrateRouteComposer` escapes with a single loop where **both gates share one counter**:

```ts
while (NamingConvention.variable(key) === false || usedParameterKeys.has(key))
  key = `_${key}`;
```

Widening the identifier gate moves `let` from the second iteration to the first — precisely where an off-by-one collision would appear. The issue's acceptance criteria name *"duplicate escaping, numeric-leading segments"* as boundaries, and nothing in the branch covered them. Adding `test_http_migrate_route_parameter_key_escape`, pinning verified behavior:

| route | keys | note |
| --- | --- | --- |
| `/a/{let}/b/{_let}` | `_let`, `__let` | reserved word beside its escaped spelling — must not both settle on `_let` |
| `/e/{eval}/f/{_eval}` | `_eval`, `__eval` | restricted binding through the same loop |
| `/c/{9foo}` | `_9foo` | numeric-leading, predates #2111, unchanged |
| `/d/{connection}` | `_connection` | collides with the receiver `escapeDuplicate` reserves, unchanged |

Each declaration is compiled, so a collision fails as a duplicate-parameter `SyntaxError` rather than a silent equality mismatch. Verified fail-first (see table above).

### Verified as *not* defects

**Only path parameters are affected — scope confirmed by construction.** The issue lists *"path, header, query, cookie, and body keys"*. Traced: `HttpMigrateRouteComposer` assigns `key: "body"` (L612/633/651/667) and `key: type` for `headers`/`query`/`cookies` (L131) as **fixed literals** — they never reach `variable()`. Only path parameter keys (L427-435) and accessor segments (`HttpMigrateRouteAccessor:24`) are gated by it. Those non-path keys flow through `escapeDuplicate` only, which this change does not touch. So the affected surface is exactly path keys + accessor segments, both covered.

**`prettier --check` warnings on this checkout are a Windows `autocrlf` artifact, not unformatted output.** `core.autocrlf=true` and there is no `.gitattributes`, so any `git checkout` materializes CRLF while prettier defaults to `endOfLine: "lf"`. Files this branch never touched (`MapUtil.ts`, `EndpointUtil.ts`) warn identically, and every committed blob is LF and content-conformant:

```
NamingConvention.ts / _isLegalBinding.ts / _isLegalDeclaration.ts
test_naming_convention_variable.ts / test_naming_convention_reserved.ts
test_http_migrate_route_accessor_identifier.ts / ..._parameter_key_escape.ts
  => all "conformant" when prettier is diffed against the committed blob
```

Noting it so a later round does not "fix" a phantom, and so Post-Campaign Cleanup is not surprised.

**Diagnostic consumers behave as intended.** Re-read all six call sites. `OpenApiObjectValidator` (L27/48/83) and `stringifyValidationFailure` (L176/196) build human-readable access paths, and `JsonDescriptor` (L58) builds a `{@link ...}` doc string. For the eight words these switch from `x.let` to `x["let"]` — correct in both spellings, and already the behavior for `module`/`case`/`class` before this change, so no new class of output appears. Left untouched per the issue's consumer split.

**Broader consumer suites pass.** Beyond the batch's required commands, every suite that consumes `@typia/utils` was run: `test-langchain`, `test-mcp`, `test-vercel`, and `test-typia-schema` — all `Success`.

samchon:
## Self-Review — Round 3 (clean round; Self-Review complete)

Fresh complete round over the whole `origin/master...HEAD` diff from the current state: source, both internal helpers, all four tests, consumers, docs, packaging, CI, platform behavior, and compatibility. **No improvement survived verification, so this round is clean and Self-Review ends here.**

Round 3 spent its effort trying to *break* the change rather than re-reading Rounds 1-2, so the two claims below are now evidence rather than assertion.

### The predicate is exhaustively correct, not just correct on the words I chose

A curated 69-word sweep only proves the predicate on words I already thought of — precisely the bias that produced #2111 (the original set omitted `implements`/`interface`/`let`/`static` while shipping their four siblings). So `variable()` was fuzzed against Node's parser across **every lowercase-alpha string of length 1-4**:

```
exhaustive lowercase-alpha probes (len 1-4): 475,254
mismatches (excluding the deliberate `module` policy): 0
module policy still asserted: true   engine says legal: true
```

**475,254 probes, zero mismatches.** Every short keyword lives in that space; the longer restricted words (`implements`, `interface`, `protected`, `arguments`, `constructor`, …) are covered by the curated sweep in `test_naming_convention_variable`. `module` remains the single, intentional, documented departure. The fuzz is a one-off verification, not a committed test — 475k `vm.Script` compiles do not belong in a suite that runs in seconds.

### The `EndpointUtil.pascal` consequence is real, and it is a generated-artifact change

The PR body claims `pascal` derives `_Let` where it derived `Let`. Verified against a real migration rather than reasoned about — these are **generated OpenAPI component schema names**:

| path | generated component | status |
| --- | --- | --- |
| `/let/items` | `IApi_LetItems.PostBody` | **changed** (was `IApiLetItems.PostBody`) |
| `/package/items` | `IApi_PackageItems.PostBody` | unchanged (pre-existing) |
| `/foo/items` | `IApiFooItems.PostBody` | unchanged |

Flagging this explicitly under the development skill's Change Integrity rule: a generated baseline is part of the specification. The product reason is that `reserved()` must be correct for the strict dialect typia generates, which #2111 establishes as an oversight rather than a policy; the result makes `let` consistent with the `package` treatment that already shipped.

**Deliberately not pinned by a test.** No test pins this surface today — `IApi_PackageItems` has no coverage either — so adding a pin only for the newly-changed case would leave an odd asymmetry on an adjacent, out-of-scope surface. Verified and disclosed is the honest disposition; pinning generated component names is its own concern.

### Also checked in this round, no action

- **`new vm.Script()` compiles but never executes.** Both helpers parse only; neither calls `runInContext`. Building code strings introduces no execution path.
- **`NamingConvention.localize("")` still crashes** (`str[0]!.toLowerCase()` on an empty string). Untouched on purpose — the issue names it an independent defect tracked separately in this campaign, and folding it in here would blur two fixes.
- **Packaging is untouched:** no `packages/utils/package.json` change, no version bump, no new dependency (`node:vm` is a builtin used only by test helpers). `pnpm --filter ./packages/utils build` passes.
- **Platform:** pure string logic plus `node:vm`; nothing platform-specific. Verified on Windows.
- **Test isolation:** every test builds its own `OpenApi.IDocument`; both helpers are pure.
- **The four intermediate word arrays** (`RESERVED_WORDS`, `STRICT_RESERVED_WORDS`, `POLICY_RESERVED_WORDS`, `STRICT_RESTRICTED_BINDINGS`) exist only to compose two `Set`s. Kept: they *are* the abstraction fix — each names a distinct restriction class with a distinct rationale, and collapsing them back into one bag is what made `eval`/`arguments` unclassifiable in the first place.

### Final local verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils start` | **pass** — 288 tests, all 4 new ones executing |
| `pnpm --filter ./packages/utils build` | **pass** |
| `pnpm --filter ./tests/test-typia-automated start` | **pass** (`variable()`-coupled) |
| `pnpm --filter ./tests/test-utils-automated start` | **pass** |
| `pnpm --filter ./tests/test-typia-schema start` | **pass** |
| `pnpm --filter ./tests/test-langchain start` | **pass** |
| `pnpm --filter ./tests/test-mcp start` | **pass** |
| `pnpm --filter ./tests/test-vercel start` | **pass** |

All four tests re-confirmed fail-first against the `origin/master` implementation using `git checkout origin/master -- <path>`, with a `grep` sanity check proving the revert landed.

Ready for the lead. Not merging — the lead owns the merge for this batch.

samchon:

## Rebase

Rebased onto current `master` `8ea0011add` (#2113). All four commits replayed with no conflict. **No version bump**, correctly: the diff contains no native Go, so the development skill's coordinated-bump rule does not apply, and this batch merges outside the campaign's native-version serial chain.

## Independent Self-Review of the whole diff: clean

The change repairs the abstraction rather than topping up a word list, which is what #2111 actually asked for:

- `variable()` is now identifier shape ∧ ¬`RESTRICTED`, where `RESTRICTED` = `RESERVED` ∪ `{eval, arguments}`. `reserved()` stays `RESERVED` only. That split is the whole point: `eval` and `arguments` are **not** reserved words, so no reserved-word lookup could ever have expressed "legal binding identifier". Keeping them out of `reserved()` while rejecting them in `variable()` keeps both predicates semantically honest instead of making one lie to serve the other.
- The word lists are split into four separately named and separately justified classes — `RESERVED_WORDS`, `STRICT_RESERVED_WORDS` (correct for typia, which emits always-strict ES modules), `POLICY_RESERVED_WORDS` (just `module`, with the CommonJS `module.exports` shadowing reason stated), and `STRICT_RESTRICTED_BINDINGS`. A future reader can tell why each word is in the set, which is exactly what the old flat 41-word bag could not say.
- `IDENTIFIER` is declared without `/g`, with the reason recorded on the constant: hoisting a `/g` pattern out of a per-call literal would carry `lastIndex` across `test()` calls and make the predicate alternate answers for the same input. Worth noting that the identical `lastIndex` trap independently bit the #2113 batch, where an accept/reject pair could not distinguish `/i` from `/gi` because a failed `test()` resets `lastIndex`. Two batches, one hazard.
- The JSDoc states the binding-versus-property-access boundary (`x.let` is legal, so callers escaping a property key are held to a stricter rule than they need) and that the ASCII-only shape is a deliberate bias that can only over-escape, never leak an illegal binding. Both are the kind of constraint the code cannot show by itself.

## Lead verification on `dc585b51ed`

| Gate | Command | Result |
| --- | --- | --- |
| Focused | `pnpm --filter test-utils start -- --include naming_convention` | `test_naming_convention_variable` and `test_naming_convention_reserved` pass |
| Owning workspace | `pnpm --filter test-utils start` | Success, 5,011 ms |
| Package build | `pnpm --filter @typia/utils build` | pass |

The batch's own evidence is stronger than a suite pass and I am recording it rather than repeating it: a real `HttpMigration.application()` probe emitted `SyntaxError`-producing declarations on 8 of 11 routes before and 0 after, and an exhaustive differential of 475,254 probes against Node's own parser found zero mismatches.

## What this batch's Self-Review caught about itself

Round 2 found that **its own verification was unsound**: a post-commit `git stash push` on a clean file silently no-ops, so one fail-first claim had actually been observed against the *fixed* code. All four were then re-established properly with `git checkout origin/master --`. An agent auditing the integrity of its own evidence, rather than its code, is the failure mode that a green suite can never surface — this campaign's CI is deliberately cancelled, so unsound local evidence is the one thing that could carry a defect through.

## Recorded for follow-up, not fixed here

1. **Generated component names shift** for the six newly-reserved words (`/let/items` -> `IApi_LetItems.PostBody`, mirroring the pre-existing `IApi_PackageItems`). That rename *is* the fix, but it is user-visible. Deliberately not pinned, since that surface has no coverage today and a one-sided pin would be asymmetric.
2. **A second, divergent copy of this predicate** lives at `packages/typia/src/internal/_accessExpressionAsString.ts` (36 words). It already disagreed with `@typia/utils` on five words; this widens the gap to thirteen. Not a defect today — it is a diagnostic path, and `x.let` is legal there — but `test-typia-automated` predicts typia's paths through utils' `variable()`, so the two are coupled and pass only because no `@typia/template` structure uses a restricted word as a key. Carried into the campaign's next discovery round as a latent-divergence candidate rather than widened into this batch.

PR #2115, opened 2026-07-16
fix(tests): repair feature-test function identity

### LINKED ISSUE #2112: Feature-test files export test functions that do not match their filenames, breaking selection and report identity
## Problem

Five feature-test files export a `test_*` function whose name does not match the file's basename. Because `DynamicExecutor` keys both its `--include`/`--exclude` filter and its report entries on the **exported name** rather than the filename, these files cannot be selected by their own names, and three exported names are defined twice inside one suite — two of them colliding with a real, correctly-named sibling file in the same directory.

This is a repository-integrity defect in the test suites, not a product bug. No test body is silently dropped (see Evidence), but test selection and reporting identity are broken, and it violates the development skill's rule that a test file contains one test named after what it asserts.

Affected users: contributors running or filtering the `test-typia-schema` and `test-utils` suites, and anyone reading a name-keyed test report.

## Evidence

Scanning all 426 tracked `tests/*/src/features/**/test_*.ts` files for `export const|function test_*` whose name differs from the basename yields exactly five:

| File | Exported identity |
| --- | --- |
| `tests/test-typia-schema/src/features/llm.application/test_llm_application_custom_valiidate.ts` | `test_llm_application_custom_validate` |
| `tests/test-typia-schema/src/features/llm.application/test_llm_controller_custom_valiidate.ts` | `test_llm_application_custom_validate` |
| `tests/test-utils/src/features/llm/coerce/test_llm_coerce_json_stringify_overlapped.ts` | `test_llm_coerce_json_strignify_overlapped` |
| `tests/test-utils/src/features/llm/http/test_http_llm_application_human.ts` | `test_http_llm_application` |
| `tests/test-utils/src/features/llm/http/test_http_llm_function_tags.ts` | `test_http_llm_function_deprecated` |

Note the two directions of error: `valiidate`/`strignify` are typos on the **filename** side, while `_human`/`_tags` are wrong on the **export** side (the export names a different case than the file asserts).

Three exported names are consequently defined by two files **within one suite**:

| Export name | Defined by | Suite |
| --- | --- | --- |
| `test_llm_application_custom_validate` | `test_llm_application_custom_valiidate.ts` + `test_llm_controller_custom_valiidate.ts` | test-typia-schema, same directory |
| `test_http_llm_application` | `test_http_llm_application.ts` + `test_http_llm_application_human.ts` | test-utils, same directory |
| `test_http_llm_function_deprecated` | `test_http_llm_function_deprecated.ts` + `test_http_llm_function_tags.ts` | test-utils, same directory |

The last two collide with a real, correctly-named sibling. (Three further cross-suite duplicates — `test_llm_schema_empty_required`, `test_llm_schema_enum`, `test_llm_schema_nullable` — are legitimate: matching filenames in two different suites and processes.)

Reproduction (no build required):

```bash
git ls-files 'tests/*/src/features/**/test_*.ts' | while read f; do
  base=$(basename "$f" .ts)
  grep -qE "export (const|function|async function) $base\b" "$f" || echo "MISMATCH: $f"
done
```

Observed: the five files above are reported. Expected: no output.

**Severity, verified against the installed runner.** `@nestia/e2e@10.0.2` `lib/DynamicExecutor.js` iterates `Object.entries(modulo)` per file, filters on `key.substring(0, prefix.length) !== props.prefix` and `props.filter(key)`, then does `report.executions.push({ name: key, location, ... })`. Because executions are **pushed to an array** rather than keyed into a map, no test body is dropped or shadowed. The real defects are:

1. **Filter identity is keyed on the export name, not the filename.** `--include test_llm_controller_custom_valiidate` selects nothing; `--include test_http_llm_application` silently runs two tests, one of which is the `_human` case.
2. **Report identity collides.** Two executions report the same `name` at different `location`s, indistinguishable in any name-keyed view.

## Consequence surface

- **Suites:** `test-typia-schema` and `test-utils` only — both function-per-file `DynamicExecutor` suites. The generated matrix suites (`test-typia-automated`, `test-utils-automated`) are unaffected.
- **Consumers:** anyone using `--include`/`--exclude` selection, and any name-keyed report or dashboard over the execution list.
- **Not affected:** product source, generated artifacts, and test outcomes. Every test body still runs today.
- **Case matrix for the cause:** filename-side typo (2 files), export-side wrong name (2 files), export-side typo (1 file). All five are hand-written mistakes with no shared code path.

## Approach

The invariant: **within a suite, each `tests/*/src/features/**/test_*.ts` exports exactly one `test_*` function whose name equals the file's basename, and no two files in one suite export the same name.**

Rename each of the five to satisfy the invariant, correcting the misspelled side (the filename for `valiidate`/`strignify`; the export for `_human`/`_tags`, which must name the case the file actually asserts). Then add a cheap backstop that enforces the invariant for the whole class permanently, so this cannot regress silently again. The backstop belongs to the feature-suite tree, not to product source.

- **Positive:** every `tests/*/src/features/**/test_*.ts` exports a `test_*` function matching its basename.
- **Negative:** no two files within one suite export the same `test_*` name; the three legitimate cross-suite duplicates remain allowed.
- **Boundary:** the backstop must not fire on the cross-suite pairs, and must fail on a deliberately introduced mismatch.
- **Regression:** `--include <name>` selects exactly the file of that basename for each of the five repaired cases.
- **Commands:** the reproduction loop above (must print nothing); `pnpm --filter ./tests/test-typia-schema start` and `pnpm --filter ./tests/test-utils start` for the renamed suites.

## Coordination

- **Complementary to, not a duplicate of, #2035.** #2035 (*Automated tests skip direct matrices and mask asynchronous failures*) cites only `tests/test-typia-automated/src/build/TestAutomationController.ts`, `tests/test-typia-automated/src/index.ts`, and `tests/test-utils-automated/src/index.ts`, and owns the **generated**-tree naming/count backstop. None of the five files here lives in a generated suite, the causes are unrelated (generator control flow vs hand-written typos), and #2035's harness backstop cannot detect a typo in a function-per-file suite. Folding this into #2035 would strand a minutes-long fix behind that issue's thirteen prerequisite batches.
- No product change and no package version bump.
- Safe to parallelize with any product work; touches only the two test suites.

samchon:
## Intent

Repairs the five feature-test files whose exported `test_*` function name does not match their filename, and adds a permanent backstop so the class cannot regress silently.

Closes #2112.

## Scope

Tests only. No product source, no package version bump.

### The five repairs

The error is on a different side per file, so the repair differs per file. `_human` and `_tags` were copy-paste of a sibling's export, so the corrected name comes from reading what each body actually asserts:

| File | Misspelled side | Repair |
| --- | --- | --- |
| `test-typia-schema/.../test_llm_application_custom_valiidate.ts` | filename | rename to `..._custom_validate.ts`; export already correct |
| `test-typia-schema/.../test_llm_controller_custom_valiidate.ts` | filename **and** export subject | rename to `..._custom_validate.ts`; export becomes `test_llm_controller_custom_validate` — the body exercises `typia.llm.controller`, not `llm.application`, which is why it collided with its sibling |
| `test-utils/.../coerce/test_llm_coerce_json_stringify_overlapped.ts` | export (`strignify`) | export becomes `test_llm_coerce_json_stringify_overlapped` |
| `test-utils/.../http/test_http_llm_application_human.ts` | export | body asserts an `x-samchon-human` operation is dropped ⇒ `test_http_llm_application_human` |
| `test-utils/.../http/test_http_llm_function_tags.ts` | export | body asserts `func.tags` ⇒ `test_http_llm_function_tags` |

No test body, assertion, or import changed. Nothing referenced the five names outside their own files.

### The backstop: `tests/test-feature-identity`

A new workspace holding the invariant: **a `test_`-named file exports exactly one `test_*` function equal to its basename, a helper file exports none, and no two files of one suite export the same name.**

**Why a workspace, and why there.** It had to run in CI unchanged and had to attribute failures correctly:

- `tests/*` is a pnpm workspace glob and root `test:packages` starts every `./tests/test-*`, so it runs under `pnpm test` and the `test` workflow (whose `paths` already include `tests/**`) with no root or workflow edit.
- One implementation covers **all** suites, including ones added later. A per-suite backstop file would duplicate the scanner five times and silently lose coverage the day someone adds a suite without remembering to copy it.
- Putting it in an existing suite would make `test-utils` fail because `test-vercel` had a typo. Its own workspace owns the repository-level concern.
- It reads sources as **text** via `git ls-files` and imports nothing from another suite, so it respects the "don't reach into another suite's internals" rule.

**Why git, not a directory walk.** The generated matrix suites gitignore `src/features` and rebuild it per run, and #2035 owns their generator-side naming. Reading the tracked tree excludes them **by construction rather than by a hardcoded suite name**, and makes the result independent of which suites ran first.

The scan is fail-closed: only the top-level `export const|let|var|function` forms count, so an exotic export reads as "this file exports nothing" — a reported violation — rather than passing unchecked.

Cases: `test_feature_identity_repository` (the live tree), `_collect` (a throwaway git repo holding a non-ASCII name, a helper, a `.d.ts`, a file outside `features`, and a staged-then-deleted file), `_filename_mismatch`, `_duplicate_export`, `_helper_file`, `_source_parse`. It runs on explicit imports rather than `DynamicExecutor`, since a discovery-driven suite that tests discovery could hide its own failure.

Both skill docs' suite inventories are updated to match.

## Verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-feature-identity start` **before** the repairs | **Fails** — the 5 mismatches + the 3 intra-suite collisions, and nothing else |
| `pnpm --filter ./tests/test-feature-identity start` **after** the repairs | **Success** — 6 cases |
| `pnpm --filter ./tests/test-typia-schema start` | **Success** |
| `pnpm --filter ./tests/test-utils start` | **Success** |

`--include` selection is measured in [this comment](https://github.com/samchon/typia/pull/2115#issuecomment-4995088828), including one honest deviation from #2112's wording: each suite's filter is a **substring** match, not equality, so `--include test_http_llm_application` still selects its four `_human`/`_version`/`_function_name_length` siblings. What the repair delivers is that all four now report **distinct** names, and each of the five repaired names resolves to exactly one file — four of which previously selected **nothing**.

The three legitimate cross-suite duplicates (`test_llm_schema_empty_required`, `test_llm_schema_enum`, `test_llm_schema_nullable`) keep working; the uniqueness rule is scoped per suite and `test_feature_identity_duplicate_export` pins that boundary.

## Campaign CI

Issue-campaign implementation pull request. Automatic CI is deliberately suspended per `.agents/skills/issue-campaign/development.md`: every pushed SHA's runs are cancelled by the exact-SHA gate and local verification is recorded here instead. `spell-check` is fast enough to settle before the cancel lands; it caught a real `honours`/`honors` violation of `typos.toml`'s `locale = 'en-us'` on `3319b7e`, which is fixed, and it passes on the final SHA. `pnpm format` is deliberately not run; Post-Campaign Cleanup owns the repository-wide formatter result.


samchon:
## Self-Review — round 1 findings

Complete round over the `origin/master...HEAD` diff (backstop suite, five repairs, skill docs). Findings first; remediation follows in the next push.

All six findings are in the **backstop I wrote**, not in the five repairs. That is the right place for the scrutiny: the repairs are four one-line renames whose correctness the backstop attests, so a hole in the backstop is the only way this PR can claim more than it delivers.

### 1. `collect()` silently skips non-ASCII feature paths — fail-open (correctness)

`git ls-files` applies `core.quotePath` by default. Verified in a scratch repo:

```
$ git ls-files -- tests
"tests/test-x/src/features/test_\303\251.ts"
$ git ls-files -z -- tests | tr '\0' '\n'
tests/test-x/src/features/test_é.ts
```

`FEATURE_PATH` anchors on `^tests/`, so a quoted line never matches and that file **drops out of the scan entirely**. No such file exists today, so this is latent — but a check whose entire value proposition is "no feature file escapes the invariant" must not ship with a silent escape hatch. It also directly contradicts the fail-closed property the `parse` doc comment advertises.

**Fix:** `git ls-files -z` (which never quotes) and split on `\0`.

### 2. `collect()` throws ENOENT on an unstaged deletion (correctness)

`git ls-files` reads the **index**; `readFileSync` reads the **working tree**. Delete a feature file without staging it and the suite dies with an unhandled ENOENT instead of reporting anything. That is a normal mid-refactor state.

The working tree is what `DynamicExecutor` actually requires, so a file absent from disk is genuinely out of scope rather than a violation.

**Fix:** skip index entries with no working-tree file.

### 3. `execFileSync` inherits the 1 MB default `maxBuffer` (robustness)

`git ls-files -- tests` emits ~180 KB today. Growth past 1 MB would surface as an obscure ENOBUFS inside a naming check. **Fix:** set an explicit `maxBuffer`.

### 4. `describe()` prints "1 test functions" (diagnostic quality)

The helper-file branch can report a single export. A backstop's whole output is its diagnostic; ungrammatical output is a defect in the deliverable. **Fix:** singular/plural.

### 5. A `test_*.d.ts` would be a false positive (robustness)

`FEATURE_PATH` matches `.+\.ts$`, so `test_foo.d.ts` would yield basename `test_foo.d` and be reported as "exports no test_ function". None exist; excluding declaration files is cheap and stops a future contributor hitting a nonsense diagnostic.

### 6. Three lines exceed the 80-column prettier width (style)

`test_feature_identity_duplicate_export.ts:49`, `test_feature_identity_filename_mismatch.ts:24`, `test_feature_identity_source_parse.ts:57`.

Campaign rules forbid `pnpm format` here, so these get hand-formatted to keep the Post-Campaign Cleanup diff empty for the new files. The long `index.ts` imports are **not** a finding: prettier leaves single-specifier imports unbroken, matching the existing `test-typia-exact-optional/src/index.ts`.

### Not findings (checked, deliberately kept)

- **Cross-suite duplicates stay legal.** `test_llm_schema_empty_required`, `test_llm_schema_enum`, `test_llm_schema_nullable` exist in two suites; the uniqueness rule is scoped per suite and stays silent on them. Pinned by `test_feature_identity_duplicate_export`.
- **The generated matrices are out of scope by construction,** not by a hardcoded suite name: they gitignore `src/features`, so reading through git excludes them regardless of run order. #2035 owns their generator-side backstop.
- **`PlainNativeClone.ts`** is a real non-`test_` helper in a feature tree; the rule permits it and pins the permission.
- **The suite uses explicit imports, not `DynamicExecutor`.** A discovery-driven suite that tests discovery could hide its own failure.

samchon:
## Self-Review — round 2 findings

Round 1's six findings are remediated, and its `collect()` fixes are now pinned by a new `test_feature_identity_collect` case that builds a throwaway git repository holding the awkward shapes the live tree does not have (non-ASCII name, declaration file, unstaged deletion, file outside `features`). Fresh full round over the updated `origin/master...HEAD` diff follows.

### 1. The new collect case leaks git's warnings into the suite log (observed)

Its `git init` / `git add` calls inherit stderr, so the run prints:

```
  - test_feature_identity_repository: 305 ms
warning: in the working copy of 'tests/test-alpha/src/features/Helper.ts', LF will be replaced by CRLF the next time Git touches it
... (5 more)
  - test_feature_identity_collect: 300 ms
```

Six warning lines interleaved between two passing cases, from a **temporary repository that is deleted moments later**. A test log is read to tell pass from failure; noise that looks like a problem but is not trains the reader to skim past the lines that are. **Fix:** `stdio: "pipe"` on the fixture's git calls.

### 2. `FeatureIdentity.git()` discards git's stderr on failure

Its catch rethrows `(error as Error).message`, which for `execFileSync` is the exit-status line — the actual git diagnostic sits in `error.stderr` and is dropped. The helper exists to fail loudly and comprehensibly; today it fails loudly and mutely. **Fix:** pipe stderr and include it in the thrown message.

### 3. The collect case's mismatch assertion can pass for the wrong reason

Both `test_alpha_ordinary.ts` and `test_alpha_é.ts` take the `write` default body, so both export `test_alpha_stub`. That makes them an intra-suite **duplicate**, and the duplicate diagnostic names both paths — so `diagnose(files).some(line => line.includes("test_alpha_é.ts"))` would hold even if the mismatch rule never fired. The assertion does not test what its title claims.

**Fix:** give the ordinary file its own matching export. The fixture then yields exactly **one** diagnostic — the unicode mismatch — so the case can assert that count exactly, which is both stricter and simpler than a substring probe.

### 4. Three lines still exceed 80 columns (style)

`FeatureIdentity.ts:171`, `test_feature_identity_collect.ts:33` and `:60`. Same reasoning as round 1 finding 6.

### Verification status

- `test-feature-identity`: fails on the unrepaired tree with the 8 expected diagnostics, passes after the repairs.
- `test-typia-schema`: **Success**, 142 cases.
- `test-utils`: running; `--include` selection checks follow.

samchon:
## Self-Review — round 3 findings

Round 2's four findings are remediated. Fresh full round over the updated diff.

### 1. `spell-check` failed on this branch — real, and fixed

The suspended-CI gate cancels campaign runs, but `spell-check` is fast enough to finish before the cancel lands, so it reported for real on `3319b7e`:

```
error: `honours` should be `honors`
  ╭▸ ./tests/test-feature-identity/src/FeatureIdentity.ts:117:56
```

`typos.toml` pins `locale = 'en-us'`. Fixed, and the rest of the new suite is clean of British spellings.

Worth recording for the campaign: **`typos.toml` carries no allowlist entry for `valiidate` or `strignify`.** They survived on master because `typos` only reports words in its correction corpus and neither is in it — so removing them strands no dead config, and spell-check was never going to catch this class. The backstop is the only thing that does.

### 2. The collector is documented as reading *committed* sources — it reads the *index* (accuracy)

`git ls-files` lists **tracked** files, which includes a file staged but never committed. My own new fixture proves the contradiction: it calls `git add -A` and never commits, and `collect()` returns its files.

The behavior is right — a staged file is about to land and should be checked — but the word is wrong in 10 places across `FeatureIdentity.ts`, `test_feature_identity_repository.ts`, and both skill docs. A comment that misdescribes the trust boundary of a check is worse than none: the next reader reasons from it. **Fix:** say "tracked".

### 3. Non-findings, checked this round

- **`test_alpha_é` and Unicode normalization.** On macOS the filesystem stores NFD; git's `core.precomposeunicode` (default on) hands back NFC, so the comparison holds. If it ever did not, the `collected` path-list equality fails first with a readable diff rather than the `!` assertion throwing — the failure mode is diagnosable, so the fixture stays.
- **`git init` needs no `user.name`/`user.email`**: the fixture stages and never commits.
- **`FEATURE_PATH.exec("")`** on the trailing empty element of the `\0` split returns null and is filtered.
- **`FeatureIdentity.ts:171`** stays over 80 columns: it is a single regex literal, which prettier cannot break.

## Verification (complete)

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-feature-identity start` (before repairs) | **Fails** — 8 diagnostics: the 5 mismatches + the 3 intra-suite collisions |
| `pnpm --filter ./tests/test-feature-identity start` (after repairs) | **Success** — 6 cases |
| `pnpm --filter ./tests/test-typia-schema start` | **Success** |
| `pnpm --filter ./tests/test-utils start` | **Success** |

### `--include` selection, measured

`test-typia-schema -- --include test_llm_application_custom_validate test_llm_controller_custom_validate`:

```
  - test_llm_application_custom_validate: 1 ms
  - test_llm_controller_custom_validate: 0 ms
```

Two distinct names, one file each. Before: `test_llm_controller_custom_valiidate` selected **nothing**, and `test_llm_application_custom_validate` selected **two executions reporting the same name**.

`test-utils -- --include test_http_llm_application test_llm_coerce_json_stringify_overlapped test_http_llm_function_tags`:

```
  - test_llm_coerce_json_stringify_overlapped: 5 ms
  - test_http_llm_application: 43 ms
  - test_http_llm_application_function_name_length: 772 ms
  - test_http_llm_application_human: 8 ms
  - test_http_llm_application_version: 5 ms
  - test_http_llm_function_tags: 4 ms
```

**One nuance worth stating plainly rather than claiming the issue's wording verbatim.** #2112's regression criterion reads "`--include <name>` selects exactly the file of that basename". That is not literally reachable for a name that prefixes its siblings, because each suite's filter is `include.some((str) => name.includes(str))` — a **substring** match, not equality. So `--include test_http_llm_application` still selects four tests. That is a pre-existing property of the suites' filter and is outside this issue.

What the repair actually delivers, and what the run above proves:

| Name | Before | After |
| --- | --- | --- |
| `test_llm_controller_custom_validate` | 0 (unreachable) | 1 |
| `test_llm_coerce_json_stringify_overlapped` | 0 (unreachable) | 1 |
| `test_http_llm_function_tags` | 0 (unreachable) | 1 |
| `test_http_llm_application_human` | 0 (unreachable) | 1 |
| `test_llm_application_custom_validate` | 2 executions, **identical** names | 1 |
| `test_http_llm_application` (substring) | 4, two reporting the **same** name | 4, all **distinct** |

Every one of the five is now reachable by its own name, and no execution shares a name with another in its suite — which is the invariant, and what the backstop enforces.

samchon:
## Self-Review — round 4: clean

Round 3's two findings are remediated (`honours` → `honors`; "committed" → "tracked" across ten comments and both skill docs). A fresh complete round over the whole `origin/master...HEAD` diff produced **no finding that survives verification**. Self-Review is therefore complete: three rounds with findings, one clean round.

Checked again this round, nothing to change:

- **Fresh-checkout CI.** `actions/checkout` leaves a populated index, and `git ls-files` reads the index, so `fetch-depth: 1` is irrelevant. The lockfile carries the new importer, and `tests/*` is a workspace glob, so `pnpm install` and `test:packages` pick the suite up with no root or workflow edit.
- **`ttsx` with no `ttsc` dependency** matches every sibling suite and is exercised by the runs above.
- **`tsconfig.json` sets `plugins: []`** because the suite declares no `typia` dependency and never calls `typia.*`; it would otherwise fail to resolve the shared config's `typia/lib/transform` entry.
- **Parallel-safe.** `pnpm -r` may start suites concurrently; the fixture uses `mkdtempSync`, so each run owns its scratch repository.
- **Windows cleanup.** `git` marks objects read-only, which is the platform where `fs.rmSync` most plausibly fails. The fixture was run on Windows and cleans up; CI is Linux, so the harsher case is the verified one.
- **Duplicate diagnostics per file are intentional.** A mismatched file that also collides reports both the identity and the uniqueness violation — two real consequences of one typo, and the pair is exactly what the pre-repair run printed.

### Change Integrity disclosures

- **New package wiring:** adds the `tests/test-feature-identity` workspace and the corresponding `pnpm-lock.yaml` importer (`@nestia/e2e`, `@types/node`; both already in the `utils` catalog). No product dependency and no version bump.
- **Skill docs edited:** `.agents/skills/project/SKILL.md` and `.agents/skills/development/SKILL.md` enumerate the test workspaces, so leaving the new suite out would have made them wrong. While there, "seven shapes covering all twelve suites" became "these shapes" — that count was **already stale before this branch** (`test-typia-bundler-cache` is a thirteenth suite the list never picked up), and a total that must be edited on every suite addition rots faster than the list it introduces. `test-typia-bundler-cache` is deliberately **not** documented here: it is not this issue's to describe.

Ready for the lead. Not merging — #2112 is a lead-merge batch.

samchon:

## Independent Self-Review of the whole `origin/master...HEAD` diff: clean

- All five repairs verified directly: each file's exported `test_*` now equals its basename, and the two collisions with a real same-directory sibling (`test_http_llm_application`, `test_http_llm_function_deprecated`) are gone. The misspelled side was corrected per case rather than uniformly — filename for `valiidate`, export for `strignify`/`_human`/`_tags`.
- No file under `packages/` is touched; `pnpm-lock.yaml` gains only the new workspace importer with catalog deps. No product change and no version bump, so this batch is correctly outside the campaign's native-version serial chain.
- The skill-doc edits are sound and overdue: the removed *"seven shapes covering all twelve suites"* count was already stale on `master` before this branch.

## Backstop, run by the lead on `2d94219878`

```
pnpm --filter test-feature-identity start
  - test_feature_identity_repository: 947 ms
  - test_feature_identity_collect: 232 ms
  - test_feature_identity_filename_mismatch: 1 ms
  - test_feature_identity_duplicate_export: 0 ms
  - test_feature_identity_helper_file: 1 ms
  - test_feature_identity_source_parse: 0 ms
Success
```

The design choices hold up on review: reading `git ls-files` rather than the disk excludes the generated matrices **by construction** instead of by a hardcoded suite name, so a future generated suite cannot silently fall into scope; and an independent workspace means one scanner covers every suite including ones that do not exist yet.

## Corrections this batch made to the issue, both accepted

1. **The reproduction glob in #2112's body under-reached.** It scanned 426 files across two suites, not the full 479 across six, because it required a subdirectory under `features/`. A corrected scan finds exactly the same five mismatches and three intra-suite duplicates, so the finding set and acceptance criteria stand. Recorded on the issue as a chronology comment.
2. **One acceptance line was not literally reachable.** `--include <name>` is a `name.includes(str)` substring match, so it can never select exactly one sibling by name. That is pre-existing and outside this issue. The delivered and measured criterion is the one the rest of the body states: all five names now resolve to exactly one file and every execution reports a distinct name.

Round 3 of the author's Self-Review found the most valuable defect in this batch, and it was in the backstop itself rather than the repairs: `git ls-files` quotes non-ASCII paths, so the anchored pattern would have silently dropped those files — a fail-open hole in a check whose entire value is that nothing escapes it. Fixed with `-z` and pinned by a fixture.

PR #2118, opened 2026-07-16
feat(website): retheme the site to a deep blue and white two-tone

samchon:
Retheme the website to a blue/white two-tone, following [ttsc.dev](https://ttsc.dev)'s composition and implementation one shade deeper, so the sibling sites read as related rather than identical.

## The palette

| Token | Value | Role |
|---|---|---|
| `--typia-blue` | `#235fa9` | primary accent, navbar, sidebar active, links |
| `--typia-blue-deep` | `#154279` | gradient start, hover, emphasis |
| `--typia-blue-mid` | `#2d73be` | gradient end |
| `--typia-blue-ink` | `#0d1f36` | body text, and code-panel background |
| `--typia-blue-soft` | `#e5effa` | tinted section background |
| `--typia-blue-wash` | `#f6fafd` | subtle fills, cards |
| `--typia-blue-border` | `#c0d4ed` | hairlines |
| `--typia-muted` | `#50647c` | secondary text |

ttsc's blue is `#3178c6` (hsl 211/60%/48%); typia's is `#235fa9` (hsl 213/66%/40%) — same family, clearly deeper.

`--typia-blue` is the **exact hex of the HSL triple** passed to Nextra's `<Head color={{hue: 213, saturation: 66, lightness: 40}}>`. Nextra derives its own primary ramp from that HSL while our CSS uses the hex, and the two must agree. ttsc maintains these independently and they have already drifted there — `hsl(211,60%,48%)` is actually `#3178c4`, not the `#3178c6` it declares. Deriving one from the other avoids inheriting that failure mode; the note in `global.css` says so.

Contrast (WCAG): ink on white 16.6:1, blue on white 6.4:1, white on blue 6.4:1, muted on white 6.1:1, navbar text on the band 9.6:1.

## Not a hue swap — an inversion

The site was **forced dark** (`defaultTheme: "dark"`, `--nextra-bg: 17,17,17`). It is now forced light, and the landing sections — which hardcoded `rgba(255,255,255,*)` for every piece of text — moved onto the palette. Section rhythm: blue hero band → white → soft → white → soft → white.

**Code panels deliberately stay dark ink.** That contrast against the light page is what makes the site read as *two*-tone; ttsc does the same. `HomeCodeHighlight` is already tuned for dark panels and is untouched.

## Fixes found along the way

Each of these was a real defect surfaced by the retheme, not incidental cleanup:

- **The landing never had a real full bleed.** Nextra lays the page out as `[sidebar spacer | sidebar | toc | article]`; the spacer is a `w-64 h-0` div that keeps its 256px even when the sidebar is hidden, and the toc balanced it on the right. Hiding only the toc left the article 256px off-centre and overflowing 240px past the viewport. Both gutters now go, and `HomeLayout`'s `-15rem` negative-margin pull — which existed only to escape the old content column — goes with them.
- **No MUI `ThemeProvider` existed at all.** MUI ran its stock light palette while the page rendered dark, which is why every movie hardcoded `rgba(255,255,255,*)` and why the hero's contained button was MUI blue `#1976d2` rather than the site blue. It now binds to the same tokens as `global.css`.
- **The hero logo would have vanished.** `logo.png` is a blue-gradient wordmark on transparency; its ink measures **1.05–1.96:1** against the blue band. It is knocked out to white with `filter: brightness(0) invert(1)` — the pixel-dissolve silhouette is what carries the mark and survives the flatten. A purpose-made white asset would be better; this is CSS-only and needs no new file.
- **The OpenCollective backer strip** draws near-white avatar plates (`#e6f3ff` letter circles, `#f0f1f2` guest circles) that relied on the old dark background; they measure **~1.13:1** on white. The strip gets a tinted plate.
- **`blog.css` mixed its card background from `--nextra-bg`**, which is now `255,255,255` — so `color-mix(rgb(var(--nextra-bg)) 92%, white 8%)` mixed white into white. Its six `.dark`-prefixed rules were dead under `forcedTheme` and are removed.
- **`HomeInnovationMovie` built a header tint as `` `${props.borderColor}15` ``**, an 8-digit hex-alpha suffix that only yields valid CSS for hex literals. With the old `rgba(...)` values it produced `rgba(255,100,100,0.25)15`, which was silently dropped — those header bars had been rendering with no tint at all. Hex values now make the suffix resolve.

## Verification

Driven in a real browser (Playwright) against `next dev`, at 1440×900 and 390×844:

- **Landing, docs, blog index, blog post, playground** — all render on the palette; computed values confirm `--nextra-primary-hue: 213deg`, `--nextra-bg: 255,255,255`, `html.light`, and the navbar gradient resolving to `rgb(21,66,121) → rgb(35,95,169) → rgb(45,115,190)`.
- **Layout geometry probed, not eyeballed** — `article` is x=0 w=1440 (was x=256 w=1184 off-centre); no horizontal overflow at 390px on either landing or docs.
- `npx tsc --noEmit -p website/tsconfig.json` — clean.
- `npx next build` — compiles and prerenders every route successfully.

**Honestly reported gap:** the documented website prerequisite (`pnpm build` inside `website/`) could **not** be completed locally. It fails at the WASM step with a Go interface mismatch:

```
have Run(string, []string) int
want Run(*host.PluginInvocation) int
go build ... ./cmd/playground
```

This is **pre-existing local drift, not this change** — I reproduced it identically on pristine `master`, and this branch touches no Go, no build scripts, and no package wiring (diff vs master is 15 files, all under `website/src`). The sibling `../ttsc` checkout has uncommitted edits and a newer host API than typia's playground command implements. Root `pnpm install`, `pnpm build`, and `pnpm package:tgz` all succeeded; only the wasm/playground compile is blocked. CI checks out a matching ttsc, so it should pass there — worth watching.

## Deferred

- `HomeBenchmarkMovie.tsx`, `HomeStrengthSectionMovie.tsx`, and `ProductHeroLayout.tsx` still hold dark-theme colors. All three are **dead code** — nothing imports them (verified by grep) — so they are left alone rather than adding diff noise. They will need converting if ever revived.
- The playground itself stays dark; it is `@ttsc/playground`'s own IDE surface and consistent with code panels staying dark.
- A purpose-made white/knockout `logo.png` variant would beat the CSS filter on the hero.

https://claude.ai/code/session_01Pq8sozdFV7CKziiApEbPjq


samchon:
CI: `build` passed (12m15s) — which resolves the open question in the PR body. The local WASM failure was confirmed to be sibling-`../ttsc`-checkout drift only; CI's matching ttsc compiles the playground wasm fine, so the full website build is green on the runner.

`spell-check` failed on two comments this change added — the repo runs `crate-ci/typos` with `locale = 'en-us'` and I wrote the British "off-centre" in `global.css` and `HomeLayout.tsx`. Fixed in 25cc50cbbb; swept the rest of the diff for other Briticisms (colour/behaviour/initialise) and found none.

samchon:
Two more changes landed on this branch.

**MUI is gone** — `@mui/material`, `@mui/icons-material`, `@mui/system`, `@emotion/react`, `@emotion/styled`, and 535 lockfile lines. The landing moves to Tailwind v4 (already installed, and how the sibling ttsc site builds its landing); the docs move to Nextra's own `Callout`. Rendering is unchanged — it's a dependency swap. `HomeBenchmarkMovie`, `HomeStrengthSectionMovie` and `ProductHeroLayout` are deleted: nothing imported them, and they existed only to hold MUI code that never rendered.

**Added `public/images/concepts/transform.svg`** — the landing's input → output comparison as a static figure, embedded in the README and the docs introduction. `build/aot-svg.js` generates it from the same code constants and tokenizer the landing uses, so it can't quietly drift from the page.

Two traps worth recording for future conversions:

- MUI resolves `borderRadius` against `shape.borderRadius` (4px), not the 8px spacing scale — so `borderRadius: 2` is 8px (`rounded-lg`), and a Card with no explicit radius is 4px (`rounded-sm`). The ×2 spacing rule does **not** apply there.
- A CSS grid item defaults to `min-width: auto`, so the long code lines in the LLM panel set the track's floor and pushed the page 55px past a 390px viewport. MUI's `Grid` capped that via flex-basis/max-width. Only caught by measuring `scrollWidth` in a real browser — `tsc` and `next build` were both green with the bug present.

Verified: tsc clean, `next build` prerenders + exports all 74 pages, and landing/docs/blog/playground driven in Chromium at 1440×900 and 390×844 with no horizontal overflow.

PR #2119, opened 2026-07-16
fix(typia): skip zero-length unknown length-delimited protobuf fields exactly

### LINKED ISSUE #2116: Protobuf decoders reject valid messages containing an unknown zero-length length-delimited field
## Problem

`typia.protobuf.decode` rejects **valid** Protocol Buffer messages that contain an unknown length-delimited field of length zero. `_ProtobufReader.skip(0)` is overloaded: `0` is the sentinel meaning "skip one varint", but `skipType` forwards the decoded length directly for the `LEN` wire type. A legal zero-length unknown field therefore triggers the varint-skip branch, which consumes a byte belonging to the **next** field and desynchronizes the parser.

This is the mirror image of #2083: that issue is about accepting truncated payloads; this one is about rejecting well-formed ones. Unknown fields are the core of Protobuf schema evolution, so this breaks forward compatibility — the property unknown-field skipping exists to provide.

Affected users: anyone calling `typia.protobuf.decode` / `assertDecode` / `createDecode` on a message produced by a newer schema that added an optional string, bytes, message, or repeated field left empty.

## Evidence

`packages/typia/src/internal/_ProtobufReader.ts` on master (`22f00400b5`):

```ts
public skip(length: number): void {
  if (length === 0) while (this.u8() & 0x80);   // sentinel: skip one varint
  else {
    if (this.index() + length > this.size())
      throw new Error("Error on typia.protobuf.decode(): buffer overflow.");
    this.ptr += length;
  }
}

public skipType(wireType: ProtobufWire): void {
  switch (wireType) {
    case ProtobufWire.VARIANT:
      this.skip(0);                 // deliberate sentinel use
      break;
    case ProtobufWire.LEN:
      this.skip(this.uint32());     // defect: uint32() legally returns 0
      break;
    ...
  }
}
```

`skip(0)` has two callers with two different intents. `VARIANT` means the sentinel; `LEN` means "advance by this many bytes", and zero is a legal length.

Reproduction with a faithful replica of the reader's arithmetic (no build required):

```js
class R {
  constructor(buf){ this.buf=buf; this.ptr=0; }
  index(){ return this.ptr; } size(){ return this.buf.length; }
  u8(){ return this.buf[this.ptr++]; }
  uint32(){ let r=0,s=0,b; do { b=this.u8(); r |= (b&0x7f)<<s; s+=7; } while(b&0x80); return r>>>0; }
  skip(length){
    if (length === 0) while (this.u8() & 0x80);
    else { if (this.index()+length > this.size()) throw new Error("buffer overflow"); this.ptr += length; }
  }
  skipType(w){ switch(w){ case 0: this.skip(0); break; case 1: this.skip(8); break; case 2: this.skip(this.uint32()); break; case 5: this.skip(4); break; } }
}
const r = new R(Uint8Array.of(0x00, 0x0a));  // zero-length LEN field, then the next field's tag
r.skipType(2);
console.log(r.index());   // observed: 2   expected: 1
```

Observed `index() === 2`; spec-correct is `1`. The reader consumed `0x0a`, the tag byte of the following field.

End-to-end consequence, on a **valid** message:

```ts
typia.protobuf.decode<{ value?: string }>(Uint8Array.of(0x12, 0x00, 0x0a, 0x01, 0x61));
```

Wire content: field `2`/`LEN` length `0` (an unknown, empty field), then field `1`/`LEN` length `1` = `"a"`. Expected `{ value: "a" }`. Observed: throws `Error on typia.protobuf.decode(): buffer overflow.` — because the zero-length skip ate the `0x0a` tag, after which `0x01` is misread as a tag and the parse runs off the end.

The `VARIANT` case is correct and must keep working: a varint has no length prefix, so the sentinel is the only way to skip it.

## Consequence surface

- **Owner:** `packages/typia/src/internal/_ProtobufReader.ts` — `skip` / `skipType`. TypeScript runtime only; the generated decoders call this reader.
- **Wire types reaching the defect:** `LEN` only. `VARIANT` (sentinel, correct), `I64` (`skip(8)`), `I32` (`skip(4)`) pass non-zero constants. `START_GROUP` recurses through `skipType`, so a zero-length `LEN` member inside a group inherits the same defect.
- **Field kinds that legally encode a zero length:** empty `string`, empty `bytes`, an empty embedded message, and an empty packed repeated field. All are legal proto3 and all are reachable through explicit presence and schema evolution.
- **Public API:** `protobuf.decode`, `protobuf.assertDecode`, `protobuf.validateDecode`, `protobuf.isDecode`, and `protobuf.createDecode` inherit it, since every generated decoder skips unknown fields through `skipType`.
- **Compatibility:** strictly a repair. Messages that decode today keep decoding identically; only currently-rejected valid payloads begin to succeed. No payload that should be rejected becomes accepted.
- **Relationship to #2083:** distinct root cause and opposite direction. #2083 (PR #2084) makes truncated length-delimited records fail at their declared boundary; this defect makes a correctly-bounded zero-length record corrupt the stream. Verified behaviourally unchanged by PR #2084 — master and that branch reach the same throw by the same path.

| Case | Expected |
| --- | --- |
| unknown zero-length `LEN` field followed by another field | skip exactly the field; decode the rest |
| unknown zero-length `LEN` as the last field | decode succeeds |
| unknown zero-length `LEN` nested in a group | skip exactly the field |
| unknown non-zero `LEN` field | unchanged (already correct) |
| unknown `VARIANT` field | unchanged — sentinel varint skip must keep working |
| truncated `LEN` field (length exceeds the buffer) | reject (owned by #2083) |

## Approach

The invariant: **`skipType` must advance the cursor by exactly the bytes the wire type occupies, for every legal length including zero.**

The architectural fault is the overloaded parameter, not the arithmetic. `skip` currently multiplexes "skip a varint" and "advance N bytes" through one `number`. Separate the two intents so a legal length of zero can never be read as a sentinel — for example an explicit varint-skipping operation for the `VARIANT` case, leaving `skip(n)` to mean only "advance n bytes" with its bounds check intact for every `n >= 0`. Do not special-case zero at the `LEN` call site; that would leave the trap in place for the next caller.

- **Positive:** unknown zero-length `LEN` fields — empty string, empty bytes, empty embedded message, empty packed repeated — are skipped exactly, both mid-message and as the final field, and the surrounding fields decode correctly.
- **Negative:** a `LEN` field whose length exceeds the remaining buffer is still rejected (the #2083 contract must not regress).
- **Boundary:** unknown `VARIANT` fields still skip one varint, including multi-byte continuations; groups containing a zero-length `LEN` member skip correctly.
- **Regression:** a round-trip through `_ProtobufWriter` of a message carrying an empty string/bytes field, decoded by a reader that does not know the field.
- **Commands:** `pnpm --filter ./tests/test-typia-schema start` (narrow); `pnpm --filter ./tests/test-typia-automated start` for the generated decode matrix; `pnpm test:go:public` for the contract tree.

## Coordination

- TypeScript runtime only. Precedent #2066 (Protobuf varints) was likewise runtime-only, so no `packages/typia/package.json` version bump is required unless the final diff touches native Go.
- Serialize against PR #2084 (#2083), which owns `_ProtobufReader` bounds checking and touches the same file. Land after it and rebase; the two fixes are complementary and must not relabel each other's errors.
- Found while reviewing PR #2084; deliberately excluded from that batch because it is pre-existing on master, behaviourally unchanged by it, and a distinct root cause.

samchon:
## Intent

Fixes #2116.

`typia.protobuf.decode` rejects **valid** Protocol Buffer messages containing an unknown length-delimited field of length zero. `_ProtobufReader.skip(0)` is overloaded: `0` is a sentinel meaning "skip one varint" (used by the `VARIANT` case), but `skipType` forwards the decoded length directly for `LEN`, where zero is a legal length. The sentinel branch then consumes a byte belonging to the *next* field and desynchronizes the parser, so a well-formed payload throws `buffer overflow`.

Unknown fields are the mechanism of Protobuf schema evolution, so this breaks the exact forward-compatibility property that unknown-field skipping exists to provide.

## Scope

The invariant: **`skipType` must advance the cursor by exactly the bytes the wire type occupies, for every legal length including zero.**

The fault is the overloaded parameter, not the arithmetic. Per the issue's Approach, the two intents are separated so a legal length of zero can never be read as a sentinel — rather than special-casing zero at the `LEN` call site, which would leave the trap for the next caller.

Sits on top of the `_ProtobufReader` design merged by PR #2084 (`end` boundary, `take()`, `validate()`, `atomic()`, `fork()`, `close()`) and does not disturb PR #2090's `string()` decoder boundary.

## Verification

Recorded on the thread once implementation lands. Campaign CI is deliberately suspended; the exact-SHA cancellation gate is passed after every push, and local verification is the gate.

### DIFF
--- a/packages/typia/src/internal/_ProtobufReader.ts
+++ b/packages/typia/src/internal/_ProtobufReader.ts
@@ -104,10 +104,15 @@ export class _ProtobufReader {
     this.end = previous;
   }
 
+  /** Advance by exactly `length` bytes, for every length including zero. */
   public skip(length: number): void {
+    this.take(length);
+  }
+
+  /** Advance by exactly one varint, which carries no length prefix. */
+  public skipVarint(): void {
     this.atomic(() => {
-      if (length === 0) while (this.u8() & 0x80);
-      else this.take(length);
+      while (this.u8() & 0x80);
     });
   }
 
@@ -115,7 +120,7 @@ export class _ProtobufReader {
     this.atomic(() => {
       switch (wireType) {
         case ProtobufWire.VARIANT:
-          this.skip(0);
+          this.skipVarint();
           break;
         case ProtobufWire.I64:
           this.skip(8);
--- a/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_zero_length_unknown_field.ts
+++ b/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_zero_length_unknown_field.ts
@@ -1,0 +1,178 @@
+import typia from "typia";
+
+/**
+ * Verifies generated Protobuf decoders accept unknown zero-length
+ * length-delimited fields.
+ *
+ * An empty string, an empty bytes value, an empty embedded message, and an
+ * empty packed repeated field all serialize to the same length-delimited record
+ * with a declared length of zero. Skipping such a record must consume exactly
+ * its two bytes, otherwise a reader that does not know the field desynchronizes
+ * and rejects a well-formed payload, which is the forward compatibility that
+ * unknown-field skipping exists to provide.
+ *
+ * 1. Round-trip messages whose trailing field is an empty string, bytes value, or
+ *    embedded message, and decode them through a schema missing that field.
+ * 2. Decode hand-built payloads carrying the same record first, last, and between
+ *    two known fields, through every direct and factory decoder.
+ * 3. Assert the known fields survive, and that truncated records are still
+ *    rejected.
+ */
+export const test_protobuf_decode_zero_length_unknown_field = (): void => {
+  // an unknown zero-length record is what a newer schema's empty field emits
+  const evolved: Array<readonly [string, Uint8Array]> = [
+    ["empty string", typia.protobuf.encode<IString>({ value: "a", extra: "" })],
+    [
+      "empty bytes",
+      typia.protobuf.encode<IBytes>({ value: "a", extra: new Uint8Array(0) }),
+    ],
+    [
+      "empty embedded message",
+      typia.protobuf.encode<INested>({ value: "a", extra: {} }),
+    ],
+  ];
+  for (const [label, encoded] of evolved) {
+    if (encoded.join() !== "10,1,97,18,0")
+      throw new Error(
+        `${label} did not encode to a trailing zero-length record, but to ${encoded.join()}.`,
+      );
+    if (typia.protobuf.decode<IValue>(encoded).value !== "a")
+      throw new Error(`${label} did not survive a decoder missing the field.`);
+  }
+
+  // an empty packed repeated field encodes to the identical record, which
+  // typia's own writer never emits because it omits an empty repeated field
+  const packed: typia.Resolved<IPacked> = typia.protobuf.decode<IPacked>(
+    Uint8Array.of(0x0a, 2, 1, 2, 0x12, 0),
+  );
+  if (packed.values.join() !== "1,2")
+    throw new Error(
+      "a trailing empty packed record broke a known packed repeated field.",
+    );
+
+  const decoders: Array<
+    readonly [string, (input: Uint8Array) => typia.Resolved<IValue>]
+  > = [
+    ["decode", (input) => typia.protobuf.decode<IValue>(input)],
+    ["createDecode", typia.protobuf.createDecode<IValue>()],
+    ["assertDecode", (input) => typia.protobuf.assertDecode<IValue>(input)],
+    ["createAssertDecode", typia.protobuf.createAssertDecode<IValue>()],
+    [
+      "isDecode",
+      (input) => must("isDecode", typia.protobuf.isDecode<IValue>(input)),
+    ],
+    [
+      "createIsDecode",
+      (
+        (closure) => (input: Uint8Array) =>
+          must("createIsDecode", closure(input))
+      )(typia.protobuf.createIsDecode<IValue>()),
+    ],
+    [
+      "validateDecode",
+      (input) =>
+        unwrap("validateDecode", typia.protobuf.validateDecode<IValue>(input)),
+    ],
+    [
+      "createValidateDecode",
+      (
+        (closure) => (input: Uint8Array) =>
+          unwrap("createValidateDecode", closure(input))
+      )(typia.protobuf.createValidateDecode<IValue>()),
+    ],
+  ];
+  for (const [label, decode] of decoders) {
+    if (decode(Uint8Array.of(0x12, 0, 0x0a, 1, 0x61)).value !== "a")
+      throw new Error(
+        `${label} desynchronized on a leading zero-length record.`,
+      );
+    if (decode(Uint8Array.of(0x0a, 1, 0x61, 0x12, 0)).value !== "a")
+      throw new Error(
+        `${label} desynchronized on a trailing zero-length record.`,
+      );
+  }
+
+  // optional presence and a record standing between two known fields
+  if (
+    typia.protobuf.decode<IOptional>(Uint8Array.of(0x12, 0, 0x0a, 1, 0x61))
+      .value !== "a"
+  )
+    throw new Error(
+      "an optional field lost its value to a zero-length record.",
+    );
+  if (
+    typia.protobuf.decode<IOptional>(Uint8Array.of(0x12, 0)).value !== undefined
+  )
+    throw new Error("a lone zero-length record produced a phantom value.");
+  const pair: typia.Resolved<IPair> = typia.protobuf.decode<IPair>(
+    Uint8Array.of(0x0a, 1, 0x61, 0x1a, 0, 0x12, 1, 0x62),
+  );
+  if (pair.first !== "a" || pair.second !== "b")
+    throw new Error(
+      "a zero-length record between two known fields desynchronized.",
+    );
+
+  // #2083: a declared length beyond the buffer is still rejected
+  assertOverflow("truncated unknown record", () =>
+    typia.protobuf.decode<IValue>(Uint8Array.of(0x0a, 1, 0x61, 0x12, 2, 0x62)),
+  );
+};
+
+interface IValue {
+  value: string;
+}
+
+interface IOptional {
+  value?: string;
+}
+
+interface IString {
+  value: string;
+  extra: string;
+}
+
+interface IBytes {
+  value: string;
+  extra: Uint8Array;
+}
+
+interface INested {
+  value: string;
+  extra: {
+    text?: string;
+  };
+}
+
+interface IPacked {
+  values: Array<number & typia.tags.Type<"uint32">>;
+}
+
+interface IPair {
+  first: string;
+  second: string;
+}
+
+const must = <T>(label: string, value: T | null): T => {
+  if (value === null) throw new Error(`${label} rejected a valid payload.`);
+  return value;
+};
+
+const unwrap = <T>(label: string, result: typia.IValidation<T>): T => {
+  if (result.success === false)
+    throw new Error(`${label} rejected a valid payload.`);
+  return result.data;
+};
+
+const assertOverflow = (label: string, closure: () => unknown): void => {
+  try {
+    closure();
+  } catch (error) {
+    if (
+      error instanceof Error &&
+      error.message === "Error on typia.protobuf.decode(): buffer overflow."
+    )
+      return;
+    throw new Error(`${label} reported ${String(error)} instead of overflow.`);
+  }
+  throw new Error(`${label} accepted a truncated Protobuf payload.`);
+};
--- a/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_zero_length_unknown_field.ts
+++ b/tests/test-typia-schema/src/features/protobuf.decode/test_protobuf_decode_zero_length_unknown_field.ts
@@ -15,13 +15,13 @@ import typia from "typia";
  *    embedded message, and decode them through a schema missing that field.
  * 2. Decode the same record hand-built as the leading and the trailing field,
  *    through every direct and factory decoder.
- * 3. Assert the known fields survive that record, an optional one, a packed one,
- *    and one standing between two known fields, and that truncated records are
- *    still rejected.
+ * 3. Assert the record leaves a required, optional, packed, and surrounding known
+ *    field intact, and that a truncated record is still rejected.
  */
 export const test_protobuf_decode_zero_length_unknown_field = (): void => {
-  // field 1 = "a", then field 2 length-delimited declaring a length of zero
+  // field 2 length-delimited declaring a length of zero, around field 1 = "a"
   const trailing: Uint8Array = Uint8Array.of(0x0a, 1, 0x61, 0x12, 0);
+  const leading: Uint8Array = Uint8Array.of(0x12, 0, 0x0a, 1, 0x61);
 
   // an unknown zero-length record is what a newer schema's empty field emits
   const evolved: Array<readonly [string, Uint8Array]> = [
@@ -92,7 +92,7 @@ export const test_protobuf_decode_zero_length_unknown_field = (): void => {
     ],
   ];
   for (const [label, decode] of decoders) {
-    if (decode(Uint8Array.of(0x12, 0, 0x0a, 1, 0x61)).value !== "a")
+    if (decode(leading).value !== "a")
       throw new Error(
         `${label} desynchronized on a leading zero-length record.`,
       );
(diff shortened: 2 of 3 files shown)


samchon:
## Self-Review round 1

Complete round over the whole `origin/master...HEAD` diff (`_ProtobufReader.ts`, the two new test files), plus the consequence surface: reader callers, generated decoders, the `#2083` and `#2090` contracts, docs, packaging, and versioning.

### To apply

1. **`assertOverflow` in `test_protobuf_reader_zero_length_skip.ts` does not assert `size()` is preserved.** The sibling `test_protobuf_reader_bounds.ts` pins both `index()` and `size()` on a rejected read, because `atomic()` restores both. Mirroring the established peer convention makes the rollback assertion complete rather than partial. Remediation: capture and assert `size()` alongside `index()`.

### Verified, no change

2. **`skip()` dropping its `atomic()` wrapper is safe.** `take()` calls `validate()` before mutating `ptr`, so `skip` stays atomic by construction for every rejected length — including `NaN`, negatives, and lengths past the `fork` boundary. The wrapper was only load-bearing for the multi-byte sentinel loop that no longer lives there.
3. **`skipVarint()` keeps its own `atomic()`** even though `skipType` already wraps one. It is public and standalone-callable, so it owns its rollback rather than borrowing its caller's.
4. **`skipVarint` is public, mirroring the existing public `skip`.** Both are peer wire-advance operations on the same internal reader. Neither is called by generated code today — the emitted decoders reach the reader only through `skipType` (`ProtobufDecodeProgrammer.go:141` is the sole reference). Making one private and leaving the other public would be arbitrary.
5. **No documentation is owed.** No page under `website/src/content/docs/**` describes unknown-field skipping or Protobuf schema evolution semantics, and no documented behavior changes — this restores the spec behavior the docs already imply.
6. **Single owner confirmed.** `packages/typia/src/internal/_ProtobufReader.ts` is the only reader in the repository; no second copy or shim exists to drift.
7. **No version bump.** The diff touches no native Go, matching the #2066 runtime-only precedent. The lead owns the version at merge.

### Observed, deliberately out of scope

These are pre-existing on master, behaviourally untouched by this diff, and have root causes distinct from #2116. Reporting rather than fixing, to keep the batch atomic — the lead may want them as candidates:

- `skipType`'s `default:` clause throws `Invalid wire type ${wireType} at offset ${this.ptr}.`, which lacks the `Error on typia.protobuf.decode():` prefix every other reader error carries.
- `skipVarint`'s loop has no 10-byte cap, unlike `varint32`/`varint64`. A malformed varint with excess continuation bytes is skipped rather than rejected. It cannot run away — the buffer end bounds it — and this change preserves the behavior exactly.

samchon:
## Self-Review round 2

Fresh complete round over the current `origin/master...HEAD` state. Round 1's finding is verified applied. Three new findings, all in `test_protobuf_decode_zero_length_unknown_field.ts`; the reader change itself survives a second from-scratch inspection unchanged.

### To apply

1. **The round-trip oracle is a decimal magic string.** `encoded.join() !== "10,1,97,18,0"` asserts the same bytes the hand-built payloads two blocks below write as `0x0a, 1, 0x61, 0x12, 0` — but a reader has to decode the decimals to see it. Worse, the test's whole argument depends on the writer-produced and hand-built payloads being *identical*, and nothing states that. Remediation: hoist the record to one `Uint8Array` in the file's hex vocabulary and compare both halves against it, making the link explicit rather than coincidental.
2. **The doc comment overclaims step 2.** It says the record is decoded "first, last, and between two known fields, through every direct and factory decoder". The between-two-known-fields case runs through `decode` alone. Remediation: reword so the spec matches what the case actually pins.
3. **`IString` / `IBytes` / `INested` collide with the sibling `test_protobuf_decode_truncated_length_delimited.ts`.** That file's `IString` is `{ value: string }`; mine is `{ value: string; extra: string }`. Same directory, same feature area, same name, different shape. The names also say nothing about their role — they model a *newer schema* that added a field left empty, which is the entire scenario. Remediation: rename to `IEvolvedString` / `IEvolvedBytes` / `IEvolvedNested`, matching the local `evolved` array.

### Verified, no change

4. **The empty-packed claim is oracle-backed, not assumed.** I confirmed against the writer that `typia.protobuf.encode` of a message with an empty repeated field emits `0x0a 0x01 0x61` only — it omits the field rather than emitting a zero-length record. That is why the packed case is hand-built at the wire level, and the comment says so.
5. **`must`/`unwrap` do not hide the pre-fix failure.** Pre-fix, `isDecode` and `validateDecode` *throw* the reader's framing error rather than returning `null`/`success: false` — consistent with #2090's boundary. Both helpers and the propagated throw fail the test; neither can mask a regression.
6. **No platform, concurrency, or packaging surface.** The reader is a per-decode instance with no shared state, and the diff carries no path handling, no generated artifact, and no package wiring.

samchon:
## Self-Review round 3

Fresh complete round. Round 2's three findings are verified applied. Two findings remain, both consequences of round 2's own edit; the reader change survives a third from-scratch inspection unchanged.

### To apply

1. **Round 2 left the `leading` payload asymmetric with `trailing`.** Hoisting the trailing record to a named const made its twin the odd one out: `Uint8Array.of(0x12, 0, 0x0a, 1, 0x61)` is still an inline literal, repeated at two call sites. The leading/trailing pair is the point of the case matrix, so both halves should be named. Self-inflicted by the previous round's fix.
2. **Doc comment step 3 is ambiguous.** "Assert the known fields survive that record, an optional one, a packed one, and one standing between two known fields" reads as though "an optional one" and "a packed one" are kinds of *record*, when they are kinds of *known field*. Remediation: name the axis explicitly.

### Verified, no change

3. **`prettier --check` flags `_ProtobufReader.ts`, but this is a pre-existing CRLF artifact, not a defect in this diff.** The untouched `_ProtobufWriter.ts` and `_ProtobufSizer.ts` fail the identical check in this Windows working copy. Piping the committed blobs through prettier confirms both master's reader and this branch's reader are byte-for-byte prettier-clean, as are the two new test files. No formatter run is owed, and per the campaign rule none was performed — only the files this branch authors were formatted individually, so the diff carries no repository-wide churn.
4. **Verification re-confirmed at the current state:** `pnpm --filter ./tests/test-typia-schema start` (Success), `pnpm exec ttsc --noEmit` in that suite (clean), `pnpm --filter ./tests/test-feature-identity start` (Success, covering the two new feature files' naming).

samchon:
## Self-Review round 4 — clean, and the verification record

Fresh complete round over `origin/master...HEAD` at `f393a039d2`. Round 3's findings are verified applied. **This round produced nothing that survives verification, so Self-Review stops here.**

### The change

`skip()` multiplexed two intents through one `number`: zero meant "consume one varint", while `skipType` forwards a decoded length for `LEN`, where zero is a legal length. Per the issue's Approach the intents are separated rather than zero being special-cased at the `LEN` call site:

- `skip(n)` now means only "advance n bytes", keeping its bounds check for every `n >= 0`.
- `skipVarint()` owns the varint wire type, which carries no length prefix and genuinely needs the consuming loop.

Dropping `skip`'s `atomic()` wrapper is safe: `take()` validates before mutating `ptr`, so `skip` stays atomic by construction. The wrapper was only load-bearing for the sentinel loop that moved to `skipVarint`, which keeps its own.

### Fail-then-pass

The regression matrix was written first and observed failing on unmodified master, reproducing the issue's exact symptom:

```
Error: zero-length LEN before another field left index 2 instead of 1.
  - test_protobuf_decode_zero_length_unknown_field: Error
  - test_protobuf_reader_zero_length_skip: Error
Failed
```

### Non-vacuity by mutation

Reverting only the fix — the reader restored byte-identical to master, tests untouched, package rebuilt — fails **exactly** the two new tests while `test_protobuf_decode_truncated_length_delimited` (#2083) and `test_protobuf_decode_invalid_utf8` (#2090) keep passing. The new coverage is specific to this fix and does not overlap the neighbouring contracts. Restoring the fix turns them green.

### Verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-schema start` | Success — incl. the 2 new cases and the #2083/#2090 regressions |
| `pnpm --filter ./tests/test-typia-automated start` | Success — the generated decode matrix |
| `go -C packages/typia/test test ./...` | ok — all packages |
| `pnpm --filter ./tests/test-feature-identity start` | Success — the 2 new feature files' naming |
| `pnpm exec ttsc --noEmit` (test-typia-schema) | clean |

The heavy suites ran at `da8b885b`; `git diff da8b885b..HEAD -- packages/` is empty, so every input they consume is byte-identical at `f393a039d2` and their results hold. Rounds 1–3 touched only the two new test files.

### Case matrix against the issue

| Case | Covered by | Result |
| --- | --- | --- |
| unknown zero-length `LEN` then another field | both tests, all 8 decoders | skipped exactly |
| unknown zero-length `LEN` as the last field | both tests, all 8 decoders | decodes |
| empty string / bytes / embedded message | round-trip through `_ProtobufWriter` | decodes |
| empty packed repeated | hand-built wire form + known packed field | decodes |
| zero-length `LEN` nested in a group | reader test, incl. a nested group | skipped exactly |
| zero-length `LEN` at a `fork` boundary | reader test | consumes nothing past the frame |
| unknown non-zero `LEN`, `I64`, `I32` | reader test | unchanged |
| unknown `VARIANT`, single- and multi-byte | reader test | still skips one varint |
| truncated `LEN` / `VARIANT` / group member | both tests | still rejected, atomically |

### Scope

TypeScript runtime only — no native Go, so no `packages/typia/package.json` bump per the #2066 precedent; the lead owns the version at merge. No documentation is owed: no page describes unknown-field skipping, and this restores the spec behavior the docs already imply. Per the campaign rule `pnpm format` was not run; only this branch's own files were formatted individually, so the diff carries no repository-wide churn.

Left as a draft: rebase and merge are lead-owned. Local verification is the gate — campaign CI is deliberately suspended, and every push passed the exact-SHA cancellation gate.

samchon:

## Rebase

Rebased from `40cfb302f3` onto current `master` `7c5a3aa493` (#2114). All five commits replayed with no conflict. **No version bump**, correctly: the diff contains no native Go, matching the #2066 precedent for a TypeScript-runtime-only Protobuf repair.

## Independent Self-Review: clean

The fix is at the right level. `skip(n)` now means only "advance n bytes" for every `n >= 0`, and a new `skipVarint()` owns the `VARIANT` case that needed the sentinel. That removes the overload rather than special-casing zero at the `LEN` call site — which is what #2116's Approach section asked for, and the difference matters: a call-site guard would have left the same trap armed for the next caller.

Dropping `skip`'s `atomic()` wrapper is also correct, and the reasoning holds under review: `take()` runs `validate()` **before** it mutates `ptr`, so `skip` is atomic by construction and the wrapper was load-bearing only for the sentinel loop — which moved into `skipVarint` and kept its own. The loop needs it because `u8()` advances `ptr` incrementally, so a mid-loop throw would otherwise leave the cursor moved.

The diff sits cleanly on #2084's boundary design and never touches #2090's `string()` boundary.

## Lead verification on `55f5c14b76`

| Gate | Command | Result |
| --- | --- | --- |
| Owning workspace | `pnpm --filter test-typia-schema start` | Success, 3,382 ms |
| Public Go contract tree | `go -C packages/typia/test test ./...` | pass, 0 failures |

All three Protobuf reader contracts now coexist in one suite run and none relabels another:

```
test_protobuf_decode_truncated_length_delimited   (#2083, PR #2084)
test_protobuf_decode_invalid_utf8                 (#2088, PR #2090)
test_protobuf_reader_invalid_utf8                 (#2088, PR #2090)
test_protobuf_decode_zero_length_unknown_field    (#2116, this PR)
```

## The mutation evidence is what makes this trustworthy

Reverting only the fix — reader byte-identical to master, tests untouched — fails **exactly** the two new tests while `test_protobuf_decode_truncated_length_delimited` and `test_protobuf_decode_invalid_utf8` keep passing. That proves two things at once: the new coverage is non-vacuous, and it is specific rather than overlapping its neighbours. Campaign CI is deliberately cancelled here, so an oracle that cannot fail is the one thing that could carry a defect through; this batch proved its own oracles can.

The batch also added a case the issue did not list: a zero-length skip at a `fork` boundary must not consume the enclosing message's tag.

## Recorded for the next discovery round, not fixed here

Two pre-existing defects surfaced adjacent to the fix and were correctly left alone for batch atomicity:

1. `skipType`'s `default:` throws `Invalid wire type ${wireType} at offset ${this.ptr}.`, missing the `Error on typia.protobuf.decode():` prefix every other reader error carries — an error-contract inconsistency.
2. `skipVarint`'s loop has no 10-byte cap, unlike `varint32`/`varint64`, so a malformed varint with excess continuation bytes is skipped rather than rejected. It cannot run away — the buffer end bounds it — and this diff preserves the behavior exactly.

`55f5c14b76` settled after four polls; `29530661784` (spell-check) reached terminal `success` before the cancel landed and is recorded as `raced_terminal`, every other run `cancelled`. Record in `.wiki/typia-repository-2026-07/pr-2119-rebase-push-gate-55f5c14b76.json`.

Issue #2117, opened 2026-07-16
ttsc-typia transform single-file host ignores typia diagnostics and publishes untransformed output


samchon:
## Problem

`ttsc-typia transform` reports success and publishes output when transforming a **single file**, even when the typia analysis produced diagnostics. The project host correctly returns exit `3` and encodes the diagnostics; both single-file paths never consult them. Because `defaultTransformOutput()` returns `"ts"`, the broken path is the **default** single-file behavior.

The published artifact is worse than the wrong exit code: the untransformed `typia.is<T>(input)` call survives into the emitted output. A caller that trusts exit `0` therefore ships a file containing a call that cannot work at runtime, with no signal that anything failed.

This is the same invariant #2079 owns — validation and artifact-commit semantics must not diverge between hosts — on a code path #2079's batch (PR #2096) does not touch.

Affected users: `typia generate` via `packages/typia/src/executable/generate/ttsc.ts`, the ttsc transform stage, and the wasm playground host (`main_wasm.go` → `jsTransform`).

## Evidence

`packages/typia/native/cmd/ttsc-typia/transform.go` on master (`22f00400b5`).

The project path collects and honours diagnostics:

```go
out.Dependencies = collector.ToJSON()
for _, diag := range *transformDiags {
    out.Diagnostics = append(out.Diagnostics, transformDiagnosticToCompilerDiagnostic(diag))
}
if err := json.NewEncoder(stdout).Encode(out); err != nil { ... return 3 }
if len(out.Diagnostics) > 0 {
    return 3
}
return 0
```

Both single-file paths do not:

```go
if *output == "js" {
    return runTransformSingleJS(prog, typiaTransform, target, absFile, *out)  // -> writeSingleOutput(captured, outPath)
}
text := transformFileToTypeScript(prog, typiaTransform, target)
return writeSingleOutput(text, *out)                                          // transformDiags never read
```

`transformDiags` is populated by the same collector for every host — the single-file branches simply never read it before writing.

The default is the broken path:

```go
func defaultTransformOutput() string {
    if os.Getenv("TYPIA_TTSC_TRANSFORM_OUTPUT") == "js" {
        return "js"
    }
    return "ts"
}
```

Reproduction: build `ttsc-typia` from master and run it against a source file whose typia call produces a transform diagnostic.

```bash
go -C packages/typia/native build -o /tmp/ttsc-typia ./cmd/ttsc-typia
[...]

PR #2120, opened 2026-07-16
fix(typia): make ttsc-typia single-file transform honor typia diagnostics

samchon:
Implements #2117.

## Intent

`ttsc-typia transform` has two hosts. The project host (no `--file`) collects typia's transform diagnostics and returns exit `3`. Both single-file hosts — `--output ts` (the **default**, per `defaultTransformOutput()`) and `--output js` — never read `transformDiags` at all. They exit `0` and publish an artifact containing the **untransformed** `typia.is<T>(input)` call, so a caller that trusts exit `0` ships a file that cannot work at runtime, with no signal that anything failed.

This is the sibling of #2079 / PR #2096 (`40cfb302f3`), which repaired the identical root cause on the **build** host: collect outputs into `pending`, check `len(transformDiags) > 0` **before** draining them to disk, and withhold the artifact on failure. This PR tells the same story for the single-file transform hosts.

## Approach

The invariant: **every transform host consults the same `transformDiags` before deciding its exit code and before publishing any artifact.**

Rather than duplicate the check into two branches — which is exactly how the divergence arose — both single-file modes are routed through one shared commit decision. Producing the artifact text and publishing it become separate steps, so a future output mode cannot reintroduce the bug by forgetting a check.

## Public CLI contract change

**This deliberately flips `ttsc-typia transform --file` from exit `0` to exit `3`** for inputs that currently "succeed" while emitting an untransformed stub. Per #2117 this is a correction, not a regression: the previous `0` asserted a successful transform that did not happen. Consumers checked:

- `packages/typia/src/executable/generate/ttsc.ts` — propagates the child status verbatim (`process.exitCode = result.status ?? 1`), so `typia generate` now fails loudly instead of writing a stub.
- the ttsc transform stage — drives the project host (JSON envelope), which already returned `3`; unchanged.
- `packages/typia/native/cmd/ttsc-typia/main_wasm.go` -> `jsTransform` -> `runTransform` — inherits the fix; its `{ code, stdout, stderr }` response now carries `3` and the diagnostic.

## Verification

Pending; recorded on this thread as it completes. Campaign CI is suspended per `.agents/skills/issue-campaign/development.md`; verification is local.

## Deferred / coordination

- **The `typia` version bump is pending lead assignment.** This changes native Go under `packages/typia/native`, so Change Integrity requires a bump; the campaign lead serializes the version across the chain, so this branch deliberately does not pick one and is not rebased onto master.
- Serialized after PR #2096 (#2079); branched from `40cfb302f3`, which already contains it.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/non_trailing_rest_tuple_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/non_trailing_rest_tuple_transform_test.go
@@ -18,20 +18,22 @@ import (
 // addresses tuples by fixed leading positions, so the honest resolution is
 // a compile-time rejection, mirroring the json.schemas bigint precedent.
 //
-//  1. Require leading-rest and middle-rest tuples to be rejected: the call
-//     stays untransformed and a transform diagnostic is recorded (the same
-//     convention the duplicate-Sequence protobuf fixture pins; in build and
-//     project modes these diagnostics fail the compilation).
+//  1. Require leading-rest and middle-rest tuples to be rejected: the transform
+//     records a diagnostic naming the offending tuple, which every host --
+//     single-file included since samchon/typia#2117 -- reports while refusing to
+//     publish an artifact.
 //  2. Require trailing-rest tuples to keep compiling and validating
 //     correctly at every length.
 func TestNonTrailingRestTupleTransform(t *testing.T) {
   t.Run("leading rest rejected", func(t *testing.T) {
     nonTrailingRestExpectRejection(t, "leading-",
-      "export const validate = typia.createValidate<[...unknown[], string]>();")
+      "export const validate = typia.createValidate<[...unknown[], string]>();",
+      "[...unknown[], string]")
   })
   t.Run("middle rest rejected", func(t *testing.T) {
     nonTrailingRestExpectRejection(t, "middle-",
-      "export const validate = typia.createValidate<[number, ...boolean[], string]>();")
+      "export const validate = typia.createValidate<[number, ...boolean[], string]>();",
+      "[number, ...boolean[], string]")
   })
   t.Run("trailing rest keeps working", func(t *testing.T) {
     nonTrailingRestTrailingControl(t)
@@ -66,7 +68,12 @@ func nonTrailingRestProject(t *testing.T, prefix string, body string) string {
   return dir
 }
 
-func nonTrailingRestExpectRejection(t *testing.T, prefix string, body string) {
+// nonTrailingRestExpectRejection requires the rejection to reach the caller as a
+// diagnostic naming `tuple`, with no artifact published. This previously asserted
+// code 0 and inspected the untransformed call in stdout; that success code was
+// the samchon/typia#2117 defect, and in `js` mode the surviving `.createValidate()`
+// was only type erasure, so the diagnostic is the load-bearing evidence anyway.
+func nonTrailingRestExpectRejection(t *testing.T, prefix string, body string, tuple string) {
   t.Helper()
   project := nonTrailingRestProject(t, prefix, body)
   out, errText, code := ttscTypiaTestCapture(func() int {
@@ -77,11 +84,16 @@ func nonTrailingRestExpectRejection(t *testing.T, prefix string, body string) {
       "--output", "js",
     })
   })
-  if code != 0 {
-    t.Fatalf("single-file transform should stay inspectable: code=%d stderr=\n%s", code, errText)
+  if code != 3 {
+    t.Fatalf("non-trailing rest tuple should be rejected with code 3: code=%d stdout=\n%s\nstderr=\n%s", code, out, errText)
   }
-  if !strings.Contains(out, ".createValidate()") {
-    t.Fatalf("non-trailing rest tuple call should stay untransformed:\n%s", out)
+  if !strings.Contains(errText, "error TS(typia.createValidate):") ||
+    !strings.Contains(errText, tuple) ||
+    !strings.Contains(errText, "non-trailing rest element in tuple type is not supported.") {
+    t.Fatalf("rejection diagnostic did not name the tuple and cause:\n%s", errText)
+  }
+  if out != "" {
+    t.Fatalf("rejected transform published an artifact:\n%s", out)
   }
 }
 
--- a/packages/typia/native/cmd/ttsc-typia/protobuf_sequence_field_numbers_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/protobuf_sequence_field_numbers_transform_test.go
@@ -14,22 +14,18 @@ import (
 // explicit protobuf field numbers were silently replaced by auto-numbering and
 // duplicate-sequence validation never fired.
 //
+// The two messages live in separate fixtures because a duplicate-sequence
+// diagnostic now withholds the whole file's artifact (samchon/typia#2117), so one
+// combined source could no longer show the valid message's emitted numbers.
+//
 //  1. Transform a message fixture carrying explicit `Sequence<5>` / `Sequence<7>`
 //     field numbers plus an untagged property (auto-numbered after the highest
-//     explicit number).
-//  2. Require the emitted proto message to carry the explicit numbers.
-//  3. Require a message whose two properties share `Sequence<1>` to fail the
-//     transform, leaving the call untransformed.
+//     explicit number), and require the emitted proto message to carry them.
+//  2. Require a message whose two properties share `Sequence<1>` to fail the
+//     transform with a diagnostic naming the duplicate.
+//  3. Require that rejected transform to publish no artifact.
 func TestProtobufSequenceFieldNumbersTransform(t *testing.T) {
-  project := protobufSequenceFieldNumbersProject(t)
-  out, errText, code := ttscTypiaTestCapture(func() int {
-    return runTransform([]string{
-      "--cwd", project,
-      "--tsconfig", "tsconfig.json",
-      "--file", "src/main.ts",
-      "--output", "js",
-    })
-  })
+  out, errText, code := protobufSequenceFieldNumbersTransform(t, "sequenced-", protobufSequenceFieldNumbersSource)
   if code != 0 {
     t.Fatalf("protobuf sequence transform failed: code=%d stderr=\n%s", code, errText)
   }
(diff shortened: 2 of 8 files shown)


samchon:

Complete round over the whole `origin/master...HEAD` surface (7 files). Posting findings before acting on them.

### 1. `js` mode's diagnostic scope is whole-program and undocumented — will fix (comment)

`transformSingleToJavaScript` drives `EmitWithPluginTransformers`, which iterates `GetSourceFilesToEmit(host, nil, false)` — every file in the program, not just `--file`. So in `js` mode `transformDiags` can hold a diagnostic raised by an *unrelated* source, and after this change `--file a.ts --output js` exits 3 when `b.ts` has a typia error. `ts` mode transforms only the target, so its scope is exactly the named file.

This asymmetry is a consequence of each mode's emit shape, not of the guard, and it matches both existing precedents (the build host fails the whole build on any transform diagnostic; the website playground port checks diagnostics after a whole-program emit the same way). I am keeping the behavior — it is the conservative reading of the issue's invariant — but the code should say so rather than leave the next reader to rediscover it. **Remediation: document the scope on `transformSingleToJavaScript`.**

### 2. Missing boundary: a failed transform must not create the `--out` directory — will fix (test)

`writeSingleOutput` does `os.MkdirAll(filepath.Dir(outPath))` before writing. The fix returns before that call, so a rejected transform creates no directory — but nothing pins it. The atomicity test seeds an already-existing `dist/`, so it would not catch a regression that mkdir'd the parent before consulting diagnostics. That is a real filesystem side effect of a "published nothing" run. **Remediation: add a boundary case asserting an unwritten nested `--out` parent is never created.**

### 3. Pre-existing `--output js` defect found while testing — reporting, not fixing here

`sameSourceStem` compares whole path stems:

```go
strings.TrimSuffix(filepath.ToSlash(jsPath), filepath.Ext(jsPath)) == strings.TrimSuffix(tsPath, filepath.Ext(tsPath))
```

With `rootDir: src` / `outDir: dist`, the emitted `<p>/dist/main.js` never matches `<p>/src/main.ts`, so `--output js` reports `no output produced` and exits 3 for *any* project with an `outDir` — including on master. Its own comment says "match on basename stem", which the code does not do. This is outside #2117's scope (it is neither the diagnostic decision nor artifact atomicity) so I am not expanding into it; flagging for the lead as a separate candidate.

It does constrain this PR's fixtures: `transformSingleFileProject` deliberately omits `outDir` so `js` mode is genuinely exercisable rather than passing because it produced nothing. That reason is recorded in the fixture's doc comment.

### Checked and clean

Guard placement (diagnostics outrank the producer's own code, matching the playground port's ordering); `cwd` threading into `writeTypiaTransformDiagnostics` matches the build host; rename of `runTransformSingleJS` has no lingering references; the dropped unused `target` parameter was already unused; project host stdout/stderr/exit byte-identical vs. a master binary in both success and failure; `go vet` clean; `GOOS=js GOARCH=wasm` build succeeds; no `pnpm format` run (campaign rule); Windows path handling unchanged (`filepath.ToSlash` boundaries untouched).

samchon:

Fresh complete round over the current `origin/master...HEAD` surface. No correctness defects this round; three clarity/consistency items, all in test code.

### 1. `seeds` map literal is not aligned like its peer — will fix

`build_transform_failure_atomicity_test.go` aligns its `seeds` values to the longest key. My new `seeds` map does not, so two sibling atomicity fixtures read differently. Work Rules require mirroring a nearby peer's code style.

### 2. Dead assertion in the stdout test — will fix

```go
if outText != "" { t.Fatalf(...) }
if strings.Contains(outText, "typia.is<") { t.Fatalf(...) }
```

The second check is unreachable: the first already fatals on any non-empty stdout, and its failure message prints the content. Remove it rather than leave an assertion that can never fire.

### 3. Pointless `filepath.ToSlash` — will fix

`filepath.ToSlash(errText)` in the stdout test normalizes separators for an assertion that contains no path (`error TS(typia.is):`). It is harmless but implies a path-sensitivity that is not there. The atomicity test's `ToSlash` is load-bearing (`src/main.ts:3:9`) and stays.

### Re-verified clean this round

Documentation: the `transform` verb and its exit codes are not described anywhere user-facing (`website/src/content/docs/**`, READMEs) — it is an internal backend command driven by `ttsc` and `typia generate` — so the development skill's documentation obligation does not attach. `--rewrite-mode none` still publishes at exit 0 because it records no diagnostics, which is its intent. The `code` reassignment in the second atomicity run is not a shadow. Guard ordering, `cwd` threading, Windows path handling, the rename, and the fixture's deliberate `outDir` omission all re-checked against the current state.

samchon:
## Self-Review round 3 — clean; verification complete

Fresh complete round over the whole `origin/master...HEAD` surface at `b0849a24ec`. **Nothing survived verification — Self-Review is complete** (rounds 1 and 2 each applied improvements and triggered this round).

Re-checked this round: the guard cannot withhold a good artifact (diagnostics enter only through `extras.AddDiagnostic` from `nativetransform.Transform`, so a non-empty set always means a call typia could not lower); `--rewrite-mode none` still publishes at exit 0; the `js`-mode `typia.is<` assertion is redundant there (JS emit erases type args) but the `"number"` assertion carries the lowering proof; imports, vet, and the peer-mirrored fixture style all clean.

## Final local verification (campaign CI suspended)

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia` | pass |
| `go -C packages/typia/test test -p 2 ../native/...` | pass |
| `go -C packages/typia/test vet ../native/cmd/ttsc-typia` | pass |
| `GOOS=js GOARCH=wasm go build ../native/cmd/ttsc-typia` | pass (wasm host compiles) |
| `pnpm --filter ./tests/test-typia-cli start` | pass |
| `pnpm --filter ./tests/test-typia-generate start` | pass |

### Out-of-process proof (real binary, real typia package, live project)

Built `ttsc-typia` and a second binary from master (`40cfb302f3`), and ran both against one live project whose `node_modules/typia` is the real built package.

| Case | master binary | this branch |
| --- | --- | --- |
| `transform --file src/invalid.ts --out dist/invalid.ts` (default `ts`) | **exit 0**, published an artifact containing the untransformed `return typia.is<T>(input);` | **exit 3**, `src\invalid.ts:3:9 - error TS(typia.is): ... non-specified generic argument.` on stderr, **no artifact**, pre-existing `dist/invalid.ts` byte-intact |
| `--output js` with `--out` | — | exit 3, diagnostic, no artifact published |
| default `ts`, no `--out` (stdout is the artifact) | — | exit 3, **0 bytes** on stdout, no stub |
| valid file, default `ts` | — | exit 0, artifact published, call lowered (no raw `typia.is<`) |
| project host, failing source | exit 3 | exit 3 — stdout envelope and stderr **byte-identical** |
| project host, valid source | exit 0 | exit 0 — envelope **byte-identical** |

### Fail-then-pass and mutation evidence

Tests were written first and observed failing on unmodified source: `ts` mode returned **exit 0** with `return typia.is<T>(input);` on stdout; the positive twin already passed, proving the fixture sound.

Three mutations, each caught:

1. guard never fires (`len > 0` -> `len < 0`) — all three failure tests fail.
2. guard always fires (`len > 0` -> `len >= 0`) — the positive twin fails.
3. guard moved **after** `writeSingleOutput` — still exits 3 and still prints the diagnostic, so an exit-code-only test would pass it; the atomicity assertions catch it. This is why the artifact checks are separate from the exit-code checks.

### Change Integrity call-outs

- **Version bump still pending lead assignment** (native Go changed; branch deliberately not rebased and no version picked).
- **Two existing tests pinned the defective contract** and were updated: `TestNonTrailingRestTupleTransform` and `TestProtobufSequenceFieldNumbersTransform` both asserted exit 0 while requiring a *rejected* call to stay untransformed — that success code is exactly what #2117 declares a defect. Their subject is the rejection, so they now assert the diagnostic text and the withheld artifact, and the protobuf fixture splits into valid/duplicated sources because one diagnostic now withholds the whole file's artifact. Both intents are preserved and strengthened; neither was weakened to pass.

samchon:
Final rebase, version assignment, and lead certification.

## Rebase and version

Rebased from `40cfb302f3` onto `e0c2df02f9` (#2105). All four commits replayed with no conflict. Assigned **`13.1.15`**, the next patch on master's `13.1.14` — mandatory because the diff changes native Go.

## Lead certification on `c40fc81af6`

| Gate | Command | Result |
| --- | --- | --- |
| Focused transform host | `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` | ok, 37.7s |
| Vet, changed package | `go -C packages/typia/native vet ./cmd/ttsc-typia/` | clean |
| **Wasm host builds** | `GOOS=js GOARCH=wasm go build ./cmd/ttsc-typia` | ok — this is the host the fix reaches |
| CLI consumer | `pnpm --filter test-typia-cli start` | Success |
| CLI consumer | `pnpm --filter test-typia-generate start` | exit 0 — `generate/ttsc.ts` propagates `result.status` verbatim, so the exit-code flip lands here first |

## Spec change: two existing tests pinned the defective contract

Flagged by the batch rather than slipped through, and I agree with the resolution after checking it.

`TestNonTrailingRestTupleTransform` and `TestProtobufSequenceFieldNumbersTransform` both asserted **exit 0** while requiring a *rejected* call to remain untransformed. They used single-file mode as an "inspectable" mode **precisely because it swallowed diagnostics** — that is, they depended on the defect this PR removes.

Their subject was never the exit code; it is that typia rejects the construct. The old assertion verified that indirectly, through output the fix now correctly withholds. Each now asserts the diagnostic names the offending tuple or the duplicate sequence collision — a more direct statement of the same subject, and a stronger one. The Protobuf fixture split into valid and duplicated sources because file-level atomicity means a valid case and an invalid case can no longer share one file and have the valid one's artifact inspected. Intent preserved, evidence strengthened.

## The mutation evidence is what makes the atomicity assertions worth having

Three mutations were caught, and one is instructive: **moving the guard after `writeSingleOutput` still exits 3 and still prints the diagnostic.** An exit-code-only test would pass that mutant while the artifact leaked anyway. That is why the artifact checks are separate from the exit-code checks, and it is the same lesson #2096 established — the exit code and the published state are two different contracts.

## Out-of-process proof

Binaries built from this branch **and from master**, run against one live project with `node_modules/typia` pointing at the real built package:

| Case | master | this branch |
| --- | --- | --- |
| `--file src/invalid.ts --out dist/invalid.ts` (default `ts`) | exit 0, published untransformed `return typia.is<T>(input);` | exit 3, diagnostic on stderr, no artifact, prior `dist/invalid.ts` byte-intact |
| `--output js` | — | exit 3, diagnostic, nothing published |
| default `ts`, no `--out` | — | exit 3, 0 bytes on stdout |
| valid file | — | exit 0, published, call lowered |
| project host, fail and success | 3 / 0 | 3 / 0 — envelope and stderr byte-identical |

## A correction to this issue's premise, which strengthens it

I wrote in #2117 that the wasm host backs the online playground. That is wrong: the playground is served by `website/compiler/cmd/playground/typia_plugin.go`, a separate port that **already had** the diagnostic check in both single-file branches. `ttsc-typia`'s own wasm host (`main_wasm.go` -> `jsTransform` -> `runTransform`) genuinely had the bug and is fixed here.

The correction makes the case stronger rather than weaker: the correct behavior was already established in a sibling port, so this was one host drifting from a contract the repository had already written down elsewhere — the same two-owners-diverging shape as the `uri-reference` quote and the duplicated identifier predicate this campaign already repaired.

## Recorded for the next discovery round, not fixed here

`sameSourceStem` compares **whole path stems**, so `--output js` reports "no output produced" and exits 3 for any project with an `outDir` (`<p>/dist/main` never equals `<p>/src/main`) — on master too. Its own comment says "match on basename stem", which it does not do. It forced this batch's fixture to omit `outDir`, documented in the fixture comment.

The first gate attempt threw on `29535912095` (spell-check): the cancel lost the race and the readback caught the run mid-transition, before it published its terminal state. That is a timing artifact, not a refusal — the run is `completed`/`success`. I hardened the gate to re-read a few times before concluding a run refused to cancel, cancelled the one genuinely active run (`29535912100`, website), and re-ran: settled in three polls with all runs terminal, five `cancelled` and spell-check `success`.

PR #2099, opened 2026-07-16
test(automation): fail on fatal asynchronous events

samchon:
## Intent

Make fatal asynchronous events in both generated automation runners fail closed instead of logging an error and allowing a successful process exit.

## Scope

- Replace the `uncaughtException` and `unhandledRejection` success-masking behavior in `test-typia-automated` and `test-utils-automated`.
- Cover normal completion plus both fatal event kinds through subprocess tests at lifecycle boundaries.
- Preserve worker/servant cleanup where possible while guaranteeing a nonzero final status.

## Deferred

- Direct/factory matrix restoration remains a later #2035 unit after its product-fix prerequisite PRs merge.
- `resolved_equal_to` and `protobuf_equal_to` hardening remains a later #2035 unit for the same dependency reason.
- This PR is only part of #2035 and intentionally does not close the issue.
- No repository-wide formatting; Post-Campaign Cleanup owns formatting.

## Verification

- Tests-first subprocess regressions demonstrated the old zero-exit `Success` behavior for fatal events.
- Both runner prestart matrices now pass normal completion and both fatal event kinds before, during, and after the lifecycle.
- Both complete affected automated workspaces and root `pnpm test` passed on the implementation's original base.
- Both affected prestart matrices pass on the final repaired base.
- Full `test-typia-automated` and full `test-utils-automated` pass on the final head.
- Root `pnpm test` passes across all twelve TypeScript workspaces, public Go, and native Go.
- Findings-first solo Self-Review hardened the subprocess oracle against hangs, signals, and lost listener state; fresh full-diff rounds, including one after the final repaired-base rebase, found no remaining issue.

Campaign Actions remain enabled and are cancelled only for this branch's exact pushed SHA after every push or pull-request mutation. Local verification is the implementation gate.

Part of #2035

PR #2121, opened 2026-07-16
fix(tests): restore direct/equality matrices in the automated suites

### LINKED ISSUE #2035: Automated tests skip direct matrices and mask asynchronous failures

samchon:
## Intent

Close out the remaining units of #2035. The fatal-event unit landed separately in #2099; this pull request owns the other three, and it contains test infrastructure and generated baselines — no compensating product special case.

**B — the direct matrix.** `TestAutomationController` asked for the direct visit of every runtime template and then threw the request away: `generateFeatureSet()` returned immediately when `create === false`. Twenty public operation families shipped only their `createX` half, and nothing failed to say so. Removing that early return restores **1,879 direct cases** (3,929 generated files across 41 families, up from 2,050 across 21).

**C1 — the synchronous oracles.** `resolved_equal_to` and `protobuf_equal_to` both answered `true` for any fixture whose name contained `Class`; the former also approved any comparison involving a function. Six generated class cases executed with no oracle at all. Both walked only the input's own keys, hiding unexpected output properties, and their native comparisons were shallow enough that same-length buffers, a different `DataView` window, and a swapped typed-array brand all compared equal.

**C2 — FormData content.** `Blob.arrayBuffer()` is asynchronous, so byte equality lives in `resolved_equal_to_async`, used only by the four FormData internals. Their generated cases now declare `Promise<void>`, so a rejected oracle reaches `DynamicExecutor` instead of becoming an unhandled rejection.

## Why `protobuf_equal_to` is deleted rather than hardened

This is the campaign's signature defect, not scope creep. Two near-copies of one contract had already drifted: the Protocol Buffer copy never grew a `Date`, `Set`, or `ArrayBuffer` branch, so any such value fell through to its object walk, where `Object.entries(new Date())` is `[]` and `.every()` on an empty array is `true` — **it compared equal to anything**. Nothing caught that, because no Protocol-Buffer-encodable fixture happens to carry one of those types today.

Once the name bypass and the shallow native checks were gone, the copy had no semantics left that `resolved_equal_to` did not. One `strict_equal_to` engine now owns the walk, and the two remaining wrappers hold only their fixture projection. Keeping a second oracle alive would preserve the exact mechanism that produced the defect.

## Fixture knowledge belongs to the fixture

The generic comparator must never recognize fixtures by name again, so `TestStructure.RESOLVE` declares the shape a resolving operation must produce. Real transformed output — not the readiness note's table — decided the policy: `AtomicClass`, `ClassMethod`, and `ClassPropertyAssignment` reproduce their input exactly under an own-enumerable-key walk, because prototype methods and wrapper identity never appear in `Object.keys()`. **Only `ClassNonPublic` needs a projection**: its codec emits exactly `implicit` and `shown`, while `Object.keys()` still reports the constructor-owned `heritage` and `hidden`.

## Change Integrity: the `ObjectHttpFormData` fixture

Its binary parts were `new Blob()` and `new File([], "file")` — empty, untyped, unnamed. A FormData oracle that lost every byte and every media type still compared them equal, so the restored content oracle had nothing it could catch. They now carry random bytes, media types, names, and modification times.

This is a deliberate fixture change under the development skill's Change Integrity rule. The product reason is that the issue's own acceptance criterion — "DataView visible bytes/range and Blob/File content/metadata are checked, with negative-twin tests proving each oracle can fail" — is unsatisfiable against empty blobs. An oracle that cannot fail is what this issue exists to end.

## The restored matrices exposed no product defect

All twelve previously bypassed class cases — `plain.clone`/`createClone` for AtomicClass, ClassMethod, ClassPropertyAssignment, and `protobuf.encode`/`decode` plus their factories for ClassMethod, ClassNonPublic, ClassPropertyAssignment — now execute against a real oracle and pass, as do their newly restored direct twins. #2035's thirteen prerequisite batches were sequenced precisely so its oracles would come back to a repaired product; they did.

## Every restored oracle is mutation-proven

Each is non-vacuous under a deliberate reversion:

| Mutation | Fails |
| --- | --- |
| reinstate `if (create === false) return` | `the direct "is" feature set is missing` |
| reinstate the top-level function blanket | function reference-identity twins |
| one-sided `Object.keys(x)` traversal | unexpected-output-property twins |
| drop the intrinsic brand check | typed-array and array-like twins |
| `ArrayBuffer` compares only `byteLength` | same-length different-bytes twin |
| `DataView` compares only backing `byteLength` | all four window twins |
| `Blob`/`File` compare constructors only | every metadata twin |
| asynchronous byte pass compares nothing | same-size different-content twins |

## Verification

Recorded on the thread.

## Campaign CI

Campaign CI for each pushed SHA is cancelled at the exact-SHA gate; local verification is the gate.

Closes #2035


samchon:

Surface: `origin/master...HEAD` (`358d438de9`), every changed file and hunk, plus the generated trees the suites rebuild.

### Findings

**1. `blob_equal_to` skips byte comparison silently when no accumulator is supplied.**
`strict_equal_to`'s synchronous entry passes no `blobs` accumulator, so `ctx.blobs?.push(...)` is a no-op and blob content is never compared. Today that is harmless — every `Blob`-bearing fixture (`ObjectHttpFormData`) reaches the four FormData internals, which all use `resolved_equal_to_async`. But the shape is exactly the one this issue exists to end: a `Blob` added to a headers, query, or plain-clone fixture tomorrow would lose byte coverage with no signal at all. Plan: make the synchronous path throw a harness error naming `resolved_equal_to_async`, so the gap is loud instead of silent.

**2. `brand()` is computed up to four times per object node.**
`recursive_equal_to` calls it twice in the native-brand condition and twice again in the `trace` call, and `Object.prototype.toString.call` allocates. The walk now runs across roughly 4,000 generated cases over deep structures. Plan: hoist `bx`/`by` into single consts. Behavior-preserving; the oracle regression is the guard.

**3. The projection rule is spelled out twice.**
`resolved_equal_to` and `resolved_equal_to_async` each carry `factory.RESOLVE !== undefined ? factory.RESOLVE(input) : input`. Two copies of one rule is how `resolved_equal_to` and `protobuf_equal_to` drifted apart in the first place. Plan: extract it to a single helper both wrappers call.

### Not changed, and why

- **`tests/test-utils-automated/src/regressions/test_process_fatal_events.ts` still self-invokes** while the typia twin no longer does. The asymmetry is intentional: the typia suite has four regressions and needs an aggregating `index.ts`, the utils suite has one and its `prestart` points straight at it. Adding an aggregator there would be churn with no coverage.
- **The suppression's sibling cruft is left alone.** `tests/test-typia-automated/src/utils/headers_to_string copy.ts` is a byte-identical, unreferenced duplicate introduced by the same monorepo migration `a7cbc4f1ae` as the direct-matrix suppression. It is dead-file cruft, not a matrix or oracle defect, so it is being reported to the lead as a separate issue candidate rather than swept into this pull request.

Verification for each fix will be the prestart regression matrix plus the mutation matrix already recorded; the full workspace runs follow.

samchon:

Fresh full round over `origin/master...HEAD` (`5c9ab4587c`), every changed file and hunk again from scratch.

### Finding

**1. `resolved_equal_to_async` miscounts the consumers it keeps synchronous.**
Its doc says "keeps the twelve synchronous consumers synchronous". Fourteen is right — eight HTTP query/header internals, four plain-clone internals, and the two Protocol Buffer internals, which joined the same oracle when `protobuf_equal_to` was removed. The wrong number came from the readiness note's count of *HTTP* consumers, which included the four FormData ones that this change made asynchronous. Verified by enumerating the imports. Plan: correct to fourteen.

### Checked and sound

- `brand()` is now computed at most twice per object node, and adds two `Object.prototype.toString.call` to the plain-object path that were not there before. Left as is: the suite's cost is dominated by compiling roughly four thousand generated files, and the intrinsic is what makes the check realm-safe.
- `TestAutomationTemplate.method` (namespace function) reads next to `tpl.method` (interface property) through declaration merging. Both names are accurate for what they are; renaming either would be worse.
- The `TestStructure.RESOLVE` addition and both fixture edits land only on structures carrying `JSONABLE = false`, so `test-utils-automated`'s structure set is untouched.
- `regressions/index.ts` mirrors `src/index.ts`'s failure-listener shape, so a fatal event arriving after the last regression still exits nonzero via `process.exitCode`.

### Two issue candidates for the lead, deliberately not fixed here

1. **`tests/test-typia-automated/src/utils/headers_to_string copy.ts`** is a byte-identical, unreferenced duplicate of its neighbour, introduced by the same monorepo migration `a7cbc4f1ae` as the direct-matrix suppression. Dead-file cruft, not a matrix or oracle defect.

2. **`tests/test-utils-automated/src/TestAutomation.ts:getStructures()` selects fixtures by raw source-substring match** — `content.includes("never")`, `"toJSON"`, `'<"uint32">'`, `"[key: "`. A doc comment or identifier containing any of those words would silently drop a structure from the validate/validateEquals matrix with no signal: the same hidden-coverage-loss family as this issue, in the sibling suite. It is latent, not live — the only two structures matching today (`DynamicNever`, `ObjectUndefined`) match on real `never` types. Out of this issue's declared scope, so it belongs in its own unit rather than being swept in here.

samchon:
## Self-Review round 3 — findings and remediation plan

Fresh full round over `origin/master...HEAD` (`174836cec2`).

### Finding

**1. The backstop proves parity but not distinctness.**
#2035's acceptance reads "every eligible template generates **distinct** direct and factory tests; a regression assertion proves parity". `test_direct_factory_matrix` proves the parity half properly, but it takes distinctness for granted. If `directory(tpl, false)` ever equaled `directory(tpl, true)`, both halves would generate into one directory, the second `rm`-and-regenerate would win, and the backstop would still pass: one visited name satisfies both `has()` checks, and a directory trivially has the same structure set as itself. Not reachable today — the factory half is `create${Capitalize(method)}` — but the assertion the issue asks for is one line, and its absence is precisely the kind of gap this batch exists to close. Plan: require the two names to differ.

### Checked and sound

- The "unexpected file" guard in `structures()` looked dead, since the controller composes file names from the same `directory()` the backstop reads and `generateFeatureSet` removes the directory before writing. Keeping it: it is the guard against the controller reverting to its own inline name composition, which is exactly the drift `directory()` was extracted to prevent.
- No family-count assertion, deliberately. Parity is proved per template in `DATA`; a hardcoded "twenty" would be a magic number that a legitimate new operation breaks, and removing a template from `DATA` is a deliberate registry edit, not a silent suppression.
- `TestStructure.RESOLVE` uses method shorthand, matching its neighbour `generate()`.
- `write_common`'s `asynchronous` flag only annotates the return type; the generated arrow already returns the internal's promise, and `DynamicExecutor` awaits it.

## Verification of `174836cec2`

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-automated start` | **pass** — 3,929 generated cases across 41 families, prestart matrix green |
| `pnpm --filter ./tests/test-utils-automated start` | **pass** |
| mutation matrix (8 reversions) | each fails its matching regression; recorded in the body |
| `pnpm test` | **`tests/test-error` fails — #2122, pre-existing on master and not mine** |

`test-error` dies on an unhandled `ENOENT` reading `bin/examples/llm.schema.additionalProperties.js`, the #2096 emit-withholding fallout the lead bisected to `14ea77e03c`. It is independent of this branch by construction: this diff touches only `tests/template` and `tests/test-typia-automated`, and `test-error` depends on neither — `@typia/template` has exactly three consumers, the two automated suites and itself.

`pnpm --filter=./tests/test-* -r start` bails at the first failure, so the remaining suites never ran. Re-running them with `test-error` excluded, plus the Go toolchain, and will record the result here.

samchon:
## Self-Review round 4 — findings and remediation plan

Fresh full round over `origin/master...HEAD` (`cb81cb4a0b`).

### Finding

**1. The blob-guard case in `test_resolved_equal_to_oracle` accepts any throw.**
It is `try { ... } catch { return; }`, so *any* exception counts as the guard firing — including a `TypeError` from a genuine bug in the helper (`undefined is not a function`, a bad property read). An oracle that passes for a reason other than the one it names is exactly the defect class this batch exists to end, and I wrote a fresh instance of it into the regression meant to end it. Plan: require the caught error to be the blob guard, by matching the message that names `resolved_equal_to_async()`, and fail loudly on any other throw.

### Checked and sound

- `instanceof` dispatch (Blob, Date, RegExp, ArrayBuffer, DataView, Set, Map) is realm-local, and a cross-realm value would fall through to the object walk where `Object.keys(new Date())` is `[]` and everything compares equal. Not reachable: every value here is created and compared inside one servant process. The *identity rule* — does `y` reproduce `x`'s brand — is realm-independent via `Object.prototype.toString`, which is what the hardening required.
- `bx`/`by` are computed before the Blob branch, which does not use them when both sides are blobs. One redundant intrinsic call on a rare path; not worth a second dispatch.
- `total` in the backstop counts direct files across creatable and non-creatable templates alike, and the `createOnly` branch skips before counting. Correct: the create-only family has no direct half to count.
- `optional()` treats an empty `Map` but not an empty `Set` as satisfying a missing value. Pre-existing, preserved deliberately: it mirrors what the Protocol Buffer and HTTP encodings actually drop, and no resolved fixture carries a `Set`.

samchon:
## Self-Review rounds 5 and 6 — complete, clean

**Round 5** over `1d8396f1d3` found one item: the blob-guard rationale had two ideas merged into one comment paragraph. Split. No behavioral change.

**Round 6** over `f794aa91d5` — a fresh full pass over all 35 changed files — produced nothing that survives verification. Re-enumerated every oracle call site as the closing check: 18 consumers, 14 synchronous and 4 asynchronous, none passing a fixture name. **Self-Review is complete.**

## Final verification of `f794aa91d5`

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-automated start` | **pass** — 3,929 generated cases, 41 families |
| `pnpm --filter ./tests/test-utils-automated start` | **pass** |
| `pnpm --filter=./tests/test-* --filter=!./tests/test-error -r start` | **pass** — 13 suites |
| `pnpm run test:toolchain` (both Go trees) | **pass** |
| `pnpm test` | `tests/test-error` fails — **#2122, pre-existing on master, not this branch** |
| mutation matrix, 10 reversions | each fails its matching regression |

The suites that ran clean under the recursive command: `test-feature-identity`, `test-interface`, `test-langchain`, `test-mcp`, `test-typia-automated`, `test-typia-bundler-cache`, `test-typia-cli`, `test-typia-exact-optional`, `test-typia-generate`, `test-typia-schema`, `test-utils`, `test-utils-automated`, `test-vercel`.

`test-error` is independent of this diff by construction: this branch touches only `tests/template` and `tests/test-typia-automated`, and `@typia/template` has exactly three consumers — the two automated suites and itself. Worth recording that `pnpm --filter=./tests/test-* -r start` **bails at the first failure**, so `test-error` takes the other nine suites down before they ever run; that is why the exclusion above was needed to prove anything at all.

Two mutations were added after the rounds above and are folded into the body's table:

| Mutation | Fails |
| --- | --- |
| `directory()` ignores `create`, collapsing both halves | `the direct and factory halves of "is" share one feature set` |
| the blob guard throws a different message | `a Blob threw, but not from the guard` |

## Campaign CI

GitHub's Actions **list** endpoint has been returning HTTP 503 throughout this window — confirmed server-side with raw `curl`, not a `gh` artifact, while `repos/…` returns 200. `cancel-exact-sha.ps1` therefore cannot enumerate, and it treats that as total failure.

The gate's substance was met anyway, because only *enumeration* is down — *cancellation* works:

- enumerate via GraphQL `Commit.checkSuites`, which reaches Actions data without the REST list endpoint;
- cancel via `gh run cancel`, which succeeded first try on every attempt;
- read back with two consecutive stable polls, as the gate requires.

| SHA | `test` | `spell-check` |
| --- | --- | --- |
| `5c9ab4587c` | cancelled (script) | — |
| `174836cec2` | cancelled (script) | — |
| `cb81cb4a0b` | **cancelled** (29539421313) | success (29539421416) |
| `1d8396f1d3` | **cancelled** (29540201948) | success (29540201886) |
| `f794aa91d5` | **cancelled** (29540536304) | success (29540536360) |

Every `spell-check` completing `success` before the cancel lands is the documented benign race. No campaign run is left uncancelled. The queued third-party check suites on these SHAs (GitHub Pages, Vercel, Socket Security, pkg.pr.new) are integrations, not Actions runs, and are outside the gate's scope by construction.

samchon:
Lead certification before merge. **This is the campaign's final product batch.**

## Rebase

Rebased onto `647426a623`, which now carries #2123 — so `test-error` is repaired and `pnpm test` can complete for the first time since `40cfb302f3`. Seven commits replayed with no conflict. **No version bump**, correctly: the diff is `tests/template` and `tests/test-typia-automated` only, with zero files under `packages/` and zero Go.

## Lead certification on `352f8e09b4`

Measured in two halves with controlled parallelism, for a reason recorded below:

| Half | Command | Result |
| --- | --- | --- |
| TypeScript | `pnpm run test:packages` | **exit 0**, zero hard failures, all 15 suites — `test-error` reports `50 files, 845 typia diagnostics`, `test-typia-automated` Success, `test-utils-automated` Success |
| Go | `go -C packages/typia/test test -p 2 ./...` and `../native/...` | **0 failures**, both trees |

**A first pass at `pnpm test` reported 31 hard failures, and every one was an environment artifact.** The Go tree failed with `exit status 0xc0000142` — `STATUS_DLL_INIT_FAILED` — across a dozen packages: Windows refusing to start `compile.exe` and `vet.exe` when Go spawns them at full parallelism alongside other work. A tests-only diff cannot break the Go build. Re-run at `-p 2` with nothing else running, both trees pass with zero failures. Recording it because a failure that names a dozen packages looks exactly like a real regression, and this campaign has already lost hours to environment failures misfiled as code failures.

## Why this batch's result is the point, not an anticlimax

**The restored matrices exposed no product defect.** 3,929 cases green, with 1,879 direct cases restored across 41 families (up from 21), and all twelve previously bypassed class cases now executing against a real oracle. #2035 was deliberately sequenced behind thirteen prerequisite batches so its oracles would expose defects that were *already fixed* rather than rediscover them. That it finds nothing to paper over is the evidence the sequencing worked — and nothing was special-cased to get there.

## Ten mutations, each proving an oracle can fail

| Reversion | Regression that fails |
| --- | --- |
| reinstate `if (create === false) return` | `the direct "is" feature set is missing` |
| `directory()` ignores `create` | `direct and factory halves of "is" share one feature set` |
| reinstate the function blanket | function reference-identity twins |
| one-sided `Object.keys(x)` traversal | unexpected-output-property twins |
| drop the intrinsic brand check | typed-array / array-like twins |
| `ArrayBuffer` compares only `byteLength` | same-length different-bytes twin |
| `DataView` compares only backing `byteLength` | all four window twins |
| `Blob`/`File` compare constructors only | every metadata twin |
| async byte pass compares nothing | same-size different-content twins |
| blob guard throws a different message | `a Blob threw, but not from the guard` |

Rounds 3 through 5 of Self-Review each found a real oracle that could not fail — **including one the batch had written itself in round 2**, where `catch { return; }` accepted any throw, so a `TypeError` from a genuine helper bug would have passed. Round 6 clean. With campaign CI deliberately cancelled, an oracle that cannot fail is the only thing that can carry a defect through, and this batch found its own.

## Two Change Integrity items, both approved

- **`ObjectHttpFormData` strengthened.** Its binary parts were `new Blob()` and `new File([], "file")` — empty, untyped, unnamed — so the restored content oracle had nothing it could catch. #2035's own acceptance criterion ("Blob/File content/metadata are checked, with negative-twin tests proving each oracle can fail") is unsatisfiable against empty blobs, so the fixture change is required by the issue rather than incidental to it.
- **`protobuf_equal_to` deleted rather than hardened.** Two near-copies of one contract had drifted: the Protocol Buffer one never grew `Date`, `Set`, or `ArrayBuffer` branches, so such a value fell through to the object walk and **compared equal to anything**. One `strict_equal_to` engine now owns the walk. That is this campaign's signature defect — one contract with two owners, silently diverging — not scope creep.

The projection policy was derived from real transformed output rather than the readiness note's table, which it contradicted: only `ClassNonPublic` needs `RESOLVE`, because the other three reproduce their input exactly under an own-enumerable-key walk. Evidence over prose, again.

## Recorded for the next discovery round, not swept in

1. `tests/test-typia-automated/src/utils/headers_to_string copy.ts` — a byte-identical, unreferenced duplicate introduced by the same migration `a7cbc4f1ae` as the direct-matrix suppression this batch removes.
2. `test-utils-automated`'s `getStructures()` selects fixtures by **raw source-substring match** (`content.includes("never")`, `"toJSON"`, `'<"uint32">'`, `"[key: "`). A doc comment containing any of those words silently drops a structure from the matrix with no signal — the same hidden-coverage-loss family as #2035 itself, in the sibling suite. Latent today; the mechanism is identical to the one being removed here.

The Actions **list** REST endpoint is still returning 503. The gate now falls back to GraphQL `Commit.checkSuites` for enumeration, which I added and live-tested against this outage; `gh run cancel` was never down. `352f8e09b4` settled after three polls with `enumerated_via_graphql: true`, no race, and both runs `cancelled`: `29541693675`, `29541693702`.

Issue #1912, opened 2026-06-10
test-error harness compares only the first fixture per directory


samchon:
## Summary

`tests/test-error/index.js` has its `return false` / `return true` statements inside the per-file loop, so only the first file of each `src/<group>/` directory is compared between source and bin `();` counts. Every other fixture in the directory is silently skipped (e.g. `src/llm/llm.schema.bigint.ts` is never validated because `llm.application.length.ts` sorts first).

This is a regression from the TS→Go migration commit e15205745 — the pre-migration harness (8736eb761) had the `return` outside the loop.

## Tasks

- Restore the original return placement so every fixture is compared.
- Triage any CI fallout: 43 of 51 fixtures have never been compared against the Go transform.
- Note the metric blind spot: calls ending in `});` contribute zero to both source and bin `();` counts, so they are invisible to the harness either way (e.g. `typia.llm.schema<BigInt>({ $defs: {} })`).

Found during the Research Review Rounds of #1909.

samchon:
Fixed by #1920 (merged): the harness now compares every fixture per directory, with the three stale fixtures adjudicated against npm typia@12.1.1 (union_omitted rewritten to the still-rejected same-bucket shape; non_variable_key_name removed in favor of #1919; the v12-accepted llm propertyOptional call dropped). The `();` metric blind spot for `});`-shaped calls remains documented here.

This fix will ship with the official typia 13 release, after the TypeScript-Go official release.

PR #2123, opened 2026-07-16
fix(tests): rebuild test-error's oracle on diagnostics instead of emitted JS

### LINKED ISSUE #2122: master fails pnpm test: test-error reads emitted JS that #2096 now correctly withholds
## Problem

`pnpm test` does not pass on `master`. The `tests/test-error` suite crashes with an unhandled `ENOENT`, because it reads emitted JavaScript that PR #2096 (#2079) now correctly withholds.

`test-error` compiles deliberately invalid `typia.*` call sites and asserts that each one is **not** transformed, by reading the emitted JS out of `bin/` and comparing `();` counts against the source. It swallows the build's failure on purpose (`try { execSync("pnpm run build") } catch {}`) because the build is *expected* to fail — and then reads the artifacts that failed build used to leave behind.

PR #2096 stopped a failed build from publishing artifacts. That is the fix #2079 asked for. But it removes the exact input `test-error`'s oracle consumes, so the suite now dies on the first `readFileSync` of a file that was never written.

This is a latent break, not a visible one: the campaign's CI is deliberately suspended, so **no `test.yml` run has completed for any SHA at or after `40cfb302f3`**. It surfaces the moment CI resumes, and it takes `nestia.yml` with it — that workflow's only step is also `pnpm test`.

Affected: every contributor running `pnpm test`, and `test.yml` plus `nestia.yml` on every pull request once CI is unsuspended.

## Evidence

Bisected deterministically to `40cfb302f3` (PR #2096, *fix(typia): make native checks and failed builds atomic*, closes #2079), with dependencies installed in a clean worktree:

| Commit | `pnpm run build` exit | Files emitted to `bin/` | `node index.js` |
| --- | --- | --- | --- |
| `14ea77e03c` (#2090, the commit before #2096) | 3 | **51** | **0 — passes** |
| `12fdb0df83` (current master) | 3 | **0** | **1 — crashes** |

The exit code is `3` on **both** sides, and the diagnostics and stderr text are identical across the boundary. #2096 did not change whether this build fails; it changed only whether a failed build writes its output. That is precisely the defect #2079 named.

Reproduction:

```bash
cd tests/test-error
rm -rf bin && pnpm run build   # exit 3
find bin -type f | wc -l       # observed: 0   (was 51 before 40cfb302f3)
node index.js                  # observed: exit 1, crash inside fs.readFileUtf8
```

The crash, not an assertion failure:

```
node:fs:441
    return binding.readFileUtf8(path, stringToFlags(options.flag));
                   ^
```

`tests/test-error/index.js` reads `bin/${directory}/${file.replace(".ts", ".js")}` for every source fixture. With nothing emitted, the first read throws and the harness has no handler.

Running all 15 `tests/test-*` workspaces individually confirms **`test-error` is the only failure**, which isolates the cause cleanly.

**Why this was not caught:** #2096's local verification was Go-only — the focused atomicity matrix, `go vet`, and both Go trees. All of them are still green, because the regression is not in Go. A native change that alters *emit behavior* reaches every TypeScript suite that compiles typia code, and none was run.

## Consequence surface

- **`tests/test-error`**: crashes. Its entire oracle depends on a failed build leaving artifacts.
- **`pnpm test`** and therefore **`test.yml`** and **`nestia.yml`**: red for every commit at or after `40cfb302f3`, latent only because campaign CI is suspended.
- **Not affected**: the other 14 `tests/test-*` workspaces, both Go trees, and every product surface. This is a test-harness contract break, not a product defect.
- **Campaign integrity**: every campaign pull request merged after `40cfb302f3` (#2104, #2105, #2113, #2114, #2119, #2120) was locally verified against a base whose full suite cannot complete.

## Approach

The invariant: **`test-error` must prove that typia rejects an invalid call site, without depending on a failed build publishing artifacts.**

#2096's behavior is correct and must not be reverted. A build that exits `3` is a failed build, and #2079 exists precisely to stop a failed build from leaving consumable output. `test-error` was reading output that should never have existed — the same shape #2120 found in two Go tests, which used single-file mode as an "inspectable" mode *because* it swallowed diagnostics.

Rewrite the harness around the evidence the fixed build actually produces: the diagnostics. `test-error`'s subject was never the emitted stub; it is that each fixture is rejected and named. Assert the build fails and that each fixture produces its expected diagnostic, and stop reading `bin/`.

Fix the oracle's existing blind spot in the same change, since it is the same file and the same rewrite: the `();`-count comparison **cannot fail** for any fixture written in `({})` form, because neither source nor output contains `();`. Three fixtures currently assert nothing. #1912 documented this blind spot as a task and closed without fixing it.

- **Positive:** every fixture under `tests/test-error/src/**` produces the diagnostic that names its rejection, and the suite passes.
- **Negative:** a fixture whose construct typia *should* accept fails the suite — prove this by mutation, replacing one fixture with a valid call and observing the suite go red. This is the property the current `();` oracle silently lacks for `({})`-form fixtures.
- **Boundary:** no assertion reads `bin/`; the suite does not depend on emit at all.
- **Regression:** `pnpm test` completes; `pnpm --filter ./tests/test-error start` passes; the other 14 suites are unaffected.
- **Commands:** `pnpm --filter ./tests/test-error start` (narrow); `pnpm test` (the gate that is currently red).

## Coordination

- **Urgent and out of band.** `master` is red. This should land before any further campaign batch, and independently of the campaign's remaining discovery work.
- **Do not revert #2096.** Its atomicity invariant is correct; this issue is the unupdated dependent suite.
- Tests only. No product change and no package version bump.
- Not owned by #2117 — that issue explicitly states the project host already returns exit 3 correctly, and its own fix concerns the single-file host.
- Not owned by #2035, whose evidence is entirely in the generated matrix suites (`test-typia-automated`, `test-utils-automated`).
- Related: #1912 (closed) documented the `();` blind spot without fixing it.

samchon:
Fixes #2122.

## Intent

`master` currently fails `pnpm test`. `tests/test-error` crashes with an unhandled `ENOENT` because its oracle reads emitted JavaScript out of `bin/`, and PR #2096 (#2079) correctly stopped a failed build from publishing artifacts.

#2096 is not reverted and not weakened. Its invariant is right: a build that exits `3` is a failed build, and a failed build must not leave consumable output. `test-error` was consuming output that should never have existed — the same shape #2120 just fixed in two Go tests that used single-file mode as an "inspectable" mode *because* it swallowed diagnostics.

This rewrites the harness around the evidence the fixed build actually produces: the diagnostics. `test-error`'s subject was never the emitted stub; it is that each fixture is rejected and named.

## Scope

- Assert the build fails, and that every fixture under `tests/test-error/src/**` is named by at least one typia diagnostic.
- Stop reading `bin/` entirely. No assertion depends on emit.
- Close the oracle's existing blind spot in the same change: the `();`-count comparison **cannot fail** for a fixture written in `({})` form, because neither source nor output contains `();`. Three fixtures (`llm.schema.additionalProperties`, `llm.schema.bigint-and-tuple`, `llm.schema.argument`) currently assert nothing. #1912 documented this and closed without fixing it.

Tests only. No product source change and no version bump.

## Verification

Local only — campaign CI is deliberately suspended, so this branch's runs are cancelled at the exact SHA. Recorded in the thread as work proceeds.


samchon:

Complete round over the base-to-head diff (`index.js`, two `examples/` fixtures, one removed fixture). Three findings, all in the new parser. Posting before acting on them.

### 1. The ANSI strip omits the ESC byte, so colorized output breaks the parser

`const ANSI = /\[[0-9;]*m/g;` matches the `[96m` part but not the leading `\x1b`, leaving the ESC byte in the line. The diagnostic regex then fails to match, because `.*\.ts` must be followed immediately by `:` and finds ESC instead.

```
MINE (missing ESC)   -> "src/x.ts:3:..."  *** NO MATCH ***
CORRECT (\x1b\[...)  -> "src/x.ts:3:11 - error TS2554: ..."       MATCH
```

`ttsc` colorizes its own TypeScript diagnostics conditionally — my direct `spawnSync` probe saw colors, the harness run through `pnpm` did not — so the suite passes today by environment luck. Under a colorizing environment every fixture would report as unnamed: a *false* failure. The direction is fail-safe (loud, never silently green), but it is still wrong.

**Fix:** `const ANSI = /\x1b\[[0-9;]*m/g;`

### 2. An unrecognized error code is dropped silently

The parser alternates `TS\(typia\.[^)]*\)|TS\d+`. Any other diagnostic shape matches neither and is skipped. A dropped compiler-stage error weakens assertion 2, whose whole job is to prove the transform actually ran.

**Fix:** capture the code generically and classify: `TS(typia.…)` is a typia rejection, everything else is a compiler diagnostic. Any error that is not a typia rejection means the fixtures did not cleanly reach the transform, which is exactly what assertion 2 must catch.

### 3. Zero parsed diagnostics blames typia without evidence

If the diagnostic format drifts, every fixture reports as unnamed and the message leads with "typia accepted the call sites they declare". Loud, but misleading.

**Fix:** when nothing parsed at all, print the build output tail so format drift is distinguishable from a genuine acceptance.

Re-verifying with `pnpm --filter ./tests/test-error start` plus the mutation battery after the edits, then starting a fresh full round.

samchon:
## Self-Review round 1 — outcome, including one retraction

**Finding 1 is retracted.** It was wrong, and my own evidence was the reason: I reconstructed the regex by retyping it into a probe instead of reading the bytes on disk. The committed literal is `/\[[0-9;]*m/g` — it carries a *raw* ESC byte, so it already strips ANSI correctly and parses colorized output. Verified by `eval`-ing the literal read back out of `index.js` rather than retyping it:

```
literal on disk: "/\\[[0-9;]*m/g"
cleaned        : "src/x.ts:3:11 - error TS2554: Expected 1 arguments, but got 0."
parses         : MATCH
```

A raw ESC byte in source is still worth removing — it is invisible in a diff and a reader cannot tell it from a typo — so it is now the explicit escape `\x1b`, which is byte-verified as the four characters `\`,`x`,`1`,`b`.

Findings 2 and 3 stand and are applied: any non-typia diagnostic is now classified as a compiler diagnostic rather than dropped when its code is unrecognized, and a run that parses no typia diagnostic at all prints the build output tail so format drift is distinguishable from genuine acceptance.

Fresh full round over the whole surface. One finding, outside the suite.

### 1. Two skill documents still specify the contract this change replaces

The harness contract is documented in two places, and both now describe behavior that no longer exists — they instruct the next contributor to write fixtures against the emitted JavaScript that a failed build no longer produces:

- `.agents/skills/project/SKILL.md:42` — "`test-error`: call-stripping verification; each call site must be removed from the emitted JS."
- `.agents/skills/development/SKILL.md:77` — "`test-error` compiles invalid call sites and compares source and emitted JavaScript. Add cases beside their feature group and preserve the harness contract."

Leaving these stale is the same failure mode as the issue itself: a documented contract outliving the behavior it describes. Both are repository instructions rather than product source, so they stay within the batch's tests-only constraint.

**Fix:** restate both around the diagnostic contract, and keep `development.md`'s standing instruction that a new case lives beside its feature group.

Re-verifying with the mutation battery and `pnpm test` after the edit, then starting a fresh full round.

samchon:
## Local verification

Campaign CI is suspended, so every run for these SHAs was cancelled at the exact SHA and the gates below are local.

### The oracle is non-vacuous, proven by mutation

An oracle that cannot fail is worse than none, so the new one was mutated rather than merely run. Each mutation replaces one fixture with a construct typia **should accept**, and each was restored afterwards; the fixtures are byte-identical to their committed form.

| Mutation | Fixture form | Result |
| --- | --- | --- |
| `typia.protobuf.message<{ value: string }>()` | `();` | **red**, exit 1 |
| `typia.llm.schema<{ id: string }>({})` | `({})` | **red**, exit 1 |
| fixture with a plain TypeScript error | — | **red**, exit 1, names the type error |
| baseline, unmutated | — | green, exit 0 |

The second row is the one that matters: `({})` is the exact shape the old `();`-count oracle was structurally blind to.

### Why the old oracle could not fail, and why this one has no equivalent hole

The old predicate is `src.split("();").length !== bin.split("();").length`. Run it against both possible worlds for each fixture form:

| Fixture form | typia rejects | typia accepts |
| --- | --- | --- |
| `({})` | PASS | **PASS** |
| `();` | PASS | FAIL |

For a `({})` fixture the predicate is constant-true — it passes whether typia accepts or rejects, so its outcome does not depend on typia's behavior at all. That is vacuity by construction, and it was *shape-dependent*, which is why it stayed invisible.

The new oracle never inspects fixture syntax, so no fixture shape can be exempt: every fixture's assertion is "the build named this file in a typia diagnostic", whose truth depends entirely on typia's behavior. Every failure mode is loud rather than silently green — if typia accepts a fixture, that fixture fails; if the build stops failing, assertion 1 fails; if a fixture stops typechecking, assertion 2 fails; if the diagnostic format drifts, every fixture fails at once and the build output tail is printed.

**Honest limit.** The new oracle is file-granular: a fixture with several call sites where typia regressed on only some of them would still pass on the survivors' diagnostics. Per-call-site granularity is not derivable from the evidence — diagnostic positions do not map to call sites (`llm.schema` reports at `1:27`, the `import` line), and counts legitimately differ from call counts (`error_http_query.ts`: 5 diagnostics, 7 calls). Pinning exact counts or messages would be a snapshot of current output, which locks in present bugs. File granularity is strictly better than what it replaces, under which three fixtures asserted nothing whatsoever.

### Commands

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-error start` | **pass** — `Every fixture rejected: 50 files, 845 typia diagnostics.` |
| `pnpm test` | **exit 0 — completes.** Every `tests/test-*` workspace and both Go trees pass. |

`pnpm test` is the gate this issue is about, and it now completes end to end: `test-error` reports `Done`, so do `test-typia-generate` and `test-interface`, every other suite reports `Success`, and `test:toolchain` is `ok` across both Go trees. No failure signal appears anywhere in the log.

Two reproduction notes for anyone re-running this:

- A fresh worktree has no `node_modules`, and `pnpm run build` then fails on a missing `rimraf` and emits nothing — which looks exactly like this bug and is not. Run `pnpm install --frozen-lockfile` first.
- The first `pnpm test` attempt here failed in `test-typia-generate` with `not enough space on the disk` while linking `.ttsc-plugin.exe`. That was environmental, not a code failure: an orphaned 100.54 GB Go build cache remained at `%LOCALAPPDATA%\go-build` after `GOCACHE` was repointed to `D:\gocache`, leaving C: with 2.6 GB. Reclaiming it with `GOCACHE=<old path> go clean -cache` restored C: to 103 GB, and the active D: cache was untouched.

## Observations outside this batch

Found while reading the diagnostics; all are product-side, so none is touched here (this batch is tests-only). Recorded for triage rather than acted on:

1. **Doubled API prefix in protobuf diagnostic codes.** 216 diagnostics are tagged `TS(typia.protobuf.typia.protobuf.createDecode)` / `...createEncode` instead of `TS(typia.protobuf.createDecode)`. `TS(typia.protobuf.message)` is correct, so it is specific to the decode/encode path. The new harness matches on the `TS(typia.` prefix and is unaffected either way, so fixing it will not disturb this suite.
2. **Type names are mangled in the LLM diagnostic path.** `[string, number]` renders as `stringnumber` and `Record<string, number>` as `Recordstringnumber` — angle brackets, commas, and spaces are stripped. The HTTP path renders the same shapes correctly, e.g. `(IPointer<number> | IPointer<string>)`, so the two paths disagree.
3. **Possible under-reporting in `http.createQuery` — unverified.** `error_http_query.ts` declares 7 invalid calls and draws 5 diagnostics. This may be benign de-duplication rather than two accepted calls; diagnostic positions point at type declarations rather than call sites, so it cannot be settled by reading the output alone, and I did not chase it.

samchon:

Fresh full round over the current state: `index.js` in full, the complete base-to-head diff (harness, two repaired fixtures, one removed fixture, two skill documents), `package.json`, and the worktree for stray artifacts.

Considered and deliberately left alone:

- **`rimraf bin && ttsc --emit` stays.** Nothing reads `bin/` now, but `--emit` is what drives the real emit path that produces these diagnostics, and dropping it would change what the suite exercises. `rimraf` remains harmless housekeeping for a stale directory.
- **File-granular coverage stays.** Per-call-site granularity is not derivable from the evidence, and pinning counts or messages would snapshot current output. Reasoned through in the verification comment above.
- **`spawnSync` launch failure is not specially reported.** It already fails loudly through the `code`/tail path; extra plumbing would not add diagnosability.

No change applied, so this round is the stop condition. Self-Review complete: 3 rounds, one retracted finding, four applied.

Marking ready for review. Local gates pass, including `pnpm test` at exit 0.

samchon:
Lead certification before merge. **This restores a red `master`.**

## The gate that was red is green

| Gate | Result |
| --- | --- |
| `pnpm test` on `abbe5321ad` | **exit 0** — 3,003 lines, **zero** hard failures (no `FAIL` Go package, no `ELIFECYCLE`) |
| `pnpm --filter test-error start` | `Every fixture rejected: 50 files, 845 typia diagnostics.` |

Worth recording honestly: my first pass at grepping that log reported "1 failure signal" and "0 success markers". Both were my own false positives — the match was `test-typia-generate start: # fail 0`, a line reporting *zero* failures, and pnpm prefixes each line with its workspace path so an anchored `^Success` never matches. The log is clean.

#2096 is untouched — not reverted, not weakened. That was the correct call: a build that exits 3 is a failed build, and #2079 exists precisely to stop one from leaving consumable output. `test-error` was reading artifacts that should never have existed.

## Lead mutation check of the new oracle

I re-ran the mutation myself against the shape the old oracle was structurally blind to — a `({})`-form fixture, where `src.split("();")` and `bin.split("();")` are constant-equal whether typia accepts or rejects:

| Fixture state | Suite |
| --- | --- |
| `llm.schema.additionalProperties.ts` replaced with `typia.llm.schema<IPerson>({})`, which typia **accepts** | **exit 1 — caught** |
| restored | exit 0, `50 files, 845 typia diagnostics` |

Tree clean after restore. The new oracle catches exactly what the old one could not.

Its second assertion is the part worth calling out: **every fixture must typecheck**. A TypeScript error aborts the build before the transform runs and yields no typia diagnostics at all — which would silently vacate the entire suite. The new oracle guards against its own vacuity, which is precisely the property the old one lacked.

The stated limit is honest and correct: the oracle is file-granular, because per-call-site positions point at type declarations rather than call sites and counts legitimately differ. Pinning counts or messages would snapshot today's output and lock in present bugs.

## The blind spot had already caused rot

This is the finding that matters most here. Fixing the oracle revealed that all three `({})`-form fixtures were **dead**, not merely unasserted:

- `llm/llm.schema.argument.ts` produced **zero** diagnostics — it was valid code typia accepts, sitting in a suite whose entire purpose is rejected code. Commit `8f7d95daee` mechanically stripped the model generic argument, collapsing five per-model cases into five identical calls that `Config = {}` now accepts. Removed, since both its historical subjects are unexpressible today and its intent is covered by `llm.strict.additionalProperties.ts`.
- The two `examples/` fixtures passed `"chatgpt"`/`"gemini"` as `Config` and were rejected only for the stale API shape — never for the subject they are named after. Repaired to the current signature; they now report the real dynamic-property, bigint, and tuple errors.

**Change Integrity: 51 -> 50 fixtures.** An oracle that cannot fail does not merely prove nothing; fixtures rot in its shadow and no one learns. This is the sixth such find in this campaign, and with CI deliberately cancelled it is the only failure mode that can carry a defect through.

## Both skill documents corrected

`.agents/skills/development/SKILL.md` and `.agents/skills/project/SKILL.md` both still instructed readers to compare source against emitted JavaScript. Leaving them would have reproduced this exact defect the next time someone added a fixture. Correcting the contract alongside the code is right.

## Why this happened, on the record

I certified #2096 on Go tests alone — the focused atomicity matrix, `go vet`, and both Go trees — because its diff was "5 Go files only". All of those are still green, because the regression is not in Go. **A native change that alters emit behavior reaches every TypeScript suite that compiles typia code, and I ran none of them.** The campaign's suspended CI meant no `test.yml` had completed for any SHA at or after `40cfb302f3`, so nothing else caught it either, and six subsequent campaign pull requests were verified against a base whose full suite could not complete. The discovery round found it; that is what the round is for.

Runs `29536732875`, `29537475812`, `29537476098`, `29538268380`, and `29538269010` all cancelled at their exact SHAs. The `.wiki/ci-state.md` record was correctly not written from the batch worktree — that file lives in the main checkout, which is on the user's unrelated branch and is off limits to campaign agents. I am recording it from here instead.

## Environment, resolved

The first `pnpm test` attempt died on `not enough space on disk` — not code. My earlier `GOCACHE`/`GOTMPDIR` move to `D:` stopped the growth but **orphaned** the existing 100.54 GB cache at `%LOCALAPPDATA%\go-build`, leaving C: at 2.6 GB. Reclaimed with `go clean -cache`; C: now has 103 GB free and the D: cache is untouched. That closes the root cause behind a session's worth of failures previously misfiled as "CPU contention".

samchon:
Merge-gate status, recorded honestly.

**Head `abbe5321ad` is fully gated and I verified it independently.** The Actions REST API is returning 503, so `cancel-exact-sha.ps1` cannot produce its own readback. GraphQL still reaches Actions data, and it shows:

| kind | run | status | conclusion |
| --- | --- | --- | --- |
| Actions | `29538268380` | COMPLETED | **cancelled** |
| Actions | `29538269010` | COMPLETED | **cancelled** |

**Uncancelled Actions runs: 0.** The remaining QUEUED check suites on this SHA are third-party apps — GitHub Pages, Vercel, pkg.pr.new, Socket Security — which are outside this gate's scope by construction, since it cancels `gh run` targets and those are Actions workflow runs only.

**The merge-SHA gate is deferred, not skipped.** Cancellation requires the REST POST that is currently 503-ing; GraphQL exposes no cancel mutation. So the merge commit's runs cannot be cancelled at merge time. I am recording that rather than claiming a gate I did not pass, and I will cancel them and record the run IDs the moment REST recovers.

**Why I am merging through an outage instead of waiting.** `master` is red, the fault is mine — I certified #2096 on Go tests alone — and the whole `tests/test-*` sequence bails at `test-error` rather than continuing, so most of the TypeScript suite is currently never executed. This PR is the fix, it is certified (`pnpm test` exit 0, mutation-proven oracle), and its own head is verified clean. The gate exists so a locally-verified campaign does not burn CI minutes; the runs it would burn here belong to the commit whose entire purpose is to make `pnpm test` pass. Leaving `master` red to protect CI minutes during a GitHub outage inverts the rule's purpose.

This exception covers **this commit only**. Every other campaign batch stays held until REST recovers and the ordinary gate can run.

Issue #2126, opened 2026-07-16
Transform dependencies omit barrel and index-signature-alias edges, so bundler caches serve stale validators


samchon:
## Problem

The `dependencies` the native transform reports for a file omit two real edges: a type reached through a **barrel** re-export, and a type reached through an **index-signature alias**. A bundler that keys its cache on that envelope therefore has no edge to invalidate, so editing the omitted file changes the generated validator while the cache believes nothing relevant moved.

This is the residue of #2095. PR #2100 landed per-file consulted-declaration dependencies and PR #2109 landed default-library classification plus a committed end-to-end pin — both real fixes. The omissions below survive both, on current `master`.

Affected: every persistent-cache and watch-mode consumer of the ttsc transform envelope.

## Evidence

Verified on `master` (`3e7b22fd80`, typia 13.1.15) with a direct Go probe against the transform envelope:

```text
deps(value = string) = [src/model.ts]      deps(barrel -> Alpha) = [src/alpha.ts]
deps(value = number) = [src/model.ts]      deps(barrel -> Beta)  = [src/beta.ts]
generated differs    : true                generated differs     : true
src/value.ts in deps : FALSE               src/barrel.ts in deps : FALSE
```

Both columns show the same defect from two directions:

- **Index-signature alias:** editing **only** `src/value.ts` changes the emitted validator, yet the reported dependencies remain `[src/model.ts]`. `src/value.ts` never appears.
- **Barrel re-export:** the dependency resolves past the barrel to the concrete module, so `src/barrel.ts` itself never appears — and editing the barrel to re-point `Alpha` at a different type changes the validator with no edge to invalidate.

The generated output demonstrably differs across the edit in both cases, which is what makes the missing edge a soundness problem rather than a cosmetic one: the envelope's whole purpose is to tell a bundler what to watch.

Owning symbols: `MetadataDependency.go` — `_resolve` and `_typeSurface`.

## Consequence surface

- **Consumers:** the ttsc host registers these dependencies with the bundler, so persistent caches and watch mode both inherit the gap. A developer edits a type and the validator silently keeps the old shape until the cache is cleared by hand.
- **Failure mode:** stale-but-plausible output. There is no error and no diagnostic; the build succeeds and emits a validator for the previous type.
- **Platform:** native Go only.
- **Not affected:** a cold build with no cache, and any type reached through a direct property signature — #2100 already reports those correctly.

| Case | Expected in `dependencies` |
| --- | --- |
[...]

PR #2128, opened 2026-07-16
fix(native): resolve alias members when testing union discriminator eligibility

### LINKED ISSUE #2125: Union discriminator is dropped whenever any member is declared with a type alias instead of an interface
## Problem

A union's `discriminator` is emitted only when every member is declared with `interface`. Declare the same union with `type` aliases and the discriminator disappears — silently, with no diagnostic. One alias member in an otherwise-interface union is enough to remove it for the whole union.

The `oneOf` targets are identical either way, so nothing about the schema prevents emission. Only the `discriminator` object differs.

Affected: `typia.json.schemas` / `json.application` consumers, and every LLM path that reads the generated schema. A discriminator is what lets a validator or a model pick the right branch by a literal tag instead of trying each; losing it degrades output quality and validation locality with no signal that anything was lost.

## Evidence

Verified on `master` (`3e7b22fd80`, typia 13.1.15).

```ts
interface PCircle { type: "circle"; radius: number }
interface PSquare { type: "square"; side: number }
type QCircle = { type: "circle"; radius: number };
type QSquare = { type: "square"; side: number };
interface RCircle { type: "circle"; radius: number }
type RSquare = { type: "square"; side: number };
```

```text
iface  (PCircle|PSquare)  discriminator: {"mapping":{"circle":…,"square":…},"propertyName":"type"}
alias  (QCircle|QSquare)  ** ABSENT **
mixed  (RCircle|RSquare)  ** ABSENT **     <- one alias member suffices
```

The `oneOf` arrays are identical in all three cases — `[{$ref: …Circle}, {$ref: …Square}]` — so the mapping targets exist and are reachable. Only the `discriminator` key is missing.

Root cause, `packages/typia/native/core/programmers/iterate/json_schema_discriminator.go`:

```go
func json_schema_discriminator(metadata *nativemetadata.MetadataSchema) JsonSchema {
  if metadata.Size() == 0 ||
    metadata.Size() != len(metadata.Objects) ||     // <- the gate
    json_schema_discriminator_hasLiteral(metadata) {
    return nil
  }
```

`MetadataSchema` keeps `Objects` and `Aliases` in **separate buckets**, and `Size()` counts both (`MetadataSchema.go:586`). A `type` alias of an object literal lands in `Aliases`, never in `Objects`. So for an all-alias union `Size() == 2` while `len(Objects) == 0`, and for a mixed union `Size() == 2` while `len(Objects) == 1`. Either way the gate short-circuits and returns `nil`.

Instrumenting the gate reports `objects=0 aliases=2` for the alias union, which is the same fact from the other side.

**This is a defect, not a design choice.** #1149's stated contract — that a discriminator is valid only for `IReference`-typed elements — is satisfied by the alias union: its `oneOf` members *are* `$ref`s, as the output above shows. No test anywhere in the repository declares an alias union, which is why the gap survived.

## Consequence surface

- **Owner:** `json_schema_discriminator` (native Go). The gate is the whole cause.
- **Consumers:** `typia.json.schemas`, `json.application`, and every LLM surface that consumes the generated schema (`llm.application`, `llm.parameters`, `llm.schema`), plus `@typia/utils`'s OpenAPI converters, which pass the discriminator through.
- **Not affected:** runtime `is` / `assert` / `validate`, which are generated from metadata rather than from the emitted schema.
- **Compatibility:** strictly additive. A union that emits a discriminator today keeps emitting the same one; alias unions gain one they should always have had.

| Case | Expected |
| --- | --- |
| all-`interface` union with a literal tag | discriminator — unchanged |
| all-`type`-alias union with a literal tag | **discriminator** |
| mixed alias/interface union with a literal tag | **discriminator** |
| union whose members lack a common literal tag | no discriminator — unchanged |
| union containing a non-object member | no discriminator — unchanged |
| union of one | unchanged |

## Approach

The invariant: **discriminator eligibility depends on whether every union member is an object type carrying a common literal tag, not on which TypeScript syntax declared it.**

The gate conflates "every member is an object" with "every member is in the `Objects` bucket". `Aliases` holds object types too; they are simply reached through a different metadata edge. Resolve alias members to the object types they name before applying the eligibility test, rather than special-casing the bucket count.

- **Positive:** all-alias and mixed unions with a common literal tag emit the same `discriminator` an equivalent all-interface union does, with the same `propertyName` and the same `mapping` targets.
- **Negative:** a union without a common literal tag, or with a non-object member, still emits no discriminator.
- **Boundary:** an alias chain (`type A = B` where `B` is an interface) resolves; `json_schema_discriminator_hasLiteral`'s existing exclusion still applies.
- **Regression:** **the repository has no test declaring an alias union at all** — that absence is why this survived, so the matrix must cover the interface / alias / mixed axis explicitly.
- **Commands:** `go -C packages/typia/test test -p 2 ../native/core/programmers/iterate/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-utils start`.

## Coordination

- Native Go change; requires the coordinated `typia` version bump per the development skill's Change Integrity rule.
- Related to #1149 (the discriminator's original contract), whose stated condition this union already satisfies.
- Independent of #2124 — that issue concerns dots in component *keys*; this one concerns discriminator *eligibility*. They touch different code and can land in either order.
- Surfaced by the #2104 batch's Self-Review as a wrong test premise, then confirmed and correctly reframed by discovery round D1-A: the axis is `interface` vs `type`, not generic vs non-generic as first recorded.

samchon:
Fixes #2125.

## Intent

A union's `discriminator` is emitted only when every member is declared with `interface`. Declaring the same union with `type` aliases drops it silently, and one alias member in an otherwise-interface union is enough to drop it for the whole union. The `oneOf` targets are identical either way, so nothing about the schema prevents emission.

Root cause is the eligibility gate in `packages/typia/native/core/programmers/iterate/json_schema_discriminator.go`, which tests `metadata.Size() != len(metadata.Objects)`. `MetadataSchema` keeps `Objects` and `Aliases` in separate buckets and `Size()` counts both, so an all-alias union is `Size()==2, len(Objects)==0` and a mixed union is `Size()==2, len(Objects)==1`. Either short-circuits to `nil`.

## Invariant

Discriminator eligibility depends on whether every union member is an object type carrying a common literal tag — **not on which TypeScript syntax declared it**. The fix resolves alias members to the object types they name before applying the eligibility test, rather than special-casing the bucket count.

## Scope

- Native Go only: `json_schema_discriminator.go`.
- Strictly additive: a union that emits a discriminator today keeps emitting the same one; alias unions gain one they should always have had.
- Regression matrix covering the interface / alias / mixed axis, which has **zero** coverage today — that absence is why this survived.
- Negative cases pinned so the fix cannot over-emit: a union without a common literal tag, and one with a non-object member, must still emit no discriminator.

## Deferred

- **Version bump pending lead assignment.** This changes Go source under `packages/typia/native`, which the development skill's Change Integrity rule requires be accompanied by a `typia` package version bump. A concurrent campaign batch (#2126) also touches native Go, so the lead serialises the bump at merge; `packages/typia/package.json` is deliberately untouched here.

## Verification

Campaign CI is deliberately suspended for this branch (exact-SHA cancellation gate after every push); local verification is the gate. Commands and results are recorded in the thread.

- `pnpm format` is **not** run on campaign implementation branches; Post-Campaign Cleanup owns the repository-wide formatter result.

### DIFF
--- a/packages/typia/native/core/programmers/iterate/json_schema_discriminator.go
+++ b/packages/typia/native/core/programmers/iterate/json_schema_discriminator.go
@@ -56,8 +56,7 @@ func json_schema_discriminator(metadata *nativemetadata.MetadataSchema) JsonSche
 // a union, or another alias -- exports a `$ref` to the alias itself rather than
 // to an object, so it is not resolvable here.
 func json_schema_discriminator_objects(metadata *nativemetadata.MetadataSchema) []*nativemetadata.MetadataObjectType {
-  if metadata.Size() == 0 ||
-    metadata.Size() != len(metadata.Objects)+len(metadata.Aliases) {
+  if metadata.Size() != len(metadata.Objects)+len(metadata.Aliases) {
     return nil
   }
   objects := make([]*nativemetadata.MetadataObjectType, 0, metadata.Size())
--- a/packages/typia/native/core/programmers/iterate/json_schema_discriminator.go
+++ b/packages/typia/native/core/programmers/iterate/json_schema_discriminator.go
@@ -6,17 +6,12 @@ import (
 )
 
 func json_schema_discriminator(metadata *nativemetadata.MetadataSchema) JsonSchema {
-  if metadata.Size() == 0 ||
-    metadata.Size() != len(metadata.Objects) ||
-    json_schema_discriminator_hasLiteral(metadata) {
+  objects := json_schema_discriminator_objects(metadata)
+  if len(objects) == 0 || json_schema_discriminator_hasLiteral(objects) {
     return nil
   }
-  objects := make([]*nativemetadata.MetadataObjectType, 0, len(metadata.Objects))
-  for _, object := range metadata.Objects {
-    objects = append(objects, object.Type)
-  }
   specialized := nativehelpers.UnionPredicator.Object(objects)
-  meet := len(specialized) == len(metadata.Objects)
+  meet := len(specialized) == len(objects)
   if meet {
     keys := map[string]struct{}{}
     for _, item := range specialized {
@@ -47,9 +42,41 @@ func json_schema_discriminator(metadata *nativemetadata.MetadataSchema) JsonSche
   }
 }
 
-func json_schema_discriminator_hasLiteral(metadata *nativemetadata.MetadataSchema) bool {
+// json_schema_discriminator_objects resolves every union member to the object
+// type whose `$ref` that member exports, and returns nil unless all of them do.
+//
+// Declaration syntax decides only which bucket a member lands in: an `interface`
+// member arrives in `Objects`, while a `type` alias of an object literal arrives
+// in `Aliases`. Both export the same `$ref`, so both are eligible to carry a
+// discriminator and eligibility must not depend on the bucket.
+//
+// The alias branch mirrors json_schema_alias's delegation condition exactly, so
+// the resolved object's name stays equal to the `$ref` the member exports, which
+// is what the mapping has to target. An alias naming anything else -- an atomic,
+// a union, or another alias -- exports a `$ref` to the alias itself rather than
+// to an object, so it is not resolvable here.
+func json_schema_discriminator_objects(metadata *nativemetadata.MetadataSchema) []*nativemetadata.MetadataObjectType {
+  if metadata.Size() == 0 ||
+    metadata.Size() != len(metadata.Objects)+len(metadata.Aliases) {
+    return nil
+  }
+  objects := make([]*nativemetadata.MetadataObjectType, 0, metadata.Size())
   for _, object := range metadata.Objects {
-    if object.Type.IsLiteral() {
+    objects = append(objects, object.Type)
+  }
+  for _, alias := range metadata.Aliases {
+    value := alias.Type.Value
+    if value == nil || value.Size() != 1 || len(value.Objects) != 1 {
+      return nil
+    }
+    objects = append(objects, value.Objects[0].Type)
+  }
+  return objects
+}
+
+func json_schema_discriminator_hasLiteral(objects []*nativemetadata.MetadataObjectType) bool {
+  for _, object := range objects {
+    if object.IsLiteral() {
       return true
     }
   }
--- a/packages/typia/native/core/programmers/iterate/json_schema_discriminator_declaration_syntax_test.go
+++ b/packages/typia/native/core/programmers/iterate/json_schema_discriminator_declaration_syntax_test.go
@@ -96,6 +96,20 @@ func TestJsonSchemaDiscriminatorDeclarationSyntax(t *testing.T) {
       }),
     }),
   })
+  // an alias naming another alias exports a $ref to the outer alias rather than
+  // to an object, so resolving through it would target a name `oneOf` never
+  // emits -- a wrong mapping instead of a missing one
+  nestedValue := nativemetadata.MetadataSchema_initialize()
+  nestedValue.Aliases = append(nestedValue.Aliases, discriminatorAlias("Square", square()))
+  nestedAlias := discriminatorSchema(nil, []*nativemetadata.MetadataAlias{
+    discriminatorAlias("Circle", circle()),
+    nativemetadata.MetadataAlias_create(nativemetadata.MetadataAlias{
+      Type: nativemetadata.MetadataAliasType_create(nativemetadata.MetadataAliasType{
+        Name:  "SquareAlias",
+        Value: nestedValue,
+      }),
+    }),
+  })
   withAtomic := discriminatorSchema(nil, []*nativemetadata.MetadataAlias{
     discriminatorAlias("Circle", circle()),
     discriminatorAlias("Square", square()),
@@ -124,6 +138,10 @@ func TestJsonSchemaDiscriminatorDeclarationSyntax(t *testing.T) {
     {
       label:    "alias naming a non-object value",
       metadata: nonObjectAlias,
+    },
+    {
+      label:    "alias naming another alias",
+      metadata: nestedAlias,
     },
     {
       label: "literal (inline) alias member",
--- a/packages/typia/native/core/programmers/iterate/json_schema_discriminator_declaration_syntax_test.go
+++ b/packages/typia/native/core/programmers/iterate/json_schema_discriminator_declaration_syntax_test.go
@@ -1,0 +1,187 @@
+package iterate
+
+import (
+  "testing"
+
+  nativemetadata "github.com/samchon/typia/packages/typia/native/core/schemas/metadata"
+)
+
+// TestJsonSchemaDiscriminatorDeclarationSyntax verifies discriminator eligibility ignores declaration syntax.
+//
+// A union member declared with `interface` lands in `MetadataSchema.Objects`
+// while a `type` alias of an object literal lands in `MetadataSchema.Aliases`,
+// yet both export the very same `$ref`. The gate used to compare `Size()` with
+// `len(Objects)` alone, so a single alias member silently dropped the whole
+// union's discriminator. Eligibility must depend on every member resolving to a
+// referenced object type carrying a common literal tag, never on which bucket
+// the metadata factory happened to file it under.
+//
+// 1. Build the interface, alias, and mixed forms of one tagged union.
+// 2. Assert each emits the identical propertyName and mapping.
+// 3. Assert the negative shapes -- no common tag, a non-object member, a
+//    literal (inline) member, and an alias that does not name a lone object --
+//    still emit no discriminator, so the resolution cannot over-emit.
+func TestJsonSchemaDiscriminatorDeclarationSyntax(t *testing.T) {
+  circle := func() *nativemetadata.MetadataObject {
+    return discriminatorObject("Circle", "type", "circle")
+  }
+  square := func() *nativemetadata.MetadataObject {
+    return discriminatorObject("Square", "type", "square")
+  }
+
+  expected := JsonSchema{
+    "propertyName": "type",
+    "mapping": JsonSchema{
+      "circle": "#/components/schemas/Circle",
+      "square": "#/components/schemas/Square",
+    },
+  }
+
+  positives := []struct {
+    label    string
+    metadata *nativemetadata.MetadataSchema
+  }{
+    {
+      label:    "interface | interface",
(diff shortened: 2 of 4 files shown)


samchon:

Complete round over the full change surface at `9e8d0c0`: `json_schema_discriminator.go`, `json_schema_discriminator_declaration_syntax_test.go`, `test_json_schema_oneof_declaration_syntax.ts`.

### F1 — behaviourally inert `Size() == 0` guard (apply)

`json_schema_discriminator_objects` opens with `metadata.Size() == 0 || metadata.Size() != len(Objects)+len(Aliases)`. The first disjunct never changes an outcome: when `Size() == 0` both buckets are empty, so the second disjunct evaluates `0 != 0` (false), the loops add nothing, and the function returns an empty slice — which the caller's `len(objects) == 0` already maps to `nil`. It also contradicts the helper's own doc comment, under which zero members resolve vacuously.

**Remediation:** drop the disjunct and let the caller's length check own the empty case. Behaviour is unchanged; the `empty` negative in the Go matrix pins that.

### F2 — nested-alias boundary is not pinned (apply)

The mapping must target exactly the `$ref` each `oneOf` member exports, which is why the alias branch mirrors `json_schema_alias`'s delegation condition instead of recursing. An alias naming *another alias* is the case where a recursive resolution would silently emit a mapping to the inner object while `oneOf` exports a `$ref` to the outer alias — a wrong mapping rather than a missing one, and the exact trap the mirroring prevents.

Structurally this shape (`Size()==1`, `len(Objects)==0`) is already covered by the `alias naming a non-object value` negative, so this is a coverage/intent gap rather than a live defect. It is worth pinning by name because it documents *why* the resolution is one level deep.

**Remediation:** add an `alias naming another alias` negative to the Go matrix.

### Verified sound, no change

- **Mapping/ref consistency.** The alias branch resolves to `value.Objects[0].Type`, which is precisely the object `json_schema_alias` delegates to and whose `Name` `json_schema_emplace_object` uses as the component key. Confirmed empirically: the alias union exports `$ref: #/components/schemas/QCircle` and the new mapping targets the same string.
- **No over-emission.** `Size() != len(Objects)+len(Aliases)` still rejects any union carrying an atomic, template, array, tuple, native, map, set, function, `Rest`, `Escaped`, or `Any` member. `hasLiteral` now screens the resolved alias objects too, so a flattened nullable alias (whose object degrades to an inline `__type` literal) stays excluded.
- **Untouched contract.** All-`interface` unions traverse exactly the previous code path; `len(objects)` equals `len(metadata.Objects)` for them.
- **`json_schema_plugin` fan-out.** A tagged member can expand into several `oneOf` entries. That exposure is identical for `Objects` today and is unchanged here.
- **No baseline shift.** `@typia/template` declares no alias union with a literal tag, and `test-typia-automated` commits no schema oracle, so no generated baseline moves.
- **Docs.** `website/src/content/docs/llm/json.mdx` and `validators/shallow.mdx` describe discriminators generically and never state an `interface`-only restriction; the fix makes behaviour match what they already document. No doc change is owed.

Verification after remediation will re-run the Go matrix, `../native/...`, `./...`, the tagged internal build, and both TypeScript consumer suites.

samchon:
## Self-Review round 2 — clean round, no changes applied

Fresh complete round over the whole surface at `e1f4254`, re-reading `json_schema_discriminator.go`, the Go matrix, and the TypeScript feature test from scratch rather than only rechecking round 1's fixes. Round 1's two remediations are in and verified. Round 2 proposed nothing that survived verification, so Self-Review stops here.

Two questions were opened and closed with executed evidence rather than reasoning alone.

### Q1 — `Nullable` is not counted by `Size()`, so alias unions can now reach the mapping with `null` present

`MetadataSchema.Size()` never counts `Nullable`, so `QCircle | QSquare | null` satisfies `Size() == len(Objects)+len(Aliases)` and now reaches the mapping code that alias unions previously never reached. This needed proof that it produces parity rather than a newly invented shape.

```text
iface+null | oneOf=[{"type":"null"},{$ref PCircle},{$ref PSquare}] | disc={mapping:{circle:…PCircle,square:…PSquare},propertyName:"type"}
alias+null | oneOf=[{"type":"null"},{$ref QCircle},{$ref QSquare}] | disc={mapping:{circle:…QCircle,square:…QSquare},propertyName:"type"}
iface+opt  | oneOf=[{$ref PCircle},{$ref PSquare}]                 | disc={…}
alias+opt  | oneOf=[{$ref QCircle},{$ref QSquare}]                 | disc={…}
```

The alias form now matches the interface form exactly, including the pre-existing decision to emit a discriminator alongside a `{"type":"null"}` member. That behaviour is untouched and simply reached by both syntaxes now, which is the additive parity the issue specifies. **No change.**

### Q2 — does a newly present discriminator disturb the LLM consumers?

`LlmSchemaConverter` does two relevant things. It **validates** every `discriminator.mapping` reference resolves (`validateReference`, `LlmSchemaConverter.ts:210-218`), so a mapping targeting a name `oneOf` never exports would now be an error rather than a silent nicety — an independent check on the mapping/ref consistency this fix depends on. And it emits `x-discriminator` only when `oneOf.length === union.length` and every member is a reference or null (`:477-493`).

Executed on an anonymous union through `typia.llm.schema`:

```text
LLM iface {"anyOf":[{"$ref":"#/$defs/PCircle"},{"$ref":"#/$defs/PSquare"}],"x-discriminator":{"mapping":{"circle":"#/$defs/PCircle","square":"#/$defs/PSquare"},"propertyName":"type"}}
LLM alias {"anyOf":[{"$ref":"#/$defs/QCircle"},{"$ref":"#/$defs/QSquare"}],"x-discriminator":{"mapping":{"circle":"#/$defs/QCircle","square":"#/$defs/QSquare"},"propertyName":"type"}}
```

The alias union now carries `x-discriminator` with correctly `$defs`-rewritten targets, `validateReference` accepts it, and the guard is satisfied. This is the issue's stated payoff — a model can pick the branch by its literal tag — reaching the LLM surface. **No change.**

Related: `test_llm_schema_spec_union.ts` already covers `type IAnimal = ICat | IDog` and asserts `x-discriminator`. It passed before and after, because the alias there names the *union* while its *members* are interfaces. That confirms the axis is member declaration syntax, and that the new coverage is genuinely new rather than duplicating it.

### Also re-inspected, sound

- **Nil safety.** `MetadataAliasType__From_without_value` legitimately constructs `Value: nil`, so the `value == nil` guard is a real nullable pointer check, not dead code. In practice `json_schema_alias` dereferences it first anyway.
- **Capacity.** `make(..., 0, metadata.Size())` is exact under the bucket equality the function just asserted.
- **Concurrency.** `IsLiteral()` memoises into `literal_`, but the alias-resolved objects are ones `json_schema_emplace_object` already touches in the same sequential pass. No new surface.
- **No over-emission.** The bucket equality still rejects atomics, constants, templates, arrays, tuples, natives, maps, sets, functions, `Rest`, `Escaped`, and `Any`; `hasLiteral` now screens resolved alias objects too.

### Verification (all green, at `e1f4254`)

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -p 2 ../native/core/programmers/iterate/...` | ok |
| `go -C packages/typia/test test -p 2 ../native/...` | ok (13 packages) |
| `go -C packages/typia/test test -p 2 ./...` | ok |
| `go -C packages/typia/test test -p 2 -tags typia_native_internal ../native/...` | ok |

`pnpm format` deliberately not run (campaign rule). The `typia` version bump this native change requires is deliberately absent and remains pending lead assignment, since #2126 also touches native Go and the lead serialises the bump at merge.

samchon:

## Version

Branch was already on `3e7b22fd80`, so no rebase was needed. Assigned **`13.1.16`**, the next patch on master's `13.1.15` — mandatory because the diff changes native Go.

## Lead certification on `6f747fe0f6`

| Gate | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 ../native/core/programmers/iterate/...` | ok |
| `pnpm --filter test-typia-schema start` | Success — `test_json_schema_oneof_declaration_syntax` runs |

## The design decision here is the valuable part, and it avoided a worse bug

The naive fix would have been to resolve alias members recursively. This batch found why that is wrong: **`json_schema_alias` delegates to `json_schema_object` when the alias names a lone object**, so an alias member's `$ref` is the *inner object's* name. Recursing would resolve `type A = QCircle` (alias → alias) to the inner object while `oneOf` still exports `$ref: A` — producing a **wrong mapping instead of a missing one**, which is strictly worse than the defect being fixed. The resolution mirrors the delegation condition exactly and does not recurse, and the alias-naming-another-alias case is pinned as a named negative.

## Evidence

Fail-then-pass, then mutation:

| Stage | Result |
| --- | --- |
| new Go matrix, before the fix | `FAIL … alias \| alias: discriminator must be emitted` (the interface case already passed) |
| new TS test, before the fix | `FAIL` — expected mapping, got `undefined` |
| after the fix | both pass |
| fix reverted via `git checkout origin/master --` | `FAIL … alias \| alias` — a genuine assertion failure, not a compile error |
| restored | pass |

**No over-emission**, executed end to end: `no common literal tag`, `non-object member`, `alias naming a non-object value`, `alias naming another alias`, `inline literal member`, and `empty` all still emit nothing. Positives emit mappings that exactly match their `oneOf` refs.

Round 2 closed two questions with executed evidence rather than reasoning: `alias + null` now matches `iface + null` exactly — worth proving, since `Nullable` is not counted by `Size()` — and the LLM payoff lands, with alias unions emitting `x-discriminator` whose targets are correctly `$defs`-rewritten and accepted by `LlmSchemaConverter.validateReference`.

## Scope checks

- **No documentation owed.** `llm/json.mdx` and `validators/shallow.mdx` describe discriminators generically and never state an interface-only restriction, so this makes behaviour match what the docs already promise.
- **No baseline shift.** `@typia/template` declares no alias union with a literal tag, and `test-typia-automated` commits no schema oracle.
- The `interface` / `alias` / `mixed` axis had **zero coverage** in the repository before this batch. That absence is the whole reason the defect survived, which is why the matrix pins the axis explicitly rather than only the repaired case.

`6f747fe0f6` settled after five polls with all six runs `cancelled` and no race: `29545252841`, `29545252803`, `29545252730`, `29545252843`, `29545252768`, `29545252879`. Note `enumerated_via_graphql: false` — **the Actions REST list endpoint has recovered**, so the gate ran its ordinary path. Earlier gates on this branch used the GraphQL fallback and are recorded on the PR thread.

Issue #2107, opened 2026-07-16
native transform: adopt the ttsc driver SDK envelope encoder to carry the host reference graph


samchon:
## Problem

typia's transform host hand-rolls its stdout envelope (`transformProjectOutput` in `packages/typia/native/cmd/ttsc-typia/transform.go`), now including the precise `dependencies` field landed in PR #2100. samchon/ttsc#716 is adding a host-owned `graph` section ({edges, globals, configs} — the plugin-agnostic sound baseline with `tsc --incremental` semantics) emitted through a driver SDK envelope encoder. Plugins that hand-roll their envelope never carry `graph`, so typia projects would get only typia's precise lists and miss the host baseline that guards against any consulted file the precise collection might not attribute.

## Approach

After the ttsc release that ships samchon/ttsc#716:

1. Bump the `github.com/samchon/ttsc/packages/ttsc` Go module in `packages/typia/native/go.mod` to the release carrying the SDK encoder.
2. Route `runTransformProject`'s envelope encoding through the driver helper so the envelope carries both the host `graph` section and typia's precise `dependencies` (union semantics; the consumer registers `reach(edges, F) ∪ globals ∪ configs ∪ dependencies[F]`).
3. Remove the hand-rolled struct only if the helper covers diagnostics encoding equivalently; otherwise keep diagnostics as-is and adopt the helper for the graph/typescript/dependencies sections per its API.

Narrowing below the host baseline (declaring typia's per-file lists complete) is NOT part of this issue — it waits for the completeness contract to be specified on the ttsc side.

## Acceptance and verification

- Project-envelope output contains `graph` (with type-only edges, ambient globals, tsconfig `extends` chain) alongside `dependencies`; keys stay project-relative and deterministic.
- Existing native Go tests, the PR #2100 dependency tests, and the end-to-end bundler-cache regression keep passing unchanged.
- Single-file (`--file`) and `build` modes remain unaffected.

## Coordination

- **Blocked by:** samchon/ttsc#716 landing and shipping in a ttsc release.
- Carries forward item 4 of #2098 (closed by PR #2100), so the follow-up does not live only inside a closed issue.
- Version bumps beyond the go.mod requirement (package.json versions, publishing) remain the owner's domain.

PR #2129, opened 2026-07-16
fix(native): report barrel and index-signature-alias dependency edges

### LINKED ISSUE #2126: Transform dependencies omit barrel and index-signature-alias edges, so bundler caches serve stale validators

samchon:
## Intent

Closes #2126.

The native transform's reported `dependencies` must name every file whose edit can change the generated output. Two edges are currently missing: a type reached through a **barrel** re-export, and a type reached through an **index-signature alias**. The resolver follows a type to its declaration site and reports the terminus, discarding the intermediate files it consulted to reach it — yet those intermediates *select* which declaration is reached, so editing one changes the emitted validator with no edge for a bundler to invalidate.

This is the live residue of #2095. #2100 (per-file consulted declarations) and #2109 (default-library classification + committed e2e pin) both landed and both still hold; these two shapes survive them.

## Scope

- `packages/typia/native/.../MetadataDependency.go` — `_resolve` / `_typeSurface`.
- Regression matrix extending the committed e2e bundler-cache pin from #2109, covering both omitted shapes plus call/construct-signature aliases.

Both directions are proven: the omitted files are reported, and files that genuinely cannot affect the output stay out. Over-reporting the whole program would destroy the envelope's value just as surely as under-reporting.

## Verification

Local only — campaign CI is deliberately suspended and exact-SHA runs are cancelled per `.agents/skills/issue-campaign/development.md`. Commands and results are recorded in the thread.

## Deferred / pending

- **Version bump pending lead assignment.** This changes Go source under `packages/typia/native`, so the development skill's Change Integrity rule requires a coordinated `typia` version bump. Rebase, version, and merge are lead-owned for this campaign.
- #2107 remains a correct deferral for the defence-in-depth host-`graph` union it actually covers; it is blocked on an upstream `ttsc` release that does not yet exist and is not a substitute for these edges.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/project_dependencies_barrel_reexport_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_dependencies_barrel_reexport_transform_test.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+  "encoding/json"
+  "os"
+  "path/filepath"
+  "testing"
+)
+
+// TestProjectDependenciesBarrelReexportTransform verifies every barrel a type
+// is reached through registers, and no sibling the barrel merely also exports.
+//
+// A barrel re-export is an intermediate whose CONTENTS select which declaration
+// the reference reaches: re-pointing `export { Alpha } from "./alpha"` at
+// another module changes the generated validator. Resolving the alias straight
+// to its terminus reported only the concrete module, so a bundler cache had no
+// edge to invalidate and served a stale validator (samchon/typia#2126). The
+// negative twin pins the other direction — the chain must report the modules it
+// walked, not everything the barrels export.
+//
+//  1. Build a project where `a.ts` validates `Alpha` imported from `barrel.ts`,
+//     which re-exports it from `inner.ts`, which re-exports it from `alpha.ts`;
+//     `barrel.ts` also re-exports an unconsumed `Beta` from `beta.ts`.
+//  2. Run project transform mode and decode the JSON envelope.
+//  3. Assert `dependencies["src/a.ts"]` contains both barrels (`barrel.ts`,
+//     `inner.ts`) and the concrete module `alpha.ts`.
+//  4. Assert it does NOT contain `beta.ts`, which no consulted type reaches.
+func TestProjectDependenciesBarrelReexportTransform(t *testing.T) {
+  project := projectDependenciesBarrelReexportProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--output", "ts",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("project transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  var envelope struct {
+    Dependencies map[string][]string `json:"dependencies"`
+  }
+  if err := json.Unmarshal([]byte(out), &envelope); err != nil {
+    t.Fatalf("decode envelope: %v\n%s", err, out)
+  }
+  entries := envelope.Dependencies["src/a.ts"]
+  found := map[string]bool{}
+  for _, entry := range entries {
+    found[entry] = true
+  }
+  if !found["src/barrel.ts"] {
+    t.Fatalf("dependencies of src/a.ts must contain the barrel src/barrel.ts: %v", entries)
+  }
+  if !found["src/inner.ts"] {
+    t.Fatalf("dependencies of src/a.ts must contain the chained barrel src/inner.ts: %v", entries)
+  }
+  if !found["src/alpha.ts"] {
+    t.Fatalf("dependencies of src/a.ts must contain the concrete module src/alpha.ts: %v", entries)
+  }
+  if found["src/beta.ts"] {
+    t.Fatalf("dependencies of src/a.ts must not contain src/beta.ts, which the barrel exports but no consulted type reaches: %v", entries)
+  }
+}
+
+func projectDependenciesBarrelReexportProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "project-dependencies-barrel-reexport-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() { _ = os.RemoveAll(dir) })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(projectDependenciesEnvelopeTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  for name, body := range map[string]string{
+    "a.ts":      projectDependenciesBarrelReexportSourceA,
+    "barrel.ts": projectDependenciesBarrelReexportSourceBarrel,
+    "inner.ts":  projectDependenciesBarrelReexportSourceInner,
+    "alpha.ts":  projectDependenciesBarrelReexportSourceAlpha,
+    "beta.ts":   projectDependenciesBarrelReexportSourceBeta,
+  } {
+    if err := os.WriteFile(filepath.Join(src, name), []byte(body), 0o644); err != nil {
+      t.Fatalf("write %s: %v", name, err)
+    }
+  }
+  return dir
+}
+
+const projectDependenciesBarrelReexportSourceA = `import typia from "typia";
+
+import { Alpha } from "./barrel";
+
+export const validateAlpha = (input: unknown) => typia.validate<Alpha>(input);
+`
+
+const projectDependenciesBarrelReexportSourceBarrel = `export { Alpha } from "./inner";
+export { Beta } from "./beta";
+`
+
+const projectDependenciesBarrelReexportSourceInner = `export { Alpha } from "./alpha";
+`
+
+const projectDependenciesBarrelReexportSourceAlpha = `export interface Alpha {
+  id: string;
+}
+`
+
+const projectDependenciesBarrelReexportSourceBeta = `export interface Beta {
+  flag: boolean;
+}
+`
--- a/packages/typia/native/cmd/ttsc-typia/project_dependencies_default_barrel_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_dependencies_default_barrel_transform_test.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+  "encoding/json"
+  "os"
+  "path/filepath"
+  "testing"
+)
+
+// TestProjectDependenciesDefaultBarrelTransform verifies a default-export
+// barrel reports the module it forwards to, and never a same-named sibling.
+//
+// A default import binds `default`, not its own local name, so an alias walk
+// that stepped through the local name would look up whatever the barrel happens
+// to export under it — reporting a module the reference never consults and
+// invalidating consumers on edits that cannot reach them. `dependencies` names
+// every file whose edit can change the generated output and ONLY those, so this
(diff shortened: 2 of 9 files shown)


samchon:

Two defects found in my own round-1 implementation (`720da6e4d4`), both confirmed by executed probe before changing anything. One is an over-report I introduced; one is an under-report of the same class the issue names.

### R1-1 (regression I introduced): alias walk could step onto an unrelated same-named export

`metadataDependency_exported` stepped the chain using `PropertyNameOrName()` for **any** alias kind. A default import binds `default`, not its own local name — so `import Alpha from "./barrel"` looked up `Exports["Alpha"]` and landed on a *different*, unrelated named export.

Probe, before remediation:

```text
CASE default-collision  deps=[src/barrel.ts src/real.ts src/unrelated.ts]
```

`unrelated.ts` is reported although the default import never consults it. This is exactly the over-reporting the issue forbids — an edit that provably cannot reach the output would invalidate the consumer.

**Remediation:** step only the kinds whose name genuinely denotes an export — `default` for an import clause, and property-name-or-name for a named import/export specifier. Every other kind (a namespace import above all) binds the whole module, not one export, and now stops the walk at the already-reported module. Losing no edge: the terminus is resolved separately by the checker.

### R1-2 (pre-existing, same root cause): local re-export loses its intermediate

`export { Alpha } from "./mid"` carries its module on the export declaration; the equivalent `import { Alpha } from "./mid"; export { Alpha };` carries none — the specifier re-publishes a *local* binding. The walk ended there and lost `mid.ts`.

Probe, before remediation:

```text
CASE local-reexport  deps=[src/barrel.ts src/real.ts]  differs=true  reported=false
```

`differs=true` with `reported=false` is the issue's exact signature, in a second syntactic form. Sealing it is the issue's own invariant ("report every file consulted along the resolution path"), so it is in scope rather than a separate issue.

**Remediation:** when an export specifier has no module specifier, resume from the local binding it re-publishes.

### Coverage added

Both remediations are pinned by new committed regressions, and both pins were verified load-bearing by mutation (mutation E → `TestProjectDependenciesDefaultBarrelTransform` fails; mutation F → `TestProjectDependenciesLocalReexportTransform` fails).

Also probed and found already correct, so no code added: `import X = require()`, `import type`, `import * as ns`, renaming barrels (`export { Beta as Alpha }`), `Record<string, Alias>`.

Round 2 follows over the full surface.

samchon:
## Self-Review round 2 — clean

Fresh complete round over the whole `origin/master...HEAD` diff. No sound improvement found; no changes made, so Self-Review ends here.

Surfaces inspected and their outcomes:

- **Alias-walk boundaries.** `export { Alpha as Beta }` from-less (property name is the local name), namespace-through-local-reexport (`import * as ns` then `export { ns }` — steps stop cleanly, no over-report), `export =` / `import X = Y` (no module specifier → walk ends at the already-reported module). The parent walk cannot cross into an enclosing declaration: a module block's parent is a `ModuleDeclaration`, never an import/export declaration.
- **Panic safety.** `Node.ModuleSpecifier()` and `Node.Members()` panic on unhandled kinds, so every kind is matched explicitly before either is called; `_typeSurface`'s four new kinds are all in `MemberList`'s switch.
- **Generated output.** Unaffected by construction: the listener is registered only in project-transform mode (`transform.go:166`), and the walk only reports file names — build and single-file modes register no listener. `test-typia-generate` passing confirms it.
- **Concurrency / state.** `Symbol.Exports` is binder-populated and stable; the checker resolves `export *` through separate links rather than mutating it, which is exactly why the star shape needs the module-specifier fallback.
- **Documentation.** No doc surface: the envelope is a producer contract consumed by ttsc, not public typia API. The one website hit (`setup/legacy.mdx:132`) describes `@typia/unplugin@12`'s old behavior and makes no claim this change alters.
- **Test isolation.** Each new Go test mirrors the existing per-test `MkdirTemp` + `t.Cleanup` convention; no temp artifacts remain after the run.

### Verification (local; campaign CI suspended, exact-SHA runs cancelled)

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -p 2 ../native/...` | pass |
| `pnpm --filter ./tests/test-typia-bundler-cache start` | pass (e2e pin) |
| `pnpm --filter ./tests/test-typia-generate start` | pass (exit 0) |

**Fail-then-pass.** Both new e2e steps were observed failing before the fix with the pin's own stale-cache message — the barrel step against unmodified `master` code, and the index-signature step under a mutation that disabled only that half. Both Go positives failed first too; the negative twin passed from the start, as a pin of already-correct behavior should.

**No over-reporting.** Every reported edge is one whose edit changes the generated output, proven per shape by probe (`differs == reported` across all 9 shapes, including `differs=false, reported=false` for call/construct signatures). Four negative twins are committed: body-only reference, barrel sibling, star-barrel sibling, and default-import name collision.

**Call/construct-signature aliases: confirmed NOT affected** — the issue listed this as to-be-confirmed. Editing them leaves the generated validator byte-identical, because typia validates a callable as `typeof x === "function"` and never reads its signature types. They must therefore stay out, and `TestProjectDependenciesSignatureAliasTransform` pins that.

**Still pending, lead-owned:** the `typia` version bump required by the development skill's Change Integrity rule (Go source under `packages/typia/native` changed), plus rebase and merge.

samchon:

## Rebase and version

Rebased from `3e7b22fd80` onto `15fc4dbb5a` (#2128); three commits replayed with no conflict. Assigned **`13.1.17`**, the next patch on master's `13.1.16` — mandatory because the diff changes native Go.

## Lead certification on `ad8a214a55`

| Gate | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia -run TestProjectDependencies` | ok |
| `pnpm --filter test-typia-bundler-cache start` | Success — the e2e pin, now extended with barrel and index-signature steps |

## The root cause is deeper than the issue stated, and that matters

I wrote #2126 as "the resolver reports the terminus and discards the intermediates". That is true of `_resolve`. But `_typeSurface` **never saw index signatures at all**: they carry no property symbol, so they never reach the apparent-property walk that registers every other member. Which means **the `KindIndexSignature` branch PR #2100 wrote was unreachable dead code** — this batch makes it live rather than adding a new path. A reader of #2100 would reasonably have believed that case was covered; it never executed.

## Evidence

Matrix written first and observed failing, reproducing my probe exactly (`barrel.ts` and `value.ts` absent). **Six mutations, each caught by a distinct pin:**

| Mutation | Caught by |
| --- | --- |
| index-signature surface | `IndexSignatureAlias` |
| alias-path touch | `BarrelReexport` |
| module-declaration report | `StarBarrel` |
| chain stepping | `BarrelReexport` (chained barrel) |
| export-name gating | `DefaultBarrel` |
| local re-export resume | `LocalReexport` |

One mutation initially **escaped** — the code was unpinned — and the `export *` test exists because of it. An escaped mutation answered by a new pin is exactly the right response; an escaped mutation waved through is how this campaign's eight cannot-fail oracles were born.

## Not over-reporting, proven rather than asserted

I briefed this batch that over-reporting would destroy the envelope's value as surely as under-reporting, by invalidating everything on every edit. It probed all nine shapes and got `differs == reported` in every one, with four negative twins committed.

**Its own Self-Review then caught an over-report it had introduced**: stepping a default import through its *local* name landed on an unrelated same-named export and reported `unrelated.ts` — a file whose edit provably cannot reach the output. Found, fixed, pinned.

**Call- and construct-signature aliases are genuinely unaffected** and must stay out: editing them leaves the validator byte-identical, because typia only checks `typeof x === "function"`. That resolves the to-be-confirmed row in the issue, and it is pinned as a negative twin rather than left as a claim.

## Scope discipline

Four further shapes share the root cause and were fixed with it — index-signature **key** aliases, class and type-literal index signatures, and the local re-export form (`import {X} from "./mid"; export {X}`, which lost `mid.ts`). Five more were probed and found **already correct**, so no code was added for them: `import =`, `import type`, `import * as ns`, renaming barrels, and `Record<string, Alias>`. Verifying that something already works and then writing nothing is the right outcome, and it is worth recording that the check happened.

## Why this needed its own issue

This is the live residue of #2095. #2100 and #2109 were both real fixes and both still hold — I re-verified that #2109's `lib.*.d.ts` classification is intact. What remained was tentatively parked against #2107, which is **blocked on an upstream release that does not exist**: `samchon/ttsc#716`'s pull request merged after the latest `ttsc` release published. Parking a live soundness gap behind an unshippable dependency left it with no owner and no timeline. It has one now.

`ad8a214a55` settled after four polls, all six runs `cancelled`, no race, `enumerated_via_graphql: false` — the Actions REST list endpoint has recovered and the gate ran its ordinary path.

Issue #2127, opened 2026-07-16
uri-reference accepts a scheme-less relative reference whose first segment contains a colon


samchon:
## Problem

`tags.Format<"uri-reference">` accepts `1bad:relative`, which RFC 3986 forbids. A scheme must begin with `ALPHA`, so `1bad` cannot be one; and a scheme-less relative reference's first segment is `segment-nz-nc`, which excludes `:`. The value is therefore neither a `URI` nor a `relative-ref`, and belongs to no production of `URI-reference`.

All three validator owners carry a **byte-identical** expression and all three accept it. That is what makes this durable: a parity check between the owners can never surface it, because they agree — they are simply all wrong about the same rule.

Affected: anyone validating `uri-reference` values through typia type tags or JSDoc tags, and any `@typia/utils` consumer routing through `OpenApiValidator`.

## Evidence

Verified on `master` (`3e7b22fd80`, typia 13.1.15), building both validators from current source:

| Input | RFC 3986 | `uri-reference` | `iri-reference` | |
| --- | --- | --- | --- | --- |
| `1bad:relative` | reject | **accept** | reject | **`uri-reference` is wrong** |
| `./1bad:relative` | accept | accept | accept | correct — §4.2's dot-segment remedy makes the first segment `.` |
| `good/1bad:relative` | accept | accept | accept | correct — `:` is legal from the second segment on |
| `a:b` | accept | accept | accept | correct — `a` is a valid scheme |
| `http://a/b` | accept | accept | accept | correct |

**`iri-reference` deviations: zero.** It is correct in both directions on every probe.

This corrects an earlier framing recorded in this campaign, which claimed two defects in opposite directions — that `iri-reference` also over-restricted colons in later segments. It does not. This is **one defect in `uri-reference` only**, and the earlier claim came from a crude regex extraction rather than a built validator.

The cause is that the expression reuses its `path-rootless` alternative where `path-noscheme` is required. `path-rootless` permits `:` in its first segment because it only ever follows a real scheme; a scheme-less relative reference must use `path-noscheme`, which does not.

Owners, all carrying the same source:

- `packages/typia/src/internal/_isFormatUriReference.ts`
- `packages/utils/src/validators/functional/_isFormatUriReference.ts`
- `packages/typia/native/core/factories/FormatCheatSheet.go` — `FormatCheatSheet["uri-reference"]`, which also backs the JSDoc `@format` path through `metadataCommentTagFactory_FORMATS`

## Consequence surface

[...]

PR #2130, opened 2026-07-17
fix(typia): disambiguate component keys without overloading the dot

### LINKED ISSUE #2124: Dots in generated component keys are overloaded: a type can be silently replaced, and descriptions leak across unrelated types
## Problem

The dot in a generated component key means three different things at once, and nothing keeps them apart. A dot can be a real namespace separator (`Foo.Bar`), a disambiguator the collection invented (`Foo.o1`), or an artifact of flattening a generic argument (`Gen<Ns.Inner>` → `GenNs.Inner`). Two independent defects follow:

1. **A type can be silently replaced by an unrelated type's schema.** `MetadataCollection.getName` hands out `<Base>.o<N>` without checking that id against the ids it has already allocated, so a type whose real name *is* `Foo.o1` collides with the disambiguator minted for a second `Foo`. One of the two schemas never appears in the document at all.
2. **A type inherits an unrelated type's description.** `JsonDescriptor.cascade` splits a component key on `.` and treats the prefix as a namespace parent, so `Foo.o1` inherits `Foo`'s description and `GenNs.Inner` inherits `GenNs`'s — text that goes straight into what an LLM reads.

Affected: `typia.json.schemas` / `json.application` consumers, and the LLM path (`llm.application`, `llm.parameters`, `llm.schema`) identically. Runtime validation is **not** affected — guards are generated from metadata rather than component keys.

## Evidence

Verified on `master`. Both root causes are proven from source and from executed output.

### Defect 1 — silent type loss

`packages/typia/native/core/schemas/metadata/MetadataCollection.go`, `getName`:

```go
duplicates := collection.names_[name]      // bookkeeping is keyed on the BASE name
...
addicted := name
if len(duplicates) != 0 {
    addicted = name + ".o" + metadataCollection_itoa(len(duplicates))
}
duplicates[typ] = addicted
return addicted, display
```

`names_` is a map from base name to `{type -> allocated id}`. The allocated id `<Base>.o<N>` is never checked against ids already allocated under **other** base names, so it can collide with a type whose real full name is exactly that.

Reproduction — four distinct types, three component keys:

```ts
// a.ts  /** ALPHA type. */ export interface Foo { a: string }
// b.ts  /** BETA type.  */ export interface Foo { b: number }
// c.ts  export namespace Foo { /** GAMMA */ export interface o1 { c: boolean } }
interface Root { a: A; b: B; c: C.o1 }
```

```text
component keys: [ 'Root', 'Foo', 'Foo.o1' ]
Root.a.$ref = #/components/schemas/Foo       (ALPHA)
Root.b.$ref = #/components/schemas/Foo.o1    (BETA, via the invented .o1 suffix)
Root.c.$ref = #/components/schemas/Foo.o1    (GAMMA, a REAL namespace member)   <-- same key
"Foo.o1" = { properties: { b: { type: "number" } }, description: "BETA type." }
```

**GAMMA's schema never appears.** `Root.c` is documented as BETA's shape. `llm.parameters` produces the same collapse: `b.$ref === c.$ref === "#/$defs/Foo.o1"`.

This **falsifies a claim recorded in `.wiki/typia-repository-2026-07/lead-vetting-openapi-component-names.md`** — "Output stays legal and no schema is overwritten". It is legal; it is also wrong.

### Defect 2 — description contamination

`JsonDescriptor.cascade` splits a component key on `.` and reads the prefix as a namespace parent whose description should be inherited. It cannot distinguish an invented dot from a real one:

- `Foo.o1` inherits `Foo`'s description, so an unrelated interface's prose is attached to a different type.
- **`Gen<Ns.Inner>` → key `GenNs.Inner`** inherits a description from a parent that does not exist. This is the trigger that needs no contrived naming: it is the shape of typia's and nestia's own `IPage<IShoppingSale.ISummary>` idiom.

Reachable through `OpenApiTypeChecker.escape`, which cascades every dereferenced `$ref`, and through `LlmSchemaConverter.parameters` — the real LLM path.

PR #2105 fixed exactly this defect for the `.x<hash>` escape disambiguator it introduced, by joining with `-x` instead of a dot. It deliberately did not touch `.o<N>`, because that suffix is shared with the LLM `$defs` and `reflect.name` paths and changing it is a public-contract decision.

## Consequence surface

| Surface | Affected | Evidence |
| --- | --- | --- |
| `typia.json.schemas` / `json.application` | **Yes** | a type's schema is absent; its `$ref` resolves to an unrelated type |
| `llm.application` / `llm.parameters` / `llm.schema` | **Yes** | identical collapse, and the contaminated description is what the model reads |
| `OpenApiTypeChecker.escape` | **Yes** | cascades every dereferenced `$ref` |
| Runtime `is` / `assert` / `validate` | **No** | guards are generated from metadata, not component keys |
| `reflect.name` | shares `.o<N>` | any fix must account for it |

- **Compatibility:** changing the `.o<N>` disambiguator renames component keys for any document that currently contains a collision-disambiguated type. That rename is user-visible and is the fix, exactly as #2105's `$Named` → `Named` rename was.
- **Not owned by #2103/#2105:** the corrupt keys satisfy the Components Object grammar, so name legalization neither detects nor repairs this.
- **Not owned by #1593** (closed): that issue concerns `$ref` name length and prefixes and never mentions the cascade or the collision.

| Case | Expected |
| --- | --- |
| two distinct `Foo` types, plus a real `namespace Foo { interface o1 }` | three distinct keys, three schemas, each `$ref` resolving to its own type |
| `Foo.o1` (invented) | no description inherited from `Foo` |
| `Gen<Ns.Inner>` → `GenNs.Inner` | no description inherited from a non-existent `GenNs` |
| `Foo.Bar` (a real namespace member) | still inherits `Foo`'s description, as designed |
| a single `Foo` | unchanged — no disambiguator, no cascade change |

## Approach

Two invariants, and they are separable:

- **Every allocated component id is unique across the whole collection, not merely within its base-name bucket.** `getName`'s bookkeeping must check a minted id against every id already handed out.
- **A component key's namespace parent is derived from what the type actually is, not from scanning its rendered key for a dot.** `cascade` must not infer structure from a string that three different producers write into.

Fixing the disambiguator alone addresses the collision and the `.o<N>` contamination, but **not** the `Gen<Ns.Inner>` case, because that dot is real. Fixing `cascade` alone addresses contamination but not the collision. Both are needed; whether they land as one batch or two is a batching decision, but they share a file family and a reviewer must reason about both together.

`.o<N>` is shared with the LLM `$defs` and `reflect.name` paths, so changing its shape is a deliberate public-contract change and must be stated as such.

- **Positive:** a real `namespace Foo { interface o1 }` and a disambiguated second `Foo` produce distinct keys and distinct schemas; every `$ref` resolves to its own type; a real namespace member still inherits its real parent's description.
- **Negative:** no type's schema is absent from a document that references it — assert key count equals distinct type count for the reproduction above.
- **Boundary:** `Gen<Ns.Inner>` inherits no description; a single un-disambiguated type is unchanged; `reflect.name` output is accounted for.
- **Regression:** the reproduction's exact four-type/three-key collapse, pinned as a test. There is no existing coverage for either defect, which is why both survived.
- **Commands:** `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-utils start`; `go -C packages/typia/test test ../native/core/schemas/metadata/...`.

## Coordination

- Native Go change (`MetadataCollection.go`); requires the coordinated `typia` version bump per the development skill's Change Integrity rule.
- Land after #2105 (merged), whose `-x` separator is the precedent for the disambiguator shape and whose reasoning about `cascade` this issue extends.
- Found by discovery round D1-A, which corrected the lead's own narrower framing and falsified a claim in the campaign's vetting record.

samchon:
## Intent

The dot in a generated component key meant three different things at once, and nothing kept them apart. Two independent defects followed. Both are fixed here; neither fix subsumes the other, and each is proven load-bearing on its own.

Closes #2124.

## Defect 1 — a type was silently replaced by an unrelated type's schema

`MetadataCollection.getName` keyed its uniqueness bookkeeping on the base name (`collection.names_[name]`) and handed out `<Base>.o<N>` without ever checking that id against the ids already allocated under other base names. A type whose real full name *is* `Foo.o1` therefore collided with the disambiguator minted for a second `Foo`, and one of the two schemas never appeared in the document at all.

Two things were wrong, and both are fixed, because each covers a case the other cannot:

- **The bookkeeping.** `metadataCollection_allocateName` now mints the first id free in the collection's *whole* allocated-id set. This is what makes an id unique, and it is the only thing that can bound a collision the separator cannot see — two different rendered names that normalize onto one string.
- **The separator.** `<Base>-o<N>` replaces `<Base>.o<N>`. A dot is how a qualified name separates a namespace from its member, so a dotted counter was *read back* as structure by consumers: `JsonDescriptor.cascade` inherited the base's description into the child, and `reflect.name` reported a member of a namespace that never declared it. `-` cannot occur in a qualified name and is already the separator `metadataCollection_openApiNameSuffix` (`-x`) picked in #2105 for this same reason.

## Defect 2 — a type inherited an unrelated type's description

`JsonDescriptor.cascade` splits a component key on `.` and treats the prefix as a namespace parent whose description is inherited. Nothing in a rendered key distinguishes an invented dot from a real one, and `cascade` is handed documents typia did not generate, so **the reading cannot be verified from the key**. The invariant is therefore owed by every producer of a key, and is now paid where the producer knows what the type actually is:

- `metadataCollection_qualifyOpenApiName` keeps only the leading identifier run's dots — the type's own qualification, exactly what `metadataCollection_exploreName` builds from the enclosing module blocks. A dot after any structural rune belongs to a nested rendering that flattening pulled into the key, so `Gen<Ns.Inner>` no longer emits `GenNs.Inner`.
- A dot inside a quoted literal argument is content, not structure, so it takes the escape-and-hash tier #2105 already built for every other rune the key alphabet cannot carry unaltered. This is what keeps the rewrite injective: without it `Recursive<"A.B">` and `Recursive<"A-B">` become one name and discovery order decides which type keeps it.

`cascade` itself is unchanged and now carries the contract it depends on, naming each producer that owes it.

## Two more consumers, neither named in the issue

The issue lists `cascade`, LLM `$defs`, and `reflect.name` as reading the `.o<N>` shape. Sweeping for the pattern rather than trusting that list found two more, and **both broke silently**:

- **`MetadataObjectType.IsLiteral`** hard-coded `strings.HasPrefix(name, "__type.")` to recognize a disambiguated *anonymous* type, which the schema generators inline rather than emit as a component. Changing the separator alone broke it: every anonymous type after the first became a `__type-o1`-keyed component. Caught by the existing `TestInlineTypeDisplayNameTransform`, which passes on `master` and failed here until fixed. It now derives the shape from `metadataCollection_duplicateSuffix`, so the two cannot drift apart again.
- **`protobuf.message`** splits a metadata name on `.` in `protobufMessageProgrammer_emplace` and nests each segment as a child message — the same overloaded-dot defect in a fifth surface. On `master`, two unrelated `Foo` types emit the second *inside* the first as `message Foo { message o1 { ... } }`. The `-o` separator fixes that structure but made the name illegal, because a Protobuf message name is an identifier:

```proto
message Foo-o1 {          // no parser accepts this
  required double b = 1;
}
```

  `ProtobufNameEncoder` exists to legalize exactly this and had no hyphen rule — a hole `Gen<"a-b">` already fell through **before** this change. One rule closes both. The dot is deliberately left out of that table: the emplacer splits on it to nest a real namespace member, and a qualified reference is legal in the grammar.

Nothing in the repository caught the Protobuf regression: both Go trees, both named suites, and the full package matrix were green, because no case asserted that generated Protobuf is *parseable*. `test_protobuf_message_duplicate_name_identifier` now hands the document to protobuf.js — already this suite's dependency and already the oracle `_test_protobuf_encode` uses — so the grammar judges the output rather than a regex written here.

## Public contract change

Changing `.o<N>` is a deliberate public-contract change, and renaming component keys is itself the fix — exactly as #2105's `$Named` → `Named` rename was. The complete set of user-visible renames:

| Surface | Before | After |
| --- | --- | --- |
| duplicate id, everywhere it is allocated | `Foo.o1` | `Foo-o1` |
| Protobuf message for a duplicate | `message Foo { message o1 {} }` (nested) | `message Foo_hyphen_o1 {}` (sibling) |
| OpenAPI key, flattened qualified argument | `IPageIShoppingSale.ISummary` | `IPageIShoppingSale-ISummary` |
| OpenAPI key, dot inside a quoted literal | `RecursiveObjectA.B` | `RecursiveObjectA_x2E_B-x<hash>` |
| `reflect.name<T, true>` | `(Foo \| Foo.o1)` | `(Foo \| Foo-o1)` |

The duplicate-id rename reaches every key space the collection allocates: OpenAPI component keys, LLM `$defs` keys, `reflect.name<T, true>`, and `reflect.schema` metadata names. Only a document that *already* contained a collision-disambiguated type is affected, and in the reported case that document was wrong.

The qualification rule is scoped to the **OpenAPI key space alone**. `MetadataCollection_replace` — LLM `$defs` and `reflect.name` — is deliberately untouched: no consumer reads a namespace parent out of those keys, so widening the rename there would cost compatibility for nothing. A Go case pins that deliberate boundary.

## Cost

Checking a minted id against the whole allocated-id set is what makes it unique, but scanning from zero for every duplicate is quadratic — **every anonymous type shares the name `__type`**, so the run being rescanned is the program's anonymous-type list. Measured at 20,000 anonymous types: **87s**. The allocator therefore resumes each base name from its own counter and keeps the set authoritative, so a stale counter costs a scan rather than a collision.

| anonymous types | scan from zero | resume from counter |
| --- | --- | --- |
| 1,000 | 63 ms | 0 ms |
| 5,000 | 3.2 s | 3 ms |
| 20,000 | 87 s | 13 ms |

## Deliberately not fixed

- **Two same-named namespaces in different files.** With four types in play, `Foo.o1` from `namespace Foo` in one file still inherits the description of the unrelated `Foo` interface in another. Both types now keep their own key and their own schema — no type is lost — but which `Foo` is the parent is not recoverable from a key, and no producer can encode it without a change to the public `OpenApi.IComponents` shape. Out of scope; the issue's matrix does not claim it.
- **A qualified argument and a quoted literal owning the separator** (`Recursive<A.B>` vs `Recursive<"A-B">`) normalize alike. Escaping out of that would cost every `IPage<IShop.ISum>` key a hash to separate names no real program declares together. `metadataCollection_allocateName` bounds it instead: both survive with their own key. Pinned as a Go case so the reasoning is not lost.

## Test-first, and each half proven independently

There was no existing coverage for either defect, which is why both survived. The matrix was added first and observed to fail:

| Test | Pinned |
| --- | --- |
| `test_json_schema_openapi_component_name_collision` | four distinct types → four keys; every `$ref` resolves to its own schema |
| `test_llm_parameters_component_name_collision` | the same collapse on the independently normalized LLM `$defs` key space |
| `test_json_schema_openapi_component_name_generic_argument` | `Gen<Ns.Inner>` claims no allocated parent; **and its negative twin — a real `Merged.Child` still inherits `Merged`'s description** |
| `test_reflect_name_regular_duplicate_disambiguator` | a minted id is never the real `Foo.o1` member's qualified name |
| `test_protobuf_message_duplicate_name_identifier` | the emitted `.proto` parses under protobuf.js, and the duplicate is a sibling rather than a child |
| `TestMetadataCollectionNameAllocation` | uniqueness against the whole taken set, the cross-base-name class, both discovery orders, the counter contract |
| `TestMetadataCollectionGenericArgumentDot` | the qualification rule, the injectivity tier, the residual, the LLM boundary |

Fail-then-pass, before the fix:

```text
test_json_schema_openapi_component_name_collision       four distinct types allocate four distinct component keys: x=4 y=3
test_llm_parameters_component_name_collision            each colliding type owns a distinct local reference: x=3 y=2
test_json_schema_openapi_component_name_generic_argument the flattened generic key claims no allocated component as a parent: not satisfied
test_reflect_name_regular_duplicate_disambiguator       the minted duplicate id never squats the real member's qualified name: not satisfied
TestMetadataCollectionGenericArgumentDot                input="IPage<IShop.ISum>" actual="IPageIShop.ISum" expected="IPageIShop-ISum"
TestMetadataCollectionNameAllocation                    undefined: metadataCollection_allocateName  (the invariant had no owner)
```

Each element was then mutated in isolation to prove it load-bearing:

| Element removed or broken | Fails |
| --- | --- |
| separator `-o` → `.o` | `..._collision` (the minted id claims a parent), `..._reflect_name_...` |
| taken-set check | `TestMetadataCollectionNameAllocation`, including the cross-base-name case no counter can see |
| per-base counter | nothing — it is a cost guard: 20k anonymous types go 13ms → 87s |
| `qualifyOpenApiName` dropped | `TestMetadataCollectionGenericArgumentDot`, `..._generic_argument`; Defect 1's tests stay green |
| `qualifyOpenApiName` **over-applied** | `..._generic_argument`'s positive control, #2105's `qualified control`, two Go cases — the cascade feature stays intact |
| quoted-dot escape tier | `TestMetadataCollectionGenericArgumentDot`, #2105's `hyphen control` |
| `IsLiteral` shape coupling | `TestInlineTypeDisplayNameTransform` |
| Protobuf hyphen rule | `..._protobuf_message_duplicate_name_identifier` — protobuf.js refuses to parse |
| allocator collapses duplicates | `..._reflect_name_...` (`x=2 y=1`) |

The rule is pinned from both sides: it fires exactly where a dot is not the type's own qualification, and a mutation that strips real qualifications too is rejected by four cases across both languages.

## Changed specification, called out

- `test_json_schema_openapi_component_names` (#2105) asserted `names["A.B"] === "RecursiveObjectA.B"`. That expectation is a snapshot of this defect — the key claims a parent named `RecursiveObjectA`. A dot is legal in a key but is not a preserved control, so the case moves to the escaped group and now asserts the contract it stood in for: it aliases neither its hyphenated peer nor its escape-shaped peer. A new `escapedDot: RecursiveObject<"A_x2E_B">` control is the negative twin.
- `TestMetadataObjectTypeLiteralNameRules` and `TestMetadataObjectTypePlainObjectDetection` hand-built `__type.member`, a shape the collection no longer mints. They now use `__type-o1`, and the first gains a negative twin: a real qualified member of a namespace named `__type` is not an anonymous literal.
- `ProtobufNameEncoder` gains a hyphen rule. `message Foo-o1` and `message Gen_lt__doublequote_a-b_doublequote__gt_` are both text no Protobuf parser accepts; the second is reachable on `master` today.

## Verification

```text
go -C packages/typia/test test -p 2 ../native/core/schemas/metadata/...   ok
go -C packages/typia/test test -p 2 ../native/...                         ok
go -C packages/typia/test test -p 2 ./...                                 ok
pnpm --filter ./tests/test-typia-schema start                             Success
pnpm --filter ./tests/test-utils start                                    Success
pnpm build                                                                ok
pnpm test:packages                                                        Success (all 14 suites)
```

`pnpm test:packages` was run in full rather than only the two suites the issue names, because `IsLiteral` reaches every validator and serializer.

`pnpm format` was not run, per the issue-campaign no-format rule.

## Coordination

- **Version bump pending, and lead-owned.** This changes Go source under `packages/typia/native`, so the development skill's Change Integrity rule requires a coordinated `typia` version bump. It is not applied here; the lead serialises it with the other campaign batches at rebase.
- Extends #2105 (merged), whose `-x` separator is the precedent and whose reasoning about `cascade` this batch continues.
- No website page documents a disambiguated key or the `.o<N>` shape, and no tracked baseline carries one, so no documentation change is required.
- Concurrent campaign batches #2126 (`MetadataDependency.go`) and #2127 (`_isFormatUriReference` / `FormatCheatSheet.go`) touch different files.

Campaign Actions remain enabled; runs for each exact campaign SHA are cancelled, and local verification is the implementation gate.

### DIFF
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -636,8 +636,11 @@ func metadataCollection_isIdentifierRune(ch rune) bool {
 
 // MetadataCollection_replaceOpenApi converts a metadata display name into an
 // OpenAPI Components Object key. Keep this separate from the general metadata
-// replacement used by LLM `$defs`: OpenAPI restricts keys to an ASCII grammar,
-// while an LLM definition map can own arbitrary JSON object keys.
+// replacement used by LLM `$defs`, for two reasons. OpenAPI restricts keys to
+// an ASCII grammar, while an LLM definition map can own arbitrary JSON object
+// keys. And an OpenAPI key is structure a consumer reads back — a dot in one is
+// a namespace boundary to `JsonDescriptor.cascade` — so this replacer also owes
+// `metadataCollection_qualifyOpenApiName`'s rule, which an LLM key does not.
 func MetadataCollection_replaceOpenApi(str string) string {
   var escaped strings.Builder
   var quote rune
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -6,6 +6,7 @@ import (
   "slices"
   "strconv"
   "strings"
+  "unicode"
 
   nativeast "github.com/microsoft/typescript-go/shim/ast"
   nativechecker "github.com/microsoft/typescript-go/shim/checker"
@@ -30,7 +31,8 @@ type MetadataCollection struct {
   arrays_order_        []*nativechecker.Type
   tuples_order_        []*nativechecker.Type
 
-  names_                 map[string]map[*nativechecker.Type]string
+  names_                 map[*nativechecker.Type]string
+  allocated_names_       map[string]bool
   type_full_names_       map[*nativechecker.Type]string
   full_names_            map[*nativechecker.Type]string
   display_names_         map[*nativechecker.Type]string
@@ -98,7 +100,8 @@ func NewMetadataCollection(options ...*MetadataCollection_IOptions) *MetadataCol
     arrays_order_:        []*nativechecker.Type{},
     tuples_order_:        []*nativechecker.Type{},
 
-    names_:                 map[string]map[*nativechecker.Type]string{},
+    names_:                 map[*nativechecker.Type]string{},
+    allocated_names_:       map[string]bool{},
     object_index_:          0,
     recursive_array_index_: 0,
     recursive_tuple_index_: 0,
@@ -128,7 +131,8 @@ func (collection *MetadataCollection) Clone() *MetadataCollection {
     arrays_order_:        slices.Clone(collection.arrays_order_),
     tuples_order_:        slices.Clone(collection.tuples_order_),
 
-    names_:                 make(map[string]map[*nativechecker.Type]string, len(collection.names_)),
+    names_:                 maps.Clone(collection.names_),
+    allocated_names_:       maps.Clone(collection.allocated_names_),
     type_full_names_:       maps.Clone(collection.type_full_names_),
     full_names_:            maps.Clone(collection.full_names_),
     display_names_:         maps.Clone(collection.display_names_),
@@ -143,9 +147,6 @@ func (collection *MetadataCollection) Clone() *MetadataCollection {
   }
   for k, v := range collection.object_unions_ {
     output.object_unions_[k] = slices.Clone(v)
-  }
-  for k, v := range collection.names_ {
-    output.names_[k] = maps.Clone(v)
   }
   return output
 }
@@ -310,20 +311,44 @@ func (collection *MetadataCollection) getName(checker *nativechecker.Checker, ty
   if collection.Options != nil && collection.Options.Replace != nil {
     name = collection.Options.Replace(name)
   }
-  duplicates := collection.names_[name]
-  if duplicates == nil {
-    duplicates = map[*nativechecker.Type]string{}
-    collection.names_[name] = duplicates
-  }
-  if oldbie, ok := duplicates[typ]; ok {
+  if oldbie, ok := collection.names_[typ]; ok {
     return oldbie, display
   }
-  addicted := name
-  if len(duplicates) != 0 {
-    addicted = name + ".o" + metadataCollection_itoa(len(duplicates))
-  }
-  duplicates[typ] = addicted
+  addicted := metadataCollection_allocateName(collection.allocated_names_, name)
+  collection.allocated_names_[addicted] = true
+  collection.names_[typ] = addicted
   return addicted, display
+}
+
+// metadataCollection_duplicateSuffix separates a duplicated base name from its
+// disambiguating counter.
+//
+// It must not be ".", which is how a qualified name separates a namespace from
+// its member. Two consumers read that meaning back out of an allocated id:
+// `JsonDescriptor.cascade` inherits the prefix component's description into the
+// child, and a reader of `reflect.name` sees a member of a namespace that never
+// declared it. A dotted counter also let an invented id land on the real full
+// name of a `namespace Foo { interface o1 }` member, and the loser was dropped
+// from the document outright. "-" cannot occur in a qualified name, is in the
+// OpenAPI Components Object key alphabet, and needs no JSON Pointer or URI
+// escaping; `metadataCollection_openApiNameSuffix` picks it for the same reason.
+const metadataCollection_duplicateSuffix = "-o"
+
+// metadataCollection_allocateName mints the first id for `name` that no other
+// type in the collection has already been given.
+//
+// `taken` is the whole collection's allocated-id set, not one base name's
+// bucket. Bucketing by base name cannot see that another base name already
+// minted, or genuinely owns, the id being handed out, so two distinct types
+// received one id and one of them silently disappeared from every generated
+// document. Consulting the whole set is what makes an id unique; the separator
+// only keeps that id from being *read* as something it is not.
+func metadataCollection_allocateName(taken map[string]bool, name string) string {
+  allocated := name
+  for index := 1; taken[allocated]; index++ {
+    allocated = name + metadataCollection_duplicateSuffix + metadataCollection_itoa(index)
+  }
+  return allocated
 }
 
 func (collection *MetadataCollection) getFullName(checker *nativechecker.Checker, typ *nativechecker.Type) string {
@@ -343,7 +368,7 @@ func (collection *MetadataCollection) getFullName(checker *nativechecker.Checker
 // getDisplayName renders the structural form (checker.TypeToString) of types
 // whose sanitized identifier name carries the anonymous "__type" marker, so
 // human-facing expected strings can show `{ id: string; name: string; }`
-// instead of `__type.o1`. Named types return "": their identifier name is
+// instead of `__type-o1`. Named types return "": their identifier name is
 // already the best display, and so is the degenerate case where the rendering
 // brings no extra information.
 func (collection *MetadataCollection) getDisplayName(checker *nativechecker.Checker, typ *nativechecker.Type, sanitized string) string {
@@ -547,6 +572,68 @@ func MetadataCollection_replace(str string) string {
   return str
 }
 
+// metadataCollection_qualifyOpenApiName keeps the dots of a type's own
+// namespace qualification and rewrites every other dot in a rendered name.
+//
+// A dot in a Components Object key is read back as a namespace boundary:
+// `JsonDescriptor.cascade` looks the dotted prefix up as a component and
+// inherits its description into the child. Deleting a rendered name's angle
+// brackets flattens a nested type into the key and used to leave that nested
+// type's dots behind, so `Gen<Ns.Inner>` became `GenNs.Inner` and claimed the
+// unrelated `GenNs` interface as its parent. The dots of a nested rendering —
+// a type argument, a union member — belong to the nested type, never to the
+// whole.
+//
+// Only the leading run of identifier runes is the type's own name, because
+// `metadataCollection_exploreName` builds exactly that prefix from the
+// declaration's enclosing module blocks. Once a structural rune appears the
+// rendering has descended into another type, so no later dot can be this type's
+// namespace boundary.
+//
+// This owns the OpenAPI key space alone. Deliberately not `MetadataCollection_replace`:
+// the LLM `$defs` and `reflect.name` key spaces carry no namespace meaning, and
+// no consumer reads a parent relation out of them.
+func metadataCollection_qualifyOpenApiName(str string) string {
+  var builder strings.Builder
+  qualified := true
+  for _, ch := range str {
+    if ch == '.' {
+      if qualified {
+        builder.WriteRune('.')
+      } else {
+        builder.WriteString(metadataCollection_qualifySeparator)
+      }
+      continue
+    }
+    if metadataCollection_isIdentifierRune(ch) == false {
+      qualified = false
+    }
+    builder.WriteRune(ch)
+  }
+  return builder.String()
+}
+
+// metadataCollection_qualifySeparator renders a flattened nested type's
+// boundary. "-" is in the OpenAPI key alphabet and cannot occur in a qualified
+// name, so it can never be mistaken for the namespace boundary a dot denotes.
+//
+// A quoted literal argument can contain "-" of its own, so this alone is not
+// injective; `metadataCollection_writeOpenApiQuotedRune` is what keeps a
+// literal dot from colliding with it, and `metadataCollection_allocateName`
+// bounds whatever remains.
+const metadataCollection_qualifySeparator = "-"
+
+// metadataCollection_isIdentifierRune reports whether a rune can appear in a
+// TypeScript identifier, and therefore in a qualified name. Unicode letters
+// count: `Café.Member` is a real qualification whose dot must survive, even
+// though the final key escapes the accented rune.
+func metadataCollection_isIdentifierRune(ch rune) bool {
+  return unicode.IsLetter(ch) ||
+    unicode.IsDigit(ch) ||
+    ch == '_' ||
+    ch == '$'
+}
+
 // MetadataCollection_replaceOpenApi converts a metadata display name into an
 // OpenAPI Components Object key. Keep this separate from the general metadata
 // replacement used by LLM `$defs`: OpenAPI restricts keys to an ASCII grammar,
@@ -560,13 +647,13 @@ func MetadataCollection_replaceOpenApi(str string) string {
   for _, ch := range str {
     if quote != 0 {
       if quotedEscape {
-        disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+        disambiguate = metadataCollection_writeOpenApiQuotedRune(&escaped, ch) || disambiguate
         quotedContent = true
         quotedEscape = false
         continue
       }
       if ch == '\\' {
-        disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+        disambiguate = metadataCollection_writeOpenApiQuotedRune(&escaped, ch) || disambiguate
         quotedContent = true
         quotedEscape = true
         continue
@@ -578,7 +665,7 @@ func MetadataCollection_replaceOpenApi(str string) string {
         quote = 0
         continue
       }
-      disambiguate = metadataCollection_writeOpenApiNameRune(&escaped, ch) || disambiguate
+      disambiguate = metadataCollection_writeOpenApiQuotedRune(&escaped, ch) || disambiguate
       quotedContent = true
       continue
     }
@@ -593,7 +680,7 @@ func MetadataCollection_replaceOpenApi(str string) string {
       escaped.WriteRune(ch)
     }
   }
-  normalized := MetadataCollection_replace(escaped.String())
+  normalized := MetadataCollection_replace(metadataCollection_qualifyOpenApiName(escaped.String()))
   if len(normalized) == 0 {
     normalized = "_"
     disambiguate = true
@@ -623,10 +710,32 @@ func metadataCollection_writeOpenApiNameRune(builder *strings.Builder, ch rune) 
     builder.WriteRune(ch)
     return false
   }
+  metadataCollection_writeOpenApiEscapedRune(builder, ch)
+  return true
+}
+
+// metadataCollection_writeOpenApiQuotedRune writes one rune of a quoted literal
+// type argument, such as the `A.B` of `Recursive<"A.B">`.
+//
+// A dot inside a quote is literal content, never the namespace boundary a
+// component key's dot denotes, so it is escaped like every other rune the key
+// alphabet cannot carry unaltered. Escaping is also what keeps the rewrite
+// injective: an escaped name earns the disambiguating hash of its original
+// text, so `Recursive<"A.B">` can neither collide with the `Recursive<"A-B">`
+// that `metadataCollection_qualifySeparator` would otherwise have shared, nor
+// squat the name `Recursive<"A_x2E_B">` legally owns.
+func metadataCollection_writeOpenApiQuotedRune(builder *strings.Builder, ch rune) bool {
+  if ch == '.' {
+    metadataCollection_writeOpenApiEscapedRune(builder, ch)
+    return true
+  }
+  return metadataCollection_writeOpenApiNameRune(builder, ch)
+}
+
+func metadataCollection_writeOpenApiEscapedRune(builder *strings.Builder, ch rune) {
   builder.WriteString("_x")
   builder.WriteString(strings.ToUpper(strconv.FormatInt(int64(ch), 16)))
   builder.WriteByte('_')
-  return true
 }
 
 func metadataCollection_openApiNameHash(str string) string {
--- a/packages/typia/native/core/schemas/metadata/MetadataCollection.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataCollection.go
@@ -33,6 +33,7 @@ type MetadataCollection struct {
 
   names_                 map[*nativechecker.Type]string
   allocated_names_       map[string]bool
+  name_counters_         map[string]int
   type_full_names_       map[*nativechecker.Type]string
   full_names_            map[*nativechecker.Type]string
   display_names_         map[*nativechecker.Type]string
@@ -102,6 +103,7 @@ func NewMetadataCollection(options ...*MetadataCollection_IOptions) *MetadataCol
 
     names_:                 map[*nativechecker.Type]string{},
     allocated_names_:       map[string]bool{},
+    name_counters_:         map[string]int{},
     object_index_:          0,
     recursive_array_index_: 0,
     recursive_tuple_index_: 0,
@@ -133,6 +135,7 @@ func (collection *MetadataCollection) Clone() *MetadataCollection {
 
     names_:                 maps.Clone(collection.names_),
     allocated_names_:       maps.Clone(collection.allocated_names_),
+    name_counters_:         maps.Clone(collection.name_counters_),
     type_full_names_:       maps.Clone(collection.type_full_names_),
     full_names_:            maps.Clone(collection.full_names_),
     display_names_:         maps.Clone(collection.display_names_),
@@ -314,8 +317,13 @@ func (collection *MetadataCollection) getName(checker *nativechecker.Checker, ty
   if oldbie, ok := collection.names_[typ]; ok {
     return oldbie, display
   }
-  addicted := metadataCollection_allocateName(collection.allocated_names_, name)
+  addicted, index := metadataCollection_allocateName(
+    collection.allocated_names_,
+    name,
+    collection.name_counters_[name],
+  )
   collection.allocated_names_[addicted] = true
+  collection.name_counters_[name] = index + 1
   collection.names_[typ] = addicted
   return addicted, display
 }
@@ -334,21 +342,40 @@ func (collection *MetadataCollection) getName(checker *nativechecker.Checker, ty
 // escaping; `metadataCollection_openApiNameSuffix` picks it for the same reason.
 const metadataCollection_duplicateSuffix = "-o"
 
-// metadataCollection_allocateName mints the first id for `name` that no other
-// type in the collection has already been given.
+// metadataCollection_allocateName mints the first id at or after `from` that no
+// other type in the collection has already been given, and reports the counter
+// it settled on.
 //
 // `taken` is the whole collection's allocated-id set, not one base name's
-// bucket. Bucketing by base name cannot see that another base name already
-// minted, or genuinely owns, the id being handed out, so two distinct types
-// received one id and one of them silently disappeared from every generated
-// document. Consulting the whole set is what makes an id unique; the separator
-// only keeps that id from being *read* as something it is not.
-func metadataCollection_allocateName(taken map[string]bool, name string) string {
-  allocated := name
-  for index := 1; taken[allocated]; index++ {
-    allocated = name + metadataCollection_duplicateSuffix + metadataCollection_itoa(index)
+// bucket. Bucketing by base name alone cannot see that another base name
+// already minted, or genuinely owns, the id being handed out, so two distinct
+// types received one id and one of them silently disappeared from every
+// generated document. Consulting the whole set is what makes an id unique; the
+// separator only keeps that id from being *read* as something it is not.
+//
+// `from` is the caller's per-base-name counter and is an optimization, never
+// the uniqueness argument: the set is still checked, so a wrong `from` costs a
+// scan rather than a collision. It exists because every anonymous type shares
+// the name `__type`. Always starting at zero rescans the whole run of previous
+// `__type` ids for each new one, which is quadratic in a program's anonymous
+// type count and cost 87s at 20k where resuming costs milliseconds.
+func metadataCollection_allocateName(taken map[string]bool, name string, from int) (string, int) {
+  index := from
+  allocated := metadataCollection_composeName(name, index)
+  for taken[allocated] {
+    index++
+    allocated = metadataCollection_composeName(name, index)
   }
-  return allocated
+  return allocated, index
+}
+
+// metadataCollection_composeName renders a base name at a counter. Counter zero
+// is the base name itself, so the first type of a name keeps it unchanged.
+func metadataCollection_composeName(name string, index int) string {
+  if index == 0 {
+    return name
+  }
+  return name + metadataCollection_duplicateSuffix + metadataCollection_itoa(index)
 }
 
 func (collection *MetadataCollection) getFullName(checker *nativechecker.Checker, typ *nativechecker.Type) string {
--- a/packages/typia/native/core/schemas/metadata/MetadataObjectType.go
+++ b/packages/typia/native/core/schemas/metadata/MetadataObjectType.go
@@ -165,14 +165,27 @@ func (obj *MetadataObjectType) IsLiteral() bool {
   if obj.Recursive == false {
     name := strings.ToValidUTF8(obj.Name, "__")
     name = strings.ReplaceAll(name, "\uFFFD", "__")
-    value = name == "__type" ||
-      name == "__object" ||
-      strings.HasPrefix(name, "__type.") ||
-      strings.HasPrefix(name, "__object.") ||
+    value = metadataObjectType_isAnonymousName(name, "__type") ||
+      metadataObjectType_isAnonymousName(name, "__object") ||
       strings.Contains(name, "readonly [")
   }
   obj.literal_ = &value
   return value
+}
+
+// metadataObjectType_isAnonymousName reports whether an allocated id names an
+// anonymous type literal, which the schema generators inline instead of
+// emitting as a component.
+//
+// A second anonymous type is a duplicate of the first, so its id carries the
+// collection's disambiguating counter and the bare marker is not enough to
+// recognize it. This reads that counter back out of an id, which is why it must
+// track `metadataCollection_duplicateSuffix` rather than spell the separator
+// again: when the two disagreed, every anonymous type after the first stopped
+// being recognized and was emitted as a `__type`-keyed component.
+func metadataObjectType_isAnonymousName(name string, marker string) bool {
+  return name == marker ||
+    strings.HasPrefix(name, marker+metadataCollection_duplicateSuffix)
 }
 
 func (obj *MetadataObjectType) ToJSON() IMetadataSchema_IObjectType {
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_generic_argument_dot_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_generic_argument_dot_test.go
@@ -90,20 +90,25 @@ func TestMetadataCollectionGenericArgumentDot(t *testing.T) {
   // A qualified type argument and a quoted literal owning the separator do
   // normalize alike. Escaping out of that would cost every `IPage<IShop.ISum>`
   // key a hash, to separate names no real program declares together. The
-  // allocator is the layer that bounds it: uniqueness is checked against every
-  // id already handed out, so both types survive with their own key and neither
-  // schema is dropped. Uniqueness must never rest on the separator alone.
+  // allocator is the layer that bounds it instead, so both types survive with
+  // their own key and neither schema is dropped. Uniqueness must never rest on
+  // the separator alone.
   if MetadataCollection_replaceOpenApi("Recursive<A.B>") !=
     MetadataCollection_replaceOpenApi(`Recursive<"A-B">`) {
     t.Fatalf("this case no longer documents the residual; re-derive the note above")
   }
   taken := map[string]bool{}
+  counters := map[string]int{}
   for _, input := range []string{"Recursive<A.B>", `Recursive<"A-B">`} {
-    allocated := metadataCollection_allocateName(taken, MetadataCollection_replaceOpenApi(input))
+    // Allocate exactly as `getName` does, counter and all, so this pins the
+    // real path rather than a convenient approximation of it.
+    name := MetadataCollection_replaceOpenApi(input)
+    allocated, index := metadataCollection_allocateName(taken, name, counters[name])
     if taken[allocated] {
       t.Fatalf("a normalization collision dropped a type: input=%q id=%q", input, allocated)
     }
     taken[allocated] = true
+    counters[name] = index + 1
   }
 
   // 6. THE LLM KEY SPACE IS DELIBERATELY UNTOUCHED
--- a/packages/typia/native/core/schemas/metadata/metadata_collection_generic_argument_dot_test.go
+++ b/packages/typia/native/core/schemas/metadata/metadata_collection_generic_argument_dot_test.go
@@ -1,0 +1,117 @@
+package metadata
+
+import (
+  "regexp"
+  "strings"
+  "testing"
+)
+
+// TestMetadataCollectionGenericArgumentDot verifies only a type's own namespace
+// qualification survives OpenAPI normalization as a dot.
+//
+// A dot in a Components Object key is read downstream as a namespace boundary:
+// `JsonDescriptor.cascade` inherits the prefix component's description into the
+// child. Normalization used to delete the angle brackets of `Gen<Ns.Inner>` and
+// leave the argument's dot behind, so `GenNs.Inner` presented the unrelated
+// `GenNs` interface as its parent — typia's and nestia's own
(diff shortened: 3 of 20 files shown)


samchon:
## Self-Review round 1

Complete fresh round over the whole `origin/master...HEAD` diff (15 files): correctness and boundaries, Windows/POSIX, concurrency and state, public API and compatibility, cache/clone invariants, test isolation, CI and packaging, generated output, documentation, migration.

### Verified sound, no change needed

- **`Clone()` after the field restructure.** `names_` went from `map[string]map[*Type]string` to `map[*Type]string`, plus a new `allocated_names_ map[string]bool`. Both are flat value maps, so `maps.Clone` gives the same isolation the old outer-plus-inner clone did, and the removed inner-clone loop is not a lost deep copy. `names_[name][typ]` and `names_[typ]` are equivalent memos, because a type's replaced name is deterministic — the bucket a type landed in was always a function of that type.
- **Nil-map risk.** `getName` writes `allocated_names_` unguarded. `MetadataCollection` is only ever constructed by `NewMetadataCollection` (41 call sites) or `Clone`, both of which initialize it, and `maps.Clone` of a non-nil map is non-nil. No lazy-init guard needed, matching `objects_`/`explore_cache_` rather than the nil-tolerant caches.
- **Counter numbering is unchanged.** Old `len(duplicates)` gave the 2nd type `.o1`; the new loop starting at `index := 1` gives it `-o1`. Same sequence.
- **Unicode qualification.** `qualifyOpenApiName` runs before the escape pass, so it sees raw `é`/`한`. `metadataCollection_isIdentifierRune` uses `unicode.IsLetter`, not ASCII, so `Café.Member` keeps a real namespace boundary it would otherwise have lost. Pinned as a Go case.
- **`disambiguate` is deliberately not set by a qualification rewrite.** Setting it would append a hash to every `IPage<IShop.ISum>` key. Rejected; the residual it leaves is bounded by allocation and documented.

### Applied

- **`MetadataCollection_replaceOpenApi`'s doc comment gave one reason the two key spaces are separate (ASCII grammar); there are now two.** The second — an OpenAPI key is structure a consumer reads back, an LLM key is not — is exactly what justifies `qualifyOpenApiName` living here and not in `MetadataCollection_replace`. A reader hitting the scoping decision would have had to reconstruct it. Comment extended.

### Re-verified after the edit

Comment-only; `go -C packages/typia/test test -p 2 ../native/core/schemas/metadata/...` re-run green.

Round 2 follows.

samchon:
## Self-Review round 2

Fresh complete round over the whole diff, since round 1 applied an improvement.

### Finding — an issue matrix row had no assertion

The issue's case matrix has five rows. Four were pinned. The second — **"`Foo.o1` (invented) → no description inherited from `Foo`"** — was not asserted anywhere. It was *true* (the minted id is now `Foo-o1`, which has no dot, so `cascade` structurally cannot find a parent), but nothing held it true. The reflect.name case proves the id is not a qualified name, and the generic-argument case proves the *flattened* key claims no parent; neither covers the *minted* key on the `json.schema` path, which is the row's actual subject.

That is the exact shape of gap that let this defect survive: the behavior was right, and no test would have noticed it going wrong.

### Remediation, applied

`test_json_schema_openapi_component_name_collision` gains a step 4: the minted id claims no allocated component as a namespace parent, its escaped description carries none of the unrelated ALPHA interface's prose, and it still keeps its own. Doc comment updated to match.

Proven load-bearing rather than assumed — re-running the separator mutation (`-o` → `.o`):

```text
- test_json_schema_openapi_component_name_collision: Error
Error: Bug on the minted id claims no allocated component as a namespace parent: expected condition is not satisfied.
```

Before this step, that same mutation left this test green and was caught only by `test_reflect_name_regular_duplicate_disambiguator`. The `json.schema` surface the issue actually reports now pins it directly.

### Re-verified

`pnpm --filter ./tests/test-typia-schema start` and the metadata Go package, both green. Round 3 follows.

samchon:
## Self-Review round 4

Fresh complete round. One finding, and it is the most serious of this batch — a performance regression I introduced, not a defect of the issue.

### Finding — the uniqueness check made allocation quadratic

Replacing the base-name bucket with a scan over the whole allocated-id set traded `O(1)` allocation for a scan from zero on every duplicate. **Every anonymous type in a program shares the name `__type`**, so that run is not a rare edge — it is the length of the program's anonymous-type list, and the scan reruns it for each new one.

Measured, not estimated:

| anonymous types | before this round |
| --- | --- |
| 1,000 | 63 ms |
| 5,000 | 3.2 s |
| 20,000 | **87 s** |

The old code was `O(1)` here. Shipping this would have traded a silent correctness bug for a silent compile-time bug, on a path nothing in the suite times. It surfaced only because the layering claim in round 1 — "the set is what makes an id unique" — invited the question of what the set costs.

### Remediation, applied

Reinstate a per-base-name counter as the allocator's *starting point*, and keep the taken set authoritative:

- `metadataCollection_allocateName(taken, name, from)` starts at the caller's counter and returns the index it settled on;
- `getName` resumes each base name where it left off;
- the set is still consulted, so a stale or wrong counter costs a scan, never a collision.

| anonymous types | after |
| --- | --- |
| 1,000 | 0 ms |
| 5,000 | 3 ms |
| 20,000 | **13 ms** — all 20,000 ids distinct |

### The set is still load-bearing, and now provably so

With the counter restored, the counter alone handles the issue's own repro, so the earlier mutation no longer failed the TypeScript cases. That is not the set becoming redundant — it is the repro being covered twice. The set is the only thing that can see a collision **across** base names, so a case was added that isolates exactly that: a second `GenNs` mints `GenNs-o1` from its counter, while `Gen<Ns.o1>` normalizes *to* the base name `GenNs-o1` and mints it from a different counter. Neither counter can see the other.

Isolated, with the set:

```text
order=[GenNs GenNs Gen<Ns.o1>] -> allocations=[GenNs GenNs-o1 GenNs-o1-o1] distinct=3
order=[Gen<Ns.o1> GenNs GenNs] -> allocations=[GenNs-o1 GenNs GenNs-o2]    distinct=3
```

and with the set check removed, counter intact:

```text
order=[GenNs GenNs Gen<Ns.o1>] -> allocations=[GenNs GenNs-o1 GenNs-o1] distinct=2   <- a type dropped
order=[Gen<Ns.o1> GenNs GenNs] -> allocations=[GenNs-o1 GenNs GenNs-o1] distinct=2   <- a type dropped
```

Three elements, each now proven load-bearing by an isolated mutation: the **set** (cross-base-name uniqueness), the **counter** (cost), the **separator** (not read back as a namespace boundary).

Round 5 follows.

samchon:
## Self-Review round 5

Fresh complete round. One finding.

### Finding — an assertion whose title outran its predicate

`test_reflect_name_regular_duplicate_disambiguator` asserted "both duplicates are named, and distinctly" with:

```ts
() => duplicated.split(" | ").length === 2
```

That counts members. It never checks they differ. An allocator that handed both `Foo` types the identical name would report `(Foo | Foo)`, split into two parts, and pass — and the sibling assertion would pass too, since neither part is `Foo.o1`. The exact regression this file exists to catch could have gone green under its own title.

### Remediation, applied

Split once, strip the union parentheses once, and assert `new Set(members).size === 2`.

Proven by mutating the allocator so every duplicate keeps the base name:

```text
- test_reflect_name_regular_duplicate_disambiguator: Error
Error: Bug on both duplicates are named, and distinctly: x=2 y=1
```

The previous predicate was green under that same mutation.

### Also reviewed, no change

- `getName`'s counter write (`name_counters_[name] = index + 1`) resumes the next allocation of that base after the id just taken; `composeName(name, 0)` is the base name itself, so the first type of a name keeps it unchanged.
- `Clone` copies `name_counters_` alongside `allocated_names_`; a clone that allocates and is discarded leaves the parent untouched, matching the existing snapshot contract.
- Both new maps are initialized in both constructors, so `getName`'s unguarded writes cannot hit a nil map.
- Allocation still depends only on discovery order, so #2105's "discovery-order independent names" cases stay meaningful and green.

Round 6 follows.

samchon:
## Self-Review round 6

Fresh complete round. Two findings, both accuracy rather than behavior.

### Finding 1 — a doc comment stated something false about the code

`test_llm_parameters_component_name_collision` claimed the LLM generator "allocates its definition keys **from the same collection** as the OpenAPI generator". It does not: `LlmSchemaTransformer` constructs its own `MetadataCollection` with `MetadataCollection_replace`, where the JSON transformers construct theirs with `MetadataCollection_replaceOpenApi`. The distinction is the entire reason this file earns its place — it is a genuinely independent surface reproducing the same root cause, not a second view of one document. Corrected.

### Finding 2 — a Go case modelled a convenient approximation, not `getName`

`TestMetadataCollectionGenericArgumentDot`'s residual case called the allocator with `from = 0` on every iteration. `getName` never does that; it passes the base name's counter. Since both inputs in that case normalize to the *same* base name, `from = 0` was exercising a stale-counter path while claiming to show the real one. It now threads a counters map exactly as `getName` does.

Worth being explicit about what this changed: with the counter modelled faithfully, this case no longer fails when the taken-set check is removed — the counter bounds a same-base-name collision on its own. That is correct, and it is why round 4 added `TestMetadataCollectionNameAllocation`'s cross-base-name case, which is the one class no counter can see. Each layer is pinned by the case that actually needs it, and neither test now claims credit for the other's coverage.

### Also reviewed, no change

- **Tuples get fixed for free, and the reasoning holds.** `[Ns.A, Ns.B]` normalized to `Ns.ANs.B`, claiming fake parents `Ns` and `Ns.ANs`. The qualification rule covers it as the same class as the union case already pinned (`Array<A.B | C.D>`).
- **Leading dots, repeated dots, and `...rest` display names.** A dot never clears the qualified run (`A.B.C` depends on that), and `metadataCollection_exploreName` cannot produce a leading dot. Rest-tuple display strings are built by `RandomJoiner` for diagnostics and never reach `getName`.
- **`IsLiteral`'s `readonly [` branch** only ever matches when no replacer is configured, because `MetadataCollection_replace` deletes `[` and the space. Pre-existing, untouched, and out of scope.
- **JSON Pointer decode order** in the LLM case is `~1` → `/` then `~0` → `~`, per RFC 6901 §4 and matching the repo's existing `decodePointerToken`.

Round 7 follows.

samchon:
## Self-Review round 7

Fresh complete round. **One finding, and it is a regression I introduced that no round before this caught.**

### Finding — the separator made `protobuf.message` emit text no parser accepts

Sweeping for *every* consumer that reads structure out of a name — not just the ones the issue lists — turned up a fifth: `protobufMessageProgrammer_emplace` splits a metadata name on `.` and nests each segment as a child message.

So `protobuf.message` has the same overloaded-dot defect. On `master`, two unrelated `Foo` types emit the second one **inside** the first:

```proto
message Foo {
  required string a = 1;
  message o1 {              // BETA, nested inside ALPHA's message
    required double b = 1;
  }
}
```

The `-o` separator fixes that structure — and broke the output, because a Protobuf message name is an identifier:

```proto
message IArguments {
  required Foo a = 1;
  required Foo-o1 b = 2;    // illegal
}
message Foo-o1 {            // illegal
  required double b = 1;
}
```

Structurally right, lexically invalid. `protobuf.message` is a public API, and this would have shipped as "typia emits a .proto that protoc rejects" — strictly worse than the bug being fixed. Nothing caught it: the Go suites, both named TypeScript suites, and the full package matrix were all green, because no case asserts that generated Protobuf is *parseable*.

### Root cause and remediation, applied

`ProtobufNameEncoder` exists precisely to make a metadata name a legal Protobuf identifier, and its table had no `-` rule. That is not only my problem — it is a pre-existing hole: `Gen<"a-b">` already reaches it and already emits an illegal `-` on `master`. One rule closes both:

```go
{Before: "-", After: "_hyphen_"},
```

The dot is deliberately left out of that table: the emplacer splits on it to nest a real namespace member in its parent's message, and a qualified reference is legal in the grammar. Result:

```proto
message Foo            { required string a = 1; }
message Foo_hyphen_o1  { required double b = 1; }
```

Legal, distinct, and siblings rather than parent-and-child.

### Coverage for the gap that let it through

`test_protobuf_message_duplicate_name_identifier` asserts every declared message name is a legal identifier, every field type resolves to a declared message, and the duplicate is a sibling rather than a child. Proven load-bearing by removing the new rule:

```text
- test_protobuf_message_duplicate_name_identifier: Error
Error: Bug on message name "Foo-o1" is a legal Protobuf identifier: expected condition is not satisfied.
```

This makes **five** consumers of the disambiguator's shape, against the three the issue names: `cascade`, LLM `$defs`, `reflect.name`, plus `MetadataObjectType.IsLiteral` (round 0) and `protobuf.message` (here). Both extras were found by sweeping for the pattern rather than by trusting the issue's list, and both broke silently.

Round 8 follows; full Go trees already re-run green.

samchon:
## Self-Review round 8

Fresh complete round. One finding, on the round-7 remediation itself.

### Finding — the new Protobuf case asserted my opinion of the grammar, not the grammar

`test_protobuf_message_duplicate_name_identifier` checked message names against a hand-written `/^[A-Za-z_][A-Za-z0-9_]*$/`. That is a snapshot of what I believe Protocol Buffers accepts, which is exactly the failure mode the development skill's "oracle-derived expectations" rule names — the test is only ever as right as the reviewer who wrote its regex, and it silently answers a different question ("does this match my pattern?") than the one that matters ("would a consumer's parser accept this?").

The oracle was already in the repository and already a dependency of this suite. `tests/test-typia-automated/src/internal/_test_protobuf_encode.ts` hands generated `.proto` text to **protobuf.js** and encodes against it; `tests/test-typia-schema` declares `protobufjs` too. Nothing in that automated path caught this regression only because no `@typia/template` structure declares a duplicate name — the harness was right, its fixtures just never reached the case.

### Remediation, applied

The case now parses the emitted document with protobuf.js and resolves through the parsed tree: `root.lookupType("IArguments")`, then each field's message, asserting the first `Foo` keeps its `string` field, the disambiguated one keeps its `double` field, and the two resolve to different messages. The hand-written pattern is gone.

Re-running the mutation that removes the hyphen rule now fails against the real grammar rather than my approximation of it:

```text
- test_protobuf_message_duplicate_name_identifier: Error
Error: Bug on protobuf.js parses the emitted document: expected condition is not satisfied.
```

That is an independent implementation refusing the output — the strongest available evidence that the round-7 regression was real and is now fixed.

### Also reviewed, no change

- **`ProtobufNameEncoder.Decode` has no callers**, in Go or TypeScript. The new rule is symmetric with the table's existing entries, so it stays correct if a caller ever appears. The dead method predates this batch and is left alone.
- **No TypeScript consumer decodes Protobuf name encoding**, so the rename is confined to the emitted `.proto`.
- The round-4 package matrix (validating the allocator change) returned exit 0 with no failure lines; the final matrix over the complete change is running.

Round 9 follows.

samchon:
## Self-Review round 9

Fresh complete round. No new finding; one gap in the *evidence* rather than the code, now closed.

### The fix was proven to fire, never proven not to over-fire

Every mutation so far removed something and watched a test fail — which proves the rule is *necessary*. None proved it is *bounded*. A qualification rule that stripped every dot, real ones included, would have satisfied all of them: `Gen<Ns.Inner>` would claim no parent, the collision cases would still hold, and the whole batch would look green while having quietly deleted the cascade feature the issue explicitly requires preserved.

That is the negative twin the development skill asks for. It was written into the tests but never demonstrated.

### Demonstrated

Mutating `qualifyOpenApiName` to rewrite every dot, real qualifications included:

```text
--- FAIL: TestMetadataCollectionGenericArgumentDot
    input="Merged.Child" actual="Merged-Child" expected="Merged.Child"
--- FAIL: TestMetadataCollectionOpenApiName          (#2105's Qualified.Member control)

- test_json_schema_openapi_component_names: Error
  Bug on qualified control: expected condition is not satisfied.
- test_json_schema_openapi_component_name_generic_argument: Error
  Bug on a real namespace member still inherits its real parent's description
```

Four cases across both languages, including #2105's own controls and this batch's positive control. Paired with the round-4 mutation that drops the rule entirely — which fails `..._generic_argument` and the Go dot case — the rule is now pinned from both sides: it fires exactly where a dot is not the type's own qualification, and nowhere else.

### Mutation ledger, complete

| Element removed / broken | Fails |
| --- | --- |
| `-o` separator → `.o` | `..._collision` (minted id claims a parent), `..._reflect_name_...` |
| taken-set check | `TestMetadataCollectionNameAllocation`, incl. the cross-base-name case |
| per-base counter | 20k anonymous types: 13ms → 87s |
| `qualifyOpenApiName` dropped | `TestMetadataCollectionGenericArgumentDot`, `..._generic_argument` |
| `qualifyOpenApiName` over-applied | the four above — the cascade feature stays intact |
| quoted-dot escape tier | `TestMetadataCollectionGenericArgumentDot`, #2105's `hyphen control` |
| `IsLiteral` shape coupling | `TestInlineTypeDisplayNameTransform` |
| Protobuf hyphen rule | `..._protobuf_message_duplicate_name_identifier` (protobuf.js refuses to parse) |
| allocator collapses duplicates | `..._reflect_name_...` (`x=2 y=1`) |

Every element of this change is load-bearing, each proven by an isolated mutation, and no element can be over-applied without a test objecting.

Round 10 follows once the final full package matrix reports.

samchon:
## Self-Review round 10

Fresh complete round over the final diff. **No finding.** Nothing applied.

Re-read every hunk: `MetadataCollection.go` (fields, `NewMetadataCollection`, `Clone`, `getName`, the allocator, the qualification rule, the OpenAPI replacer and its rune writers), `MetadataObjectType.go`, `ProtobufNameEncoder.go`, `JsonDescriptor.ts`, both public Go cases, and all six TypeScript cases.

Checks that produced nothing this round:

- **Boundaries.** A leading dot (`metadataCollection_exploreName` cannot produce one), repeated dots (`A.B.C` — a dot never clears the qualified run), an empty name (`replaceOpenApi` guarantees non-empty; `replace`'s fallback is non-empty for non-empty input), `composeName(name, 0)` returning the base name unchanged, and `Options.Replace == nil` (the `reflect.schema` path, which correctly gets `-o` but not the qualification rule).
- **Nil maps.** `names_`, `allocated_names_`, and `name_counters_` are initialized in both constructors — the only two places a `MetadataCollection` is built — and `maps.Clone` of a non-nil map is non-nil, so `getName`'s unguarded writes are safe.
- **Determinism.** Allocation depends only on discovery order, as before; #2105's discovery-order-independence cases stay green and meaningful.
- **Windows/POSIX.** Pure string logic, no paths.
- **Documentation.** No website page documents a disambiguated key or the `.o<N>` shape, and `git grep` finds no tracked baseline carrying one, so no guide change is owed.
- **Isolation.** The main checkout is untouched and still on the user's `feat/website-theme` with only their `website/` changes; all work is in the `typia-ck-dots` worktree.

### Round ledger

| Round | Finding |
| --- | --- |
| 1 | `replaceOpenApi`'s comment gave one reason the key spaces differ; there are two |
| 2 | an issue matrix row (`Foo.o1` inherits nothing) held true but nothing asserted it |
| 3 | a negative assertion would pass vacuously if the reference stopped escaping |
| 4 | **the uniqueness check made allocation quadratic — 87s at 20k anonymous types** |
| 5 | an assertion titled "distinctly" only counted members |
| 6 | a doc comment stated something false; a Go case modelled an approximation of `getName` |
| 7 | **the separator made `protobuf.message` emit text no parser accepts** |
| 8 | the new Protobuf case asserted my regex, not the grammar |
| 9 | the rule was proven necessary, never proven bounded |
| 10 | — clean |

Stopping here per the Non-Negotiable Review Law: a complete round produced nothing that survives verification.

Final verification and READY status to follow once the full package matrix reports.

samchon:
## Final verification — READY for rebase, version bump, and merge

Self-Review is complete: ten rounds, six of which found something, and round 10 produced nothing that survived verification.

### Commands and results

```text
go -C packages/typia/test test -p 2 ../native/core/schemas/metadata/...   ok
go -C packages/typia/test test -p 2 ../native/...                         ok   (all packages)
go -C packages/typia/test test -p 2 ./...                                 ok   (all packages)
pnpm build                                                                exit 0
pnpm test:packages                                                        exit 0 — 13 Success, 0 failures
pnpm --filter ./tests/test-typia-schema start                             Success
pnpm --filter ./tests/test-utils start                                    Success
```

The full matrix was run rather than only the two suites the issue names, because `IsLiteral` reaches every validator and serializer and the Protobuf encoder reaches every generated `.proto`. All 14 workspaces ran: `test-typia-schema`, `test-utils`, `test-typia-automated`, `test-utils-automated`, `test-interface`, `test-error`, `test-feature-identity`, `test-langchain`, `test-mcp`, `test-vercel`, `test-typia-cli`, `test-typia-generate`, `test-typia-exact-optional`, `test-typia-bundler-cache`.

Worth noting: `test-typia-automated` ran 103 Protobuf cases, and `_test_protobuf_encode` parses each generated document with protobuf.js. Every one passed, which independently confirms the new hyphen rule is inert for existing fixtures — it fires only for the newly-possible `Foo-o1`. No `@typia/template` structure declares a type name containing a hyphen, so no existing `.proto` output changes.

### Pending, and deliberately not done here

- **`typia` version bump — required and not applied.** Go source under `packages/typia/native` changed, so the development skill's Change Integrity rule requires a coordinated bump. Left to the lead to serialise with the other campaign batches at rebase.
- **No rebase onto `master`, no merge** — both lead-owned.
- **`pnpm format` not run**, per the issue-campaign no-format rule. Post-Campaign Cleanup owns the formatter result, so the new files are unformatted by design.

### Campaign CI

Every push and pull-request mutation passed the exact-SHA cancellation gate. One run for `ec6eaa90` (id `29550977246`) was already terminal and green when first observed, so it was recorded rather than cancelled, per the gate's rule; every other run for every campaign SHA ended `cancelled`. No repository Actions or workflow setting was changed.

samchon:
## Self-Review round 11

Round 10 said "clean" and it was — about the code. It was not a complete round, because I reviewed the diff and never checked **what the branch actually contains**. Doing that found a real defect in this pull request.

### Finding — two suite artifacts rode in on `git add -A`

`git diff origin/master...HEAD --name-status` listed 19 files. Two do not belong to this change:

```text
A  tests/test-typia-generate/project-link/src/generate_project.ts
A  tests/test-typia-generate/project-link/tsconfig.json
```

`project-link` is a **temporary junction** that `tests/test-typia-generate/scripts/check-file-errors.cjs` creates to `project-dir` and removes in a `finally`, to exercise "implicit project through aliased cwd". I staged with `git add -A packages tests` while a package matrix was running, git followed the live junction, and committed `project-dir`'s contents under the alias path. Neither file is tracked on `master`, neither is gitignored, and no source owns them.

This is the development skill's "keep local outputs local" and the pull-request skill's "never include unrelated user changes silently", and it happened precisely because I reached for `-A` instead of naming paths. The first two commits staged explicit paths and are clean; the ones that used `-A` are where it crept in.

The symptom that exposed it was the deletion showing up afterwards: the suite's `finally` removed the junction, so a later `git status` reported the two files as deleted — files I had never knowingly added.

### Remediation, applied

Removed from the branch (`git rm -r --cached` plus deleting the stray real files the earlier restore had materialized). The branch now touches exactly the 17 intended files, and the worktree is clean. The suite creates and removes the junction itself; nothing needs it committed.

### Verified

- `git diff origin/master...HEAD --name-status` → 17 files, all intentional, no `project-link`.
- `git status --porcelain` → empty.
- No other generated tree leaked: `src/output` is gitignored, and no `.tgz` or `.env` is present.

Round 12 follows: a fresh round over the corrected branch.

samchon:
## Self-Review round 12 — clean

Fresh complete round over the corrected branch: the 17-file diff, the branch contents, and the worktree.

Re-verified after the round-11 removal:

```text
go -C packages/typia/test test -p 2 ../native/...   ok
go -C packages/typia/test test -p 2 ./...           ok
pnpm --filter ./tests/test-typia-schema start       Success
pnpm --filter ./tests/test-utils start              Success
pnpm --filter ./tests/test-typia-generate start     Done
```

`test-typia-generate` was re-run specifically because it owns the junction that leaked: it passes, and `git status --porcelain` is empty afterwards — the suite creates and removes `project-link` itself, so nothing was needed in the tree. The full `pnpm test:packages` result from the previous SHA still stands; the only change since is the removal of two files no source references and no suite reads.

No finding. Stopping here per the Non-Negotiable Review Law.

### Self-Review summary — 12 rounds, 7 findings

| Round | Finding |
| --- | --- |
| 1 | `replaceOpenApi`'s comment gave one reason the key spaces differ; there are two |
| 2 | an issue matrix row held true but nothing asserted it |
| 3 | a negative assertion would pass vacuously if the reference stopped escaping |
| 4 | **the uniqueness check made allocation quadratic — 87s at 20k anonymous types** |
| 5 | an assertion titled "distinctly" only counted members |
| 6 | a doc comment stated something false; a Go case modelled an approximation of `getName` |
| 7 | **the separator made `protobuf.message` emit text no parser accepts** |
| 8 | the new Protobuf case asserted my regex, not the grammar |
| 9 | the rule was proven necessary, never proven bounded |
| 10 | clean — but only of the code |
| 11 | **two suite artifacts had ridden into the branch on `git add -A`** |
| 12 | clean |

Two of those (4 and 7) were regressions I introduced that every test in the repository was green on. Both were found by asking what the change costs and who else reads the thing it changed, rather than by a failing test.

**READY** for rebase, version bump, and merge — all three lead-owned. The version bump is required (native Go source changed) and is not applied here.

samchon:
Lead certification before merge. **This is the last of discovery round D1's findings.**

## Rebase and version

Rebased onto `1117b993fa` (#2133); six commits replayed with no conflict. Assigned **`13.1.19`** — mandatory because the diff changes native Go. Branch contains exactly the 18 intended files; I re-checked for stray artifacts specifically, given round 11's finding below.

## Lead certification on `1ae99c0acf`

| Gate | Result |
| --- | --- |
| `go -C packages/typia/test test ./...` | 0 failures |
| `pnpm run test:packages` | **exit 0**, 0 hard failures, 28 Success/Done markers |

The full TypeScript suite was the right gate here rather than the two named workspaces, because `IsLiteral` reaches every validator.

## The two regressions this batch caught in itself are the campaign's deepest lesson

The batch introduced two regressions, and **every existing test stayed green on both**:

- the new uniqueness scan was quadratic over `__type` — 20,000 anonymous types went from **13 ms to 87 s**;
- joining the counter with `-o` made `protobuf.message` emit `message Foo-o1`, text no Protocol Buffer parser accepts.

Neither was caught by a test. Both were caught by asking two questions no suite asks on your behalf: **what does this change cost**, and **who else reads this**. A suite answers "is the output still correct for the cases I thought of". It does not answer "is this still fast enough" or "does a downstream grammar accept this text". This campaign found eight oracles that could not fail; these two are the other half of that lesson — an oracle that *can* fail still only covers what it was pointed at.

## The issue's consequence surface was a hypothesis, and the batch treated it as one

I listed three consumers of the `.o<N>` shape. **Sweeping for them instead of trusting my list found two more**, and both were live defects on master:

- `MetadataObjectType.IsLiteral` — every anonymous type after the first became a component;
- `protobuf.message` — which reads the dot as *nesting*, so master emitted BETA **inside** ALPHA's message.

## Both defects fixed at the right layer

**Defect 1** takes two fixes, each covering what the other cannot: `metadataCollection_allocateName` now checks the collection's **whole** allocated-id set — the only thing that can see a collision *across* base names — and the counter joins with `-o` so it is not read back as structure.

**Defect 2** is paid at the **producer**, not at `cascade`: a consumer cannot verify a key it did not produce, so `metadataCollection_qualifyOpenApiName` keeps only the type's own leading qualification. A dot inside a quoted literal takes #2105's escape-and-hash tier — without that, `Recursive<"A.B">` and `Recursive<"A-B">` merge, which #2105's own order-independence contract caught and forced a design correction.

## Evidence

Fail-then-pass on all six new cases, then **nine isolated mutations**, including one proving the rule cannot *over*-fire: stripping real qualifications fails both the positive control and #2105's `qualified control`. The protobuf case lets **protobuf.js** judge the emitted document rather than the batch's own regex — an independent oracle, which is exactly right for a defect whose whole nature was a downstream grammar rejecting our text.

## Self-Review round 11 is worth remembering on its own

**Round 10 declared clean after reviewing only the code.** Round 11 checked what the branch actually *contained* and found two `test-typia-generate` junction artifacts that a `git add -A` had swept in during a concurrent test run. Reviewing the diff and reviewing the branch are different acts. Twelve rounds, seven findings, round 12 clean.

## User-visible renames — and they are the fix

- `Foo.o1` → `Foo-o1`, across every key space including `reflect.name` and `reflect.schema`
- `IPageIShoppingSale.ISummary` → `IPageIShoppingSale-ISummary` — OpenAPI only; the LLM `$defs` space is deliberately untouched
- `RecursiveObjectA.B` → `RecursiveObjectA_x2E_B-x<hash>`
- protobuf: nested → sibling `Foo_hyphen_o1`

Changing `.o<N>` is a deliberate public-contract change, stated as such, exactly as #2105's `$Named` → `Named` rename was.

`1ae99c0acf` settled after five polls, all six runs `cancelled`, no race.

PR #2075, opened 2026-07-16
fix(typia): align format and float parity

samchon:
## Intent

Restore one authoritative scalar contract so float32 ranges and advertised JSON Schema formats agree across runtime helpers, `@typia/utils`, native generated checks, schemas, and random generation.

## Scope

- Correct `Type<"float">` to the finite float32 range, including symmetric negative values and exact finite/infinite boundaries.
- Align `byte`, `ipv6`, `iri-reference`, `date`, and `date-time` validation across runtime, utils, and native-generated code.
- Add authoritative positive, negative, repeated-call, and boundary matrices covering the full consequence surface.
- Bump the `typia` package version because native Go source changes.

## Deferred

- String length, numeric `multipleOf`, unrelated formats, and unrelated native behavior remain outside this batch.
- Repository-wide formatting remains reserved for campaign Post-Campaign Cleanup.
- The final package version is selected only after rebasing the latest `master`, because concurrent native branches share that release line.

## Verification

Pending. This implementation-free draft reserves the batch before code changes. Focused runtime/utils/native/generated regressions, package builds, broader test suites, and complete solo Self-Review rounds will be recorded before merge.

Automatic GitHub Actions for campaign commits are deliberately cancelled by exact SHA. Local verification is the implementation gate under the issue-campaign procedure.

Closes #2038
Closes #2039

PR #2131, opened 2026-07-17
fix(typia): reject scheme-less uri-reference whose first segment contains a colon

### LINKED ISSUE #2127: uri-reference accepts a scheme-less relative reference whose first segment contains a colon

samchon:
Fixes #2127.

## Intent

`tags.Format<"uri-reference">` accepts `1bad:relative`, which RFC 3986 forbids: `1bad` cannot be a `scheme` (a scheme must begin with `ALPHA`), and a scheme-less relative reference's first segment is `segment-nz-nc`, which excludes `:`. The value belongs to no production of `URI-reference`.

The cause is that the expression reuses its `path-rootless` alternative where `path-noscheme` is required. `path-rootless` permits `:` in its first segment because it only ever follows a real scheme; a scheme-less relative reference must use `path-noscheme`, whose first segment is `segment-nz-nc`. Every *later* segment is `*pchar`, which does include `:`.

Three owners carry a byte-identical expression and all three accept it:

- `packages/typia/src/internal/_isFormatUriReference.ts`
- `packages/utils/src/validators/functional/_isFormatUriReference.ts`
- `packages/typia/native/core/factories/FormatCheatSheet.go` — `FormatCheatSheet["uri-reference"]`, which also backs the JSDoc `@format` path via `metadataCommentTagFactory_FORMATS`

## Scope

- Split the shared optional-scheme prefix into three RFC-aligned top-level alternatives so the scheme-ful and scheme-less paths no longer share `path-rootless`, in all three owners, keeping them byte-identical. Fixing only the TypeScript side would recreate the runtime/generated divergence class that #2039 and #2075 repaired and that #2113 avoided.
- Extend the `uri-reference` negative matrix that #2113 introduced, rather than starting a new one, and keep its `iri-reference` ⊇ `uri-reference` coverage-relation oracle — that oracle is what made #2110 visible and it flags this case too.

**The fix stays conditional.** `:` is forbidden only in the first segment and only when no scheme precedes it. `http:1bad:relative` and every URI whose path legitimately carries colons remain accepted; a blanket ban would break them.

A parity suite could never have caught this: all three owners agree, they are simply all wrong. Every expectation is taken from RFC 3986, never from current output.

Strictly narrowing, and narrow: exactly one shape stops being accepted.

## Version

**Pending lead assignment.** This changes native Go (`FormatCheatSheet.go`), which per the development skill's Change Integrity rule requires a coordinated `packages/typia/package.json` version bump. The lead serialises the native version chain across the concurrent native pull requests of this campaign and assigns the version at rebase time, so `packages/typia/package.json` is deliberately untouched here.

## Verification

Local verification only; campaign CI is deliberately suspended and this commit's runs are cancelled under the exact-SHA gate. Commands recorded in the thread as increments land.


samchon:
## Implementation, verification, and Self-Review

Final SHA `79e8d9adfec5db7385729fccf0a6adec895e7252`. Campaign CI is suspended; every run for each pushed SHA passed the exact-SHA cancellation gate.

### The change

The optional scheme prefix was shared by all three path alternatives, so `path-rootless` served the scheme-less side too. Split into three RFC 3986 alternatives, keeping the authority sub-expression written once:

```
^(?: (?:scheme:)?(?: //authority path-abempty | path-absolute )?   # scheme optional; neither can start with a bare segment
   | (?:scheme:) path-rootless                                     # first segment *pchar — ":" legal after a real scheme
   | path-noscheme                                                 # first segment segment-nz-nc — no ":"
 )(?:\?query)?(?:#fragment)?$
```

Only the first segment of the scheme-less branch loses `:`; every later segment stays `*pchar`.

### Fail-then-pass

The regression cases were added and observed failing **before** the fix, built from source:

```
Bug on uri-reference rejects "1bad:relative" directly: found different values
{ "x": false, "y": true }
```

After the fix: `test_openapi_validator_format_uri_reference: 8 ms — Success`.

### Mutation evidence

Each owner is independently load-bearing — reverting one alone fails the matrix through that owner's own path:

| Mutation | Failure |
| --- | --- |
| A — revert `FormatCheatSheet.go` only | `rejects "1bad:relative" **through comment tag**` |
| B — revert `@typia/utils` helper only | `rejects "1bad:relative" **through OpenApiValidator**` |
| C — revert typia runtime helper only | `rejects "1bad:relative" **directly**` |
| D — **blanket** colon ban instead of the conditional fix | `accepts "urn:example:animal:ferret:nose"` |

Mutation A is the point of the issue: a TypeScript-only fix leaves the JSDoc/native-generated path wrong, which is the #2039/#2075 divergence class. Mutation D shows the matrix pins the fix's *conditional* nature — the plausible wrong fix is caught, not just the missing one.

### Owner parity

Mechanically extracted from all five points (Go string unescaped back to the regex literal it carries), SHA-256 of each:

```
typia src  len=1274 sha=5963267b7a74c965 IDENTICAL
typia lib  len=1274 sha=5963267b7a74c965 IDENTICAL
utils src  len=1274 sha=5963267b7a74c965 IDENTICAL
utils lib  len=1274 sha=5963267b7a74c965 IDENTICAL
native go  len=1274 sha=5963267b7a74c965 IDENTICAL
```

`lib` was proven rebuilt against `src` before any comparison, since a stale `lib` would silently invalidate the test's `typia/lib/internal/...` import.

### Correctness oracle

Expectations come from RFC 3986, never from current output — all three owners agreed while all three were wrong, so agreement proves nothing here. 26 hand-derived cases pass, including `1bad:` `:` `:/a` `a%3ab:c` `+foo:bar` `9http://a/b` rejected and `a%3ab` `foo+bar:baz` `?q=1:2` `#f:1` `/1bad:relative` `//a/1bad:relative` accepted.

Differential fuzz against the master expression over a 26-character alphabet (sub-delims, brackets, uppercase, pct, scheme chars), all strings up to length 4 — **475,254 strings, 37,217 disagreements, 0 outside the target class**, every one old-accept → new-reject. The target class is classified independently of the regex: no valid scheme, no leading `/`, and a `:` in the first segment. This is the evidence for "strictly narrowing".

### Consequence surface

- **Sibling `uri` format is immune** and correctly untouched: its scheme group is mandatory, so `path-rootless` there always follows a real scheme.
- **`typia.random`** routes `uri-reference` to `_randomFormatUrl`, which only emits `https://…`; 20,000 generated samples all still validate. The `@typia/template` `TypeTagFormat.uriReference` / `CommentTagFormat` fixtures therefore stay green — confirmed by running both generated matrices rather than arguing it.
- **No ReDoS**: the added top-level alternation stays linear on adversarial inputs up to n=1600.
- **Docs** need no change: `website/.../tags.mdx` only lists format names, with no behavioural claim.

### Verification

| Command | Result |
| --- | --- |
| `pnpm test:go:native` | 13 packages ok, 0 FAIL |
| `pnpm --filter ./tests/test-typia-automated start` | Success (consequence check) |
| `pnpm --filter ./tests/test-utils-automated start` | Success (consequence check) |

### Self-Review

Solo, per the review skill.

- **Round 1** — no defect in the fix. One improvement applied: the note beside the newly accepted colon cases described the rule as "wherever a scheme precedes it or the first segment ends", which does not cover the rooted-path, query, and fragment entries directly beneath it. Replaced with the four legal positions and named `segment-nz-nc` as the sole exclusion (`c3a810a1`).
- **Round 2** — fresh full round, nothing further.
- **Follow-up** — a run raced to completion ahead of the cancellation gate and legitimately failed `spell-check`: typos reads the coined "scheme-ful" as a misspelt "full". Reworded and rewrapped (`79e8d9ad`); spell-check now passes on CI at the final SHA. Also verified the changed files are Prettier-clean once line endings are normalized — `prettier --check` flags every file in a Windows worktree because the checkout is CRLF, untouched files included, so that signal is environmental. `pnpm format` was not run, per the campaign rule. `gofmt` lists `FormatCheatSheet.go` on master too; the native tree uses 2-space indent by convention, so the Go edit matches its neighbours.

### Pending, deliberately not done here

`packages/typia/package.json` is untouched: the native Go change requires a coordinated version bump that the lead serialises across this campaign's concurrent native pull requests at rebase time. Rebase, version, and merge are all lead-owned. **READY.**

samchon:

## Rebase and version

Rebased onto `208b3fbf7b` (#2129); four commits replayed with no conflict. Assigned **`13.1.18`** — mandatory because `FormatCheatSheet.go` is native Go.

## Lead verification on `9214eb892b`

**Parity, checked mechanically rather than by eye:** the typia and utils expressions now hash identically (`44e130c94a48792e`, len 1275). The batch's own check covers all five extraction points including the Go entry, and it proved `lib` was rebuilt against `src` **before** comparing — this file family has already burned an agent with a stale `lib/` artifact once this campaign.

**Runtime proof, built from the shipped source:**

| Input | Result |
| --- | --- |
| `1bad:relative`, `9x:y`, `1:2` | **rejected** — the defect is gone |
| `http:1bad:relative`, `urn:example:animal:ferret:nose`, `a:b`, `mailto:a@b.c` | accepted — a colon after a real scheme still works |
| `./1bad:relative`, `good/1bad:relative`, `/1bad:relative` | accepted — a colon from the second segment on still works |
| `?q=1:2`, `#f:1`, `%22`, `a/b`, empty | accepted |

| Gate | Result |
| --- | --- |
| `pnpm --filter test-utils start` | Success |

## Why this batch's evidence is unusually strong

**Each owner is proven to have its own covering surface.** Reverting one owner at a time fails through *that owner's own path*: the Go entry through `rejects "1bad:relative" through comment tag`, utils through `through OpenApiValidator`, and typia src `directly`. Three owners, three independent failures — not one test standing in for three.

**The matrix pins the fix's conditionality, not merely the rejection.** A fourth mutation applying a *blanket* colon ban fails on `urn:example:animal:ferret:nose`. That is the mutation that mattered: a blanket ban would have "fixed" the issue while breaking every URI whose path legitimately carries colons, and a matrix that only asserted `1bad:relative` is rejected would have waved it through.

**The narrowing is measured, not claimed.** A differential fuzz against master over 26 characters at length ≤ 4 — **475,254 strings, 37,217 disagreements, zero outside the target class**, every one old-accept → new-reject, with the class classified independently of the regex under test.

## Two consequence checks worth recording

- **`typia.random` was a live risk and it is closed.** `@typia/template`'s `TypeTagFormat.uriReference` and `CommentTagFormat` feed the generated matrices, so a narrowed validator could have started rejecting its own generator's output. `_randomFormatUriReference` delegates to `_randomFormatUrl`, which only emits `https://…` and is therefore immune — verified over 20k samples and by running both automated matrices.
- **The sibling `uri` format is immune** because its scheme group is mandatory. The failure class is genuinely confined to `uri-reference`.

## Notes

A `spell-check` run raced ahead of the cancel gate and **legitimately failed** — `typos` read the coined "scheme-ful" as a misspelt "full". That is the second time this session a spell-check failure was real rather than the benign race; it was fixed and passes at the final SHA. Worth repeating: read the conclusion before dismissing it.

`prettier --check` is unusable in a Windows worktree — a CRLF checkout fails every file, including untouched ones. `pnpm format` was correctly not run, per the campaign rule.

`9214eb892b` settled after four polls, all six runs `cancelled`, no race.

PR #2133, opened 2026-07-17
fix(tests): restore the tests/debug ad-hoc repro runner

### LINKED ISSUE #2132: tests/debug cannot run: the documented ad-hoc repro runner fails to compile its own sample
## Problem

`tests/debug` — the workspace the project skill documents as "a one-off `ttsx` runner for ad-hoc local repros" — cannot run. Its own committed sample fails to compile, so the tool a contributor reaches for when reproducing a bug is itself broken.

Two independent causes, either of which is sufficient on its own.

Affected: anyone following the documented workflow for an ad-hoc reproduction, including every agent and contributor onboarding through `.agents/skills/project/SKILL.md`. This campaign hit it repeatedly and worked around it with `test-typia-schema --include`, which is a poor substitute — it requires adding a fixture to a real suite just to look at one output.

## Evidence

Verified on `master` (`3e7b22fd80`, typia 13.1.15). Confirmed independently by two discovery agents and a third verification pass, and the two causes were separated by a single-variable test rather than assumed.

```
pnpm --filter ./tests/debug start
```

Observed: exit 2 with 24–25 `TS6196` / `TS6205` diagnostics against the committed `llm.application` sample. Expected: the sample runs and prints its output.

**Cause A — the `noUnusedLocals` conflict.** `tests/config/tsconfig.json` sets `noUnusedLocals: true`, while `packages/typia/tsconfig.json` sets it `false`. Combined with `exports["."] -> ./src/index.ts`, `tests/debug` compiles typia's own sources under a stricter setting than typia compiles itself with, and typia's sources do not satisfy it. This produces the `TS6196`/`TS6205` wall.

**Cause B — the missing `rootDir`.** `tests/debug/tsconfig.json` is **the only one of the sixteen `tests/*` configs without `rootDir: "src"`** (15 of 16 have it). Without it, module identity breaks at runtime: `_ProtobufReader_1._ProtobufReader is not a constructor`.

**The two are independent, proven by single-variable test:** adding `rootDir: "src"` fixes the runtime module-identity failure — that error disappears — while the committed sample **still** emits 24 `TS6196`/`TS6205` and exits 2 with the override in place. Neither cause masks the other, and fixing only one leaves the workspace unusable.

## Consequence surface

- **Owner:** `tests/debug/tsconfig.json` (cause B) and the interaction between `tests/config/tsconfig.json` and `packages/typia/tsconfig.json` (cause A).
- **Consumers:** the documented ad-hoc repro workflow only. No product surface, no other suite, no generated artifact.
- **Not affected:** the other fifteen `tests/*` workspaces, all of which carry the `rootDir` override.
- **Why it survived:** `tests/debug` is a developer tool with no assertions, so nothing fails when it stops working. It is excluded from `pnpm test:packages`' pass/fail accounting by not being a `test-*` workspace.

| Case | Expected |
| --- | --- |
| `pnpm --filter ./tests/debug start` on the committed sample | runs, prints output, exit 0 |
| a sample importing `typia` and calling any `typia.*` | compiles under the same settings typia compiles itself with |
| module identity across `src/` | one instance — `instanceof` and constructor checks hold |
| the other fifteen `tests/*` workspaces | unchanged |

## Approach

The invariant: **`tests/debug` compiles a sample under settings compatible with the sources it imports, and resolves modules to one identity.**

Cause B is a one-line alignment with the fifteen sibling configs, and the fact that it is the lone exception is itself the argument. Cause A is a policy question the fix must answer deliberately rather than by suppression: `tests/config` asserts a strictness that `packages/typia` does not hold itself to, and `tests/debug` is the only consumer that compiles typia's sources directly through `exports["."]`. Decide where the setting belongs rather than adding a local override that hides the mismatch from the next reader.

- **Positive:** the committed sample runs and prints. A fresh sample importing `typia` and calling `typia.llm.application`, `typia.is`, and `typia.protobuf.decode` compiles and runs.
- **Negative:** a genuine type error in the sample still fails — the fix must not be `skipLibCheck`-style suppression that also hides the user's own mistakes.
- **Boundary:** module identity holds (`_ProtobufReader` constructs); the other fifteen workspaces are unchanged.
- **Regression:** the config drift that caused this — one workspace differing from fifteen siblings — is worth a check, since nothing failed when it drifted.
- **Commands:** `pnpm --filter ./tests/debug start`; `pnpm test:packages` to prove no sibling regressed.

## Coordination

- Test infrastructure only. No product change and no package version bump.
- Independent of every open issue; nothing blocks it and it blocks nothing.
- Adjacent trivial cleanup that the same batch may sweep: `tests/test-typia-automated/src/utils/headers_to_string copy.ts` is a **SHA-256-identical, unreferenced duplicate** of its neighbour (664 bytes each, zero importers — all four consumers use `../utils/headers_to_string`). It was introduced by migration `a7cbc4f1ae`, the same commit that introduced the direct-matrix suppression #2035 removed. Deleting it is a one-line change with no risk; keep it out of the batch if it complicates the verification story.

samchon:
## Intent

Repair `tests/debug` — the `ttsx` runner the project skill documents as the tool for ad-hoc local repros — so that it compiles and runs its own committed sample again, and remove the shared-config trap that broke it.

Closes #2132.

> **This body was rewritten after implementation disproved the issue's premise.** The original body restated #2132's two-independent-causes framing; that framing is withdrawn. The investigation is on the thread ([round 1](https://github.com/samchon/typia/pull/2133#issuecomment-4998036479)), and #2132's body has been corrected by the lead. What follows is the verified picture.

## What was actually wrong

**`tests/debug` was not broken by config drift. It was left holding invalid code, with no assertion to notice.**

Its committed sample was abandoned scratch that could never compile: `typia.llm.application<IApplication>()` where `array(i: number[])` takes a bare `number[]`, violating the documented keyworded-object restriction (`website/src/content/docs/llm/application.mdx` § Restrictions; `LlmApplicationProgrammer.go:438`). A deliberate product rule, not a typia bug. Because the workspace is a scratchpad with no assertions, and is excluded from `pnpm test:packages` by not being a `test-*` package, nothing ever failed when it stopped working.

Underneath that, a real second defect: **`tests/debug` was the only one of sixteen `tests/*` workspaces that never pinned its own `rootDir`**, so it inherited one from the shared base and broke module identity at runtime.

The two were not merely independent — **the invalid sample masked the `rootDir` bug**. The sample never reached runtime, so the module-identity break could not surface. Had the sample been valid, it would have caught the drift immediately.

### The `TS6196` wall is not a cause

`ttsx` runs a full project check **only after something has already failed**, and that check reports diagnostics for typia's own sources — which carry unused transform-marker type parameters by design, which is why `packages/typia/tsconfig.json` sets `noUnusedLocals: false`.

Control: **`test-typia-schema`, which already pins `rootDir: "src"`, emits the identical 24-line wall** when given any deliberate error. The noise is universal to all sixteen workspaces, appears only on the failure path, and never hides the real error — `Found 1 error in src/index.ts` prints first, formatted, with a code frame.

**So no compiler setting is relaxed anywhere.** That is load-bearing: relaxing `noUnusedLocals`, as the issue originally proposed, would turn the sample's own `TS6133` green by suppression — breaking the very negative criterion the issue requires.

## Changes

| File | Change |
| --- | --- |
| `tests/config/tsconfig.json` | remove `rootDir: "../.."` — the trap |
| `tests/debug/tsconfig.json` | pin `rootDir: "src"`, as all fifteen siblings do |
| `tests/debug/src/index.ts` | correct the sample to the documented shape; print its output |
| `tests/test-typia-automated/src/utils/headers_to_string copy.ts` | delete (SHA-256-identical, zero importers) |

`fabff61c89` is the provenance: it added `rootDir: "../.."` to the shared base. Every workspace that already pinned its own was unaffected; `tests/debug`, which never had one, silently inherited the repository root. **A `rootDir` in `tests/config/tsconfig.json` resolves against that file's own directory, never against the extending workspace, so any value there is the wrong root for every extender** — it served zero consumers correctly.

## Verification

| Check | Result |
| --- | --- |
| `pnpm --filter ./tests/debug start` | prints the application JSON, **exit 0** |
| Fresh sample: `llm.application` + `is` + `protobuf.decode` | all run; `is` returns `true`/`false`; decode round-trips |
| **Negative** — ordinary type error (`TS2322`) | still fails, exit 2 |
| **Negative** — sample's own unused local (`TS6133`) | **still fails, exit 2** |
| **Negative** — typia transform error | still fails, exit 2 |
| **Boundary** — module identity | holds; reverting `rootDir` fails loudly, restoring passes |
| **No-op proof** — 16 workspaces' resolved configs (`tsc --showConfig`) | **semantically identical before/after**, every key |
| **Trap disarmed** — hypothetical 17th workspace omitting the override | before: `rootDir: "../.."` (repo root); after: none imposed, inferred from `include` |
| **Regression** — `pnpm test:packages` | all 14 `test-*` suites Done, 0 failures, exit 0 |

Test infrastructure only: no product source, no `packages/typia/native` change, no package version bump. Campaign CI is cancelled per the exact-SHA gate; local verification is the gate.


samchon:
## Self-Review — Round 1 (findings before remediation)

Complete round over the full base-to-head diff: `tests/debug/tsconfig.json`, `tests/debug/src/index.ts`, and the deleted `tests/test-typia-automated/src/utils/headers_to_string copy.ts`.

### Correction to this PR's opening body — Cause A is misattributed

The body I wrote at open (and issue #2132) describe two independent causes, the first being a `noUnusedLocals` config conflict that makes `tests/debug` fail to compile. **Investigation disproves that.** Recorded here rather than by rewriting the body, per the pull-request skill.

The `TS6196`/`TS6205` wall is not a cause and is not specific to `tests/debug`. `ttsx` runs a full project check **only after something has already failed**, and that check reports diagnostics for typia's own sources — which legitimately carry unused transform-marker type parameters, hence `packages/typia/tsconfig.json` setting `noUnusedLocals: false`.

Truth table (variables: sample validity, `rootDir`, `noUnusedLocals`):

| Sample | `rootDir` | `noUnusedLocals` | Result |
| --- | --- | --- | --- |
| broken `llm.application` | inherited `../..` | `true` (inherited) | 24 TS6196 + transform error, exit 2 |
| broken `llm.application` | `src` | `true` (inherited) | **24 TS6196 still**, exit 2 |
| valid protobuf | inherited `../..` | `true` (inherited) | **0 TS6196**, module-identity error |
| valid protobuf | `src` | `true` (inherited) | **runs and prints, exit 0** |
| ordinary `TS2322` error | inherited `../..` | `true` (inherited) | 24 TS6196 **+ the real error, reported first** |

The lead's single-variable test held the *sample* broken in both arms, so it isolated `rootDir` but never isolated the sample — mistaking noise that accompanies failure for an independent cause.

Decisive control: **`test-typia-schema`, which already has `rootDir: "src"`, emits the identical 24-line wall** when given any deliberate error. The noise is universal to all sixteen workspaces, appears only on the failure path, and never hides the real error (`Found 1 error in src/index.ts` is printed first, formatted, with a code frame).

Consequently **no compiler setting is relaxed anywhere** — not in `tests/debug`, not in `tests/config`, not in `packages/typia`. Blast radius outside `tests/debug` is zero. The negative criterion is satisfied *because* of this: a sample's own unused local still fails with `TS6133`, which a `noUnusedLocals: false` override — the fix the issue proposed — would have silently swallowed.

The real second defect is that **the committed sample was abandoned scratch that could never compile**: its `llm.application` function parameter was `number[]`, violating the documented keyworded-object restriction (`website/src/content/docs/llm/application.mdx`, "Restrictions"). A deliberate product rule, not a typia bug.

### Findings to remediate in this round

1. **`tsconfig.json` comment over-claims an unverified mechanism.** It asserts `ttsx` "remaps them alongside `src` and loads a second copy". I verified the *symptom* and that `rootDir: "src"` cures it; I did not verify ttsx's internal remapping. Asserting an unproven mechanism in a comment is a liability for the next reader. Reword to the observed facts.

2. **The committed sample does not defend the invariant this PR restores.** `llm.application` alone does not construct typia's internal classes, so a future revert of `rootDir: "src"` would leave `pnpm --filter ./tests/debug start` passing — exactly how Cause B hid behind the old `llm.application` sample in the first place. Since `tests/debug` is excluded from `test:packages` (not a `test-*` workspace), the sample is the only local signal there is. Extend it to round-trip `protobuf`, which is what exposed the break.

### Not remediating (recorded)

- **The base `rootDir: "../.."` in `tests/config/tsconfig.json` serves zero consumers correctly** — all sixteen workspaces override it to `"src"`. It is a latent trap rather than a useful default. Removing it is arguably the deeper fix, but it changes the inherited default for sixteen workspaces; out of scope for this batch and offered to the lead.

samchon:
## Self-Review — Round 2

Fresh complete round over the full diff after the Round 1 remediation.

### Finding: Round 1's own rationale was wrong, and the code it produced was gold-plating

Round 1 extended the sample to a three-call `llm.application` + `is` + `protobuf` round-trip, justified by this claim in its comment:

> The protobuf round-trip is deliberate: it is the only call below that constructs typia's internal classes

**That is false.** Verified by reverting `rootDir` and running each call in isolation:

| Sample under reverted `rootDir` | Result |
| --- | --- |
| `typia.is` only | `TypeError: utils_1.Singleton is not a constructor` |
| `typia.llm.application` only | `TypeError: utils_1.Singleton is not a constructor` |
| `typia.protobuf` round-trip only | `TypeError: utils_1.Singleton is not a constructor` |

Every `typia.*` call breaks under a module-identity split, so protobuf is in no way privileged. I caught this only because the Round 1 canary proof fired `utils_1.Singleton`, not the `_ProtobufWriter` I had predicted — the evidence contradicted the comment I had just written.

Two consequences:

1. The extra `is` and `protobuf` calls add **zero** canary value over the single `llm.application` call that was already there. Keeping them would be gold-plating a scratchpad on a disproven premise.
2. It also corrects the record on the original bug: the old committed sample **would** have caught the `rootDir` drift had it been valid. Cause B stayed invisible purely because the sample never compiled — the two defects were not merely independent, the second actively masked the first.

**Remediation:** reverted the sample to the minimal `llm.application` form it always had, with the parameter corrected to the documented keyworded-object shape, and a comment that claims only what is verified.

**Re-verified after remediation:**

- `pnpm --filter ./tests/debug start` → prints the application JSON, exit 0.
- Canary intact — reverting `rootDir: "src"` on the minimal sample still fails loudly: `TypeError: _llmApplicationFinalize_1._llmApplicationFinalize is not a function`, exit 1.

### Also reworded

The `tsconfig.json` comment no longer asserts ttsx's internal remapping mechanism (Round 1 finding 1). It now states the observed failure and the fix, which is all I verified.

### Clean on re-inspection

`rootDir: "src"` is isolated — no config extends `tests/debug/tsconfig.json`, and the only reference to `@typia/debug` outside its own `package.json` is the `.agents/skills/project/SKILL.md` line describing the workspace, which stays accurate. The deleted `headers_to_string copy.ts` is SHA-256-identical to its neighbour, tracked, has zero importers, and lives in the hand-maintained `src/utils` tree rather than the regenerated `src/features` tree, so the suite's generator does not reference it.

samchon:
## Self-Review — Round 3

Fresh complete round over the final diff.

### Finding: both comments still understated the failure

Round 2's comments claimed the break surfaces on "any `typia.*` call", and that it makes "typia's internal classes stop being constructible". The stack traces contradicted both — they point at `packages/utils/src/index.ts:42`, i.e. **module top-level evaluation**, not a call site.

Verified with a sample that imports typia and calls nothing at all, under a reverted `rootDir`:

```
TypeError: _llmApplicationFinalize_1._llmApplicationFinalize is not a function
    at Object.<anonymous> (packages/utils/src/index.ts:42:54)
```

So the break is unconditional at import, and it is not confined to classes — `_llmApplicationFinalize` is a plain function ("is not a function"), while `_ProtobufWriter` is a class ("is not a constructor"). Both comments were narrower than the truth.

**Remediation:** both now say that merely importing typia is enough to surface it, and the tsconfig comment cites `_ProtobufWriter_1._ProtobufWriter is not a constructor` as an example rather than as the shape of the failure.

This also strengthens the canary conclusion from Round 2: because the failure triggers on import, *any* sample in this workspace catches a `rootDir` regression. The sample's content is irrelevant to that guarantee, which is a further argument for keeping it minimal.

**Re-verified:** `pnpm --filter ./tests/debug start` → prints the application JSON, exit 0.

### Everything else clean

No further findings. The diff is three files: `rootDir: "src"` in `tests/debug/tsconfig.json`, the corrected scratchpad sample, and the deleted duplicate. No compiler setting is relaxed, no product source is touched, and no package version needs a bump (nothing under `packages/typia/native` changed).

samchon:
## Self-Review — Round 4 (clean, final)

Fresh complete round over the final diff. **No findings.** Rounds 1–3 each produced a finding and a remediation; this round produced none, so the Self-Review stops here.

### Final surface

Three files, test infrastructure only:

| File | Change |
| --- | --- |
| `tests/debug/tsconfig.json` | `rootDir: "src"` — the missing 16th override |
| `tests/debug/src/index.ts` | sample corrected to the documented keyworded-object shape; prints its output |
| `tests/test-typia-automated/src/utils/headers_to_string copy.ts` | deleted (SHA-256-identical, zero importers) |

No product source, no shared config, no compiler setting relaxed, and nothing under `packages/typia/native` — so no package version bump is required.

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter ./tests/debug start` | prints the application JSON, **exit 0** |
| Fresh sample: `llm.application` + `is` + `protobuf.decode` | all run; `is` returns `true` and `false` correctly; decode round-trips |
| **Negative** — ordinary type error (`TS2322`) | still fails, exit 2 |
| **Negative** — sample's own unused local (`TS6133`) | **still fails, exit 2** |
| **Negative** — typia transform error | still fails, exit 2 |
| **Boundary** — module identity | holds; reverting `rootDir` fails loudly, restoring it passes |
| **Regression** — `pnpm test:packages` | **all 14 `test-*` suites Done, 0 failures, exit 0** (clean run, no concurrent activity) |

The `TS6133` row is the important one. The fix the issue proposed — relaxing `noUnusedLocals` for this workspace — would have turned that row green-by-suppression and broken the issue's own negative criterion. Because no setting was relaxed, a contributor's own unused local is still caught.

### Offered to the lead, deliberately not done here

1. **`tests/config/tsconfig.json` sets `rootDir: "../.."`, which serves zero consumers correctly** — all sixteen workspaces override it to `"src"`. It is a latent trap, not a useful default, and it is the actual provenance of this bug: `fabff61c89` added it to the base, every workspace that already pinned `rootDir` was unaffected, and `tests/debug` — which never had one — silently inherited the repository root. Deleting it from the base would remove the trap by design rather than detect it by check, with provably no effect today. It touches shared surface, so it is the lead's call.
2. **A config-drift check.** The only *behavioural* seal that would have caught this is running `tests/debug` in CI, since the workspace is excluded from `test:packages` by not being a `test-*` package — that exclusion is precisely why it rotted. It is non-vacuous (it fails on pre-fix `master` for both defects), but it converts a scratchpad whose purpose is holding broken code into a CI-gated artifact, which is a policy fork rather than a fix. A tsconfig-shape check ("every `tests/*` pins `rootDir`") is also non-vacuous today but encodes a convention and has no chartered home — `test-feature-identity` is chartered for feature naming. Recommending (1) over either.

samchon:
## Self-Review — Round 5 (clean, final) — `tests/config` rootDir removal

Per the lead's Decision 1, `rootDir: "../.."` is removed from `tests/config/tsconfig.json`. That is a new change after Round 4 closed, so this is a fresh complete round over the four-file surface. **No findings.**

The body has been rewritten to the verified picture, per the lead's point 2 — normally the opening body stays as the historical intent statement, so the rewrite is noted inline there with a link back to the round-1 investigation, keeping the chronology on the thread.

### The no-op is proven, not argued

`tsc --showConfig` fully resolves `extends`, so it is the oracle. Captured every `tests/*` workspace's resolved config before and after, compared **order-independently across every key** (not just `rootDir`):

```
IDENTICAL  debug            IDENTICAL  test-typia-automated
IDENTICAL  template         IDENTICAL  test-typia-bundler-cache
IDENTICAL  test-error       IDENTICAL  test-typia-cli
IDENTICAL  test-feature-identity  IDENTICAL  test-typia-exact-optional
IDENTICAL  test-interface   IDENTICAL  test-typia-generate
IDENTICAL  test-langchain   IDENTICAL  test-typia-schema
IDENTICAL  test-mcp         IDENTICAL  test-utils / test-utils-automated / test-vercel

RESULT: all 16 resolved configs semantically identical -- PROVEN NO-OP
```

A first pass compared raw `--showConfig` text and reported all sixteen as *changed*. That was the oracle's fault, not the code's: the resolved `rootDir` value stayed `./src` everywhere and merely **moved position** in the emitted JSON, because it now resolves from the workspace rather than from the base. Comparing normalised, order-independent objects shows the true result. Worth recording — a raw-text diff here would have produced sixteen false alarms.

The same oracle also caught a real defect in my first edit: a `/*` block opened inside a `//` line never opens a block comment, so my multi-line rationale parsed as raw JSON and broke the config (`TS1136`). Fixed to `//`-prefixed lines before the proof was re-run.

### The trap is disarmed, demonstrated

Built a hypothetical seventeenth workspace that extends the base and forgets the override:

| | resolved `rootDir` |
| --- | --- |
| before | `"../.."` — the repository root. **Trap armed.** |
| after | none imposed; TS infers `./src` from `include`. **Safe.** |

This is the point of the change: `tests/debug` did not drift away from a correct default, it inherited a base setting that is wrong for everyone. A `rootDir` in `tests/config/tsconfig.json` resolves against `tests/config`, never against the extender, so no value there can be right.

### Blast radius bounded by inspection

Exactly **15** workspaces extend `tests/config` (`debug`, `template`, 13 `test-*`); `test-interface` extends the root `config/` instead. Nothing outside `tests/` extends `tests/config` — the `packages/*` configs extend the **root** `config/tsconfig.json`, a different file, untouched here. All sixteen are covered by the proof above.

### Regression

`pnpm test:packages` re-run after the change: **all 14 `test-*` suites Done, 0 failures, exit 0.** `pnpm --filter ./tests/debug start` still prints, exit 0.

### CI gate — one honest deviation

Runs for `e16d169e3a`: the expensive `test` workflow (`29548588837`) was **cancelled** as required. `spell-check` (`29548588853`) completed in **12 seconds — faster than cancellation could land — and passed**. It was observed active and did not end `cancelled`, so it is a deviation from the exact-SHA gate's letter, recorded rather than glossed. No repository Actions or workflow settings were touched.

samchon:

## The issue's premise was wrong and this batch disproved it. That is the headline.

I published #2132 with two causes. **Cause A does not exist**, and I have corrected the issue body accordingly. A valid sample compiles clean under the inherited `noUnusedLocals: true` with zero `TS6196`, even with no `rootDir`. The wall appears only on the failure path, because `ttsx` runs a full project check when something already failed and that check reports typia's own source diagnostics. The control that settles it: `test-typia-schema`, which already has `rootDir: "src"`, emits the identical wall on any deliberate error.

**My "single-variable test" was not single-variable** — it isolated `rootDir` while leaving the sample broken in both arms, so I read noise that accompanies failure as a cause of it.

**And my proposed Approach contradicted my own Acceptance criterion.** Relaxing `noUnusedLocals` would have turned the sample's own `TS6133` green by suppression, which is exactly the negative the issue says must keep failing. I confirmed that from my own hands below.

The real second defect: the committed sample was **abandoned scratch that could never compile** — an `llm.application` parameter typed `number[]`, violating a documented restriction. `tests/debug` did not drift out of a working state; it was left holding invalid code, with no assertion to notice, in a workspace excluded from `pnpm test:packages` by not being a `test-*`.

## Lead verification on `45c84a7e48`

| Check | Result |
| --- | --- |
| `pnpm --filter ./tests/debug start` | **exit 0** — prints the application JSON. The documented runner runs. |
| deliberate `TS2322` planted in the sample | **exit 2** — a genuine type error still fails |
| tree after restoring the sample | clean |

The negative check also reproduced the batch's disproof from my side: my planted `TS2322` triggered the full project check, which then emitted `TS6196`/`TS6205` **alongside** it. Those diagnostics are the symptom that accompanies any failure, not a cause of one.

## Removing `tests/config`'s `rootDir` is proven, not argued

I approved deleting the base `rootDir: "../.."` because it serves zero consumers correctly — all sixteen workspaces override it — and it is the actual provenance: `fabff61c89` added it, every workspace that already pinned `rootDir` was unaffected, and `tests/debug` silently inherited the repository root.

The batch proved the no-op with `tsc --showConfig`, which fully resolves `extends`, comparing every workspace's resolved config order-independently: **all sixteen semantically identical.** Blast radius bounded by inspection — fifteen workspaces extend `tests/config`; `test-interface` and every `packages/*` config extend the **root** `config/tsconfig.json`, a different file, untouched.

Two things the oracle caught that the author did not, both worth recording:

1. **The first proof pass reported all sixteen as CHANGED** — a false alarm, because `rootDir` stayed `./src` everywhere and merely moved position in the emitted JSON once it resolved from the workspace rather than the base. A raw-text diff produces sixteen phantoms here. Trusting that first result would have produced a wrong report back to me.
2. **The first edit broke the config**: a `/*` opened inside a `//` line never starts a block comment, so the rationale parsed as raw JSON (`TS1136`).

## No drift check, deliberately

The trap is removed rather than detected. Putting `tests/debug` in CI would convert a scratchpad whose *purpose* is holding broken code into a gated artifact — a policy change, not a fix — and a tsconfig-shape check would encode a convention with no chartered home. This campaign has found eight oracles that could not fail; it does not need a ninth that fails for the wrong reason. Removing the trap is the better fix, and the batch argued for that over its own alternatives.

## Regression

`pnpm test:packages` re-run after the config change: **all 14 `test-*` suites Done, 0 failures, exit 0.** Self-Review reached round 5 clean, and rounds 1–3 each retracted something — including the author's own round-1 protobuf rationale, which was false: the module-identity break triggers on *import* and reaches plain functions, not only classes. The round-1 sample expansion was reverted as gold-plating.

The dead `headers_to_string copy.ts` was swept in: SHA-256-identical, zero importers, in hand-maintained `src/utils` rather than a regenerated tree, and already covered by the same `test:packages` run.

`45c84a7e48` settled after five polls. `29549022228` (spell-check) reached terminal `success` before the cancel could land and is recorded as `raced_terminal`; the expensive `test` run was cancelled.

The batch flagged this same race on its earlier SHA as an honest gate deviation rather than reporting a clean gate. That instinct is right, and it is now the script's recorded behaviour rather than a fatal error — a twelve-second job finishing before a cancellation request lands is physics, not a violation. What matters is that the record says which it was.

Issue #2134, opened 2026-07-17
transform --output js reports "no output produced" for any project with an outDir


samchon:
## Problem

`ttsc-typia transform --file <f> --output js` reports `no output produced` and exits 3 for **any project with an `outDir`**, even when the transform succeeded and produced output. The mode is unusable for any realistic project layout.

`sameSourceStem` compares **whole path stems**, so it asks whether `<p>/dist/main` equals `<p>/src/main` — which it never does once `outDir` redirects the emit. Its own comment says it matches "on basename stem", which is what would have worked.

Affected: any consumer of `--output js`. Because `--output` defaults to `ts`, no default workflow breaks — this is opt-in via the flag or `TYPIA_TTSC_TRANSFORM_OUTPUT=js`, which is why it has gone unnoticed.

## Evidence

Verified on `master` with a single-variable control — same source, only `outDir` differing:

| Project | Result |
| --- | --- |
| valid source **with** `outDir` | **exit 3, `no output produced`** |
| identical source **without** `outDir` | exit 0, 500 bytes emitted |

The transform itself succeeds in both. The failure is entirely in the post-emit identity check.

Owning symbol: `sameSourceStem`, `packages/typia/native/cmd/ttsc-typia/transform.go:253-256`. The comment at `transform.go:233` states the intent — match on **basename** stem — and the implementation compares the full path stem instead.

**The repository already documents this bug as a workaround reason.** The `transform_single_file_*` fixture comments added by PR #2120 explicitly note that the fixture omits `outDir` because of it (lines 91-95). A defect that a fixture has to route around, in a comment, is a defect the repository already knows about and has been living with.

## Consequence surface

- **Owner:** `sameSourceStem` in `packages/typia/native/cmd/ttsc-typia/transform.go`. Native Go.
- **Consumers:** the `--output js` path only — `runTransformSingleJS`. The `ts` path (the default) does not use this check.
- **Reachable via:** `--output js` explicitly, or `TYPIA_TTSC_TRANSFORM_OUTPUT=js`.
- **Not affected:** the default `ts` single-file path, the project host, and every workflow that does not opt into `js`.
- **Interaction with #2117:** PR #2120 made both single-file paths consult diagnostics before publishing. That fix is correct and unrelated; this is the *identity* check that runs after a successful transform, not the diagnostic gate.

| Case | Expected |
| --- | --- |
| `--output js`, project with `outDir` | exit 0, output published |
[...]

PR #2137, opened 2026-07-17
fix(typia): resolve the single-file js transform output path from the compiler

samchon:
Implements #2134.

## Intent

`ttsc-typia transform --file <f> --output js` reports `no output produced` and exits `3` for **any project with an `outDir`**, even though the transform succeeded and the emit wrote the artifact. The mode is unusable for any realistic project layout.

The cause is entirely in the post-emit identity check. `sameSourceStem` compares **whole path stems**, so it asks whether `<p>/dist/main` equals `<p>/src/main` — which it never does once `outDir` redirects the emit. The check's own comment at `transform.go:233` says it matches "on basename stem", which is what would have worked.

Verified with a single-variable control — same source, only `outDir` differing:

| Project | Result on master |
| --- | --- |
| valid source **with** `outDir` | **exit 3, `no output produced`** |
| identical source **without** `outDir` | exit 0, 500 bytes emitted |

## Approach

The invariant: **the check identifies the emitted artifact that corresponds to the requested source file, regardless of where `outDir` places it.**

The fix does not stop at making the comment true. A bare basename comparison would still be wrong: `rootDir`, `rootDirs`, and nested source directories all reposition the emitted path independently of `outDir`, and two sources with the same basename in different directories would collide into one another under a basename match. Instead, `transformSingleToJavaScript` asks the **compiler's own output-path resolution** — the same `outputpaths` computation the emitter itself uses to choose where to write — for the JS path belonging to the target source, and matches the emit callback against that resolved path.

That makes the check exact rather than heuristic: it is right for every `outDir` / `rootDir` / `rootDirs` / nested-directory combination by construction, because it is the emitter's own answer, and two same-basename sources in different directories resolve to different output paths and cannot be confused.

The check keeps its real job. When the emit genuinely produces no artifact for the requested file — a declaration-only source, for instance — no write ever arrives for the resolved path, `found` stays false, and the host still exits `3` with `no output produced`.

## Scope

- `packages/typia/native/cmd/ttsc-typia/transform.go` — resolve the expected output path from the compiler; `sameSourceStem` is removed.
- Regression matrix pinning the issue's full case matrix, including the negative twin and the same-basename collision boundary.
- **Removes the workaround this unblocks.** #2120's `transform_single_file_*` fixture omitted `outDir` *because of this bug*, with a comment saying so. The fixture now carries an `outDir`, and the comment is gone — otherwise the coverage would stay shaped around the defect.

## Verification

Pending; recorded on this thread as it completes. Campaign CI is suspended per `.agents/skills/issue-campaign/development.md`; verification is local.

## Deferred / coordination

- **The `typia` version bump is pending lead assignment.** This changes native Go under `packages/typia/native`, so Change Integrity requires a bump; the campaign lead serializes the version across the chain, so this branch deliberately does not pick one and is not rebased onto master.
- #2117's diagnostic gate (PR #2120) is untouched: it is the check that runs *before* publishing; this is the identity check that runs after a successful transform.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/command_routes_and_diagnostics_coverage_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/command_routes_and_diagnostics_coverage_test.go
@@ -175,6 +175,8 @@ func TestCommandRoutesAndDiagnosticsCoverage(t *testing.T) {
 				"--rewrite-mode", "none",
 			})
 		}, 0},
+		// Publishes, rather than the 3 this expected while samchon/typia#2134
+		// made `--output js` miss its own artifact under the fixture's `outDir`.
 		{"transform_js_out", func(t *testing.T) int {
 			return runTransform([]string{
 				"--cwd", primitives,
@@ -184,7 +186,7 @@ func TestCommandRoutesAndDiagnosticsCoverage(t *testing.T) {
 				"--out", filepath.Join(t.TempDir(), "main.js"),
 				"--rewrite-mode", "none",
 			})
-		}, 3},
+		}, 0},
 		{"transform_ts_stdout_write_error", func(t *testing.T) int {
 			return commandRoutesWithFailingStdout(func() int {
 				return runTransform([]string{
--- a/packages/typia/native/cmd/ttsc-typia/transform.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform.go
@@ -110,7 +110,7 @@ func runTransform(args []string) int {
   }
   return runTransformSingle(cwd, *out, &transformDiags, func() (string, int) {
     if *output == "js" {
-      return transformSingleToJavaScript(prog, typiaTransform, absFile)
+      return transformSingleToJavaScript(prog, typiaTransform, target)
     }
     return transformFileToTypeScript(prog, typiaTransform, target), 0
   })
@@ -221,17 +221,35 @@ func transformFileToTypeScript(
 // matches the build host, which fails the whole build on any transform diagnostic:
 // the emit that produced this text is the same one that could not lower the other
 // call, so its success is not independently trustworthy.
+//
+// The emit writes every project output through one callback, so the target's own
+// artifact has to be identified among them. That identity comes from the
+// compiler's output-path resolution rather than from a resemblance between the
+// source and emitted paths, because the emit chooses its write path with that
+// same resolution: `outDir`, `rootDir`, and nested source directories may move
+// the artifact anywhere without the two disagreeing, and two sources sharing a
+// basename in different directories resolve to different artifacts instead of
+// collapsing onto one. Deriving the path instead of matching stems is what fixed
+// samchon/typia#2134, where a whole-path-stem match asked whether `dist/main`
+// equalled `src/main` and so reported "no output produced" for every project
+// with an `outDir`.
 func transformSingleToJavaScript(
   prog *driver.Program,
   typiaTransform driver.PluginTransform,
-  absFile string,
+  target *shimast.SourceFile,
 ) (string, int) {
+  // The resolution answers with an empty path for a source that has no
+  // JavaScript lane of its own -- a JSON module that would be emitted over
+  // itself. Nothing the emit writes is then this file's artifact, so guard the
+  // comparison rather than let an empty expectation match anything.
+  expected := ""
+  if paths := shimcompiler.GetOutputPathsFor(target, prog.TSProgram.Options(), prog.TSProgram, false); paths != nil {
+    expected = filepath.ToSlash(paths.JsFilePath())
+  }
   var captured string
   found := false
-  targetKey := filepath.ToSlash(absFile)
   writeFile := shimcompiler.WriteFile(func(fileName, text string, _ *shimcompiler.WriteFileData) error {
-    // The emitted .js maps back to its .ts; match on basename stem.
-    if sameSourceStem(targetKey, fileName) {
+    if expected != "" && filepath.ToSlash(fileName) == expected {
       captured = text
       found = true
     }
@@ -242,17 +260,10 @@ func transformSingleToJavaScript(
     return "", 3
   }
   if !found {
-    fmt.Fprintf(stderr, "ttsc-typia transform: no output produced for %s\n", absFile)
+    fmt.Fprintf(stderr, "ttsc-typia transform: no output produced for %s\n", target.FileName())
     return "", 3
   }
   return captured, 0
-}
-
-// sameSourceStem reports whether an emitted .js path corresponds to a .ts source
-// path by comparing their extension-stripped values.
-func sameSourceStem(tsPath, jsPath string) bool {
-  return strings.TrimSuffix(filepath.ToSlash(jsPath), filepath.Ext(jsPath)) ==
-    strings.TrimSuffix(tsPath, filepath.Ext(tsPath))
 }
 
 func writeSingleOutput(text, outPath string) int {
--- a/packages/typia/native/cmd/ttsc-typia/transform.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform.go
@@ -231,7 +231,7 @@ func transformFileToTypeScript(
 // basename in different directories resolve to different artifacts instead of
 // collapsing onto one. Deriving the path instead of matching stems is what fixed
 // samchon/typia#2134, where a whole-path-stem match asked whether `dist/main`
-// equalled `src/main` and so reported "no output produced" for every project
+// equaled `src/main` and so reported "no output produced" for every project
 // with an `outDir`.
 func transformSingleToJavaScript(
   prog *driver.Program,
--- a/packages/typia/native/cmd/ttsc-typia/transform_declaration_source_produces_no_javascript_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform_declaration_source_produces_no_javascript_test.go
@@ -17,9 +17,12 @@ import (
 // of returning a successful but blank result.
 //
 // The file name deliberately spells out "javascript". Go reads a trailing `_js`
-// on a source file name as a `GOOS=js` build constraint, so this case sat in the
-// tree under the name `..._produces_no_js_test.go` and compiled on no platform
-// the suite ever runs -- the tag it carries was never the reason it did not run.
+// on a source file name as a `GOOS=js` build constraint, so under the earlier
+// name `..._produces_no_js_test.go` this case built only for `GOOS=js` -- where
+// `main.go` (`//go:build !js`) drops out and the test package does not compile
+// at all. It therefore ran on no platform whatsoever, and the tag it carries was
+// never the reason. Keep a `_js`, `_windows`, `_linux`, or `_darwin` ending off
+// a Go file name unless that constraint is meant.
 //
 // 1. Create a project that includes only a declaration source file.
 // 2. Run single-file transform with JavaScript output.
--- a/packages/typia/native/cmd/ttsc-typia/transform_declaration_source_produces_no_javascript_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/transform_declaration_source_produces_no_javascript_test.go
@@ -10,17 +10,22 @@ import (
 	"testing"
 )
 
-// TestTransformDeclarationSourceProducesNoJS verifies JS transform reports empty emission.
+// TestTransformDeclarationSourceProducesNoJavaScript verifies JS transform reports empty emission.
 //
 // Declaration files can belong to the program while producing no JavaScript
 // output. The single-file transform path must detect that empty capture instead
 // of returning a successful but blank result.
 //
+// The file name deliberately spells out "javascript". Go reads a trailing `_js`
+// on a source file name as a `GOOS=js` build constraint, so this case sat in the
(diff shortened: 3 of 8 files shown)


samchon:

Complete round over the base-to-head diff (`1117b993fa..17eb6ee878`) plus uncommitted work. Recording findings before acting on them.

### F1 — spell check, uncommitted fix pending

The repository spell check (`typos`, `locale = 'en-us'`) failed twice on this branch, both times on my own en-GB spellings: `neighbour` and `equalled` on `5260d8aa89`, then `equalled` again on `17eb6ee878` after I reintroduced it in a new comment. The fix is applied locally but not yet committed.

**Plan:** commit it, and scan every added line for en-GB forms rather than round-tripping through CI a third time.

### F2 — my own comment on the revived test understates the proven fact

`transform_declaration_source_produces_no_javascript_test.go` says the old `_js` name "compiled on no platform the suite ever runs". That is weaker than what is now proven. `main.go` is `//go:build !js` and `main_wasm.go` is `//go:build js && wasm`, so under `GOOS=js` the test package does not compile at all (`run` is undefined). The file could therefore never run on **any** platform — not merely on the ones the suite targets.

**Plan:** correct the comment to the proven claim. Shipping a comment that understates its own evidence is the same defect class this pull request exists to remove.

### F3 — `_js` lane in the fixture coverage suite asserts shape, not lowering

The rewritten `_js` lane asserts exit 0 plus emitted CommonJS (`"use strict"`, `exports.`), not that the typia call was lowered.

**Plan:** keep as is, and record the reason. The lowering is pinned precisely by `TestTransformSingleFileJSLocatesOutputAcrossLayouts`, which asserts both the emitted validator and the absence of `typia.is<`. This suite's declared job is Go coverage of the command branches, and its `_ts` twin is deliberately just as loose. Relative to the previous state — where the run aborted at the identity check and never reached the printer — asserting that a real emit arrives is the transformation-direction evidence this lane owes.

### F4 — the `expected != ""` guard is defensive, not reachable as a behavior difference

Removing it would not change any observed outcome: `writePluginEmitOutput` returns early on an empty file name, so the callback never receives `""`.

**Plan:** keep it. It enforces our own invariant instead of depending on a ttsc upstream guard, and its comment claims only that, without overstating. The one route that reaches an empty resolved path — a JSON module emitted over itself — is pinned by `TestTransformSingleFileJSEmptyEmitReportsNoOutput/json_module_emitted_over_itself`.

### F5 — pre-existing unrelated failure, out of scope

`TestProjectCatalogTransformCoverage/debug` fails under `-tags typia_native_internal`. **It is not caused by this branch:** it reproduces on pristine `origin/master` sources with my diff removed. It is fallout from #2133 (`1117b993fa`), which made `tests/debug` valid so it no longer emits diagnostics, while the catalog still expects exit 3. The tagged suite is not in default CI, which is why the merge did not catch it.

**Plan:** leave it and report it to the campaign lead as a separate candidate. It is a different cause, and the fix is a judgment call this batch does not own.

### Verified clean in this round

- No website guide references `--output js` or `TYPIA_TTSC_TRANSFORM_OUTPUT`, so no documentation change is owed.
- The wasm host builds with the change (`GOOS=js GOARCH=wasm`), so `main_wasm.go` -> `jsTransform` -> `runTransform` inherits the fix without a signature break.
- `origin/master` advanced to `1d0fab689a` (#2130) mid-batch. This branch is deliberately **not** rebased; the base stays `1117b993fa` per the campaign's lead-owned rebase rule.

samchon:

Fresh complete round over `1117b993fa..59eda79c2f`. Round 1's changes are in, so this is a new full pass, not a recheck of the last fix.

### F6 — the matrix never pins the case where *several* artifacts for one source reach the callback

Every layout I added emits exactly one file for the target, so each proves only that the resolved path is *found*. It does not prove it is **discriminated**. Turn on `declaration` and `sourceMap` and one source produces three writes through the same callback — `dist/main.js`, `dist/main.js.map`, `dist/main.d.ts` — and only the first is this mode's artifact.

This is the sibling-artifact twin of the same-basename case: there I proved two *sources* are not confused, here two *outputs of one source* must not be. A `.js.map` is the sharper of the two, because its `sourcesContent` embeds the original TypeScript — so capturing it would return text containing `typia.is<`, i.e. an untransformed-looking artifact published behind exit 0.

**Plan:** add a `declaration`/`sourceMap` layout to `TestTransformSingleFileJSLocatesOutputAcrossLayouts` and confirm by mutation that a check which stops discriminating captures the wrong artifact.

### Re-verified clean in this round

- The resolution I call and the one the emit uses are provably the same computation: identical `options` (`p.TSProgram.Options()`), identical `forceDtsEmit` (`false`), the same `*ast.SourceFile` instance, and a host whose three `OutputPathsHost` methods delegate straight to `p.TSProgram` — which is what I pass. They cannot drift apart.
- `target.FileName()` reproduces the old `absFile` text in the `no output produced` report; confirmed against real CLI stderr.
- Repository spell check now verified locally with the CI binary and config (`typos v1.48.0`, `--config ./typos.toml`): repo-wide exit 0.
- No unused imports or variables after the signature change; `go vet` clean, including `GOOS=js GOARCH=wasm`.

samchon:
## Self-Review round 3 — findings and remediation plan

Fresh complete round over `1117b993fa..c0514ec0be`.

### F7 — the empty-emit path is never checked for publishing nothing

`TestTransformSingleFileJSEmptyEmitReportsNoOutput` asserts exit 3 and the report, but not that `--out` stayed unwritten.

That is a real hole rather than a duplicate of the atomicity test, because the two outcomes leave `runTransformSingle` through **different branches**: the diagnostic route returns at `len(*transformDiags) > 0`, while an empty emit returns at the later `if code != 0`. Only the first is pinned against publishing today. An exit 3 that still wrote a file would satisfy every assertion in this test.

**Plan:** require `--out` to stay absent on all three empty-emit probes.

### F8 — the pull request body overstates `rootDirs`, contradicting my own test

The body says `rootDir`, `rootDirs`, and nested source directories "all reposition the emitted path". `rootDirs` does **not**. It is a module-resolution feature and appears nowhere in the compiler's output-path resolution: `GetCommonSourceDirectory` consults `rootDir`, else the tsconfig's directory, else the computed common directory, and `getOutputPathWithoutChangingExtension` uses only `outDir` plus that common directory. My own test comment says exactly this, so the body contradicts the branch.

I inherited the phrasing from the issue's Approach section without re-deriving it. Shipping a claim I had already disproved is the same defect class this pull request exists to remove — the original `sameSourceStem` comment was wrong in precisely that way.

**Plan:** correct it here rather than rewrite the body, per the pull-request skill's chronology rule. **The operative claim is:** `outDir` and `rootDir` reposition the emitted path and nested source directories carry into it; `rootDirs` does not, and the matrix pins that the mode works when it is configured rather than asserting it moves anything. The argument against a basename match never depended on `rootDirs` — the same-basename collision and the sibling-artifact case carry it, both proven by mutation.

### Re-verified clean in this round

- `.tsx` needs no case: the resolution returns `.jsx` under `jsx: "preserve"` by construction, and that layout already worked before the fix, so it is neither a regression nor a new path.
- Concurrency: the JS lane's write loop is sequential and completes before the dts lane's mutex-guarded `Emit`; the capture closure cannot race. A `.d.ts` cannot match the resolved `.js` path regardless.
- Windows/POSIX: both sides of the comparison come from one tspath-normalized computation over the same source file, so case and separators cannot disagree; `ToSlash` on both is defensive only.
- No dangling reference to the deleted `transformSingleFileTSConfig` or `sameSourceStem`.

samchon:
## Self-Review round 4 — clean, review closed

Fresh complete round over `1117b993fa..fa01be3d5f`. **No finding survived verification, so the review stops here.**

One item examined and deliberately kept: `GetOutputPathsFor` never returns nil in the current upstream, so the nil guard is not reachable today. It stays because that call crosses into the ttsc shim over typescript-go internals — a contract this repository does not own — and the alternative to a cheap guard is a panic in a CLI. The same reasoning keeps the `expected != ""` guard (round 1, F4).

### Final verification, all re-run at `fa01be3d5f`

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` | ok, 40.1s |
| `go -C packages/typia/test test -p 2 ../native/...` | ok, no failures |
| `pnpm --filter ./tests/test-typia-cli start` | Success |
| `go -C packages/typia/test test -count=1 -p 2 ./...` (public) | ok |
| `go vet`, incl. `GOOS=js GOARCH=wasm` build | clean |
| `typos v1.48.0 --config ./typos.toml` (repo-wide, CI binary) | exit 0 — and CI `spell-check` passed on this SHA |

`-tags typia_native_internal` is green except `TestProjectCatalogTransformCoverage/debug`, which is **pre-existing and not from this branch** — reproduced on pristine `origin/master` sources with this diff removed (round 1, F5).

### End-to-end, against the real CLI binary

The issue's own single-variable control, re-run post-fix:

| Project | On master | On this branch |
| --- | --- | --- |
| valid source **with** `outDir` | exit 3, `no output produced` | **exit 0, 500 bytes** |
| identical source **without** `outDir` | exit 0, 500 bytes | exit 0, 500 bytes — unchanged |

Both artifacts are byte-identical and carry the lowered validator (`_io0`, inline `"number" === typeof`), so the mode returns the same result regardless of where `outDir` puts the emit. The 500 bytes match the issue's reported baseline exactly.

### The fix is load-bearing, by mutation

| Mutation | Caught by | Outcome |
| --- | --- | --- |
| whole-path stem (the defect) | layout matrix | 4 of 6 layouts exit 3 — the original fail-first evidence |
| **basename stem** — the repair the old comment named | same-basename case | publishes **another file's validator** behind exit 0 |
| always match | all 3 empty-emit probes | genuine empty emits flip to exit 0 |
| prefix match — finds but does not discriminate | `declaration_and_source_map` | publishes the **`.d.ts`** as the JavaScript |
| publish before checking the producer's code | empty-emit probes | exit 3 with a file written |

The second row is the answer to "why not just make the comment true": a basename match unblocks `outDir` and silently introduces a worse defect.

### Ready

Ready for the lead's rebase, version bump, and merge. **The bump is still owed and unpicked here** — native Go changed, and `origin/master` has since moved to `1d0fab689a` (#2130), which carries its own bump, so this needs one on top of that.

samchon:

## Version: deliberately untouched

`packages/typia/package.json` is **not** in this diff, and that is intentional. The maintainer owns releases and version numbers; the campaign does not touch them. `master` already differs from the last published release, so the development skill's rule is satisfied by the state of the tree, and the release commit decides the real number.

## Lead verification on `33a1d55402`

| Gate | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` | ok, 30.6s |

`TestProjectCatalogTransformCoverage/debug` still fails under `-tags typia_native_internal`, as this batch reported. I reproduced it on **pristine `master` with this diff removed** and published it as **#2141** — it is fallout from my own #2133 merge, not from this branch.

## The batch refused my brief's hint, and proved the hint was harmful

I pointed at the code's own comment — "match on basename stem" — as the likely repair. **Mutation proved that repair is actively worse than the bug**: it unblocks `outDir` and then publishes *another file's validator* under exit 0. A wrong artifact shipped silently beats a correct refusal, so the comment named a fix that would have traded a loud failure for a quiet one.

The batch went to `shimcompiler.GetOutputPathsFor` instead — the same computation the emit itself uses to choose its write path, with identical options and the same program. The two cannot drift, because they are one computation rather than two agreeing ones. That matters here specifically: this campaign's recurring defect is exactly *two implementations of one contract silently diverging*, and a basename comparison would have created a third.

## A correction to my issue's premise

I wrote that `rootDirs` repositions the emit. **It does not** — it is module-resolution only, verified in `outputpaths`. The case against basename rests on the same-basename collision and sibling artifacts, both mutation-proven, not on `rootDirs`.

## The check still does its real job

Three genuine empty-emit routes still exit 3 — declaration source, output escaping `outDir`, JSON module emitted over itself — each proven by mutation: removing the guard flips all three to exit 0, and a publish-before-check writes a file on exit 3. The batch also caught a check that *finds* the output without *discriminating* it, which publishes the `.d.ts` as the JavaScript.

## Three findings beyond the brief, two of them tests that could not work

1. **`transform_fixtures_emit_coverage_test.go` asserted the bug** across all eight fixtures. It would have failed the repair, and it left the JS emit path it exists to cover permanently unreached.
2. **`transform_declaration_source_produces_no_js_test.go` never ran anywhere.** Go reads a trailing `_js` as `GOOS=js`, and under `GOOS=js` the package cannot compile because `main.go` carries `//go:build !js`. A test file that has never executed, sitting in the tree looking like coverage. Renamed; it passes. The repo was swept — it is isolated, not a class.
3. #2141, above.

That second one is the ninth member of this campaign's cannot-fail-oracle family, and the most complete: not an assertion that could not fail, but a file that could not run.

## Workaround removed

#2120's fixture comment and its `outDir` omission are gone — the fixture now carries `rootDir`/`outDir` and shares `transformDiagnosticTSConfig`. The duplicate constant existed only to route around this bug. Coverage shaped around a defect is coverage that teaches the defect.

## #2117 untouched

`--output js` with typia diagnostics still exits 3 with the diagnostic and publishes nothing — now verified **under an `outDir`**, which was unreachable before this fix.

`33a1d55402` settled after five polls, all six runs `cancelled`, no race.

PR #2138, opened 2026-07-17
fix(utils): make NamingConvention.localize total and select automated structures by declaration

### LINKED ISSUE #2136: NamingConvention.localize throws on the empty string, and the automated selector carries two dead predicates
## Problem

Two independent gaps, each stated at the size the evidence actually supports.

1. **`NamingConvention.localize("")` throws** where all five of its siblings return `""`. `capitalize`, `camel`, `snake`, `kebab`, and `pascal` handle the empty string; `localize` does `str[0]!.toLowerCase()` on `undefined` and raises a `TypeError`. This is a public `@typia/utils` API.
2. **`TestAutomation.getStructures()` selects fixtures by raw source-substring match**, and two of its four predicates are **dead** — they exclude nothing at all.

## Evidence

Verified on `master`.

### Gap 1 — `localize("")`

```
NamingConvention.capitalize("") -> ""
NamingConvention.camel("")      -> ""
NamingConvention.snake("")      -> ""
NamingConvention.kebab("")      -> ""
NamingConvention.pascal("")     -> ""
NamingConvention.localize("")   -> TypeError
```

Owner: `packages/utils/src/utils/NamingConvention.ts:139` — `str[0]!.toLowerCase() + str.slice(1)`. The `!` asserts a value that is not there.

**No in-repo caller can reach it, and that is worth stating plainly rather than inflating.** Every call site guards on `p.method.startsWith("create")`, and `TestAutomationTemplate.ts:44` builds methods as `` `create${capitalize(tpl.method)}` ``, so `localize` always receives a non-empty capitalized remainder (`"Is"`, `"Assert"`, …). Reaching `""` would require `tpl.method === ""`, which no template has.

So this is a **public-API robustness and consistency gap**, not a live in-repo defect. It is worth fixing because `@typia/utils` is published and a caller outside this repository has no such guard, and because PR #2114 rewrote this exact file and left it.

### Gap 2 — `getStructures()`'s selector

Owner: `tests/test-utils-automated/src/TestAutomation.ts:79-89`. It filters structures with `content.includes("never")`, `content.includes("toJSON")`, `content.includes('<"uint32">')`, and `content.includes("[key: ")` — matching **raw source text**, so a doc comment or an identifier containing any of those words changes which structures enter the validate/equality matrix, silently and with no signal.

Measured across all 148 structures, rather than assumed:

- **Zero comment-only matches today.** Both `never` hits are genuine types (`never: never;` and `[key: string]: never`).
- **Two predicates are dead.** `toJSON` and `JSONABLE` drop **zero** structures each — the `ToJson*` and `Comment*` filename filters already exclude every target they were written for.

So the honest framing is **a fragile selector carrying two dead predicates**, not "structures are being silently dropped". It is latent: the mechanism is real and the same hidden-coverage-loss family as #2035, which was just repaired in the sibling suite — a doc comment containing the word "never" would drop a structure tomorrow with nothing to notice.

## Consequence surface

| Gap | Owner | Live today? | Affected |
| --- | --- | --- | --- |
| `localize("")` | `@typia/utils` public API | No in-repo caller reaches it | external `@typia/utils` consumers |
| `getStructures()` selector | `tests/test-utils-automated` | No — zero comment-only matches now | the validate/equality matrix, if a fixture's prose ever changes |

Neither touches product behaviour for any current user. Gap 1 is a published-API contract inconsistency; gap 2 is test-infrastructure fragility with two pieces of provably dead configuration.

| Case | Expected |
| --- | --- |
| `NamingConvention.localize("")` | `""`, matching its five siblings |
| `NamingConvention.localize("Is")` | `"is"` — unchanged |
| a structure whose **type** genuinely uses `never` | selected — unchanged |
| a structure whose **doc comment** mentions "never" | selected on its type, not its prose |
| the two dead predicates | removed, or made to exclude something |

## Approach

**Gap 1:** make `localize` total over the empty string, as its five siblings already are. The `!` is the tell — it asserts a value the function never checks for.

**Gap 2:** select structures by what they *are*, not by scanning their source text for words. If that is disproportionate for a test harness, then at minimum delete the two dead predicates — dead configuration is worse than none, because the next reader believes it does something. Do not simply make the substrings more specific: that leaves the same trap with a smaller mouth.

- **Gap 1 positive:** `localize("")` returns `""`; every existing call site is unchanged.
- **Gap 1 boundary:** all six `NamingConvention` string helpers agree on the empty string.
- **Gap 2 positive:** the same 148 structures resolve to the same matrix membership as today — this is a refactor of *how* they are selected, not of *which*.
- **Gap 2 negative:** a structure whose doc comment mentions a filtered word is not dropped. Prove it by adding such a comment and observing the membership unchanged — if the selector is rewritten, this is the case that must fail against the old one.
- **Gap 2:** if the dead predicates are kept in any form, prove they exclude something. If they exclude nothing, delete them.
- **Commands:** `pnpm --filter ./tests/test-utils start`; `pnpm --filter ./tests/test-utils-automated start`; `pnpm --filter ./tests/test-typia-automated start`.

## Coordination

- Gap 1 is `@typia/utils` only; gap 2 is `tests/test-utils-automated` only. No product change and no package version bump for either.
- Gap 1 is the residue of #2111 (PR #2114), which rewrote `NamingConvention.ts` to make `variable` a real identifier predicate and did not touch `localize`.
- Gap 2 is the sibling of #2035 (PR #2121), which removed the same family of hidden coverage loss from `test-typia-automated`. That batch surfaced this one and correctly declined to sweep it in.
- Both were confirmed by an independent verification pass that **narrowed their originally reported premises**; the framing above is the narrowed one.

samchon:
## Intent

Close #2136. Two independent gaps, each kept at the size its evidence supports — neither is a live defect, and this body does not imply otherwise.

## Gap 1 — `NamingConvention.localize("")` threw

`str[0]!.toLowerCase()` asserted a character that is not there, while all five siblings (`capitalize`, `camel`, `snake`, `kebab`, `pascal`) already returned `""`. `localize` is now total in the same shape as `capitalize`, one line away in the same file.

**This is a public-API robustness and consistency gap, not a live in-repo defect.** No in-repo caller can reach it: `localize`'s only callers are the three in `write_common.ts`, each guarded by `p.method.startsWith("create")`, and `TestAutomationTemplate` builds methods as `` `create${capitalize(tpl.method)}` `` — so it always receives a non-empty capitalized remainder. It is worth fixing because `@typia/utils` is published and an external caller has no such guard, and because PR #2114 rewrote this exact file and left it.

Traced past the one witness rather than patching it: every other `[0]!` in `@typia/utils` is already length-guarded, so `localize` was the sole outlier and the class is closed, not the instance.

## Gap 2 — the automated selector matched raw source text

`getStructures` chose fixtures with `content.includes("never")`, `includes("toJSON")`, `includes('<"uint32">')`, and `includes("[key: ")` over each fixture's **raw file**, so prose decided coverage. It now reads each fixture's `JSONABLE` declaration — the same source `test-typia-automated` already reads — plus an explicit, per-fixture-reasoned held-out set. No scan for a word in a comment remains.

**Latent, not live.** Measured across all 148 structures: zero comment-only matches today, both `never` hits being genuine types. Nothing was being dropped; the mechanism was simply able to.

**This is a refactor of *how* structures are selected, not *which*.** Membership is byte-identical — `validate 83 = 83`, `validateEquals 80 = 80` — verified after every change.

### The dead predicates: three, not two

Measured marginally rather than assumed, and the issue's count moved in both directions. `toJSON` and `'<"uint64">'` excluded **zero** structures each and are removed; the annotated `JSONABLE: boolean = false` variant also excluded zero and is subsumed by reading the declaration. But `"JSONABLE = false"` excludes **41** structures and is very much alive — the issue's "JSONABLE drops zero" holds only for the annotated variant. It was kept. The per-predicate table is on the thread.

### What is deliberately not fixed here

`ADDABLE` still reads source text, and the comment at that line says why: four fixtures spell it `ADDABLE: boolean = false`, which the scan misses, so they sit in the equality matrix against their own declaration. Reading the declaration would drop them — and two assert **270 superfluous paths each**, so the "faithful" read silently deletes real coverage. That is a *which* decision, so it is reported on the thread, not smuggled in behind a refactor. Four held-out entries are likewise marked `passes today` rather than quietly admitted.

## Verification

Recorded on the thread: `test-utils`, `test-utils-automated`, `test-typia-automated`, `test-feature-identity`, and `packages/utils build` all pass. Both gaps have fail-then-pass evidence, and the gap-2 negative is proven to fail against the old selector rather than merely pass against the new one.

No product change beyond the `@typia/utils` fix, and no version bump: no Go source under `packages/typia/native` changed.

## Campaign CI

Campaign CI for each pushed SHA is cancelled at the exact-SHA gate; local verification is the gate.

Closes #2136

### DIFF
--- a/packages/utils/src/utils/NamingConvention.ts
+++ b/packages/utils/src/utils/NamingConvention.ts
@@ -136,7 +136,8 @@ export namespace NamingConvention {
    * @param str Input string
    * @returns Localized string
    */
-  export const localize = (str: string) => str[0]!.toLowerCase() + str.slice(1);
+  export const localize = (str: string): string =>
+    str.length !== 0 ? str[0]!.toLowerCase() + str.slice(1) : str;
 
   /**
    * Check if string is valid JavaScript variable name.
--- a/tests/template/src/structures/TypeTagRange.ts
+++ b/tests/template/src/structures/TypeTagRange.ts
@@ -61,6 +61,14 @@ export namespace TypeTagRange {
   export const MINIMUM = 3;
   export const MAXIMUM = 7;
 
+  /**
+   * Boundary spoilers, one step outside each tag's bound.
+   *
+   * Every spoiled value is the nearest integer a correct validator must never
+   * accept: `3` against `ExclusiveMinimum<3>`, `7` against `ExclusiveMaximum<7>`,
+   * and `9` against the `equal` pair. A validator that compares with the wrong
+   * strictness passes the ordinary generated value and fails only here.
+   */
   export const SPOILERS: Spoiler<TypeTagRange>[] = [
     (input) => {
       input.value[4]!.greater = 3;
--- a/tests/test-utils-automated/package.json
+++ b/tests/test-utils-automated/package.json
@@ -4,7 +4,7 @@
   "version": "13.1.1",
   "description": "Automated test features of openapi.",
   "scripts": {
-    "prestart": "ttsx src/regressions/test_process_fatal_events.ts",
+    "prestart": "ttsx src/regressions/test_process_fatal_events.ts && ttsx src/regressions/test_structure_selection_ignores_prose.ts",
     "start": "ttsx src/index.ts"
   },
   "repository": {
--- a/tests/test-utils-automated/src/TestAutomation.ts
+++ b/tests/test-utils-automated/src/TestAutomation.ts
@@ -1,3 +1,4 @@
+import * as template from "@typia/template";
 import { dedent } from "@typia/utils";
 import fs from "fs";
 
@@ -58,10 +59,50 @@ export namespace TestAutomation {
     );
   };
 
-  const getStructures = async (equals: boolean): Promise<string[]> => {
+  /**
+   * Structures the OpenAPI validate matrix cannot cover yet.
+   *
+   * Each entry names a `typia.json.schema` output that `OpenApiValidator` does
+   * not round-trip faithfully, so the fixture is skipped rather than asserted.
+   * The reason is recorded per fixture because it is a property of the
+   * validator's coverage, not of the fixture: nothing about `DynamicSimple`
+   * makes it unfit to be a fixture, only the validator's handling of the
+   * `additionalProperties` schema its type produces.
+   *
+   * These replace the source-text scans (`"never"`, `"[key: "`, `'<"uint32">'`)
+   * this selector used to run over each fixture's raw file, where a doc comment
+   * that merely mentioned one of those words silently changed matrix membership
+   * (#2136). Membership is byte-identical to those scans.
+   *
+   * Four entries are marked `passes today`: they were excluded when the matrix
+   * was introduced (#1736) and now validate cleanly against their own spoilers.
+   * They stay listed because admitting them changes *which* structures the
+   * matrix covers, which is a separate decision from *how* they are selected.
+   */
+  const UNVALIDATABLE: Record<string, string> = {
+    // `additionalProperties` from an index signature
+    DynamicArray: "index signature; passes today",
+    DynamicComposite: "index signature",
+    DynamicNever: "index signature over a `never` member",
+    DynamicSimple: "index signature; passes today",
+    DynamicTemplate: "index signature",
+    DynamicUndefined: "index signature",
+    DynamicUnion: "index signature",
+    ObjectDynamic: "index signature; passes today",
+    // a `never` member erased from the emitted schema
+    ObjectUndefined: "`never` member",
+    // integer format tags the validator does not enforce
+    ConstantAtomicTagged: "uint32 tag; passes today",
+    TemplateInterpolationTagged: "uint32 tag",
+    TypeTagType: "uint32 and uint64 tags",
+  };
+
+  export const getStructures = async (equals: boolean): Promise<string[]> => {
     const directory: string[] = await fs.promises.readdir(
       `${TestGlobal.ROOT}/../template/src/structures`,
     );
+    const declarations: Record<string, IStructureDeclaration> =
+      template as unknown as Record<string, IStructureDeclaration>;
     const result: string[] = [];
     for (const file of directory) {
       if (
@@ -72,23 +113,43 @@ export namespace TestAutomation {
         file.startsWith("ToJson")
       )
         continue;
-      const content: string = await fs.promises.readFile(
-        `${TestGlobal.ROOT}/../template/src/structures/${file}`,
-        "utf-8",
-      );
-      if (
-        content.includes("JSONABLE = false") ||
-        content.includes("JSONABLE: boolean = false") ||
-        content.includes("[key: ") ||
-        content.includes("never") ||
-        content.includes("toJSON") ||
-        content.includes('<"uint32">') ||
-        content.includes('<"uint64">') ||
-        (equals === true && content.includes("ADDABLE = false"))
-      )
-        continue;
-      result.push(file.substring(0, file.length - 3));
+      const name: string = file.substring(0, file.length - 3);
+      const structure: IStructureDeclaration | undefined = declarations[name];
+      if (structure === undefined)
+        throw new Error(`@typia/template does not export ${name}`);
+      // Read what the fixture declares about itself, so that prose cannot
+      // decide the matrix. `JSONABLE === false` marks a type whose value has no
+      // faithful JSON form.
+      if (structure.JSONABLE === false) continue;
+      else if (UNVALIDATABLE[name] !== undefined) continue;
+      // `ADDABLE` is still read from source text, and deliberately so: four
+      // fixtures spell it `ADDABLE: boolean = false`, which this scan does not
+      // match, so they sit in the equality matrix despite declaring otherwise.
+      // Reading the declaration instead would drop them — and two of the four
+      // (ArrayRepeatedUnion, ArrayRepeatedUnionWithTuple) assert 270 superfluous
+      // paths each, so the "faithful" read silently deletes real coverage. That
+      // is a decision about *which* structures the matrix covers, so it is left
+      // to #2136's follow-up rather than smuggled in behind a refactor.
+      else if (equals === true) {
+        const content: string = await fs.promises.readFile(
+          `${TestGlobal.ROOT}/../template/src/structures/${file}`,
+          "utf-8",
+        );
+        if (content.includes("ADDABLE = false")) continue;
+      }
+      result.push(name);
     }
     return result;
   };
+}
+
+/**
+ * The part of a `@typia/template` structure this selector reads.
+ *
+ * Mirrors `TestAutomationMetadata` in `test-typia-automated`, which narrows the
+ * same namespaces to the flags its own matrix consumes. Only declared members
+ * belong here: a flag this selector does not read would be dead configuration.
+ */
+interface IStructureDeclaration {
+  JSONABLE?: boolean;
 }
--- a/tests/test-utils-automated/src/TestAutomation.ts
+++ b/tests/test-utils-automated/src/TestAutomation.ts
@@ -116,14 +116,17 @@ export namespace TestAutomation {
       )
         continue;
       const name: string = file.substring(0, file.length - 3);
-      const structure: IStructureDeclaration | undefined = declarations[name];
-      if (structure === undefined)
+      // `Object.hasOwn`, not a truthiness test: a fixture named after an
+      // `Object.prototype` member would otherwise resolve against the prototype
+      // and be held out silently — the very failure this selector is shedding.
+      if (Object.hasOwn(declarations, name) === false)
         throw new Error(`@typia/template does not export ${name}`);
+      const structure: IStructureDeclaration = declarations[name]!;
       // Read what the fixture declares about itself, so that prose cannot
       // decide the matrix. `JSONABLE === false` marks a type whose value has no
       // faithful JSON form.
       if (structure.JSONABLE === false) continue;
-      else if (HELD_OUT[name] !== undefined) continue;
+      else if (Object.hasOwn(HELD_OUT, name)) continue;
       // `ADDABLE` is still read from source text, and deliberately so: four
       // fixtures spell it `ADDABLE: boolean = false`, which this scan does not
       // match, so they sit in the equality matrix despite declaring otherwise.
--- a/tests/test-utils-automated/src/TestAutomation.ts
+++ b/tests/test-utils-automated/src/TestAutomation.ts
@@ -60,26 +60,28 @@ export namespace TestAutomation {
   };
 
   /**
-   * Structures the OpenAPI validate matrix cannot cover yet.
+   * Structures held out of the OpenAPI validate matrix, and the type feature
+   * each was held out for.
    *
-   * Each entry names a `typia.json.schema` output that `OpenApiValidator` does
-   * not round-trip faithfully, so the fixture is skipped rather than asserted.
-   * The reason is recorded per fixture because it is a property of the
-   * validator's coverage, not of the fixture: nothing about `DynamicSimple`
-   * makes it unfit to be a fixture, only the validator's handling of the
-   * `additionalProperties` schema its type produces.
+   * This replaces three source-text scans (`"never"`, `"[key: "`, `'<"uint32">'`)
+   * that the selector ran over each fixture's raw file. Matching source text
+   * means prose decides coverage: a doc comment that merely mentioned one of
+   * those words dropped a structure from the matrix silently, with nothing to
+   * observe (#2136). The set below reproduces those three scans exactly — the
+   * same 148 structures resolve to the same 83 validate and 80 equality cases.
    *
-   * These replace the source-text scans (`"never"`, `"[key: "`, `'<"uint32">'`)
-   * this selector used to run over each fixture's raw file, where a doc comment
-   * that merely mentioned one of those words silently changed matrix membership
-   * (#2136). Membership is byte-identical to those scans.
+   * The reason belongs here rather than on the fixture because it describes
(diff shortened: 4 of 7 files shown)


samchon:
## Verification

All local, campaign CI cancelled at the exact-SHA gate.

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | Success |
| `pnpm --filter ./tests/test-typia-automated start` | Success |
| `pnpm --filter ./packages/utils build` | Success |

### Gap 1 — fail-then-pass

Both new cases fail against the unfixed `localize` with the exact reported defect, and pass with it:

```
### WITHOUT the fix
  - test_naming_convention_empty: TypeError
  - test_naming_convention_localize: TypeError
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
Failed
### WITH the fix
  - test_naming_convention_empty: 0 ms
  - test_naming_convention_localize: 0 ms
Success
```

Premise re-checked, not assumed: `localize`'s only in-repo callers are the three in `write_common.ts`, each guarded by `p.method.startsWith("create")`. No in-repo caller reaches `""`. Every other `[0]!` in `@typia/utils` is already length-guarded (`EndpointUtil`, `camel`, `pascal`, `capitalize`; the `split(";", 1)[0]!` sites always yield an element), so `localize` was the sole outlier and the class is sealed rather than the one witness patched.

### Gap 2 — the negative is non-vacuous

`TypeTagRange` now carries a doc comment on `SPOILERS` that legitimately reads "must never accept". The word is real prose, and `SPOILERS` is schema-invisible, so no emitted schema changes (`test-typia-automated` is green).

Against the **old** selector the case fails on the real assertion, not the guard:

```
Error: Bug on TypeTagRange stays in the validate matrix: found different values - []
{ "x": false, "y": true }
(exit=127)
```

Against the new selector it passes (exit=0). The guard asserting the prose still exists runs first, so this oracle cannot decay into one that proves nothing.

### Membership is unchanged for all 148 structures

```
validate        old=83 new=83 IDENTICAL
validateEquals  old=80 new=80 IDENTICAL
```

1. **`ADDABLE` cannot move to the declaration read.** Chased rather than assumed: switching it dropped 4 structures. Four fixtures spell `ADDABLE: boolean = false`, which the `"ADDABLE = false"` scan does not match, so they sit in the equality matrix against their own declaration. Two of them (`ArrayRepeatedUnion`, `ArrayRepeatedUnionWithTuple`) assert **270 superfluous paths each** — measured, not estimated — so the "faithful" read silently deletes real coverage. Kept the scan, documented it at the call site, and reported it below as a follow-up. Remediation: none in this PR; it is a *which* decision.
2. **Dead local declaration.** The narrowing interface still declared `ADDABLE` after the selector stopped reading it — the same dead-configuration smell this PR removes. Removed it; the interface now declares only `JSONABLE`, and its doc says why.
3. **Doc overclaimed.** The held-out set's opening asserted every entry fails to round-trip, which four entries contradict. Rewritten so the measured truth leads instead of trailing as a caveat. Renamed `UNVALIDATABLE` → `HELD_OUT`, since four entries are demonstrably validatable and the old name asserted otherwise.
4. Verified every documentation claim against source rather than trusting it: `DynamicNever` is literally `[key: string]: never`; all 13 `TypeTagRange` spoilers really are the nearest integer outside each bound.

samchon:
## Self-Review round 2 — findings and remediation

A fresh full round over the current diff found three, all applied.

1. **My own selector could silently drop a structure — the exact class this PR removes.** `HELD_OUT[name] !== undefined` resolves through `Object.prototype`, so a fixture named `constructor`, `toString`, or `valueOf` would hit the prototype chain and be held out with no signal. Reproduced rather than theorised:

   ```
   HELD_OUT["constructor"] !== undefined  ->  true
   Object.hasOwn(HELD_OUT, "constructor") ->  false
   ```

   Both lookups now use `Object.hasOwn`, which the repository already uses (`_isUniqueItems.ts`, `_notationAssign.ts`). No fixture is named this way today, so this is a latent hazard closed rather than a live bug — worth saying plainly.

2. **A dead clause in the very guard meant to prevent vacuity.** The guard read `source.includes("never") && WITNESS.includes("never") === false`; the second half is a constant over a hardcoded name and tested nothing. It also checked the wrong thing: the point is that the witness word is *prose*, not an identifier. Now `source.includes("must never")`, a sentence that exists only in the doc comment and that the removed `includes("never")` scan matched. Re-proved non-vacuous afterwards, since changing a guard can quietly disarm its own oracle.

3. Hand-wrapped the regression's signature to prettier's 80 columns to keep Post-Campaign Cleanup churn off this file (campaign rule: no `pnpm format` here).

### Re-verification after round 2

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | Success |
| `pnpm --filter ./tests/test-typia-automated start` | Success |

Membership re-confirmed after every change: `validate 83 = 83`, `validateEquals 80 = 80`, IDENTICAL.

Negative re-proved against the old selector with the new guard — it fails on the real assertion, not the guard:

```
### NEW selector -> exit=0
### OLD selector -> Error: Bug on TypeTagRange stays in the validate matrix
```

samchon:

A complete fresh round over the final state produced nothing further to fix. Checks that found nothing: no stale reference to any removed predicate survives anywhere in `tests/` or `packages/`; the only importers of `TestAutomation` are the suite entrypoint and the new regression; every documentation claim re-checked against source (`DynamicNever` is literally `[key: string]: never`; all 13 `TypeTagRange` spoilers really are the nearest integer outside their bound); Windows and POSIX paths unchanged; no concurrency or generated-artifact surface touched.

One property was verified rather than assumed: a failing regression must actually stop the suite. It does — the `prestart` chain propagates, and `pnpm --filter ./tests/test-utils-automated start` exits nonzero when the old selector is restored, so this case cannot rot into a decorative file.

### Final verification, on the pushed SHA

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | Success |
| `pnpm --filter ./tests/test-typia-automated start` | Success |
| `pnpm --filter ./packages/utils build` | Success |

`pnpm format` deliberately not run (campaign rule); no version bump, since no Go source under `packages/typia/native` changed and the only product edit is a `@typia/utils` bug fix.

## Two findings this batch deliberately did not act on

Both change *which* structures the matrix covers. The issue scoped this PR to *how* they are selected, so both are reported rather than smuggled in.

**1. The equality matrix admits four structures against their own declaration.** The `"ADDABLE = false"` scan does not match the annotated `ADDABLE: boolean = false` spelling, so `ArrayRepeatedNullable`, `ArrayRepeatedRequired`, `ArrayRepeatedUnion`, and `ArrayRepeatedUnionWithTuple` sit in the equality matrix while declaring the opposite. This is the exact mirror of the dead `"JSONABLE: boolean = false"` predicate: someone hit the annotated spelling for `JSONABLE` and added a second scan for it, and never did the same for `ADDABLE`.

Reading the declaration is *not* obviously the fix, which is why it is escalated rather than applied:

| Fixture | superfluous paths asserted | Effect of the "faithful" read |
| --- | --- | --- |
| `ArrayRepeatedNullable` | 0 | drops a vacuous case |
| `ArrayRepeatedRequired` | 0 | drops a vacuous case |
| `ArrayRepeatedUnion` | **270** | **deletes real coverage** |
| `ArrayRepeatedUnionWithTuple` | **270** | **deletes real coverage** |

Measured, not estimated. Two are vacuous because the type is `number | string | T[]` with no object for the oracle to add a key to; two contain `IBox3D` objects and assert 270 paths each. So either `ADDABLE = false` is over-declared on the two union fixtures, or the equality matrix has been silently over-covering — worth deciding deliberately. Note `test-typia-automated` reads the declaration, so it already skips all four for its strict and prune templates.

**2. Three predicates were dead, not two.** The issue named `toJSON` and `JSONABLE`. Measured marginally across all 148 structures — removing each predicate alone and counting what re-enters:

| Predicate | Structures it alone excludes | Status |
| --- | --- | --- |
| `"JSONABLE = false"` | 41 validate / 23 equality | **alive** — the issue's "JSONABLE drops zero" holds only for the annotated variant |
| `"JSONABLE: boolean = false"` | 0 | dead |
| `"toJSON"` | 0 | dead |
| `'<"uint64">'` | 0 | **dead — not named in the issue** |
| `"[key: "` | 7 / 3 | alive |
| `"never"` | 1 | alive |
| `'<"uint32">'` | 2 | alive |

All three dead predicates are gone: `toJSON` and `<"uint64">` removed outright, and the annotated `JSONABLE` variant subsumed by reading the declaration. `"JSONABLE = false"` is emphatically *not* dead and was kept.

**3. The held-out set is over-broad by four.** `ConstantAtomicTagged`, `DynamicArray`, `DynamicSimple`, and `ObjectDynamic` validate cleanly against their own real spoilers when admitted — verified they have spoilers, so they pass non-vacuously. The other eight fail for genuine `OpenApiValidator` gaps (index signatures, a `never` member, integer format tags). They are held out here to keep membership identical, and each is marked `passes today` in the table so the staleness is visible instead of buried in a substring.

samchon:
## A false green in my own verification, caught and corrected

Worth recording, because it is the same failure this batch exists to end — an oracle that reports success without observing anything.

While confirming the final SHA I matched `^Success$` in the `test-typia-automated` log. That pattern matched the **prestart** step (`ttsx src/regressions/index.ts`, which prints its own `Success`) while the actual suite was still compiling. The check would have reported a pass over an empty run.

Re-checked with a condition that cannot pass vacuously — the verdict must be the final line *and* the log must contain executed cases:

```
verdict: Success  | cases: 3947
```

Audited the two earlier `test-typia-automated` runs for the same flaw rather than assuming they were sound. Both are genuine: 3,957 log lines, 3,947 executed cases, final line `Success`, with two `Success` tokens explained as prestart + suite. The second of them ran on code identical to the final commit, so the result stands on three independent runs.

Final state of `3108a34` — every command a real, non-vacuous pass:

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | Success |
| `pnpm --filter ./tests/test-typia-automated start` | Success — 3,947 cases |
| `pnpm --filter ./packages/utils build` | Success |

Membership: `validate 83 = 83`, `validateEquals 80 = 80`, IDENTICAL. Self-Review is three rounds, the third clean. Ready for lead review; I am not merging.

samchon:

## Version: untouched, deliberately

The only `package.json` in this diff is `tests/test-utils-automated/package.json`, adding the new prestart regression. `typia`'s version is identical on the branch and `master`. Releases and version numbers are the maintainer's; the campaign does not touch them.

## The issue's premise was wrong, and following it literally would have been harmful

I wrote that `toJSON` and `JSONABLE` each drop zero structures. **That is only true of the annotated `JSONABLE` variant.** The plain `"JSONABLE = false"` scan is **alive and excludes 41 structures**. Deleting the one my issue named would have silently admitted 41 structures into the matrix.

The batch measured instead of trusting me: **three** dead predicates, not two — `toJSON`, `<"uint64">` (which my issue never mentioned), and the annotated `JSONABLE`. All three removed; the live one kept. This is the fourth time this session an agent has disproved a premise I published, and the first where following my text would have caused the damage rather than merely missed it.

## Lead verification on `e0fd7359e5`

| Gate | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | exit 0 — prestart ran both regressions, then the suite itself ran `ttsx src/index.ts` across 172 lines of executed cases |

**A false green nearly caught me here, and it is worth recording.** My first attempt used `pnpm --filter test-utils-automated start` — the name form. It matched **nothing** and returned **exit 0**: this workspace's directory is `test-utils-automated` but its package name is `@typia/test-openapiautomated`. Had I trusted that exit code I would have certified this batch on zero executed tests. The repo's own `test:packages` uses the path form, so CI is unaffected — but `--filter <name>` silently does nothing here, and I am carrying that to the next discovery round as a candidate.

That is the same class as this batch's own catch: it found a false green in **its own** verification, where `^Success$` matched the prestart step rather than the suite, and re-checked with a condition requiring executed cases.

## Gap 2's negative is proven non-vacuous

`TypeTagRange` carries a doc comment that legitimately reads "must never accept". Against the **old** selector the new test fails on its real assertion — `Bug on TypeTagRange stays in the validate matrix`, nonzero exit — and passes against the new one. The batch also verified the failure actually blocks the suite rather than printing into a passing run, and that the anti-vacuity guard runs first.

**Membership unchanged: `validate 83 = 83`, `validateEquals 80 = 80`.** This was a refactor of *how* structures are selected, not of *which*, and it stayed one.

## Two decisions correctly reported rather than taken

1. **The equality matrix admits four structures against their own declaration.** The `"ADDABLE = false"` scan misses `ADDABLE: boolean = false`. But reading the declaration is *not* safely correct: two of the four assert **270 superfluous paths each**, so the faithful read would delete real coverage. That is a *which* decision, and the batch reported it instead of making it. `ADDABLE` still scans text, documented at the line.
2. **The held-out set is over-broad by four.** `ConstantAtomicTagged`, `DynamicArray`, `DynamicSimple`, and `ObjectDynamic` pass non-vacuously when admitted; the other eight fail for genuine `OpenApiValidator` gaps. Held out to preserve membership, each marked `passes today` so the staleness is visible rather than silent.

**Those eight `OpenApiValidator` gaps are a candidate I am carrying forward** — a structure that cannot pass the validator is either a validator defect or a fixture that never should have been in the set, and "held out" answers neither.

## Self-Review round 2 found two defects in its own fix

The new selector could silently drop a fixture named `constructor` through `Object.prototype` — fixed with `Object.hasOwn` — and the vacuity guard itself carried a constant-true clause. A guard against vacuity that is itself vacuous is the tightest version of this campaign's recurring failure, and it was caught before merge.

`e0fd7359e5` settled after five polls, all six runs `cancelled`, no race.

PR #2140, opened 2026-07-17
feat(native): adopt the ttsc driver reference-graph SDK in the transform envelope

### LINKED ISSUE #2107: native transform: adopt the ttsc driver SDK envelope encoder to carry the host reference graph

samchon:
## Scope

Adopts the ttsc host-owned reference-graph SDK (shipped in ttsc 0.19.2 by
samchon/ttsc#718) into typia's native transform envelope, so a typia project's
`ttsc-typia transform` output carries the plugin-agnostic `graph` baseline
alongside typia's precise `dependencies` (from #2100). The `@ttsc/unplugin`
consumer registers `reach(edges, F) ∪ globals ∪ configs ∪ dependencies[F]`, so
graph is added, not a replacement.

Closes #2107.

### Changes

1. **Native envelope graph emission** — `runTransformProject` now computes
   `driver.NewTransformGraph(prog, cwd)` (before the transform loop, matching
   ttsc's `api-transform` host, so the graph describes the original resolved
   references, not mutated ASTs) and stamps it into the envelope's optional
   `graph` field. All envelope sections key through `driver.TransformOutputKey`
   so `graph` / `typescript` / `dependencies` join by key. typia's precise
   `dependencies` emission is untouched.
2. **go.mod floor** — `packages/typia/native/go.mod` bumps
   `github.com/samchon/ttsc/packages/ttsc` `v0.0.0 -> v0.19.2` (documentary
   floor; the module is workspace-provided by go.work `use`, so resolution is
   unaffected).
3. **peerDependencies floor** — see the decision note below.
4. **Catalog + lockfile** — `pnpm-workspace.yaml` catalog anchor
   `ttsc ^0.18.4 -> ^0.19.2` (moves `ttsc`/`@ttsc/unplugin`/`@ttsc/wasm`/
   `@ttsc/playground` in lockstep) with `pnpm-lock.yaml` regenerated. Required:
   CI resolves `node_modules/ttsc` for `pnpm test:go:native`, and the graph SDK
   symbols only exist at >= 0.19.2. This is a dependency-range change, not a
   package `version` field change.
5. **Docs** — `website/src/content/docs/setup/index.mdx` states the ttsc
   >= 0.19.2 requirement.

### peerDependencies floor decision (owner-chosen)

typia's plugin is built from bundled Go source against the USER's installed ttsc,
and `driver.NewTransformGraph` exists only in ttsc >= 0.19.2. There is no build
tag to gate on, so referencing the SDK is a hard minimum-ttsc floor raise. The
owner chose to declare it via `peerDependencies` (npm signal), not a preflight
guard:

```jsonc
"peerDependencies": {
  "ttsc": ">=0.19.2",
  "@ttsc/unplugin": ">=0.19.2"
},
"peerDependenciesMeta": {
  "ttsc": { "optional": true },
  "@ttsc/unplugin": { "optional": true }
}
```

`optional: true` on every peer is load-bearing: an ABSENT optional peer triggers
no install/warning/error, so type-only consumers (who install neither ttsc nor
`@ttsc/unplugin`) stay completely clean; a peer PRESENT below the floor gets the
advisory version-mismatch warning. Honest tradeoff: this is a warning, not a hard
failure — a forced install on old ttsc still hits the Go compile error. Both
install topologies are covered (CLI users install `ttsc`; bundler users install
`@ttsc/unplugin`, which carries ttsc transitively).

### Deferred

Narrowing below the host baseline (declaring typia's per-file lists complete) is
NOT in scope — it waits for the completeness contract on the ttsc side (per #2107).

### Verification

Local verification (commands + results) and the empirical peer-warning check
(clean install with no ttsc; warning with old ttsc) are recorded as a thread
comment. CI runs normally on this PR.

### DIFF
--- a/experiments/esm/package.json
+++ b/experiments/esm/package.json
@@ -18,7 +18,7 @@
     "@typia/utils": "file:../tarballs/utils.tgz",
     "@typia/vercel": "file:../tarballs/vercel.tgz",
     "ai": "^6.0.116",
-    "ttsc": "^0.17.3",
+    "ttsc": "^0.19.2",
     "typescript": "^7.0.2",
     "typia": "file:../tarballs/typia.tgz"
   }
--- a/experiments/mcp/package.json
+++ b/experiments/mcp/package.json
@@ -14,7 +14,7 @@
     "@typia/interface": "file:../tarballs/interface.tgz",
     "@typia/mcp": "file:../tarballs/mcp.tgz",
     "@typia/utils": "file:../tarballs/utils.tgz",
-    "ttsc": "^0.17.3",
+    "ttsc": "^0.19.2",
     "typescript": "7.0.1-rc",
     "typia": "file:../tarballs/typia.tgz"
   }
--- a/experiments/smoke/package.json
+++ b/experiments/smoke/package.json
@@ -13,6 +13,6 @@
   },
   "devDependencies": {
     "typescript": "7.0.1-rc",
-    "ttsc": "^0.17.3"
+    "ttsc": "^0.19.2"
   }
 }
--- a/packages/typia/native/cmd/ttsc-typia/project_reference_graph_envelope_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_reference_graph_envelope_transform_test.go
@@ -1,0 +1,190 @@
+package main
+
+import (
+  "encoding/json"
+  "os"
+  "path/filepath"
+  "testing"
+)
+
+// TestProjectReferenceGraphEnvelopeTransform verifies the project transform
+// envelope carries the ttsc driver SDK's host-owned reference `graph` section
+// alongside typia's precise `dependencies`.
+//
+// `graph` is the plugin-agnostic sound invalidation baseline: the ttsc consumer
+// registers reach(edges, F) ∪ globals ∪ configs ∪ dependencies[F], so a
+// consulted file the precise `dependencies` collection might not attribute is
+// still guarded by the host baseline (samchon/typia#2107). This pins that
+// typia's hand-rolled envelope now routes through driver.NewTransformGraph:
+// direct reference edges (type-only included), ambient global-scope files, and
+// the tsconfig `extends` chain, all keyed by the same driver.TransformOutputKey
+// convention as the `typescript`/`dependencies` sections so a consumer joins
+// them by key.
+//
+//  1. Build a project where `a.ts` imports `Bee` from `b.ts` (and calls
+//     `typia.validate<Bee>()`), `b.ts` imports type-only `Cee` from `c.ts`, an
+//     ambient `globals.d.ts` contributes a global augmentation, and
+//     `tsconfig.json` extends `tsconfig.base.json`.
+//  2. Run project transform mode and decode the JSON envelope's `graph`.
+//  3. Assert the `a.ts -> b.ts` and the type-only `b.ts -> c.ts` edges, the
+//     `globals.d.ts` global entry, and the `tsconfig.json` + `tsconfig.base.json`
+//     configs chain, all under keys that join with the `typescript` section.
+//  4. Assert `dependencies` is still emitted next to `graph`, and that a
+//     non-global source is not misreported as a global.
+func TestProjectReferenceGraphEnvelopeTransform(t *testing.T) {
+  project := projectReferenceGraphEnvelopeProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--output", "ts",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("project transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  var envelope struct {
+    TypeScript map[string]string `json:"typescript"`
+    Graph      *struct {
+      Edges   map[string][]string `json:"edges"`
+      Globals []string            `json:"globals"`
+      Configs []string            `json:"configs"`
+    } `json:"graph"`
+    Dependencies map[string][]string `json:"dependencies"`
+  }
+  if err := json.Unmarshal([]byte(out), &envelope); err != nil {
+    t.Fatalf("decode envelope: %v\n%s", err, out)
+  }
+  if envelope.Graph == nil {
+    t.Fatalf("envelope must carry a graph section:\n%s", out)
+  }
+  if _, ok := envelope.TypeScript["src/a.ts"]; !ok {
+    t.Fatalf("typescript map must contain src/a.ts (graph keys join with it):\n%s", out)
+  }
+
+  contains := func(values []string, want string) bool {
+    for _, value := range values {
+      if value == want {
+        return true
+      }
+    }
+    return false
+  }
+
+  // Direct resolved edge from the transformed file to its import.
+  if !contains(envelope.Graph.Edges["src/a.ts"], "src/b.ts") {
+    t.Fatalf("graph edges of src/a.ts must contain src/b.ts: %v", envelope.Graph.Edges["src/a.ts"])
+  }
+  // Type-only edges must be included: b.ts imports Cee with `import type`.
+  if !contains(envelope.Graph.Edges["src/b.ts"], "src/c.ts") {
+    t.Fatalf("graph edges of src/b.ts must contain the type-only edge src/c.ts: %v", envelope.Graph.Edges["src/b.ts"])
+  }
+  // Ambient global-scope file appears in globals.
+  if !contains(envelope.Graph.Globals, "src/globals.d.ts") {
+    t.Fatalf("graph globals must contain src/globals.d.ts: %v", envelope.Graph.Globals)
+  }
+  // A plain module source is not a global-scope contributor.
+  if contains(envelope.Graph.Globals, "src/c.ts") {
+    t.Fatalf("graph globals must not contain the module source src/c.ts: %v", envelope.Graph.Globals)
+  }
+  // tsconfig extends chain: project config followed by its base.
+  if !contains(envelope.Graph.Configs, "tsconfig.json") {
+    t.Fatalf("graph configs must contain tsconfig.json: %v", envelope.Graph.Configs)
+  }
+  if !contains(envelope.Graph.Configs, "tsconfig.base.json") {
+    t.Fatalf("graph configs must contain the extended tsconfig.base.json: %v", envelope.Graph.Configs)
+  }
+
+  // The precise dependencies channel remains emitted alongside graph.
+  if !contains(envelope.Dependencies["src/a.ts"], "src/b.ts") {
+    t.Fatalf("dependencies of src/a.ts must still be emitted alongside graph: %v", envelope.Dependencies["src/a.ts"])
+  }
+}
+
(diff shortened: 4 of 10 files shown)


samchon:
## Local verification (all green)

| Check | Command | Result |
| --- | --- | --- |
| New graph test (native go.work) | `go -C packages/typia/native test ./cmd/ttsc-typia/ -run TestProjectReferenceGraphEnvelopeTransform` | PASS |
| New graph test (CI go.work → node_modules/ttsc@0.19.2) | `go -C packages/typia/test test ../native/cmd/ttsc-typia/ -run 'TestProjectReferenceGraphEnvelopeTransform\|TestProjectDependenciesEnvelopeTransform'` | PASS |
| Full native cmd suite | `go -C packages/typia/native test ./cmd/...` | `ok … 27.0s` |
| Native, CI path | `pnpm run test:go:native` | exit 0 (all `ok`) |
| Public contract, CI path | `pnpm run test:go:public` | exit 0 (all `ok`) |
| #2109 bundler-cache e2e (graph consumer) | `pnpm --filter ./tests/test-typia-bundler-cache start` | `Success` — plugin built from source against ttsc 0.19.2, all 5 cache-invalidation stages pass unchanged |
| `go vet` (changed pkg) | `go -C packages/typia/native vet ./cmd/ttsc-typia/` | exit 0 |

## peerDependencies floor — empirical install matrix (npm)

Packed the built typia and installed it four ways (peer block shipped verbatim; workspace deps repointed only so npm can build the tree):

| Scenario | peer state | Result |
| --- | --- | --- |
| Type-only consumer | neither `ttsc` nor `@ttsc/unplugin` installed | **clean** — `added 61 packages`, no peer warning |
| CLI user, old ttsc | `ttsc@0.19.1` pinned exact | **floor signal** — `peerOptional ttsc@">=0.19.2" from typia` |
| CLI user, at floor | `ttsc@0.19.2` pinned exact | clean, no conflict |
| Bundler user, old | `@ttsc/unplugin@0.19.1` pinned exact | **floor signal** — `peerOptional ttsc@">=0.19.2"` (transitive) |

Confirms the contract: an absent optional peer produces no install/warning/error (type-only consumers untouched); a present peer below the floor surfaces the mismatch. When the old peer is pinned so it cannot be upgraded, npm reports it as ERESOLVE naming `>=0.19.2`; a caret-ranged old peer auto-upgrades to 0.19.2 and resolves cleanly. Honest tradeoff (owner-accepted): a forced install still hits the Go compile error; no build-time preflight guard was added.

## go.mod floor — deviation from the issue text (verified)

The issue asked to bump `github.com/samchon/ttsc/packages/ttsc` `v0.0.0 → v0.19.2`, assuming the go.work `use` directive makes resolution version-agnostic. That is empirically false — with `v0.19.2` the build fails:

```
github.com/samchon/ttsc/packages/ttsc@v0.19.2: reading …/go.mod at revision
packages/ttsc/v0.19.2: unknown revision packages/ttsc/v0.19.2
```

Resolution works through the go.work `replace … v0.0.0 => <ttscRoot>`, and ttsc's `buildSourcePlugin.ts` **hardcodes** that `v0.0.0` replace in the go.work it generates on every user's machine (typia's own go.mod may not carry a ttsc replace — buildSourcePlugin throws if it does). A `v0.19.2` require would break every user's plugin build and the "build modes unaffected" acceptance. So the require stays `v0.0.0` with a documentary comment; the enforceable floor lives in `peerDependencies` + the setup doc. This preserves the issue's intent (documented + signaled floor) without breaking resolution.

CI runs normally on this PR (no runs cancelled).

samchon:
## Self-Review — Round 1

Surface reviewed (full base->head diff + working tree): `transform.go`, the new graph test, `go.mod`, `packages/typia/package.json`, `pnpm-workspace.yaml`, `pnpm-lock.yaml`, `website/.../setup/index.mdx`, plus the consequence surface — every ttsc consumer in the repo (workspace catalog members, `experiments/*` smoke fixtures, `test-typia-cli` cache sharing), CI workflows, and cross-platform key behavior.

### Findings

1. **[fixed] Frozen-lockfile mismatch.** The first `pnpm install` ran before the peer block was added, so `pnpm-lock.yaml` did not record typia's peers and CI's frozen install rejected the PR. Fix: regenerated the lockfile with the peers present; local `pnpm install --frozen-lockfile` now reports "Already up to date". (commit `5d8a2ee`)

2. **[fixed] `experiments/{smoke,esm,mcp}` fixtures pinned `ttsc@^0.17.3`.** These install the typia tarball via npm; the new `peerOptional ttsc >=0.19.2` makes npm ERESOLVE against the old pin (CI `experiments` lane failed at 3m30s). This is the intended floor signal firing on a real-consumer simulation, so the fix is to move the fixtures to the floor, not `--force`. Fix: bumped all three to `ttsc ^0.19.2`; ttsc 0.19.2 declares no peers, so no cascade. (commit `ad610d2`)

3. **[no change] Key unification.** Replacing `sourceFileKey` with `driver.TransformOutputKey` is behavior-preserving for in-root files (both yield `ToSlash(rel)`); outside-root files now key as absolute (matching the documented "absolute otherwise" contract and the graph section), and the `typescript` loop still skips them via `filepath.IsAbs`. Dependencies/self-drop stay consistent because `current` and `value` use the same function+cwd. Verified: #2100 dependencies test unchanged and passing.

4. **[no change] Graph placement.** `driver.NewTransformGraph` is computed before the transform loop, matching ttsc's `api-transform` host, so the graph describes the original resolved references. typia does not emit `volatile` (its transform has no non-file inputs) — correct.

5. **[no change] go.mod floor.** Require stays `v0.0.0` (documented) because ttsc's `buildSourcePlugin` hardcodes the `v0.0.0` replace; a real version breaks every user's plugin build. Empirically verified both directions.

No code defect found in this round; the two fixes are consequence-surface follow-ups. Proceeding to a fresh Round 2 after CI reflects the fixes.

samchon:
### CI follow-up: scoped the ttsc bump (website playground consequence)

The initial full catalog-anchor bump moved **@ttsc/wasm** to 0.19.2 as well. `website.yml` clones the ttsc sibling at `v${installed @ttsc/wasm version}` to build the in-browser playground's Go wasm host, so that pulled ttsc **v0.19.2**, whose reworked `host.Plugin` contract (`Run(*host.PluginInvocation)` + invocation-owned streams) the website playground compiler does not yet implement:

```
cmd/playground/main.go:23: typiaPlugin does not implement host.Plugin (wrong type for method Run)
  have Run(string, []string) int
  want Run(*host.PluginInvocation) int
```

That host migration is unrelated to this PR's reference-graph SDK. Rather than fold an out-of-scope wasm-host migration into #2107, I **scoped the bump**: `ttsc` and `@ttsc/unplugin` → `^0.19.2` (what the native graph SDK and the bundler consumer require); `@ttsc/wasm` and `@ttsc/playground` stay `^0.18.4`, so the website playground keeps building against ttsc's prior host API exactly as on `master`. The playground compiler links typia's `transform`/`adapter`/`core/context` (unchanged here) — not `cmd/ttsc-typia` — so it is unaffected by the graph emission. Peer graph stays consistent (`@ttsc/unplugin@0.19.2` peers `ttsc@0.19.2`; `@ttsc/playground@0.18.4` peers `@ttsc/wasm@^0.18.4`; `@ttsc/wasm` has no ttsc peer). Frozen-lockfile install verified. (commit `0e33e32`)

The `experiments` lane (fixture floor bump) is green; re-watching the full lane set now.

samchon:
## Self-Review — Round 2 (fresh full pass)

Re-read the entire base->head diff and consequence surface from scratch.

- **Graph determinism.** `driver.NewTransformGraph` sorts globals and per-file edge targets; `configs` is the ordered extends chain; Go's JSON encoder sorts map keys. Output is deterministic and project-relative.
- **JSON shape.** `Graph *driver.TransformGraph` with `graph,omitempty`; non-nil here (program is loaded). A single-file project yields `edges:{}` — harmless.
- **Windows/POSIX keys.** Keys come from `filepath.Rel` + `ToSlash`, the same construction the passing #2100 test uses on Ubuntu and I ran on Windows; the new graph test passed on both go.work paths and on Ubuntu CI.
- **No `volatile`.** typia's transform has no non-file inputs, so it correctly omits the optional `volatile` list.

No new code finding.

## Self-Review — Round 3 (fresh full pass)

Repeated the whole inspection after the Round 1 and CI-surfaced consequence fixes.

- **Every ttsc consumer accounted for.** `ttsc`/`@ttsc/unplugin` catalog entries are `^0.19.2`; `@ttsc/wasm`/`@ttsc/playground` deliberately stay `^0.18.4` (website-playground host, unrelated to the graph SDK — see the scoped-bump comment); the three `experiments/*` npm fixtures pin `^0.19.2`; `test-typia-cli` shares the workspace cache via `TTSC_CACHE_DIR` with no ttsc pin. No consumer sits below the floor, and no consumer is forced onto ttsc's reworked wasm host API by this PR.
- **Consequences chased to closure.** Frozen-lockfile mismatch, `experiments` peer-floor ERESOLVE, and the `website` playground-compiler break were each traced to the real cause and fixed at the source (not `--force`); CI now green across `test`, `nestia`, `build`, `website`, `experiments`, `spell-check`, and Socket.
- **Version fields untouched.** No package `version` field changed; only dependency ranges, the peer contract, native source, docs, a test, and fixture pins.
- **Docs.** Setup callout uses standard GFM `[!IMPORTANT]`, rendered fine by the (now green) website build.

A complete round produced no further sound improvement. Self-Review concludes; proceeding to squash-merge on green.

PR #2144, opened 2026-07-17
fix(native): assert the catalog scratchpad by purpose, not by its current contents

### LINKED ISSUE #2141: TestProjectCatalogTransformCoverage/debug asserts that the scratchpad holds broken code, and fails since #2133
## Problem

`TestProjectCatalogTransformCoverage/debug` fails on `master`. It asserts that transforming `tests/debug` reports diagnostics with exit code 3 — which was true only because the workspace happened to hold abandoned code that could not compile. PR #2133 (#2132) repaired that sample so the documented repro runner actually runs, and the assertion broke.

The test encoded an accident as a contract. `tests/debug` is a scratchpad whose contents are arbitrary by design; asserting it always produces diagnostics is asserting that the scratchpad always holds broken code.

This is a regression introduced by merging #2133, and it is stated here as such.

## Evidence

On pristine `origin/master` `1d0fab689a`, with no other diff applied:

```
go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia \
  -run TestProjectCatalogTransformCoverage
```

```
--- FAIL: TestProjectCatalogTransformCoverage (9.93s)
    --- FAIL: TestProjectCatalogTransformCoverage/debug (0.09s)
        project_catalog_transform_coverage_test.go:66: debug transform should report diagnostics with code 3, got 0
FAIL
```

The assertion, at `project_catalog_transform_coverage_test.go`:

```go
for _, project := range []string{"debug", "test-error"} {
    // ...
    if code != 3 {
        t.Fatalf("%s transform should report diagnostics with code 3, got %d", project, code)
    }
    if !strings.Contains(out, `"diagnostics"`) || !strings.Contains(out, `"typescript"`) {
        t.Fatalf("%s output should include diagnostics and TypeScript map:\n%s", project, out)
    }
}
```

`debug` and `test-error` are grouped as "projects that report diagnostics". For `test-error` that is its stated purpose — it compiles deliberately invalid call sites. For `debug` it was incidental.

**Why nothing caught it before the merge.** The test only runs under `-tags typia_native_internal`, and that tag appears in **no CI workflow and no root script**. `pnpm test` does not run it, so #2133's full-suite verification could not have seen it. It runs only when someone passes the tag by hand.

## Consequence surface

- **Owner:** `packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go`.
- **Affected:** anyone running the native suite with `-tags typia_native_internal`, which several contributors and agents do routinely.
- **Not affected:** `pnpm test`, every CI workflow, and every product surface. Nothing a user installs changes.
- **The `test-error` half of the assertion remains correct** and must stay — that workspace exists to hold rejected call sites.

| Case | Expected |
| --- | --- |
| `tests/debug` transformed while holding valid code | exit 0 — the runner works, which is #2132's contract |
| `tests/debug` transformed while holding a deliberately broken sample | exit 3 with diagnostics — still true, but not something to assert |
| `tests/test-error` transformed | exit 3 with diagnostics — unchanged, this is its purpose |
| the catalog's other projects | unchanged |

## Approach

The invariant: **the catalog asserts what a project's purpose guarantees, not what its current contents happen to be.**

`test-error` guarantees diagnostics. `tests/debug` guarantees nothing about its contents — that is what a scratchpad is. Either drop `debug` from the diagnostics group, or give it an assertion that holds regardless of what the sample currently contains. Do not restore the broken sample to satisfy the test; #2132 fixed a real defect and this test is what should move.

While here, consider whether a tagged suite that no workflow and no script ever runs can be trusted at all. This one rotted for exactly that reason, and `tests/debug` itself rotted for the same one — nothing failed when it stopped working. Whether that gap is deliberate is a maintainer decision, but it is the reason this issue exists.

- **Positive:** `go -C packages/typia/test test -tags typia_native_internal ../native/cmd/ttsc-typia` passes on current `master`.
- **Negative:** `test-error`'s half of the assertion still fails if that workspace ever stops reporting diagnostics — do not weaken it while fixing its neighbour.
- **Boundary:** the assertion must hold whether `tests/debug` currently holds valid or invalid code, since both are legitimate states for a scratchpad.
- **Regression:** the untagged suite is unaffected — `pnpm test` was already green and must stay so.
- **Commands:** `go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia`; `go -C packages/typia/test test -p 2 ../native/...`; `pnpm test`.

## Coordination

- Tests only. No product change and no package version bump.
- Direct fallout of PR #2133 (#2132). #2133 is correct and must not be reverted.
- Found by the #2134 batch, which reproduced it on pristine `master` with its own diff removed and correctly declined to sweep it in.

samchon:
## Intent

Fix `TestProjectCatalogTransformCoverage/debug`, which fails on `master` since #2133 (#2132) repaired the `tests/debug` sample so the documented repro runner actually runs.

Closes #2141.

## Scope

Tests only. No product source, no package version bump. #2133 is correct and is not reverted.

The invariant being restored: **the catalog asserts what a project's purpose guarantees, not what its current contents happen to be.**

- `tests/test-error` guarantees diagnostics — it exists to hold rejected call sites. Its half of the assertion is unchanged.
- `tests/debug` is a scratchpad and guarantees nothing about its contents. Both a valid and a deliberately broken sample are legitimate states, so its assertion must hold in both.

## Verification

Recorded in the thread. Campaign CI is deliberately suspended; every run for each pushed SHA is cancelled at the exact-SHA gate and local verification stands in for it.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
@@ -1,4 +1,4 @@
-//go:build typia_native_internal
+﻿//go:build typia_native_internal
 // +build typia_native_internal
 
 package main
@@ -16,10 +16,19 @@ import (
 // Running them through the same project transform mode gives Go coverage for
 // those native branches without duplicating that TypeScript corpus here.
 //
+// Each project is held to what its purpose guarantees, never to what its sources
+// happen to contain today. The feature projects exist to compile accepted typia
+// call sites, so they must transform without a diagnostic. `tests/test-error`
+// exists to hold call sites typia rejects, so it must report them. `tests/debug`
+// is a scratchpad whose contents are arbitrary by design (samchon/typia#2141): a
+// valid sample and a deliberately broken one are equally legitimate states, so
+// only the envelope contract itself is asserted for it. Asserting that the
+// scratchpad reports diagnostics is asserting that it always holds broken code.
+//
 // 1. Resolve existing TypeScript catalog projects from the repository root.
 // 2. Run project transform mode so every source file contributes call sites.
 // 3. Require JSON project output to include a TypeScript map.
-// 4. Fail on any native transform diagnostic from the catalog project.
+// 4. Hold each project to its own purpose's diagnostic contract.
 func TestProjectCatalogTransformCoverage(t *testing.T) {
 	root := transformCoverageRepoRoot(t)
 	for _, project := range []string{
@@ -35,14 +44,7 @@ func TestProjectCatalogTransformCoverage(t *testing.T) {
 	} {
 		project := project
 		t.Run(project, func(t *testing.T) {
-			cwd := filepath.Join(root, "tests", project)
-			out, errText, code := transformCoverageCapture(func() int {
-				return runTransform([]string{
-					"--cwd", cwd,
-					"--tsconfig", "tsconfig.json",
-					"--output", "ts",
-				})
-			})
+			out, errText, code := transformCatalogProject(root, project)
 			if code != 0 {
 				t.Fatalf("%s transform failed: code=%d stderr=\n%s", project, code, errText)
 			}
@@ -51,23 +53,48 @@ func TestProjectCatalogTransformCoverage(t *testing.T) {
 			}
 		})
 	}
-	for _, project := range []string{"debug", "test-error"} {
-		project := project
-		t.Run(project, func(t *testing.T) {
-			cwd := filepath.Join(root, "tests", project)
-			out, _, code := transformCoverageCapture(func() int {
-				return runTransform([]string{
-					"--cwd", cwd,
-					"--tsconfig", "tsconfig.json",
-					"--output", "ts",
-				})
-			})
-			if code != 3 {
-				t.Fatalf("%s transform should report diagnostics with code 3, got %d", project, code)
-			}
-			if !strings.Contains(out, `"diagnostics"`) || !strings.Contains(out, `"typescript"`) {
-				t.Fatalf("%s output should include diagnostics and TypeScript map:\n%s", project, out)
-			}
+	// tests/test-error compiles deliberately invalid call sites, and reporting them
+	// is that workspace's entire reason to exist. Its diagnostics are a contract,
+	// not an accident of what it currently holds.
+	t.Run("test-error", func(t *testing.T) {
+		out, _, code := transformCatalogProject(root, "test-error")
+		if code != 3 {
+			t.Fatalf("test-error transform should report diagnostics with code 3, got %d", code)
+		}
+		if !strings.Contains(out, `"diagnostics"`) || !strings.Contains(out, `"typescript"`) {
+			t.Fatalf("test-error output should include diagnostics and TypeScript map:\n%s", out)
+		}
+	})
+	// tests/debug is a one-off repro scratchpad, so nothing about its sources is
+	// promised: it holds valid code between investigations and invalid code during
+	// one. What survives both states is the project transform's own envelope
+	// contract, so that is what this pins. The project must resolve and reach a
+	// transform decision rather than fail to load (exit 2), every source must land
+	// in the TypeScript map, and the exit code must agree with the envelope it
+	// published.
+	t.Run("debug", func(t *testing.T) {
+		out, errText, code := transformCatalogProject(root, "debug")
+		if code != 0 && code != 3 {
+			t.Fatalf("debug transform should load and transform the project, got code=%d stderr=\n%s", code, errText)
+		}
+		if !strings.Contains(out, `"typescript"`) {
+			t.Fatalf("debug project output should include TypeScript map:\n%s", out)
+		}
+		if diagnosed := strings.Contains(out, `"diagnostics"`); diagnosed != (code == 3) {
+			t.Fatalf("debug exit code %d disagrees with its envelope (diagnostics present: %v):\n%s", code, diagnosed, out)
+		}
+	})
+}
+
+// transformCatalogProject runs project transform mode over one repository test
+// project and returns its captured stdout, stderr, and exit code.
+func transformCatalogProject(root string, project string) (string, string, int) {
+	cwd := filepath.Join(root, "tests", project)
+	return transformCoverageCapture(func() int {
+		return runTransform([]string{
+			"--cwd", cwd,
+			"--tsconfig", "tsconfig.json",
+			"--output", "ts",
 		})
-	}
+	})
 }
--- a/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
@@ -1,9 +1,10 @@
-﻿//go:build typia_native_internal
+//go:build typia_native_internal
 // +build typia_native_internal
 
 package main
 
 import (
+	"encoding/json"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -69,18 +70,30 @@ func TestProjectCatalogTransformCoverage(t *testing.T) {
 	// promised: it holds valid code between investigations and invalid code during
 	// one. What survives both states is the project transform's own envelope
 	// contract, so that is what this pins. The project must resolve and reach a
-	// transform decision rather than fail to load (exit 2), every source must land
-	// in the TypeScript map, and the exit code must agree with the envelope it
-	// published.
+	// transform decision rather than fail to load (exit 2), it must still
+	// contribute transformed sources, and its exit code must agree with the
+	// envelope it published.
+	//
+	// The envelope is decoded rather than substring-matched because `"typescript"`
+	// carries no `omitempty` and runTransformProject initializes the map before its
+	// source loop: the key is present even when the workspace resolved no source at
+	// all. That empty envelope is the shape a `rootDir` or `include` rot leaves
+	// behind, and this workspace is the one that already shipped such a rot
+	// (samchon/typia#2132), so a key-presence check would pass over the one failure
+	// worth catching here.
 	t.Run("debug", func(t *testing.T) {
 		out, errText, code := transformCatalogProject(root, "debug")
 		if code != 0 && code != 3 {
 			t.Fatalf("debug transform should load and transform the project, got code=%d stderr=\n%s", code, errText)
 		}
-		if !strings.Contains(out, `"typescript"`) {
-			t.Fatalf("debug project output should include TypeScript map:\n%s", out)
+		var envelope transformProjectOutput
+		if err := json.Unmarshal([]byte(out), &envelope); err != nil {
+			t.Fatalf("debug project output should be a transform envelope: %v\n%s", err, out)
 		}
-		if diagnosed := strings.Contains(out, `"diagnostics"`); diagnosed != (code == 3) {
+		if len(envelope.TypeScript) == 0 {
+			t.Fatalf("debug project should contribute transformed sources, got an empty TypeScript map:\n%s", out)
+		}
+		if diagnosed := len(envelope.Diagnostics) > 0; diagnosed != (code == 3) {
 			t.Fatalf("debug exit code %d disagrees with its envelope (diagnostics present: %v):\n%s", code, diagnosed, out)
 		}
 	})
--- a/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/project_catalog_transform_coverage_test.go
@@ -74,6 +74,12 @@ func TestProjectCatalogTransformCoverage(t *testing.T) {
 	// contribute transformed sources, and its exit code must agree with the
 	// envelope it published.
 	//
+	// Accepting either 0 or 3 is derived, not observed: transform mode never
+	// typechecks, so exit 2 is reachable only from a flag, cwd, or tsconfig-load
+	// failure (semantic diagnostics belong to the build route). No source content a
+	// repro leaves behind can produce it -- even an outright type error still exits
+	// 0 here -- which is what makes the band content-agnostic.
+	//
 	// The envelope is decoded rather than substring-matched because `"typescript"`
 	// carries no `omitempty` and runTransformProject initializes the map before its
 	// source loop: the key is present even when the workspace resolved no source at


samchon:

Posting before acting, per the campaign rule.

### Finding 1 (accepted): the new `debug` TypeScript-map check is vacuous, and its doc comment overclaims

The subtest I pushed asserts `strings.Contains(out, "\"typescript\"")` and the comment says *"every source must land in the TypeScript map"*. The comment claims more than the code proves.

`transformProjectOutput.TypeScript` is `map[string]string` tagged `json:"typescript"` **without** `omitempty`, and `runTransformProject` initializes it to `map[string]string{}` before the source loop. So the key is encoded unconditionally in project mode. Given the exit code is already constrained to `{0, 3}` — both of which imply the envelope was encoded — the substring check cannot fail. It proves nothing beyond the code check next to it.

That matters for this project specifically. `tests/debug` is exactly the workspace whose sources can silently stop resolving (the `rootDir`/`include` rot class behind #2132), and an envelope with `"typescript":{}` would sail through the assertion I just wrote.

**Remediation:** decode the envelope into the command's own `transformProjectOutput` type (same package, so no duplicated schema to drift) and assert the map is non-empty. "The scratchpad still contributes transformed sources" is purpose-derived and content-agnostic — it holds for a valid sample and a broken one alike — whereas "the key exists" is neither. Base the diagnostics agreement on the decoded `len(env.Diagnostics)` from the same envelope for precision.

### Finding 2 (rejected on evidence): substring `"diagnostics"` matching is content-dependent

I hypothesized that a scratchpad source containing the word `diagnostics` would collide with `strings.Contains(out, "\"diagnostics\"")` and make the assertion content-dependent — the very bug class this issue is about. **Tested and false.** Inside a JSON string every quote is escaped, so a source containing `"diagnostics"` encodes as `\"diagnostics\"`, whose closing quote is preceded by a backslash and therefore never completes the `"diagnostics"` substring. Verified with a standalone program. No change; recorded so the rejection is not re-litigated.

### Out of scope (reported, not swept in)

The same vacuous `"typescript"` check covers the nine clean-project subtests and `test-error`. It is pre-existing and equally weak there, but those projects have fixed contents and their exit-code assertions already carry the signal. `test-error`'s half stays byte-for-byte unchanged, per the issue.

Scope after remediation is still tests-only: no product source, no version bump.

samchon:

Fresh full round over the complete base-to-head diff (`26d968c6d8...19d1644a`), not just round 1's fix. Posting before acting.

### Finding 3 (accepted): the `{0, 3}` band reads as observed-behavior, not as a derived invariant

The `debug` comment justifies *why* diagnostics are not asserted, but never says why exit 2 may be excluded. A reader mid-repro whose scratchpad holds a half-written type error would reasonably expect this subtest to fail on them — and would then be tempted to loosen the band, which is how the original assertion rotted in the first place.

The band is in fact derivable, and I verified it: **transform mode does not typecheck.** `runTransform` calls `driver.LoadProgram` and exits 2 only on flag, cwd, or tsconfig-load failure; semantic and syntactic diagnostics are the `build` route's job, not this one's (`build_reports_semantic_diagnostics_test.go` owns that). I confirmed it empirically — putting `const broken: number = "not a number";` into `tests/debug/src/index.ts` still exits **0**, not 2.

So exit 2 is reachable only from workspace-structure rot, never from source content. That is exactly what makes `{0, 3}` a content-agnostic band rather than two codes I happened to observe, and it is the load-bearing rationale for the whole subtest.

**Remediation:** state it in the comment, in one sentence.

### Consequence surface (checked, no action)

- **Other consumers of `tests/debug`:** only its own `package.json`/`src`, `pnpm-lock.yaml`, and `.agents/skills/project/SKILL.md`, which already describes it as "a one-off `ttsx` runner for ad-hoc local repros" — consistent with this framing. No other test or script asserts anything about it, so nothing else needs updating.
- **Subtest ordering** now runs `test-error` before `debug` (the old loop was `{"debug", "test-error"}`). Each subtest captures and restores `stdout`/`stderr` through `transformCoverageCapture`'s `defer`, and shares no other state, so the reorder is cosmetic.
- **`project := project`** is redundant under `go 1.26` loop-var semantics, but it is pre-existing style in an otherwise untouched loop. Left alone rather than churned.

### Verification carried into this round

- Whole tagged native tree: **all 28 packages `ok`** (`go -C packages/typia/test test -count=1 -p 2 -tags typia_native_internal ../native/...`).
- `gofmt -l` clean; `go vet` (tagged) clean.

samchon:

Fresh full round over the complete base-to-head diff at `677610bd`. No finding survived verification, so this is the terminating clean round.

Re-checked from scratch rather than rechecking round 2's fix:

- **Biconditional soundness.** `diagnosed != (code == 3)` across all four states: `(0, none)` pass; `(3, some)` pass; `(0, some)` fails — an envelope carrying diagnostics that exited 0 is a real defect; `(3, none)` fails — reachable only through the encode-error branch, whose partial stdout trips the `json.Unmarshal` check first with a clearer message. Correct in all four.
- **Both string checks that remain** (the nine clean projects, `test-error`) are untouched and still compile — `strings` is still imported and used.
- **`test-error` is byte-for-byte unchanged** apart from routing through the extracted helper; mutation-proved below.
- **Envelope decode** tolerates the encoder's trailing newline and ignores the `graph`/`dependencies` sections; verified live, no unmarshal error in either scratchpad state.
- **Comment length** (three paragraphs) is deliberate: what is pinned, why `{0, 3}` is derived, why decoding replaces key-matching. Each prevents a specific mistake that has already been made once in this file.

### Verification at the final SHA

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia` | `ok` (85.8s) |
| `go -C packages/typia/test test -count=1 -p 2 -tags typia_native_internal ../native/...` | `ok`, all 28 packages |
| `go -C packages/typia/test test -count=1 -p 2 ../native/...` | green |
| `go -C packages/typia/test test -count=1 -p 2 ./...` | green |
| `gofmt -l`, `go vet` (tagged) | clean |

### A local failure worth recording, since it was not a code defect

Mid-review `pnpm test` failed in this worktree with 24 `TS6196`/`TS6205` "declared but never used" errors against `packages/typia/src/*.ts` while checking `tests/test-mcp`. It reproduced identically with my diff reverted to base, and CI's `test` workflow was green on the same base — so it was never this batch's.

Root cause: **the `D:` volume had 4.8 GB free and `D:\gocache` had grown to 104.7 GB.** Disk exhaustion also surfaced directly as `link.exe: resize output file failed: ... not enough space on the disk`. After removing ~4 GB of abandoned `D:\gotmp\go-build*` directories (11-12h old, from dead processes), `tests/test-mcp` passes with zero diagnostics. No master regression, no Windows/tsgo divergence, and nothing to fix in the repository — recorded so the next agent reads the disk before believing the diagnostics.

samchon:
## The untagged-CI gap: recommendation, not action

The lead asked for a recommendation rather than a unilateral change, so **this PR does not touch any workflow or script.** Findings below.

### What is actually true

| Claim | Status |
| --- | --- |
| No workflow passes `-tags typia_native_internal` | **Confirmed.** `test.yml` runs `pnpm test` → `test:go:public` (`go test ./...`) + `test:go:native` (`go test ../native/...`), both untagged. No other workflow runs Go tests. |
| No root script passes it | **Confirmed.** No root `package.json` script sets the tag. |
| The exclusion is accidental | **False — it is documented.** `.agents/skills/development/SKILL.md` states: "Root `pnpm test:go` does not enable that tag; run `go -C packages/typia/test test -tags typia_native_internal ../native/...` when tagged coverage is relevant." |
| The tagged suite is small | **False.** 55 tagged files across the native tree; 9 of the 84 test files in `cmd/ttsc-typia` alone. |
| The tagged suite is currently rotted | **False — it is healthy.** All 28 packages pass with this PR's fix. #2141 was the only breakage. |

### Why "it's deliberate" does not settle it

The documented decision keeps the tag out of **`pnpm test:go`** — a developer-ergonomics choice protecting the fast default loop. That is a different decision from **"no CI job should ever verify this suite,"** which is what the repository actually implements today. Only the first was ever deliberated; the second looks like its unexamined side effect.

The skill's fallback — run it "when tagged coverage is relevant" — requires a human to know their change touches tagged coverage. **#2133 is the exact counterexample:** it edited a TypeScript sample file, and no reasonable contributor would guess that breaks a Go test. That is precisely where the human heuristic fails, and it did.

The rot is measured, not theoretical: `project_catalog_transform_coverage_test.go` was born tagged in #1941 (`d713a2de84`, 2026-06-12), so **CI has never run it once in its life.** It broke five weeks later and surfaced only because the unrelated #2134 batch happened to run the tag by hand.

### The finding that changes the weight

The tag is **not** a clean "internal implementation churn" boundary. Sitting behind it in `cmd/ttsc-typia` are user-visible behavior contracts, not branch-coverage filler:

- `build_reports_semantic_diagnostics_test.go` — the build route must report TypeScript semantic errors
- `build_reports_syntactic_diagnostics_test.go` — likewise for syntax errors
- `transform_declaration_source_produces_no_javascript_test.go` — an emit contract
- `run_default_build_route_test.go` — the default CLI route
- `demo_rejects_bad_flag_test.go`, `plugin_options_missing_config_test.go` — CLI/plugin argument handling

If the compiler silently stopped reporting semantic diagnostics, **no CI job in this repository would fail.** That is a larger exposure than #2141 itself.

### Recommendation

**Wire it in — and I lean to the surgical option.**

**Option A (surgical, preferred).** Untag `project_catalog_transform_coverage_test.go` so it joins `pnpm test:go:native`, which CI already runs. It is not internal coverage at all: it drives the public transform command over 11 real workspaces. It was tagged by neighbourhood, not by nature. Cost ≈ **+13s**. Obstacle: its helpers (`transformCoverageRepoRoot`, `transformCoverageCapture`) live in the tagged `transform_fixtures_emit_coverage_test.go` and must move to an untagged file first. Seals this rot class without importing 54 unrelated internal tests into the PR gate.

**Option B (full).** Add a `test:go:internal` root script plus a `test.yml` step. Cost ≈ **+95s** (the Go portion roughly doubles: `cmd/ttsc-typia` is ~86s of it and must rebuild under the tag). Covers the behavior contracts listed above too.

**Blast radius either way — the honest cost.** `TestProjectCatalogTransformCoverage` reaches out of the Go module (`runtime.Caller` + five `..` hops) and transforms 11 real `tests/*` workspaces. Putting it in CI newly makes a **Go job fail when a TypeScript workspace changes** — a new `test-error` fixture or a `tests/*` tsconfig edit becomes a Go failure. That coupling is the strongest argument for the status quo, and it is real. It already exists locally; CI would only stop hiding it.

Option B also requires updating `.agents/skills/development/SKILL.md` in the same change, since its Go tests section describes the current split.

**What I would not do:** fold the tag into `pnpm test:go`. That would destroy the fast default loop the tag exists to protect, and no evidence here argues for it.

samchon:
## Verification: `pnpm test`, and why this diff cannot move it

Recording the full chain honestly, including two local failures that were **not** this batch's.

### This diff is not reachable from `pnpm test`

`pnpm test` → `test:packages` (TypeScript suites) + `test:toolchain` → `test:go:public` (`go test ./...`) and `test:go:native` (`go test ../native/...`). **Neither sets `-tags typia_native_internal`**, and the only file this PR touches is guarded by that constraint. Verified with the toolchain rather than by reading:

```
go list -f '{{range .TestGoFiles}}{{.}} {{end}}' ../native/cmd/ttsc-typia
  -> project_catalog_transform_coverage_test.go ABSENT

go list -tags typia_native_internal -f '{{range .TestGoFiles}}{{.}} {{end}}' ../native/cmd/ttsc-typia
  -> project_catalog_transform_coverage_test.go PRESENT
```

`pnpm test` does not compile the changed file at all, so its result is independent of this diff. That is the load-bearing non-regression argument; the runs below are corroboration.

### Two local failures, both environmental, both diagnosed

**1. `test-mcp` — 24 `TS6196`/`TS6205` "declared but never used" against `packages/typia/src/*.ts`.** Reproduced **identically with this PR's file reverted to base** (same 24 errors, byte-for-byte), and CI's `test` workflow is green on the same base commit. Root cause: **disk exhaustion** — `D:` was down to 4.8 GB free with `D:\gocache` at 104.7 GB. It also surfaced unmistakably as `link.exe: resize output file failed: truncate ...: There is not enough space on the disk`. After removing ~4 GB of abandoned `D:\gotmp\go-build*` directories (11-12h old, orphaned by dead processes), `test-mcp` passes with zero diagnostics.

**2. `test-typia-generate` — `not ok 5 - physical output aliases are excluded`.** Not an assertion failure: `spawnSync ... ETIMEDOUT` at `duration_ms: 180095`, against `runGenerate`'s hard `timeout = 180_000` in `scripts/filesystem/fixture.cjs`. The spawned CLI triggers a cold ttsc Go plugin build that exceeded 180s on this loaded machine. Re-run of that single test with a warm plugin cache:

```
node --test scripts/filesystem/output-directory-alias.test.cjs
ok 1 - physical output aliases are excluded
# pass 1
# fail 0
```

Same test, same code, warm cache — passes. A cold-build timing artifact, not a defect.

### Suites observed green in this worktree at `677610bd`

`test-feature-identity`, `test-langchain`, `test-mcp`, `test-typia-bundler-cache`, `test-typia-cli`, `test-typia-exact-optional`, `test-typia-schema`, `test-utils`, `test-utils-automated`, `test-vercel` — all `Success`; `test-typia-generate`'s failing case proven green in isolation.

`pnpm -r` fails fast, so `test-error`, `test-interface`, and `test-typia-automated` did not get to report in that run. A final full `pnpm test` is running against warm caches and I will post its verdict rather than claim it in advance.

### Go verification (the gates that actually exercise this diff)

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia` | `ok` (85.8s) |
| `go -C packages/typia/test test -count=1 -p 2 -tags typia_native_internal ../native/...` | `ok`, all 28 packages |
| `go -C packages/typia/test test -count=1 -p 2 ../native/...` | green |
| `go -C packages/typia/test test -count=1 -p 2 ./...` | green |
| `gofmt -l`, `go vet` (tagged) | clean |

samchon:
## `pnpm test` — green

Final run against warm caches, exit code captured directly (not through a pipe):

```
PNPM_TEST_EXIT=0
```

Every TypeScript suite `Success`, including the two that failed environmentally earlier:

`test-feature-identity`, `test-langchain`, `test-mcp`, `test-typia-automated`, `test-typia-bundler-cache`, `test-typia-cli`, `test-typia-exact-optional`, `test-typia-generate`, `test-typia-schema`, `test-utils`, `test-utils-automated`, `test-vercel` — plus 30 Go packages `ok`, zero `FAIL`.

This confirms the diagnosis in the previous comment: the `test-mcp` diagnostics were disk exhaustion, and the `test-typia-generate` `ETIMEDOUT` was a cold Go plugin build exceeding the fixture's 180s spawn timeout. Neither was reachable from this diff, which `pnpm test` does not even compile.

**All acceptance criteria from #2141 are met.** Ready for merge — not merging, as merge is lead-owned.

Issue #2146, opened 2026-07-17
Three small defects: a filter that silently matches nothing, a docblock wrong about all 8 of its structures, and a hardcoded int32 in integer bound messages


samchon:
## Problem

Three small, independently confirmed defects, each stated at the size the evidence supports.

1. **`pnpm --filter <name>` silently matches zero workspaces and exits 0** for `tests/test-utils-automated`, because its directory name disagrees with its package name `@typia/test-openapiautomated`. A verification command that runs nothing and reports success is worse than one that fails.
2. **The `HELD_OUT` docblock's stated reasons are wrong for all 8 structures** it attributes to "genuine validator gaps".
3. **`OpenApiIntegerValidator` hardcodes `Type<"int32">`** into all five of its bounds messages, so a plain `{ type: "integer", minimum: 0 }` reports a constraint the schema never declared.

## Evidence

Verified on `master`.

### 1 — the name/directory mismatch

```
pnpm --filter @typia/test-utils-automated start; echo $?
# EXIT=0, nothing executed
```

`tests/test-utils-automated/package.json:3` names the package `@typia/test-openapiautomated`. It is **the only such mismatch** in the repository.

CI is unaffected: every in-repo filter uses the path form, and the root `test:packages` runs `pnpm --filter=./tests/test-* -r start`. The exposure is human and agent verification — the lead's own certification of one batch nearly proceeded on zero executed tests because of this, and `TestGlobal.ts:47` reads the package name.

`--fail-if-no-match` exists in pnpm 10.6.4 and would turn this class of silent zero-match into an error.

### 2 — the held-out attribution

`tests/test-utils-automated/src/TestAutomation.ts:78` holds 12 structures out of the equality matrix and attributes 8 to "genuine validator gaps". Measured per structure by admitting all 12:

| Structures | Stated reason | Actual |
| --- | --- | --- |
| `DynamicNever`, `DynamicUndefined`, `ObjectUndefined` | genuine validator gaps | **one** real defect — #2145 — each misfiled under a different cause |
| `DynamicComposite`, `DynamicTemplate`, `DynamicUnion` | genuine validator gaps | a deliberate emitter over-approximation; the validator is correct |
| `TypeTagType`, `TemplateInterpolationTagged` | genuine validator gaps | tag/emitter contract limits |
| — | index-signature validator gaps | **zero** structures |

A wrong reason recorded in a comment outlives the code it explains. This campaign has been misled by its own records four times.

### 3 — the hardcoded `int32`

[...]

PR #2147, opened 2026-07-17
fix(utils): honour additionalProperties in OpenApiObjectValidator

### LINKED ISSUE #2145: OpenApiValidator ignores additionalProperties, so the documented defence against hallucinated LLM properties does nothing
## Problem

`OpenApiValidator` never reads `additionalProperties`' boolean value. `false`, `true`, and absent all take the same branch, so the keyword has **no effect at all** on whether an extra property is rejected — the only thing that decides is the `equals` option.

`IParameters.additionalProperties: false` is documented as existing *"to prevent LLMs from hallucinating extra properties"*. It does not do that.

A second, related divergence: with `equals: true`, the validator rejects a key whose value is `undefined`, while `typia.equals` accepts it. The two disagree about the same question.

Affected: every `@typia/utils` consumer routing through `OpenApiValidator` — `LlmJson.validate`, the MCP and Vercel adapters, and the HTTP validation path.

## Evidence

`packages/utils/src/validators/internal/OpenApiObjectValidator.ts`:

```ts
...(ctx.equals === true &&
typeof ctx.schema.additionalProperties !== "object"
  ? [validateEquals(ctx)]
  : []),
```

The only test is `typeof … !== "object"` — *is it a schema object or not*. The boolean is never inspected, so `additionalProperties: false`, `additionalProperties: true`, and an absent keyword are indistinguishable, and extra-property rejection is gated entirely on `equals`.

**No value of `equals` satisfies the published contract.** `ILlmSchema.IObject` declares three-way semantics, and on a schema mixing open and closed objects:

| | `additionalProperties: false` says | `additionalProperties: true` says |
| --- | --- | --- |
| `equals: false` | reject extras — **accepted instead** | accept extras — correct by accident |
| `equals: true` | reject extras — correct by accident | accept extras — **rejected instead** |

There is no setting that honours both halves of one document. The option cannot own this, because a single document can contain both kinds of object.

**Three owners, and they disagree:**

- `OpenApiObjectValidator` **ignores** the keyword (above).
- The sibling `OpenApiTupleValidator` **honours** its keyword unconditionally: `ctx.schema.additionalItems === false || ctx.schema.additionalItems === undefined`. It reads the boolean. #2043 edited that very file.
- `OpenApiTypeChecker.covers` implements the three-way semantics **correctly**.

So the repository already knows what the keyword means, in two places, and the object validator is the one that does not. **No test covers any of it** — which is why three owners could hold three positions without anything failing.

### The second half — `equals: true` versus `typia.equals`

With `equals: true`, `validateEquals` rejects a key whose value is `undefined`. `typia.equals` accepts it. `ObjectUndefined.generate()` — a pristine fixture from `@typia/template` — is rejected by its own validator.

### Consequences measured on the held-out set

`tests/test-utils-automated` holds 12 structures out of the equality matrix, and its docblock attributes 8 of them to "genuine validator gaps". That attribution is wrong for all 8. Of them:

- **3** — `DynamicNever`, `DynamicUndefined`, `ObjectUndefined` — are **this one defect**, each misfiled under a different cause;
- **3** — `DynamicComposite`, `DynamicTemplate`, `DynamicUnion` — are a deliberate emitter over-approximation, not validator gaps;
- **2** — `TypeTagType`, `TemplateInterpolationTagged` — are tag/emitter contract limits;
- **0** are index-signature validator gaps.

A two-line spec-conformance change moves exactly those 3 to PASS on **both** matrices with no regression — measured, then reverted.

## Consequence surface

- **Owner:** `packages/utils/src/validators/internal/OpenApiObjectValidator.ts`.
- **Consumers:** `LlmJson.validate`, `OpenApiValidator`, and through them the MCP, Vercel, and HTTP validation paths.
- **The documented purpose fails:** `additionalProperties: false` is published as the defence against LLM-hallucinated properties and provides none.
- **Not affected:** `typia.is`/`assert`/`validate`, which are generated from metadata rather than from the schema; and the tuple validator, which already reads its keyword.
- **Compatibility:** honouring the keyword changes behaviour in both directions — some documents start rejecting extras that were accepted, and some stop rejecting extras that were wrongly rejected under `equals: true`. Both changes are the fix.

| Case | Expected |
| --- | --- |
| `additionalProperties: false`, extra key present | reject — regardless of `equals` |
| `additionalProperties: true`, extra key present | accept — regardless of `equals` |
| `additionalProperties` absent | the published default for `ILlmSchema.IObject`, applied consistently |
| `additionalProperties` is a schema object | validate the extra values against it — unchanged |
| a document mixing open and closed objects | each object honours its own keyword |
| `equals: true`, key present with value `undefined` | agree with `typia.equals` |

## Approach

The invariant: **`additionalProperties` means what `ILlmSchema.IObject` publishes and what `OpenApiTypeChecker.covers` already implements, and it means that regardless of the `equals` option.**

`equals` and `additionalProperties` are different questions. `equals` asks whether the value must match the schema exactly; `additionalProperties` is part of what the schema *says*. Conflating them is why no setting satisfies a mixed document.

Two implementations of this keyword already exist in the repository and agree with each other — the tuple validator's and the type checker's. Prefer aligning the object validator with the specification they both already encode over inventing a third reading. And whatever the resolution, it must be covered: the absence of any test is what let three owners hold three positions.

- **Positive:** `additionalProperties: false` rejects an extra key with `equals: false`; `additionalProperties: true` accepts one with `equals: true`; a schema-object value still validates extras against it.
- **Negative:** a document mixing open and closed objects gets the right answer for each object under one `equals` setting — this is the case no current setting can satisfy, so it is the one that must fail against the old code.
- **Boundary:** `equals: true` with an `undefined`-valued key agrees with `typia.equals`; `ObjectUndefined.generate()` passes its own validator.
- **Regression:** the tuple validator and `OpenApiTypeChecker.covers` are unchanged — they are already right.
- **Held-out set:** `DynamicNever`, `DynamicUndefined`, and `ObjectUndefined` should return to both matrices, and the `HELD_OUT` docblock's attribution must be corrected — it is wrong for all 8, and a wrong reason recorded in a comment outlives the code it explains.
- **Commands:** `pnpm --filter ./tests/test-utils start`; `pnpm --filter ./tests/test-utils-automated start`; `pnpm --filter ./tests/test-mcp start`; `pnpm --filter ./tests/test-vercel start`.

## Coordination

- `@typia/utils` only. No native Go, and **no package version bump** — releases and version numbers belong to the maintainer.
- **Related to #2043**, which edited the sibling tuple validator and left this one. The reading that "the `equals` option intentionally owns this" cannot survive the mixed-document case above.
- Found by discovery round D2-A, which also answered per structure why the held-out set's stated reason is wrong for every one of the 8.

samchon:
Fixes #2145.

## Intent

`OpenApiObjectValidator` gated its superfluous-property check on `typeof ctx.schema.additionalProperties !== "object"` — a test of whether the keyword is a *schema*, which never reads the boolean. `true`, `false`, and an absent keyword were indistinguishable, so `equals: true` closed an object that had explicitly declared itself open.

**The invariant:** whether a stray property is superfluous is a question only a *closed* object answers. An `additionalProperties` that opens the object — `true` for any value, or a schema constraining it — declares undeclared keys to be legitimate members, so `equals` has nothing to decide. Only a closed object leaves the question to `equals`.

`typia.equals` is the oracle and already agrees:

| type | typia emits | `typia.is` | `typia.equals` |
| --- | --- | --- | --- |
| `{a: string}` | `additionalProperties: false` | ignores extra | **rejects** extra |
| `{a: string; [k: string]: any}` | `additionalProperties: {}` | ignores extra | **accepts** extra |

The old code was reaching for exactly this with `typeof !== "object"`; it just forgot that `true` is as open as `{}`.

**The case no option value could answer** — a document mixing an open and a closed object. Before: `equals: false` accepted extras on both, `equals: true` rejected them on both. After: one `equals: true` accepts on the open object and rejects on the closed one.

**Second half:** `equals: true` reported a key whose value is `undefined` as superfluous, while `typia.equals` accepts it — so a value typia generates was rejected against the schema typia emits from the same type. Both branches that walk `Object.entries` now skip keys with no JSON form.

## Scope

- `packages/utils` only. No native Go. No version bump.
- `OpenApiTupleValidator` and `OpenApiTypeChecker.covers` are **not** touched — they already read their keyword.
- `OpenApiIntegerValidator` and the docblock's other five structures belong to #2146/#2148 and are untouched.
- At the default `equals: false`, the only behaviour change is the `undefined` skip; closure still runs solely under `equals: true`.

## Two things the issue asserts that did not survive verification

Both are recorded in the thread with measurements.

1. **"`additionalProperties: false` → reject regardless of `equals`" is wrong.** typia emits `additionalProperties: false` on *every* closed object, and `typia.is` ignores extras on those. Implementing that reading broke **14 structures** in `test-utils-automated` and contradicts `website/src/content/docs/llm/application.mdx:438`. Reverted.
2. **The 3 held-out structures cannot return to the validate matrix.** `DynamicNever`'s schema is byte-identical to a plain empty interface's, yet their `typia.validate` oracles disagree — two types, one schema, two required answers. No schema-driven validator can give both. D2A's "no regression" claim came from probing only the 12 held-out entries, never the full matrix.

`ObjectUndefined` *did* carry a real validator defect — the `undefined`-valued key — and now passes the equality matrix. It stays held out only because one `HELD_OUT` set feeds both matrices (#2136's follow-up). The docblock's attribution is corrected in place; counts stay 83/80.

## Verification

Local only; campaign CI is cancelled per the exact-SHA gate. On `origin/master` = `16f574af02`:

| suite | cases | result |
| --- | --- | --- |
| `./tests/test-utils` | 296 | Success |
| `./tests/test-utils-automated` | 163 | Success |
| `./tests/test-mcp` | 19 | Success |
| `./tests/test-vercel` | 22 | Success |
| `./tests/test-langchain` | 13 | Success |
| `./tests/test-feature-identity` | 6 | Success |

Three mutations each independently caught; none is caught by the automated matrix, since typia never emits `additionalProperties: true` and `ObjectUndefined` is held out.

Self-Review: 5 rounds, last two clean.

### DIFF
--- a/packages/utils/src/utils/LlmJson.ts
+++ b/packages/utils/src/utils/LlmJson.ts
@@ -118,8 +118,10 @@ export namespace LlmJson {
    * auto-correction feedback.
    *
    * @param parameters LLM function parameters schema
-   * @param equals If `true`, reject extraneous properties not defined in the
-   *   schema. Otherwise, extra properties are ignored.
+   * @param equals If `true`, reject extraneous properties on a closed object.
+   *   Otherwise, extra properties are ignored. An object whose
+   *   `additionalProperties` opens it declares undeclared keys to be legitimate
+   *   members, so this flag does not close it.
    * @returns Validator function that checks data against the schema
    */
   export function validate(
@@ -187,8 +189,10 @@ export namespace LlmJson {
    *
    * @template T The expected output type
    * @param parameters LLM parameters schema
-   * @param equals If `true`, reject extraneous properties not defined in the
-   *   schema during validation. Otherwise, extra properties are ignored.
+   * @param equals If `true`, reject extraneous properties on a closed object
+   *   during validation. Otherwise, extra properties are ignored. An object
+   *   whose `additionalProperties` opens it declares undeclared keys to be
+   *   legitimate members, so this flag does not close it.
    * @returns Structured output interface with parse, coerce, and validate
    */
   export function structuredOutput<T>(
--- a/packages/utils/src/validators/OpenApiValidator.ts
+++ b/packages/utils/src/validators/OpenApiValidator.ts
@@ -18,7 +18,10 @@ import { OpenApiStationValidator } from "./internal/OpenApiStationValidator";
  * - {@link create}: Create reusable validator function from schema
  * - {@link validate}: One-shot validation with inline schema
  *
- * Set `equals: true` to reject objects with extra properties (strict mode).
+ * Set `equals: true` to reject extra properties on a closed object (strict
+ * mode). An object whose `additionalProperties` opens it — `true`, or a schema
+ * constraining the extra values — declares undeclared keys to be legitimate
+ * members, so `equals` does not close it.
  *
  * @author Jeongho Nam - https://github.com/samchon
  */
--- a/packages/utils/src/validators/internal/OpenApiObjectValidator.ts
+++ b/packages/utils/src/validators/internal/OpenApiObjectValidator.ts
@@ -32,11 +32,16 @@ export namespace OpenApiObjectValidator {
       ),
       ...(typeof ctx.schema.additionalProperties === "object" &&
       ctx.schema.additionalProperties !== null
-        ? Object.entries(ctx.value)
+        ? // A key whose value is `undefined` has no JSON form, so it is not an
+          // additional property to constrain. Validating it as a required value
+          // would reject what `typia.is` and `typia.equals` both accept on an
+          // index-signature type.
+          Object.entries(ctx.value)
             .filter(
-              ([key]) =>
+              ([key, value]) =>
+                value !== undefined &&
                 Object.keys(ctx.schema.properties ?? {}).includes(key) ===
-                false,
+                  false,
             )
             .map(([key, value]) =>
               OpenApiStationValidator.validate({
--- a/packages/utils/src/validators/internal/OpenApiObjectValidator.ts
+++ b/packages/utils/src/validators/internal/OpenApiObjectValidator.ts
@@ -52,8 +52,19 @@ export namespace OpenApiObjectValidator {
               }),
             )
         : []),
-      ...(ctx.equals === true &&
-      typeof ctx.schema.additionalProperties !== "object"
+      // Whether a stray property is superfluous is asked only of a closed
+      // object. An `additionalProperties` that opens the object — `true` for
+      // any value, or a schema constraining it — declares that undeclared keys
+      // are legitimate members, so they are never superfluous and `equals` has
+      // nothing to decide. `typia.equals` agrees: it accepts an extra key on a
+      // type carrying an index signature and rejects one on a type without.
+      //
+      // The previous `typeof !== "object"` test read only whether the keyword
+      // was a schema, so it excluded `{ ... }` but not `true`, and closed an
+      // object that had explicitly declared itself open. Reading the boolean is
+      // what lets one document mix an open and a closed object under a single
+      // `equals`, which no option value could otherwise satisfy.
+      ...(ctx.equals === true && !ctx.schema.additionalProperties
         ? [validateEquals(ctx)]
         : []),
     ].every((v) => v);
@@ -76,6 +87,12 @@ export namespace OpenApiObjectValidator {
     return Object.entries(ctx.value as object)
       .map(([key, value]) => {
         if (regular.has(key) === true) return true;
+        // A key whose value is `undefined` has no JSON form — `JSON.stringify`
+        // erases it — so the value does not carry it as a property and it
+        // cannot be superfluous. `typia.equals` agrees, and typia's own emitter
+        // erases `undefined` and `never` members from the schema while the
+        // generated value still spells the keys out.
+        if (value === undefined) return true;
         return ctx.report({
           ...ctx,
           path:
--- a/tests/test-utils-automated/src/TestAutomation.ts
+++ b/tests/test-utils-automated/src/TestAutomation.ts
@@ -78,16 +78,20 @@ export namespace TestAutomation {
    * Eight entries fail today if admitted. Failing is not by itself a validator
    * gap, and for `DynamicNever`, `DynamicUndefined`, and `ObjectUndefined` the
    * reasons below used to say it was — under three different causes, none of
-   * them the real one (#2145). All three fail for one shared reason, and it is
-   * the emitter's: a member or index-signature value typed `never` or
-   * `undefined` has no JSON form, so typia erases it and emits a plain closed
-   * object. `OpenApiValidator` sees only that schema. `DynamicNever`'s spoiler
-   * is a type error against the erased `[key: string]: never` signature, and
-   * `ObjectUndefined`'s spoilers are type errors against its erased `nothing`
-   * and `never` members; the emitted schema states neither, so the non-equals
-   * matrix cannot detect either and no schema-driven validator could. This is
-   * the same emitter over-approximation the other index-signature entries hit,
-   * not an index-signature gap and not a validator gap.
+   * them the real one (#2145). All three share one cause, and it is a schema
+   * collision, not a validator gap.
+   *
+   * A member or index-signature value typed `never` or `undefined` has no JSON
+   * form, so typia erases it. What is left is byte-identical to the schema of a
+   * type that never declared it: `DynamicNever` and an empty interface both
+   * emit `{properties: {}, required: [], additionalProperties: false}`. Their
+   * spoilers do not agree, though — `typia.validate<DynamicNever>` reports a
+   * stray key as a type error against `[key: string]: never`, while the same
+   * key against the empty interface is simply extra, and non-equals validation
+   * ignores extras by design. Two types, one schema, two answers: no
+   * schema-driven validator can give both, whatever it does with
+   * `additionalProperties`. `ObjectUndefined` collides the same way, through
+   * its erased `nothing` and `never` members.
    *
    * `ObjectUndefined` did also carry a real validator defect — an
    * `undefined`-valued key was reported as superfluous, which `typia.equals`
@@ -107,14 +111,14 @@ export namespace TestAutomation {
     // `additionalProperties` from an index signature
     DynamicArray: "index signature; passes today",
     DynamicComposite: "index signature",
-    DynamicNever: "index signature erased; no `never` in schema",
+    DynamicNever: "index signature erased; schema collides with `{}`",
     DynamicSimple: "index signature; passes today",
     DynamicTemplate: "index signature",
-    DynamicUndefined: "index signature erased; no `undefined` in schema",
+    DynamicUndefined: "index signature erased; schema collides with `{}`",
     DynamicUnion: "index signature",
     ObjectDynamic: "index signature; passes today",
-    // members erased from the emitted schema, so the validate matrix cannot see
-    // their spoilers; the equality matrix passes since #2145
+    // erased members, so the validate matrix cannot see their spoilers; the
+    // equality matrix passes since #2145
     ObjectUndefined: "`undefined` and `never` members erased",
     // integer format tags the validator does not enforce
     ConstantAtomicTagged: "uint32 tag; passes today",
--- a/tests/test-utils-automated/src/TestAutomation.ts
+++ b/tests/test-utils-automated/src/TestAutomation.ts
@@ -107,10 +107,10 @@ export namespace TestAutomation {
     // `additionalProperties` from an index signature
     DynamicArray: "index signature; passes today",
     DynamicComposite: "index signature",
-    DynamicNever: "index signature erased; schema states no `never` value",
+    DynamicNever: "index signature erased; no `never` in schema",
     DynamicSimple: "index signature; passes today",
     DynamicTemplate: "index signature",
-    DynamicUndefined: "index signature erased; schema states no `undefined` value",
+    DynamicUndefined: "index signature erased; no `undefined` in schema",
     DynamicUnion: "index signature",
     ObjectDynamic: "index signature; passes today",
     // members erased from the emitted schema, so the validate matrix cannot see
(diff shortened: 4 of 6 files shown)


samchon:
## Implementation: two of the issue's six rows are real defects; one row of the issue is wrong

The issue's core claim holds and is fixed. Two of its prescribed rows did not survive measurement. Recording the evidence before Self-Review.

### What was fixed

`OpenApiObjectValidator` gated its superfluous-property check on `typeof ctx.schema.additionalProperties !== "object"` — a test of whether the keyword is a *schema*, which never reads the boolean. Consequences, measured through the public API:

| schema | `equals: false` | `equals: true` (before) | `equals: true` (after) |
| --- | --- | --- | --- |
| `additionalProperties: true` + extra | accept | **reject** ✗ | accept ✓ |
| `additionalProperties: {…}` + extra | accept | accept | accept |
| `additionalProperties: false` + extra | accept | reject | reject |
| absent + extra | accept | reject | reject |

**The mixed document** — the case no option value could answer — now works: under one `equals: true`, the open object accepts an extra and the closed object rejects one. Before, `equals: false` accepted on both and `equals: true` rejected on both.

Second half: `equals: true` reported a key whose value is `undefined` as superfluous. `typia.equals` accepts it. Fixed by skipping `undefined`-valued keys — they have no JSON form.

### The row of the issue that is wrong: "`additionalProperties: false` → reject regardless of `equals`"

I implemented that reading first. It is disproven by the repository's own oracle:

| type | typia emits | `typia.is` | `typia.equals` |
| --- | --- | --- | --- |
| `{a: string}` | `additionalProperties: false` | ignores extra | **rejects** extra |
| `{a: string; [k: string]: any}` | `additionalProperties: {}` | ignores extra | **accepts** extra |

`additionalProperties: false` does not mean "reject extras" — typia emits it on *every* closed object, and `typia.is` ignores extras on those. It means "this type is closed", and `equals` selects which of typia's two semantics to apply. Truthy `additionalProperties` means the type is *open*, and an open type never has superfluous properties — under `is` **or** `equals`. That is the invariant the old code was reaching for with `typeof !== "object"`; it just forgot `true`.

Making `false` close unconditionally **broke 14 structures** in `test-utils-automated` (`ObjectTuple`, `ObjectUnionExplicit`, `ArrayRecursiveUnion*`, 8 equality cases…) whose spoiler oracles require extras to be ignored at `equals: false`, and it contradicts `website/src/content/docs/llm/application.mdx:438`. Reverted.

### The 3 held-out structures do not return, and D2A's "no regression" claim was unverifiable

The brief cited D2A: a two-line change moves `DynamicNever`, `DynamicUndefined`, `ObjectUndefined` to PASS on both matrices "with no regression". **D2A only probed the 12 held-out entries — it never ran the full matrix**, so it never observed the 14 regressions its patch caused among the 83 admitted structures.

All three fail the **validate** matrix for one shared reason, and it is the **emitter's**: a member or index-signature value typed `never`/`undefined` has no JSON form, so typia erases it and emits a plain closed object. `ObjectUndefined`'s spoiler sets `nothing = "undefined"`; the schema states no `nothing`, so at `equals: false` no schema-driven validator can detect it. Same class as `DynamicComposite`/`DynamicTemplate`/`DynamicUnion` — **not** an index-signature gap, **not** a validator gap.

`ObjectUndefined` *did* carry a real validator defect — the `undefined`-valued key — and now passes the equality matrix. It stays held out only because one `HELD_OUT` set feeds both matrices. Splitting it per matrix is #2136's follow-up (it would also admit `TemplateInterpolationTagged` and `TypeTagType`).

So the docblock's attribution is corrected in place and the three rows keep their entries with the measured cause. Counts stay 83/80.

### Justification

Not one word of this rests on provider behaviour; the fix makes no provider claim. It rests on: the unread boolean; the three-owner disagreement (`OpenApiTupleValidator` reads its `additionalItems` boolean, `OpenApiTypeChecker.covers` reads this one, this validator did not); the mixed document; and `typia.equals` as an independent code generator.

`OpenApiTupleValidator`, `OpenApiTypeChecker`, `OpenApiTypeCheckerBase`, and `OpenApiIntegerValidator` (#2146/#2148's) are provably untouched.

### Verification

| suite | cases | result |
| --- | --- | --- |
| `./tests/test-utils` | 296 | Success |
| `./tests/test-utils-automated` | 163 | Success |
| `./tests/test-mcp` | 19 | Success |
| `./tests/test-vercel` | 22 | Success |

Mutation: restoring `typeof !== "object"` fails `test_openapi_validator_object_additional_properties`; dropping the `undefined` skip fails `test_openapi_validator_object_undefined_property`. Neither is caught by the automated matrix — typia never emits `additionalProperties: true`, and `ObjectUndefined` is held out — so the two `test-utils` cases are the only guards.

samchon:

### Finding A — the same root cause survives in the sibling branch (whack-a-mole)

Fixing `undefined`-valued keys only in the closure check patched the case that surfaced and left the class. `OpenApiObjectValidator.validate` has a *second* branch that walks `Object.entries(ctx.value)` — the one validating extras against an `additionalProperties` schema — and it has the identical defect:

```
typia.equals<{a: string; [k: string]: string}>({a: "x", b: undefined})  = true
typia.is<{a: string; [k: string]: string}>({a: "x", b: undefined})      = true
OpenApiValidator additionalProperties:{type:"string"} + {a:"x", b:undefined}
  equals=false -> REJECTED    <-- wrong
  equals=true  -> REJECTED    <-- wrong
```

`b` has no JSON form, so it is not an additional property; the branch validates it as `required: true` and rejects `undefined`. Same cause, same function, both `equals` settings, and it reaches every consumer. Remediation: skip `undefined`-valued keys in that branch too, so both `Object.entries` walks agree that a key with no JSON form is not a property. Cover it in `test_openapi_validator_object_undefined_property`.

### Finding B — `additionalProperties: null` changed behaviour without coverage

`null` is outside the published `boolean | IJsonSchema` type, but this validator defensively guards `!== null` twice, so it is reachable. `typeof null === "object"` meant the old gate treated it as *open*; `!ctx.schema.additionalProperties` treats it as *closed*, matching `OpenApiTypeCheckerBase.coverObject`'s `!x.additionalProperties`. The change is right but was incidental and unpinned. Remediation: assert it deliberately as a boundary case.

Both are applied in the next commit.

samchon:
## Self-Review round 2 and 3

**Round 2** found one blemish and fixed it: a word hyphenated across a line break in a docblock, which `prettier-plugin-jsdoc` would reflow into nonsense. Round 2 also swept the consequence surface beyond the brief's command list — `@typia/langchain` is a direct `LlmJson.validateArguments` consumer and was not listed; it passes (13 cases). `packages/typia` turned out to use only `LlmJson.coerce`/`parse`, and `_llmApplicationFinalize` wires `func.validate` (native-generated) or a user validator — never `LlmJson.validate` — confirming the issue's "`typia.llm.application` not affected".

**Round 3 found a wrong reason in a comment I wrote myself**, which is precisely the failure this batch was told to correct in someone else's comment.

I had attributed the three structures to "the same emitter over-approximation the other index-signature entries hit". That is wrong. The emitted schema is not an over-approximation — it is a faithful narrowing. The real cause is a **collision**, and it is provable:

```
DynamicNever   {"type":"object","properties":{},"required":[],"additionalProperties":false}
IEmpty (plain) {"type":"object","properties":{},"required":[],"additionalProperties":false}
```

Byte-identical. Yet the oracles disagree: `typia.validate<DynamicNever>({something: "one"})` reports a type error against `[key: string]: never`, while the same key against an empty interface is merely extra, and non-equals validation ignores extras by design. **Two types, one schema, two required answers** — so no schema-driven validator can give both, whatever it does with `additionalProperties`. `ObjectUndefined` collides the same way through its erased `nothing`/`never` members.

This also independently re-confirms that "`additionalProperties: false` must reject extras regardless of `equals`" is untenable: `IEmpty` carries `additionalProperties: false` and `typia.validate` accepts extras on it.

The docblock and the three `HELD_OUT` rows now record the collision. Counts stay 83/80.

### Round 3 verification

| suite | cases | result |
| --- | --- | --- |
| `./tests/test-utils` | 296 | Success |
| `./tests/test-utils-automated` | 163 | Success |
| `./tests/test-mcp` | 19 | Success |
| `./tests/test-vercel` | 22 | Success |
| `./tests/test-langchain` | 13 | Success |
| `./tests/test-typia-schema` | 159 | Success |
| `./tests/test-feature-identity` | 4 | Success |

Round 3 changed the work, so a further complete round follows.

samchon:
## Self-Review rounds 4 and 5 — clean

**Round 4:** full re-read of every hunk; nothing to improve. Over-width lines in my files are three comments/template literals that Prettier does not reflow, and they match the surrounding tree (a dozen neighbours in the same directory exceed 80). Round 4 did surface that `origin/master` had moved to `16f574af02` (#2144, #2161); neither touches my surface. Rebased cleanly onto it and re-verified from scratch.

**Round 5:** complete round on the rebased state — worktree clean, source byte-identical to what the mutation battery ran against, diff confined to the six intended files, `MERGEABLE`. **Nothing found. Self-Review is complete.**

### Final state

Verified on `origin/master` = `16f574af02`, head `ac16edd3c4`:

| suite | cases | result |
| --- | --- | --- |
| `./tests/test-utils` | 296 | Success |
| `./tests/test-utils-automated` | 163 | Success |
| `./tests/test-mcp` | 19 | Success |
| `./tests/test-vercel` | 22 | Success |
| `./tests/test-langchain` | 13 | Success |
| `./tests/test-feature-identity` | 6 | Success |

Mutation evidence (each independently caught, none caught by the automated matrix):

| mutation | caught by |
| --- | --- |
| restore `typeof !== "object"` (the unread boolean) | `test_openapi_validator_object_additional_properties` |
| drop the closure `undefined` skip | `test_openapi_validator_object_undefined_property` |
| revert the `additionalProperties`-schema filter | `test_openapi_validator_object_undefined_property` |

`OpenApiTupleValidator`, `OpenApiTypeChecker`, `OpenApiTypeCheckerBase`, `OpenApiIntegerValidator` (#2146/#2148's), `ILlmSchema.ts`, and `packages/typia/package.json` are untouched — `git diff origin/master...HEAD --name-only` lists six files, none of them these.

Not merging — that is the lead's. Ready for review.

PR #2148, opened 2026-07-17
fix(tests,utils): resolve a silently-zero-match filter, a wrong held-out attribution, and a hardcoded int32 in integer bound messages

samchon:
Implements #2146 — three small, independent defects. Claim commit only at open; implementation follows.

## Scope

| # | Defect | Owner |
| --- | --- | --- |
| 1 | `pnpm --filter <name>` silently matches zero workspaces and exits 0 for `tests/test-utils-automated`, whose package name is `@typia/test-openapiautomated` | tests |
| 2 | The `HELD_OUT` docblock states reasons that are wrong for the structures it attributes to "genuine validator gaps" | tests |
| 3 | `OpenApiIntegerValidator` hardcodes `Type<"int32">` into all five bounds messages | `@typia/utils` |

No native Go. No package version bump — releases belong to the maintainer.

## Coordination

Item 2 overlaps #2145 (`fix/openapi-additional-properties`), which owns `DynamicNever`, `DynamicUndefined`, and `ObjectUndefined` and their return to both matrices. This pull request corrects the docblock for the other structures and leaves those three to #2145; it does not touch `OpenApiObjectValidator.ts`.

## Verification

Campaign CI is suspended, so verification is local and recorded on this thread:

- `pnpm --filter ./tests/test-utils-automated start`
- `pnpm test:packages`

Per the campaign rules, `pnpm format` is not run on this branch.


samchon:

Complete round over the base-to-head diff. Four findings, all applied in `010aedea8d`.

**1. My own new test duplicated a reusable structure.** `test_feature_identity_workspace_name.ts` carried its own `git()` / `toplevel()` wrappers, ~25 lines already present privately in `FeatureIdentity.ts`. The development skill forbids duplicating a reusable structure inside one test workspace. Extracted `src/Git.ts` (`Git.run` / `Git.toplevel`) and pointed both at it.

**2. A broken doc link.** The new file referenced `{@link FeatureIdentity.collect}` without importing that symbol — an unresolvable link, and precisely the "wrong record" class this pull request exists to correct. Replaced with prose.

**3. My collector was silently empty, and the vacuity floor caught it.** The first RED run failed on `collected test workspaces (0)`, not on the name mismatch: `git ls-files -- tests` resolves its pathspec against the working directory, and I ran it from `__dirname`. Fixed to run from the repository root. Recording this because the floor earned its place — without it the check would have passed vacuously on a tree it never read, which is the same defect the pull request is fixing.

**4. Suite scope widened, so the skill documents had to move with it.** `test-feature-identity` now also asserts workspace naming. Updated its description in `.agents/skills/project/SKILL.md` and `.agents/skills/development/SKILL.md` rather than leaving a stale contract.

### Two observations recorded, not fixed (out of this issue's scope)

Both are in `OpenApiSchemaNamingRule.getNameOfInteger`, which builds the *type-level* message, not the five bounds messages this issue owns. Neither is a regression from this change.

- It omits an integer's `format`, so `{ type: "integer", format: "int32" }` names itself `number`. That is an omission of a declared constraint, the mirror image of item 3's invention of an undeclared one.
- `exclusiveMinimum` / `exclusiveMaximum` are read as booleans nested inside the `minimum !== undefined` / `maximum !== undefined` branches, but the normalized `OpenApi.IJsonSchema.IInteger` declares them as numbers. A schema with only `{ exclusiveMinimum: 5 }` therefore names itself `number`, dropping the constraint from the message entirely.

Leaving both to the lead as possible separate candidates.

samchon:
## Self-Review round 2 — findings and remediation

Fresh complete round over the whole surface. Three findings applied in `9da6f75532`, and one candidate finding **refuted by evidence** rather than applied.

**1. The docblock contradicted its own measurement.** The paragraph above `HELD_OUT` still said the held-out reason describes what `OpenApiValidator` "does with" the schema, while the list I had just added directly below it establishes the opposite: the validator enforces every constraint its schema declares, and the miss is the emitted schema's. Rewrote it to describe the emitted schema. Left as-is, the comment would have re-seeded the same wrong attribution one paragraph from its own correction.

**2. An unpinned defensive branch.** I introduced `typeof format === "string"` in `describeType` without a case for it. The guard is reachable: the normalized schema type does not declare `format`, and `OpenApiConverter` copies an external document's value through verbatim, so `"format": 32` in a real document reaches this code. Added a boundary case asserting a non-string names no width.

**3. Dead indirection.** After extracting `Git`, `FeatureIdentity` kept `const git = Git.run` and a `toplevel` wrapper, each with one call site. Removed; the two call sites use `Git` directly.

### Refuted, not applied

I flagged `{@link OpenApiStringValidator}` in `OpenApiIntegerValidator` as a broken link, since that symbol is not imported. Checked before changing it: `OpenApiTypeChecker.ts` links `{@link OpenApiV3TypeChecker}`, `{@link OpenApiV3_1TypeChecker}`, and `{@link SwaggerV2TypeChecker}` without importing any of them. Cross-module `@link` as a prose cross-reference is the established convention in this package, so the link stays.

### Verification note

`pnpm test:packages` returned 1 on an intermediate run with `test-typia-generate` reporting `spawnSync ... ETIMEDOUT` at `duration_ms: 180029` — a 180 s CLI spawn budget exhausted while several agents and Go builds shared the machine. That suite is untouched by this diff. Re-run in isolation it passes: exit 0, `# tests 10`, `# pass 9`, `# skipped 1`, `# fail 0` (the skip predates this branch). A definitive full run on the final tree is recorded separately.

samchon:
## Self-Review rounds 3 and 4 — final

**Round 3** — one finding, applied in `2f365edccf`: the bound-message test's numbered scenario still described three steps after round 2 added a fourth (the non-string `format` boundary), so the doc comment no longer described the case. Round 3 also re-verified the consequence surface and changed nothing else:

- `pnpm install --frozen-lockfile` still succeeds after the rename — the lockfile keys workspace importers by **path** (`tests/test-utils-automated:`), not by package name, so CI's install step is untouched. No reference to the old name survives anywhere in the tree.
- `tests/test-utils-automated` really was the repository's only mismatch. `packages/typia` is deliberately named `typia` rather than `@typia/typia`, which is why the new check is scoped to `tests/*`: extending it to `packages/*` would need `typia` special-cased, and a hardcoded exception is exactly what the check exists to avoid.
- No website page documents `OpenApiValidator`'s messages, so no guide needed updating. The `Type<"int32">` hits under `website/` are all about the tag itself.

**Round 4** — a complete round over the whole diff produced nothing that survived verification. Self-Review stops here.

## Verification

All local; campaign CI is suspended and every run for each pushed SHA was cancelled under the exact-SHA gate.

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils start` | Success — **295** executed cases, 0 errors |
| `pnpm --filter ./tests/test-utils-automated start` | Success — **83** validate + **80** equality cases, 0 errors |
| `pnpm test:packages` | **exit 0** — 14 workspaces, **4,627** executed cases, 0 failures |
| `pnpm --filter ./packages/utils build` | exit 0 |
| `pnpm install --frozen-lockfile` | exit 0 |

The `HELD_OUT` set still resolves the documented 148 structures to exactly 83 validate and 80 equality cases, and the "148" was checked too (172 `.ts` files, 148 after the filename filter) rather than inherited on faith.

`spell-check` ran to completion on `9da6f75532` before the cancel landed and passed.

## Not merging

Per the batch brief, merge is the lead's. This is ready for that decision.

samchon:
## Rebased onto `5ec5c81ed9` — #2147 landed on shared ground

New tip `e9aa4632087bb0d06ef6b6431196d0ac2dfa2671`. One conflict, `TestAutomation.ts`, resolved so both corrections survive: #2147's three rows, its `// erased members` comment, and its schema-collision paragraphs are untouched; my five rows and the `// integer format tags the validator does not enforce` header are corrected. The intro paragraph — which we had both rewritten — is unified so the two halves do not contradict each other:

> every remaining miss belongs to that schema rather than to the validator: either two types emit one schema, or the schema is weaker than the type it came from.

**A claim of mine that #2147 disproved, now removed.** My earlier docblock said `DynamicNever` / `DynamicUndefined` / `ObjectUndefined` were "owned by #2145, which returns them to both matrices". #2147 measured that they *cannot* return — `DynamicNever`'s schema is byte-identical to an empty interface's, and two types with one schema need two different answers. That sentence is gone. My earlier comment on this thread carries the same wrong claim; leaving it as chronology and correcting it here.

## Re-measured: my table did not change

I did not take the "only the `additionalProperties: true` case changes" reading on faith, and it is **incomplete**. #2147 changed that validator twice: the `equals` gate, and a new `value !== undefined` filter on the **object-valued** `additionalProperties` branch — which is exactly the branch `DynamicComposite`, `DynamicTemplate`, and `DynamicUnion` go through. (The gate rewrite also differs on `additionalProperties: null`, where `typeof null === "object"` made the old test false and `!null` makes the new one true. Not reachable from any emitted schema; noted, not chased.)

So I re-measured all twelve by admitting them against merged master. Verdicts, spoilers, and mechanism are identical to my pre-rebase run:

| Structure | validate | equals | spoiler that survives | changed? |
| `DynamicComposite` | FAIL | FAIL | `$input["0"]` | no |
| `DynamicTemplate` | FAIL | FAIL | `$input.prefix_wrong` | no |
| `DynamicUnion` | FAIL | FAIL | `$input.prefix_wrong` | no |
| `TypeTagType` | FAIL | PASS | `$input.value[0].uint` | no |
| `TemplateInterpolationTagged` | FAIL | PASS | `$input.percent` | no |
| `DynamicArray` / `DynamicSimple` / `ObjectDynamic` / `ConstantAtomicTagged` | PASS | — | — | no |
| `DynamicNever` / `DynamicUndefined` (#2147's) | FAIL | — | `$input.something` | no |
| `ObjectUndefined` (#2147's) | FAIL | **PASS** | `$input[0].nothing` | **yes — FAIL→PASS** |

The single change across all twelve is `ObjectUndefined`'s equality matrix, which #2147 owns and its docblock already records. All eight failures remain mode-1 — the validator accepts the spoiler — so the mechanism is unchanged, not merely the verdict.

## Verification on the rebased tip

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils-automated start` | Success — **83** validate + **80** equality, 0 errors (agrees with #2147) |
| `pnpm --filter ./tests/test-utils start` | Success — **297** cases, 0 errors; #2147's `..._object_additional_properties` and `..._object_undefined_property` both present and passing alongside mine (296 + 1) |
| `pnpm test:packages` | **exit 0** — 14 workspaces, **4,629** executed cases (4,627 + #2147's two), 0 failures |

`#2144` (`project_catalog_transform_coverage_test.go`) and `#2161` (`website.yml`, `website/build/deploy.js`) share no file with this branch — confirmed against the diff, not assumed.

Both out-of-scope observations on `OpenApiSchemaNamingRule.getNameOfInteger` are still true on merged master; #2147 did not touch that file. Left for the lead to file.

Still not merging — lead-owned.

PR #2149, opened 2026-07-17
fix(native): read typia's transform options from ttsc's resolved plugin entry

### LINKED ISSUE #1887: native: replace regex/hardcoded tsconfig scrape for plugin options with structured --plugins-json parsing
## Context

Surfaced during the multi-agent review of the emit-factory unification work (PR #1885, branch `feat/emit`). It is a **pre-existing** issue (introduced ~commit `a3c30fc0c`), deliberately **deferred to its own PR** because changing plugin-option detection is risky and was not in scope of the emit refactor. This issue is self-contained — an agent should be able to start and finish from this text alone.

## Problem

`packages/typia/native/cmd/ttsc-typia/build.go` → `readTypiaPluginOptions(cwd, tsconfigPath)` (currently ~line 171) re-reads `tsconfig.json` from disk and **regex-scrapes** it for the typia plugin options, gated on a **hardcoded transform specifier**:

```go
func readTypiaPluginOptions(cwd, tsconfigPath string) typiaadapter.PluginOptions {
  // ...reads tsconfig.json text...
  if !regexp.MustCompile(`(?s)"transform"\s*:\s*"typia/lib/transform"`).MatchString(text) {
    return typiaadapter.PluginOptions{}   // any other valid config -> ALL options silently false
  }
  return typiaadapter.PluginOptions{
    Functional: regexp.MustCompile(`(?s)"functional"\s*:\s*true`).MatchString(text),
    Numeric:    regexp.MustCompile(`(?s)"numeric"\s*:\s*true`).MatchString(text),
    Finite:     regexp.MustCompile(`(?s)"finite"\s*:\s*true`).MatchString(text),
    Undefined:  regexp.MustCompile(`(?s)"undefined"\s*:\s*true`).MatchString(text),
  }
}
```

The same function is called from `transform.go:72` and `build.go:79`.

Why this is fragile:
1. The hardcoded `"typia/lib/transform"` literal means any other valid spelling of the same transform (absolute path, alternate specifier, multiple plugins, an `extends`-ed config) yields **all options false** — silently disabling `functional`/`numeric`/`finite`/`undefined`.
2. Regex over raw text ignores JSON structure: JSONC comments, an unrelated nested `"functional": true`, or `extends` produce wrong results.

## The structured payload is already available

ttsc already passes the real plugin config via `--plugins-json`, but the flag is **declared and discarded**: `_ = fs.String("plugins-json", "", "ordered ttsc plugin payload")` (`build.go:33`, `transform.go:36`).

Its format (from ttsc `serializeNativePlugins`): a JSON array of `{ config, name, stage }`, where `config` is the **original tsconfig plugin entry** verbatim, e.g.:

```json
[{ "config": { "transform": "typia/lib/transform", "functional": true, "numeric": false }, "name": "...", "stage": "..." }]
```

## Proposed fix

Parse `--plugins-json` (fall back to the current tsconfig read only if the flag is absent/empty), find the typia plugin entry, and read `functional`/`numeric`/`finite`/`undefined` from its `config` object as real booleans. Identify the typia entry robustly (e.g. the entry whose `config.transform` resolves to this package's transform) rather than by an exact-string match; if identification is genuinely ambiguous, document the chosen heuristic. Remove the four option regexes and the hardcoded specifier gate.

## Constraints

- **Behavior-invariant for the common case**: a standard `{ "transform": "typia/lib/transform", "functional": true, ... }` config must yield the exact same `PluginOptions` as today.
- Follow `.codex/skills/development/SKILL.md` (Work Rules, Change Integrity). Native source change ⇒ bump the `typia` package version in the same PR (lockstep, see how prior PRs do it). Run `pnpm format` (covers Go) before committing.
- Match the repo's 2-space Go style.

## Validation

- `cd packages/typia/native && GOPROXY=off go build ./... && go vet ./...`
- `pnpm test:go`
- Full `pnpm build && pnpm test` and CI. **CI gates this**: the `tests/test-typia-*` projects exercise the `functional`/`numeric` features, so a detection regression fails CI. Add/verify a test that toggles an option if coverage is thin.
- Note: if working on the `feat/emit` branch, local TS test runs can be unreliable due to a `go.work` local-ttsc override; rely on CI there. On `master` this is not an issue.

## Acceptance criteria

- [ ] Plugin options are read from the structured `--plugins-json` payload (config object), not by regex over tsconfig text.
- [ ] No hardcoded `"typia/lib/transform"` literal gate; typia entry identified structurally.
- [ ] Standard configs produce identical `PluginOptions`; CI green incl. functional/numeric coverage.
- [ ] `typia` version bumped; `pnpm format` applied.

## References
- `packages/typia/native/cmd/ttsc-typia/build.go` (`readTypiaPluginOptions`, flag at :33)
- `packages/typia/native/cmd/ttsc-typia/transform.go` (:36, :72)
- `packages/typia/native/adapter/adapter.go` (`PluginOptions`)
- ttsc `--plugins-json` producer: `ttsc/lib/compiler/internal/transformProjectInMemory.js` (`serializeNativePlugins`)
- Origin: PR #1885 review

samchon:
## Intent

Fix #1887. typia's native hosts declare ttsc's `--plugins-json` flag only to discard it (`_ = fs.String("plugins-json", ...)`, four sites) and instead re-derive the four public transform options — `functional`, `numeric`, `finite`, `undefined` — by running regexes over the raw bytes of the **leaf** `tsconfig.json`.

The scrape is wrong in both directions, silently, with no diagnostic:

| `tsconfig.json` content | typia reads `functional` today | correct |
| --- | --- | --- |
| set inline in typia's entry | `true` | yes |
| `// "functional": true` — the user commented it out | `true` | **no — typia enables what the user disabled** |
| `/* "functional": true */` | `true` | **no** |
| set on *another* plugin's entry | `true` | **no — reads a sibling's option as its own** |
| set in a base tsconfig, leaf `extends` it | `false` | **no — silently ignored** |

The `extends` row proves the mechanism: the transform **runs**, so ttsc resolved `extends`, found typia in the base entry, built the plugin, and passed the resolved entry back on `--plugins-json` — and typia ignored the very entry that caused it to be loaded.

**The invariant:** typia's effective transform options are exactly the options in typia's own resolved plugin entry — no more, no less. The host already computed that entry; the plugin's job is to find its own entry in the payload and read it, as every sibling ttsc plugin (`@ttsc/lint`, `@ttsc/paths`, `@ttsc/strip`) already does.

## Scope

Details are recorded as review comments on this thread as the work lands.

- One owner for the contract, replacing three byte-identical transcriptions of the scrape.
- All three hosts fixed: `build`, `transform`, and the playground wasm plugin.
- Regression matrix covering every row above, plus an `extends`-shaped fixture in a real suite — the coverage hole that let #1887 be closed as completed with the code untouched.

## Deferred / out of scope

- `packages/typia/package.json` is deliberately **not** touched. This changes Go source under `packages/typia/native`, which would ordinarily owe a coordinated `typia` version bump under the development skill's Change Integrity rule. Releases and version numbers belong to the maintainer, not this campaign; the bump is left for the maintainer to make.

## CI

Campaign implementation wave: automatic CI runs for this branch's commits are cancelled per the issue-campaign exact-SHA gate, and repository Actions settings are unchanged. Local verification is recorded on the thread in place of CI.

Refs #1887


samchon:

Complete round over `origin/master..b2794d15f6` (both implementation commits, every changed file and hunk), plus the consequence surface: the ttsc 0.19.2 payload producers, the wasm hosts, the suites that configure typia options, and the setup guide.

### The fallback decision, restated after re-verifying it

Recorded here because it is the one thing the issue text asks for and gets wrong.

**No tsconfig fallback. The payload is the only source.** #1887's original proposal — "fall back to the current tsconfig read only if the flag is absent/empty" — would have kept the defect alive on exactly the path that takes the fallback. I verified who actually lacks the flag before deciding:

| Host path | Sends `--plugins-json`? | Where |
| --- | --- | --- |
| `ttsc build` / `ttsc check` | yes | `runBuild.ts:767` `createNativeBuildArgs` |
| check / fix / format stage | yes | `runBuild.ts:809` `createNativeCheckArgs` |
| ttsc transform stage — drives `ttsx`, `@ttsc/unplugin`, `typia generate` | yes | `transformProjectInMemory.ts:304,325` |
| resident transform | yes | `startResidentTransform.ts:77` |
| LSP sidecar | yes | `lsp_native_plugin_source.go:405` |
| **website playground wasm** | **no** | `createTypiaWorkerCompiler.ts` → `api.plugin({…})` |

The playground was the one real gap, and it is **not** an argument for a fallback — it is an argument for the site passing the manifest. The site already builds typia's resolved entry (`createTypiaPluginEntry`) to write into its MemFS tsconfig; it now hands that same object to the plugin. The JS *is* the host there, so it is the one that resolved the entry. One contract everywhere, and no Go code re-reads a file the JS just wrote.

This mattered more than "behaviour-invariant": the playground UI has live transform toggles that flow into that entry (`website/src/compiler/index.ts`), so dropping the read *without* passing the manifest would have silently frozen those toggles at their defaults. That is a regression a fallback would have hidden rather than prevented.

A "correct" fallback was also weighed and rejected on its own merits: to be right it would have to re-implement JSONC parsing and the whole `extends` walk. This repo's own `tests/config/tsconfig.json` and `tests/test-typia-automated/tsconfig.json` use trailing commas, and the former is dense with commented-out options — so a strict `encoding/json` parse fails outright and a lenient one still has to resolve `extends`. Duplicating the host's job is what produced this bug; doing it a second time to paper over it is the wrong direction.

### Findings I am acting on

1. **`adapter.PluginOptionBoolean` is exported without a consumer.** Only `ReadPluginOptions` and `PluginName` are needed outside the package. The project skill treats the exported surface as a contract, so an export with no caller is a liability. → unexport to `pluginOptionBoolean`.

2. **The host matrix passes the flag in a form ttsc never sends.** My tests use `"--plugins-json", payload` (two argv entries); ttsc sends `--plugins-json=<json>` (one entry, `=`-joined). Go's `flag` accepts both, so this is not a bug — but the tests should drive the real wire form, since that is the shape the defect will regress in. → switch the matrix to `--plugins-json=` + payload.

3. **`website/src/compiler/index.ts`'s header comment is now half true.** It says the wrapper exists "so each compile can serialize the current transform toggles into `compilerOptions.plugins`". It now also hands that entry to the plugin as its manifest. → extend the comment.

### Verification gap I am closing in this round

`tests/test-typia-automated` is the only suite that exercises `functional` / `numeric` / `finite`, and it declares them inline — the one shape the old scrape read correctly and therefore the one most likely to regress under a payload-only reader. It is not in my briefed command list, but it is the suite those three options belong to. → running it.

### Checked and found sound (no action)

- **`packages/typia/package.json` untouched**, deliberately — see the PR body.
- **Matching typia's entry by `name`, not by `config.transform`.** `name` is the identity the host and plugin already agree on: typia's descriptor sets `name: "typia"` (`packages/typia/src/transform.ts`), the playground plugin registers `Name() = "typia"`, and `api.plugin({name})` could not have dispatched to the plugin at all under any other name. Matching the `transform` specifier instead would reintroduce the hardcoded-literal gate the issue explicitly asks to remove — a user may legally spell it as a bare specifier or an absolute path.
- **Sibling entries in the same payload.** A transform host carries every transform-stage plugin's entry (`runBuild.ts`: "all other transform plugins ride inside it via the `--plugins-json` flag"), and `createNativeCheckArgs` sends the full set. Locating our own entry by name is required, not defensive; the matrix pins both the leak case and the not-first case.
- **`undefined` tri-state survives.** `76c981ad4b` established nil / true / false; the adapter matrix pins all three plus the non-boolean case, and `tests/test-typia-exact-optional` proves it end-to-end.
- **Undecodable payload errors instead of defaulting.** Silently substituting defaults over a manifest we cannot read is the same silent-wrong-options failure this fix exists to remove. Exit 2 matches the documented config/usage code.
- **`stage` is not matched on.** typia is transform-stage; a check-stage entry named `typia` cannot exist, and typia's binary never runs the check path.
- **Docs.** `setup/index.mdx` "Transform Options" documents the inline shape and the flag semantics. Nothing there becomes false — the change makes the documented behaviour work in more places (`extends`) and stop working where it never should have (comments). No guide edit is owed. (`undefined` being undocumented is a real but separate, pre-existing gap — out of scope here.)
- **Deleted oracles.** `plugin_options_undefined_false_test.go` (both copies) and the `typia_native_internal`-tagged `plugin_options_missing_config_test.go` tested a function that no longer exists. Their contract moved to `adapter` under a strictly broader matrix that also **runs by default** — the tagged one never did. Called out as a spec change in the report.

samchon:

Fresh complete round over `origin/master..2f6deb1b8b` (every changed file and hunk re-read from the current state, not a recheck of round 1's fixes), plus the consequence surface again: every payload producer in the pinned ttsc 0.19.2, both wasm entrypoints, every workspace that configures a typia option, and the CI workflows.

### Findings I am acting on

1. **A fifth `--plugins-json` site the issue never named, and my own change created the hole in it.** `packages/typia/native/cmd/ttsc-typia/main_wasm.go` renders its JS options object into argv through an explicit allow-list — `cwd`/`tsconfig`/`file`/`output` for transform, `cwd`/`tsconfig`/`outDir`/`emit`/`noEmit` for build. There is no pass-through, so `--plugins-json` could never reach `runBuild`/`runTransform` through it. While the options were scraped back out of tsconfig that was invisible; with the scrape gone, a JS host driving this wasm would be pinned to typia's defaults forever with no channel left to say otherwise. Fixed in `b2794d15f6` by forwarding the manifest, same as the other four sites.

   Worth being precise about the blast radius, because I do not want to overstate it: nothing in this repository builds `cmd/ttsc-typia` for `GOOS=js` — `website/build/wasm.js` builds only `website/compiler/cmd/playground`, and the `apiName: "ttscTypia"` in `website/src/compiler/index.ts` binds the *playground* host's global, not this one. So this is a latent hole in an entrypoint with no current consumer, not a live regression. It is still a host, it still reads options, and leaving it as the one host that structurally cannot receive them is exactly the shape of the defect being fixed.

2. **The repository spell-check would have failed this branch.** `typos.toml` sets `locale = 'en-us'`, and `git grep` for `behaviour|recognise|honour` returns *only* files this batch added — the rest of the repository is uniformly en-us. Fixed in `2f6deb1b8b`. Also sharpened the reader's doc comment while there: re-reading the tsconfig is what *made* #1887, so a fallback is the consequence to reject, not the cause — the earlier wording had that backwards.

3. **A redundant `err` shadow I introduced.** Reading the options up front puts `err` in each host's outer scope, which left `var err error` inside the `cwd` fallback shadowing it for nothing. Cleaned in all four hosts.

### Checked and found sound (no action)

- **My test files are not silently excluded.** Go reads a trailing `_js`/`_wasm`/`_windows` before `_test.go` as a build constraint. Verified empirically rather than by inspection — `go test -list` reports all five new functions across the three packages (`TestReadPluginOptionsPayloadMatrix`, `TestBuild`/`TestTransformPluginOptionsComeFromPayload`, `TestTypiaBuild`/`TestTypiaTransformOptionsComeFromPayload`), and each was observed both failing pre-fix and passing post-fix.
- **The playground never sends an empty manifest.** `runTypia` is `runTypia && typiaPlugin !== null`, and `applyTypiaTransform` only runs under `runTypia`, so the `typiaPluginsJSON: ""` branch is unreachable from the transform call. An empty value would be read as "absent" and yield defaults regardless.
- **`api.build` needs no manifest.** `runTypiaBuildPipeline` calls `@ttsc/wasm`'s base `Build`, which compiles the already-rewritten TypeScript; typia's transform ran earlier via `api.plugin`. Only the plugin dispatch carries options.
- **One entry, two uses.** The playground resolves `createTypiaPluginEntry(...)` once and both writes it into the MemFS tsconfig and serializes it into the manifest — the two cannot drift, which was the whole failure mode.
- **The reader lost its filesystem and path handling entirely.** No `os.ReadFile`, no `filepath.Join(cwd, …)`, so the Windows/POSIX path surface it used to carry is simply gone rather than re-verified.
- **`ReadPluginOptions` is pure** — no shared state, safe under the concurrent playground invocations `host.PluginInvocation` allows.
- **Every other typia entry point still reaches the payload**: `tests/debug` and the benchmark go through `ttsx`/`ttsc`; `typia generate` goes through ttsc's `TtscCompiler`; `@ttsc/unplugin` goes through `transformProjectInMemory`/`startResidentTransform`. All four producers pass the flag.
- **Auto-discovery keeps typia's name.** A project with no `plugins` entry loads typia from `package.json`'s `ttsc.plugin`, but the descriptor module is still typia's `createTtscPlugin`, which sets `name: "typia"`. The name match holds on both discovery paths.

### Verification added this round

`tests/test-typia-automated` (inline `functional`/`numeric`/`finite` — the shape the old scrape read correctly, so the one most at risk from a payload-only reader) and `tests/test-typia-generate` are running; results will be reported. Neither was in the briefed command list; the first is the natural owner of those three options and the second exercises the `typia generate` → `TtscCompiler` producer.

samchon:
## Local verification — head `474fee5742`

Campaign CI is suspended for this branch, so this is the record that stands in for it. Every exact-SHA gate passed; the run IDs are in the campaign ledger. One run reached terminal before its cancel could land — `29574288494`, **spell-check, `success`** — recorded rather than cancelled per the gate's already-terminal rule, and a welcome piece of real CI evidence given that this batch had to correct its own en-us spellings.

### Fail-then-pass, every row of the issue's reproduction table

The matrix writes a project whose tsconfig *text* disagrees with the resolved entry the host hands over, which is precisely the distinction the scrape could not make. `functional` is the probe: with it on, typia's validator guards the function-typed property (`"function" === typeof input.cb`); with it off, it skips it.

Reverting **only** `build.go` + `transform.go` to `origin/master` (tests untouched):

| Row | Pre-fix | Post-fix |
| --- | --- | --- |
| option inline on typia's entry (control) | **pass** — behavior-invariant | pass |
| `// "functional": true` commented out | **FAIL** — "functional is not configured but the validator checked the function property" | pass |
| `/* "functional": true */` commented out | **FAIL** — same | pass |
| option owned by another plugin's entry | **FAIL** — same | pass |
| option in a base tsconfig the leaf `extends` | **FAIL** — "functional is configured on but the validator skipped the function property" | pass |

The control passing pre-fix is the point: the documented shape is unchanged, and only the four defect rows move.

### Each of the three owners has its own covering surface

Proven by reverting one owner at a time, not by inspection:

- **`build.go` + `transform.go` reverted, playground left fixed** → `TestBuildPluginOptionsComeFromPayload` and `TestTransformPluginOptionsComeFromPayload` each fail all four rows through their own host; playground tests stay green.
- **`typia_plugin.go` reverted, CLI hosts left fixed** → `TestTypiaBuildOptionsComeFromPayload` and `TestTypiaTransformOptionsComeFromPayload` each fail all four rows through the playground's own path; CLI tests stay green.

Five test functions, confirmed reachable by `go test -list` rather than assumed (a trailing `_js`/`_wasm` before `_test.go` would have silently excluded them):
`TestReadPluginOptionsPayloadMatrix`, `TestBuildPluginOptionsComeFromPayload`, `TestTransformPluginOptionsComeFromPayload`, `TestTypiaBuildOptionsComeFromPayload`, `TestTypiaTransformOptionsComeFromPayload`.

### End-to-end through a real `ttsx` compile — the coverage hole itself

`tests/test-typia-exact-optional` now inherits its `undefined: false` entry through `extends`. No suite in this repository had ever declared a typia option that way, which is why #1887 could be closed as completed with the code untouched and every suite green.

- **Pre-fix** (scrape restored, real `ttsx`, plugin rebuilt): both cases **Error** — `Bug on is rejects explicit optional undefined`, `Bug on clone root optional`.
- **Post-fix**: `test_exact_optional_undefined_validators` and `test_exact_optional_undefined_clone_random` both pass.

This is ttsc resolving `extends`, delivering the resolved entry on `--plugins-json`, and typia honoring it — the exact chain the issue says was broken.

### Commands

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` | ok, 89.8s |
| `go -C packages/typia/test test -p 2 ../native/...` | ok — all 14 packages |
| `go -C packages/typia/test test -p 2 ./...` | ok — all 15 public/contract packages |
| `go test -count=1 -p 2 ./...` in `website/compiler` | ok |
| `GOOS=js GOARCH=wasm` build, `cmd/ttsc-typia` | exit 0 |
| `GOOS=js GOARCH=wasm` build, `website/compiler/cmd/playground` (the shipped wasm) | exit 0 |
| `pnpm --filter ./tests/test-typia-cli start` | Success |
| `pnpm --filter ./tests/test-typia-exact-optional start` | Success — both cases, and proven to fail pre-fix |
| `pnpm --filter ./tests/test-typia-automated start` | Success — inline `functional`/`numeric`/`finite`, thousands of cases |
| `pnpm --filter ./tests/test-typia-generate start` | 9 pass / 0 fail / 1 environmental skip — see below |
| `npx tsc --noEmit` in `website/` | exit 0 |
| spell-check (`typos`, real CI run on this SHA) | success |

### One `test-typia-generate` failure, investigated rather than re-run away

Its first run reported `# fail 1`, so it is recorded here with the evidence rather than buried.

| Run | Conditions | Result | Duration |
| --- | --- | --- | --- |
| 1 | concurrent with `test-typia-automated` compiling ~4,000 generated files plus Go suites | `# pass 8`, **`# fail 1`** | 215.8s |
| 2 | no contention | `# pass 9`, `# fail 0` | 43.3s |
| 3 | no contention | `# pass 9`, `# fail 0` | 17.5s |

The failure tracks a 12x slowdown from load I created by running suites in parallel, not the diff. All ten cases in that file are symlink/glob filesystem tests — "mutual directory links terminate", "parent directory links terminate", "broken input links are rejected" — with termination and timing semantics, and this batch touches no filesystem, symlink, or glob code. Option delivery on this suite's path (`typia generate` → `TtscCompiler` → ttsc) is exercised by the same run; had that regressed, the symptom would be a validator behavior difference, not a link-termination test. The campaign ledger already records this same suite hitting "an unrelated generate-fixture process timeout under shared contention" during PR #2105.

The persistent `# skipped 1` — "in-root source-file links remain supported # SKIP file symbolic links require Windows developer mode" — is environmental and pre-existing on this machine.

Not run: `pnpm format`, per the campaign's no-format rule. New Go files are hand-written in the repository's 2-space style (verified: zero tab-indented lines across every added/modified `.go` file).

### Spec changes called out

- **Deleted** `plugin_options_undefined_false_test.go` (native **and** playground copies) and the `typia_native_internal`-tagged `plugin_options_missing_config_test.go`. They tested `readTypiaPluginOptions`, which no longer exists. The playground copy's own doc comment stated its purpose — "the playground **mirrors** typia's native CLI option reader … otherwise the exact optional-property behavior would **diverge**" — which is pinning one copy against the other copy rather than against the specification. There is now one owner, so divergence is structurally impossible and the test's reason for existing is gone. Their contract moved to `adapter`'s payload matrix, which is strictly broader and, unlike the tagged one, **runs by default**.
- **Changed** `tests/test-typia-exact-optional/tsconfig.json` to inherit its plugin entry from a new `tsconfig.base.json`. Same option, same assertions; the suite now also guards `extends` resolution end-to-end.

samchon:
## Self-Review rounds 3 and 4 — closing the cycle

### Round 3 — one finding, applied

**A comment left stale by my own change.** `typia_plugin.go`'s header illustrated what the stream-ownership contract does *not* cover with "Filesystem access (`os.Getwd`, `os.ReadFile`, …)" — but the file no longer calls `os.ReadFile` at all now that its options arrive as a manifest. Pointed the example at a function it still uses and named the shared manifest reader in the adapter surface list (`474fee5742`).

Found by asking the diff a question rather than re-reading it: `git grep os.ReadFile` across the four option hosts and the adapter, expecting zero hits. The one hit was this comment — which is also the cleanest available proof that the scrape is structurally gone rather than narrowed.

### Round 4 — clean

Fresh complete pass over `origin/master..474fee5742`: every changed file and hunk re-read from the current state, working tree clean, all suites green. No sound improvement found, so the cycle stops here.

### Rounds summary

| Round | Findings applied |
| --- | --- |
| 1 | Unexported `PluginOptionBoolean`; drove the matrices with the real `=`-joined wire form; extended the playground worker's header |
| 2 | Fixed the fifth `--plugins-json` site (`main_wasm.go`) my own change had holed; corrected en-us spellings CI would have failed on; removed a redundant `err` shadow |
| 3 | Corrected the playground header's stale `os.ReadFile` example |
| 4 | **Clean — no changes** |

Every round's findings were posted here before being acted on.

### What I would flag to a reader of this PR

- **The fallback decision is the load-bearing one**, and the evidence that settled it is worth keeping: the playground — the only host that lacked `--plugins-json` — generates its tsconfig through `JSON.stringify` (strict JSON, one plugin entry, no comments, no `extends`). The old scrape reads that shape *correctly*. So a fallback would have been silently taken by exactly one host, passed every test, and left the scrape alive in the tree — invisible until someone added a second plugin to the playground wasm or let users supply their own tsconfig. Passing the manifest instead deletes the scrape outright: there is no code left that can regress this way. That is a stronger property than "behavior-invariant", and it is why the issue's own proposal had to be rejected rather than implemented.
- **`packages/typia/package.json` is untouched by instruction.** Go source under `packages/typia/native` changed, which normally owes a coordinated `typia` version bump under Change Integrity. Releases belong to the maintainer, not this campaign.
- **Two pre-existing items observed and deliberately not taken**, since neither is this issue's: `setup/index.mdx` documents `functional`/`numeric`/`finite` but not `undefined`, which is a real tested public option; and `packages/typia/native/cmd/ttsc-typia` cannot be `go vet`-ed under `GOOS=js` because its test files reference `run`, which lives in the `//go:build !js` `main.go` (the *build* is clean; only vet is blocked).

samchon:
## Rebase and re-verification

Rebased onto current `master`. Clean replay, no conflicts; the seven commits are unchanged in content — `git diff <old-tip> <new-tip>` over the whole change surface is empty, so this is a base move, not a rewrite.

`master` moved twice while I worked: `5ec5c81ed9` (the requested base) and then `976bc31f78` (#2148), which another agent's fetch pulled into the shared `.git` mid-verification. I rebased onto the later one so the branch is current, and confirmed the branch's diff against it is exactly the batch's 19 files and nothing else.

### The three commits ahead, checked individually

| Commit | Touches this batch? |
| --- | --- |
| `b8fa1e816a` #2144 — rewrites `project_catalog_transform_coverage_test.go` | **Yes — same package.** See below. |
| `16f574af02` #2161 — `website.yml` + `website/build/deploy.js` | No. CI config and the deploy script; no compiler, plugin, or Go surface. |
| `5ec5c81ed9` #2147 — `OpenApiObjectValidator` + `LlmJson` + `TestAutomation.ts` | No. `@typia/utils` and its suites; no native, playground, or plugin-option surface. |
| `976bc31f78` #2148 — `OpenApiIntegerValidator`, `test-feature-identity`, `test-utils-automated`, root `package.json` | No. Verified by path: zero files under `packages/typia/native`, `website/compiler`, `website/src/compiler`, or `tests/test-typia-exact-optional`. It also leaves the typia plugin source byte-identical, so the content-keyed plugin cache key does not move and the built binary is the same one the earlier runs exercised. |

### #2144's catalog test — passes, and here is why that is not luck

`go -C packages/typia/test test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia` → **ok**, and `TestProjectCatalogTransformCoverage` specifically → **PASS** (19.4s, all 11 subtests: the 9 feature projects at exit 0, `test-error` at exit 3 with diagnostics, and `debug` against the new decoded-envelope contract).

It was right to single this out, because the interaction is real and worth stating precisely rather than just reporting green:

`transformCatalogProject` drives `runTransform` with `--cwd`, `--tsconfig`, `--output` and **no `--plugins-json`**. So under this batch every catalog project transforms at typia's defaults. Before it, the scrape read each project's leaf tsconfig. Exactly one of the eleven sets any option — `tests/test-typia-automated`, with inline `functional`/`numeric`/`finite` — so that project is the only one whose transform configuration changes here, and it changes from "options on, by accident of the scrape" to "options off, by contract".

That does not fail the test, and the reason is structural rather than lucky: those three options only make validators *stricter* by adding property checks. They cannot introduce a transform diagnostic, so they cannot move any of the three things this test asserts — exit ∈ {0,3}, a non-empty TypeScript map, and `diagnostics ⟺ exit 3`. The test's stated purpose — Go coverage of native branches across a wide API corpus — is unaffected, since options gate how call sites are validated, not which shapes are exercised.

The narrow, honest consequence: this Go harness no longer covers the `functional`/`numeric`/`finite` emit branches for that one project. Those branches remain covered by `tests/test-typia-automated` itself through real `ttsx` with the payload (re-run below, 3,945 executed cases), and by this batch's own matrices, which assert the `functional` branch directly in both directions through all three owners.

**I deliberately did not "fix" this by passing a payload in the harness.** Doing so would mean hardcoding each project's resolved plugin entry into the test — re-implementing the tsconfig resolution this batch exists to stop duplicating — and #2144 had just rewritten this very test to assert *by purpose, not by current contents*. A harness that never intended to configure typia now receives typia's defaults, which is precisely the documented "absent payload → defaults" contract. Recording it here rather than acting on it.

### Re-verification on the rebased tip

| Command | Result |
| --- | --- |
| `go test -count=1 -tags typia_native_internal ../native/cmd/ttsc-typia` (**the tagged catalog test; no CI workflow runs this tag**) | **ok**, 71.9s |
| `go test -count=1 ../native/cmd/ttsc-typia` | ok, 32.8s |
| `go test -p 2 ../native/...` | exit 0 — 14 packages ok |
| `go test -p 2 ./...` | exit 0 — 17 packages ok |
| website compiler `go test -count=1 -p 2 ./...` | ok |
| `GOOS=js GOARCH=wasm` build — `cmd/ttsc-typia` | exit 0 |
| `GOOS=js GOARCH=wasm` build — `website/compiler/cmd/playground` | exit 0 |
| `pnpm --filter ./tests/test-typia-exact-optional start` | Success — both cases executed |
| `pnpm --filter ./tests/test-typia-cli start` | Success |
| `pnpm --filter ./tests/test-typia-generate start` | `# pass 9  # fail 0  # skipped 1` |
| `pnpm --filter ./tests/test-typia-automated start` | Success — **3,945 executed cases** |
| `npx tsc --noEmit` in `website/` | exit 0 |

Case counts, not exit codes — noted that both the name and path forms of `pnpm --filter` match zero silently and still exit 0.

### Three-owner mutation proofs — re-confirmed on the rebased tip

Still hold, each through its own path:

- `build.go` + `transform.go` reverted to master → **10 failing subtests** (both CLI hosts × all four defect rows, plus their two parents); playground stays green.
- `typia_plugin.go` reverted to master → **10 failing subtests** (both playground verbs × all four defect rows, plus parents); CLI stays green.

The control row (inline on typia's entry) passes in every configuration, reverted or not — the documented shape is untouched, which is why the count is 4 rows and not 5.

### A note on scope, since it bounds how much the rebase could have changed

Between my two rebase bases (`5ec5c81ed9..976bc31f78`) there are **zero changed `.go` files**, and #2147/#2148/#2161 contain no Go source at all. Across the whole set, exactly one Go file moved: #2144's `project_catalog_transform_coverage_test.go`, in `cmd/ttsc-typia` — the single package whose full tagged *and* untagged suites I re-ran on the final tip. So the re-runs above are belt-and-braces rather than the load-bearing evidence, and I ran them anyway rather than argue from that.

Verification ran under heavy contention from four unrelated agents building Go in another repository (peaked at 33 concurrent `go`/`compile`/`link` processes). Everything reported here is green on the final tip regardless; where an earlier run flaked under self-inflicted load I said so rather than re-running quietly.

Issue #2160, opened 2026-07-17
Provider behaviour claims do not belong in @typia/interface doc comments: one is already false, contradicts a sibling comment 18 lines away, and misled two readers in opposite directions


samchon:
## Problem

**`@typia/interface`'s doc comments assert what ChatGPT, Gemini, and other providers do and do not support. Those assertions do not belong there, and at least one of them is already false and actively harmful.**

A provider's behaviour is not typia's contract. typia cannot guarantee it, cannot test it, and cannot update a comment when a provider changes. So the claim is guaranteed to rot — and a rotten claim in a doc comment does not sit quietly. It misleads.

This is not hypothetical. **The false note below misled two independent readers in opposite directions within one hour**, and `ILlmSchema.ts` now carries two mutually contradictory statements about the same keyword.

## Evidence

Verified on `master`.

### The false one

`packages/interface/src/schema/ILlmSchema.ts:378-379`:

```
* Note: ChatGPT and Gemini do not support `additionalProperties`. Use
* {@link ILlmSchema.IConfig.strict} mode for OpenAI compatibility.
```

**This is false.** OpenAI requires `additionalProperties: false` on every object type and enforces it — under `strict` ([function calling guide](https://developers.openai.com/api/docs/guides/function-calling)). Anthropic returns 400 without it. Gemini is path-dependent. The claim is unconditional; the behaviour is not. It is wrong in a way that misleads in **both** directions: read it one way and you conclude no schema can defend against extra properties; read the correction one way and you conclude the model always defends. Neither is true.

Worse, it sits **eighteen lines below** its own contradiction, in the same file:

```
* - `false` (default): No additional properties allowed
* - `true`: Any additional properties allowed
* - Schema: Additional properties must match this schema
```

One comment says the keyword has three-way meaning. The next says two major providers ignore it. Both are on `IObject.additionalProperties`. A reader cannot satisfy both.

### The full inventory

Sixteen provider references across `packages/interface/src`. They are not all the same thing, and the fix must not treat them as such:

**Behavioural claims — the defect.** These assert what a provider does:

| Location | Claim |
| --- | --- |
| `ILlmSchema.ts:378` | "ChatGPT and Gemini do not support `additionalProperties`" — **false** |
| `ILlmSchema.ts:415` | "Gemini does not support `anyOf`/`oneOf` schemas" |
| `IJsonSchemaTransformError.ts:15` | "Gemini does not support `oneOf` schemas" |
[...]

Issue #2157, opened 2026-07-17
llm.* with strict silently drops a numeric default: #2056 was fixed in @typia/utils and left in the Go emitter that typia.llm.* actually calls


samchon:
## Problem

**#2056 is half-fixed.** `typia.llm.*` with `strict` still **silently drops a numeric `default`**. This campaign's own PR #2062 fixed the `@typia/utils` copy of the contract and never touched the Go emitter — the one `typia.llm.*` actually calls.

A `string` default survives as an `@default` tag. A `number` default is deleted outright.

## Evidence

Verified on `master`, running both owners side by side in one process on the same input.

`packages/typia/native/core/programmers/llm/LlmSchemaProgrammer.go:728-741`:

```go
func llmSchemaProgrammer_shift_numeric(schema map[string]any) map[string]any {
  output := llmSchemaProgrammer_clone(schema)
  llmSchemaProgrammer_emend_exclusive(output)
  tags := []string{}
  for _, key := range []string{"minimum", "maximum", "exclusiveMinimum", "exclusiveMaximum", "multipleOf"} {
    if value, ok := output[key]; ok {
      tags = append(tags, "@"+key+" "+fmt.Sprint(value))
      delete(output, key)
    }
  }
  llmSchemaProgrammer_write_tag_with_description(output, tags)
  delete(output, "default")     // <-- deleted, no tag emitted
  return output
}
```

`"default"` is absent from the key list, and then removed outright. Its sibling twelve lines below:

```go
func llmSchemaProgrammer_shift_string(schema map[string]any) map[string]any {
  // ...
  for _, key := range []string{"minLength", "maxLength", "format", "pattern", "contentMediaType", "default"} {
```

`"default"` **is** in the string list, so a string default becomes `@default <value>` and is preserved.

The `@typia/utils` owner is correct in **both** places — `OpenApiConstraintShifter.ts:79-81` and `:133-135`:

```ts
if (schema.default !== undefined) {
  tags.push(`@default ${schema.default}`);
  delete schema.default;
}
```

That is what PR #2062 fixed. The Go emitter was not touched, and it is the one on the `typia.llm.*` path.

### Why nothing caught it

The parity test **already compares both owners correctly.** Its **fixture** simply omits `Default` — the single tag they disagree on. Adding `& tags.Default<7>` turns it red today.

[...]

PR #2162, opened 2026-07-17
fix(utils): gate description-constraint inversion on the mode that wrote the tags

### LINKED ISSUE #2150: Documenting a property disables its validation: LlmSchemaConverter.invert erases 13 constraint keywords from any leaf carrying a description
## Problem

**Documenting a property disables its validation.**

`LlmSchemaConverter.invert` strips **13 constraint keywords** from any integer, number, string, or array leaf that carries a `description`. `LlmJson.validate` then accepts data that violates the very constraints typia advertised to the model. Add a JSDoc comment to a property and its `Minimum`, `Format`, `Pattern`, `MaxLength`, `MinItems` — every tag — stop being enforced.

Affected: every `@typia/utils` consumer that inverts an LLM schema — `LlmJson.validate`, and through it the MCP, Vercel, and LangChain adapters, and the HTTP validation path.

## Evidence

Verified on `master` end to end through a real `typia.llm.controller` + `createMcpServer`, as a single-variable control. Identical types, tags, method, and return value — **only three `/** */` comments differ**:

| | `isError` | `structuredContent` shipped to the model |
| --- | --- | --- |
| with property JSDoc | `false` | `{"id":"u1","age":-5,"email":"not-an-email"}` |
| JSDoc removed | `true` | `undefined` |

`age: -5` violates its `Minimum`. `email: "not-an-email"` violates its `Format<"email">`. Both were advertised to the model and both were accepted.

### The mechanism — three gated writes, three ungated reads

The design is a pair. In **strict** mode, `OpenApiConstraintShifter` *moves* constraints out of the schema and into the `description` as `@minimum 3`-style tags, because OpenAI structured output does not accept those keywords. `invert()` reverses it: `LlmDescriptionInverter` reads the tags back out and restores them as real keywords.

The write half is gated. `LlmSchemaConverter.ts:401`, `:413`, `:422`:

```ts
props.config.strict === true
  ? OpenApiConstraintShifter.shiftNumeric({ ...input })
  : input,
```

The read half is not. `LlmSchemaConverter.ts:569`, `:643`, `:656`:

```ts
union.push({
  ...schema,
  ...LlmDescriptionInverter.numeric(schema.description),
  ...{ enum: undefined },
});
```

**`invert` cannot know.** Its signature takes no config at all — `{ components, schema, $defs }` (`LlmSchemaConverter.ts:512`). So instead of asking whether constraints were shifted, it sniffs `description === undefined`:

```ts
export const numeric = (description: string | undefined): Pick<...> => {
  if (description === undefined) return {};
  // ...
  return OpenApiExclusiveEmender.emend({
    minimum: find({ type: "number", name: "minimum", lines }),
    // ...
  });
};
```

In **non-strict** mode the constraints were never shifted, so the description holds no `@minimum` tag and every `find()` returns `undefined`. The function then returns `{ minimum: undefined, maximum: undefined, ... }` — and the spread **overwrites the real constraints with `undefined`**.

That is why the bug is guarded by the *absence* of documentation. No description, `{}` is returned, the spread is a no-op, the constraints survive. Add a description and they are erased.

The count is exactly the three functions' return shapes: `numeric` 5 (`minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf`), `string` 5 (`format`, `pattern`, `contentMediaType`, `minLength`, `maxLength`), `array` 3 (`minItems`, `maxItems`, `uniqueItems`) — **13**.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/utils/src/converters/LlmSchemaConverter.ts` — the `invert` path |
| **Consumers** | `LlmJson.validate`; the MCP, Vercel, and LangChain adapters; the HTTP validation path |
| **Trigger** | any documented integer/number/string/array leaf, in non-strict mode |
| **Not affected** | `typia.is`/`assert`/`validate`, generated from metadata rather than from an inverted schema; strict mode, where the tags really are in the description |

This is a soundness defect, not a message defect: invalid data is reported valid.

| Case | Expected |
| --- | --- |
| non-strict, documented property with `Minimum<0>`, value `-5` | rejected |
| non-strict, undocumented property with `Minimum<0>`, value `-5` | rejected — unchanged, already correct |
| strict, documented property whose description carries `@minimum 0` | rejected, constraint restored from the tag — unchanged |
| documented property with **no** constraints | description preserved, no constraints invented |
| a description that legitimately contains the literal text `@minimum 3` in non-strict mode | the author's prose is not silently promoted to a constraint |

## Approach

The invariant: **the read is gated exactly where the write is gated. A constraint is restored from a description only when it was moved there.**

`invert` cannot honour that today because it takes no config, so the fix reaches its signature and its callers. Sniffing `description === undefined` is not a substitute for knowing whether the shift happened — it is what produces the inversion between "documented" and "validated".

Whatever the shape of the fix, the spread of an all-`undefined` object over a populated schema is the immediate mechanism and must not survive it: a function that returns `{ minimum: undefined }` to mean "I found nothing" will erase a real `minimum` at every call site that spreads it, including ones added later.

- **Positive:** a documented, constrained property is enforced in non-strict mode; strict mode still restores constraints from their tags.
- **Negative:** this is the case that must fail against the current code — the documented/undocumented pair above, differing only by a comment, must produce the same verdict.
- **Boundary:** all 13 keywords across all three functions; a documented property with no constraints; a description whose prose contains tag-like text.
- **Regression:** `OpenApiConstraintShifter` and the strict round-trip are unchanged — the write half is correct.
- **Commands:** `pnpm --filter ./tests/test-utils start`; `pnpm --filter ./tests/test-utils-automated start`; `pnpm --filter ./tests/test-mcp start`; `pnpm --filter ./tests/test-vercel start`; `pnpm --filter ./tests/test-langchain start`.

## Coordination

- `@typia/utils` only. No native Go, and **no package version bump** — releases and version numbers belong to the maintainer.
- Touches `LlmSchemaConverter.ts`, which #2104 and #2105 both edited. Their contracts must not shift: `$defs` keys stay unrestricted, and component keys stay OpenAPI-legal and allocated once per conversion.
- Independent of #2145, which owns `OpenApiObjectValidator`'s `additionalProperties`, and of #2146.
- Found by discovery round D3-C and reproduced by the lead.

samchon:
Fixes #2150.

## Documenting a property disabled its validation

`LlmSchemaConverter.invert` erased 13 constraint keywords from any integer / number / string / array leaf carrying a `description`, and `LlmJson.validate` then accepted data violating the constraints typia had advertised to the model. This is a soundness defect on the **non-strict default path**. Reproduced exactly as filed:

```
typia.llm.parameters<{ /** How old the member is. */ age: number & Minimum<0> & Maximum<100> & MultipleOf<5> }>()
  →  { type: "number", minimum: 0, maximum: 100, multipleOf: 5, description: "How old the member is." }
  →  invert
  →  { description: "How old the member is.", type: "number" }
```

All three keywords gone. The undocumented twin keeps every one. Same for string and array leaves.

## The mechanism, and the invariant

The design is a pair: under `strict`, `OpenApiConstraintShifter` *moves* constraints out of the schema into the description as `@minimum 3` tags, and `LlmDescriptionInverter` reads them back. **The write was gated on `config.strict === true`; the read was not.** `invert` took no config, so it sniffed `description === undefined` instead. Outside `strict` nothing was ever shifted, every `find()` returned `undefined`, and the returned `{ minimum: undefined, ... }` was spread over the live schema — which is why the defect was guarded by the *absence* of a doc comment.

The invariant: **a constraint is restored from a description only when it was moved there.**

The gate now lives **inside** `LlmDescriptionInverter`, which takes the config that produced the schema. `LlmSchemaConverter` is its only consumer, so no call site — including one added later — can read tags the write half never wrote. `invert` gains an optional `config` defaulting through the existing `getConfig` (non-strict), and `LlmJson` threads it.

The all-`undefined` return is gone independently of the gate: the three readers drop the keys whose lookup found nothing, so a failed lookup can never delete a caller's constraint. `description` stays exempt — rewriting it to drop consumed tag lines is the point, not an accident.

### `typia.llm.application`'s `config.strict` is always `false`

The natural caller-side source of truth is `application.config`. **It is not one for `typia.llm.*`.**

```
typia.llm.application<T, { strict: true }>().config   →  {"strict":false,"validate":null}
  age description                                     →  "How old the member is.\n\n\n@minimum 0"
```

The schema *was* generated in strict mode — the constraints are shifted — while the config beside it says `strict: false`. `strict` is read in exactly two places in the Go tree, both in `LlmSchemaProgrammer`, and nothing writes it into the emitted config. `ILlmApplication.config` is declared to hold "the settings that were applied during schema generation, **including strict mode**", so the runtime contradicts its own declaration. Unfiled; **needs a follow-up issue** the maintainer owns. `HttpLlm.application` composes its config in TypeScript and reports it honestly, so the adapters are exact for HTTP controllers today.

**Consequence:** a `strict` **class** controller in MCP/Vercel does not enforce its declared output constraints, because the config it hands over is false. Shipping anyway: what is fixed is unsound behaviour on the default path with nothing else checking it; what regresses is a server-side self-check on a nearly-dead mode where the shape is already enforced at compile time. Sniffing the schema is the heuristic this issue exists to remove, `{ strict: true }` is a lie in the other direction, and fixing the emission is native Go — excluded here. Threading `application.config` is inert for class controllers today and becomes exact the moment the transform reports truthfully.

### #2157's fix is not sufficient on its own

`shift_numeric` does `delete(output, "default")` with no tag — that is #2157. But `shift_string` **does** write `@default`, and the round trip still loses it:

```
typia.llm.schema<string & Default<"abc"> & MinLength<2>, { strict: true }>
  →  { type: "string", description: "@minLength 2\n@default abc" }
  →  invert  →  { description: "@default abc", type: "string", minLength: 2 }
```

`minLength` restored; `default` lost; the tag leaked into the description as prose. The inverter restores no `default` for either type and `describe()` does not filter it. Fixing #2157 alone gives numeric this same outcome — it needs a read-side counterpart. Out of #2150's 13-keyword scope and overlapping #2157's owner, so not done here, and **no test locks the current behaviour** since that would snapshot the bug.

## Tests

Fail-then-pass on the pre-fix code for every new `test-utils` case, re-verified on the current rebased tree. The decisive pair — two properties differing only by a doc comment — and all 13 keywords across all three readers. With the fix reverted, the documented leaves invert to bare `{description, type}` (age −3, score −2, thumbnail −5, hobbies −3 = 13) and the tests Error; with the fix they pass.

One trap worth recording: **`TestValidator.equals` walks `Object.keys(x)` and skips the keys holding `undefined`.** A first draft passed the erased leaf as `x` and compared nothing at all — green against the fully-present defect. The presence cases now assert one scalar per keyword, expectation first; the prose case deliberately keeps the leaf first, because there the invented keyword is the *present* value.

`test-mcp` and `test-vercel` had zero `strict` coverage, so dropping `controller.application.config` broke nothing. Both new adapter cases are independently caught by that mutation:

| mutation | caught by |
| --- | --- |
| `LlmJson.validate(func.output, true)` in `McpControllerRegistrar` | `strict minimum violation is a tool error` |
| `LlmJson.validate(func.output, true)` in `VercelToolsRegistrar` | `strict minimum violation uses the failure branch` |

## Design note for the maintainer

`LlmJson.validate` now has three optional positionals (`parameters, equals?, config?`) — the optional-positional chain you have objected to before. Kept because converting these two to options bags is a public break outside this issue's scope. Yours to call.

## Scope

`@typia/utils`, `@typia/mcp`, `@typia/vercel`, their tests, and `website/src/content/docs/llm/json.mdx`. No native Go. No version bump. `ILlmSchema.ts` untouched (#2160). No provider behaviour claims in comments, code, or tests (#2165). No repository-wide `pnpm format`.

## Verification

Rebased onto current master `df10f34bae` (#2163 native, #2167 langchain/vercel). Executed case counts, all Success, 0 bugs:

| suite | cases |
| --- | --- |
| test-utils | 301 |
| test-mcp | 20 |
| test-vercel | 24 |
| test-langchain | 17 |
| test-typia-schema | 160 |
| test-utils-automated | 163 |

### DIFF
--- a/packages/mcp/src/internal/McpControllerRegistrar.ts
+++ b/packages/mcp/src/internal/McpControllerRegistrar.ts
@@ -42,6 +42,14 @@ export namespace McpControllerRegistrar {
               // schema. A `strict` application carries its constraints as
               // description tags instead of keywords, and only that config
               // tells the inverter to read them back.
+              //
+              // An HttpLlm application reports the config it was composed with,
+              // so this is exact for it. A typia.llm.controller reports
+              // `strict: false` whatever its Config generic said — a transform
+              // defect, not a reason to guess here: guessing is what erased the
+              // non-strict constraints in the first place. This reads the
+              // declared config and becomes exact for class controllers too the
+              // moment the transform reports it truthfully.
               LlmJson.validate(
                 func.output,
                 true,
--- a/packages/mcp/src/internal/McpControllerRegistrar.ts
+++ b/packages/mcp/src/internal/McpControllerRegistrar.ts
@@ -38,7 +38,11 @@ export namespace McpControllerRegistrar {
         validateOutput:
           func.output === undefined
             ? undefined
-            : LlmJson.validate(func.output, true),
+            : // Validate the output against the config that produced its
+              // schema. A `strict` application carries its constraints as
+              // description tags instead of keywords, and only that config
+              // tells the inverter to read them back.
+              LlmJson.validate(func.output, true, controller.application.config),
       });
     }
 
--- a/packages/mcp/src/internal/McpControllerRegistrar.ts
+++ b/packages/mcp/src/internal/McpControllerRegistrar.ts
@@ -42,7 +42,11 @@ export namespace McpControllerRegistrar {
               // schema. A `strict` application carries its constraints as
               // description tags instead of keywords, and only that config
               // tells the inverter to read them back.
-              LlmJson.validate(func.output, true, controller.application.config),
+              LlmJson.validate(
+                func.output,
+                true,
+                controller.application.config,
+              ),
       });
     }
 
--- a/packages/utils/src/converters/LlmSchemaConverter.ts
+++ b/packages/utils/src/converters/LlmSchemaConverter.ts
@@ -504,18 +504,27 @@ export namespace LlmSchemaConverter {
    * Restores constraint information from description tags and converts `$defs`
    * references to `#/components/schemas`.
    *
+   * Only `strict` conversions move constraints into the description, so only a
+   * `strict` inversion reads them back. Pass the same `config` the schema was
+   * converted with; the default is the same non-strict default {@link getConfig}
+   * applies, under which the constraint keywords are still on the schema and a
+   * description is only ever prose.
+   *
+   * @param props.config Configuration the schema was converted with
    * @param props.components Target components (mutated with definitions)
    * @param props.schema LLM schema to invert
    * @param props.$defs LLM schema definitions
    * @returns OpenAPI JSON schema
    */
   export const invert = (props: {
+    config?: Partial<ILlmSchema.IConfig>;
     components: OpenApi.IComponents;
     schema: ILlmSchema;
     $defs: Record<string, ILlmSchema>;
   }): OpenApi.IJsonSchema =>
     invertInternal({
       ...props,
+      config: getConfig(props.config),
       // One allocation per conversion, seeded with the component names the
       // caller already owns. Recursive inversion reuses it instead of
       // recomputing, so a reference and its stored component never disagree.
@@ -532,6 +541,7 @@ export namespace LlmSchemaConverter {
     });
 
   const invertInternal = (props: {
+    config: ILlmSchema.IConfig;
     components: OpenApi.IComponents;
     schema: ILlmSchema;
     $defs: Record<string, ILlmSchema>;
@@ -556,6 +566,7 @@ export namespace LlmSchemaConverter {
 
     const next = (schema: ILlmSchema): OpenApi.IJsonSchema =>
       invertInternal({
+        config: props.config,
         components: props.components,
         $defs: props.$defs,
         schema,
@@ -566,7 +577,10 @@ export namespace LlmSchemaConverter {
       if (LlmTypeChecker.isArray(schema))
         union.push({
           ...schema,
-          ...LlmDescriptionInverter.array(schema.description),
+          ...LlmDescriptionInverter.array({
+            config: props.config,
+            description: schema.description,
+          }),
           items: next(schema.items),
         });
       else if (LlmTypeChecker.isObject(schema))
@@ -640,7 +654,10 @@ export namespace LlmSchemaConverter {
         else
           union.push({
             ...schema,
-            ...LlmDescriptionInverter.numeric(schema.description),
+            ...LlmDescriptionInverter.numeric({
+              config: props.config,
+              description: schema.description,
+            }),
             ...{ enum: undefined },
           });
       else if (LlmTypeChecker.isString(schema))
@@ -653,7 +670,10 @@ export namespace LlmSchemaConverter {
         else
           union.push({
             ...schema,
-            ...LlmDescriptionInverter.string(schema.description),
+            ...LlmDescriptionInverter.string({
+              config: props.config,
+              description: schema.description,
+            }),
             ...{ enum: undefined },
           });
       else
--- a/packages/utils/src/converters/internal/LlmDescriptionInverter.ts
+++ b/packages/utils/src/converters/internal/LlmDescriptionInverter.ts
@@ -7,10 +7,9 @@ import { OpenApiExclusiveEmender } from "./OpenApiExclusiveEmender";
  * description.
  *
  * The shifter and this reader are one pair, so they share one gate. The shifter
- * only runs under `strict` — the mode whose schema may not carry the constraint
- * keywords itself — and therefore this reader only runs under `strict` too:
- * every function here takes the config that produced the schema and reports
- * nothing outside that mode.
+ * moves constraints into the description only under `strict`, so this reader
+ * reads them back only under `strict` too: every function here takes the config
+ * that produced the schema and reports nothing outside that mode.
  *
  * Without that gate an `@minimum 3` written by a documentation author would be
  * promoted to a constraint the type never declared, and, because these results
--- a/packages/utils/src/converters/internal/LlmDescriptionInverter.ts
+++ b/packages/utils/src/converters/internal/LlmDescriptionInverter.ts
@@ -1,11 +1,27 @@
-import { OpenApi } from "@typia/interface";
+import { ILlmSchema, OpenApi } from "@typia/interface";
 
 import { OpenApiExclusiveEmender } from "./OpenApiExclusiveEmender";
 
+/**
+ * Reader for the constraints `OpenApiConstraintShifter` moved into a
+ * description.
+ *
+ * The shifter and this reader are one pair, so they share one gate. The shifter
+ * only runs under `strict` — the mode whose schema may not carry the constraint
+ * keywords itself — and therefore this reader only runs under `strict` too:
+ * every function here takes the config that produced the schema and reports
+ * nothing outside that mode.
+ *
+ * Without that gate an `@minimum 3` written by a documentation author would be
+ * promoted to a constraint the type never declared, and, because these results
+ * are spread over the live schema, the lookups that found nothing would delete
+ * the real keywords a non-strict schema still carries.
+ */
 export namespace LlmDescriptionInverter {
-  export const numeric = (
-    description: string | undefined,
-  ): Pick<
+  export const numeric = (props: {
+    config: ILlmSchema.IConfig;
+    description: string | undefined;
+  }): Pick<
     OpenApi.IJsonSchema.INumber,
     | "minimum"
     | "maximum"
@@ -14,35 +30,40 @@ export namespace LlmDescriptionInverter {
     | "multipleOf"
     | "description"
   > => {
+    const description: string | undefined = read(props);
     if (description === undefined) return {};
 
     const lines: string[] = description.split("\n");
-    return OpenApiExclusiveEmender.emend({
-      minimum: find({
-        type: "number",
-        name: "minimum",
-        lines,
-      }),
-      maximum: find({
-        type: "number",
-        name: "maximum",
-        lines,
-      }),
-      exclusiveMinimum: find({
-        type: "number",
-        name: "exclusiveMinimum",
-        lines,
-      }),
-      exclusiveMaximum: find({
-        type: "number",
-        name: "exclusiveMaximum",
-        lines,
-      }),
-      multipleOf: find({
-        type: "number",
-        name: "multipleOf",
-        lines,
-      }),
+    return {
+      ...compact(
+        OpenApiExclusiveEmender.emend({
+          minimum: find({
+            type: "number",
+            name: "minimum",
+            lines,
+          }),
+          maximum: find({
+            type: "number",
+            name: "maximum",
+            lines,
+          }),
+          exclusiveMinimum: find({
+            type: "number",
+            name: "exclusiveMinimum",
+            lines,
+          }),
+          exclusiveMaximum: find({
+            type: "number",
+            name: "exclusiveMaximum",
+            lines,
+          }),
+          multipleOf: find({
+            type: "number",
+            name: "multipleOf",
+            lines,
+          }),
+        }),
+      ),
       description: describe(lines, [
         "minimum",
         "maximum",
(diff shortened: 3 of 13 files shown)


samchon:

Full surface: the five source files, the four new `test-utils` cases, `test_llm_schema_parity_invert.ts`, and `website/src/content/docs/llm/json.mdx`.

### 1. My own presence assertions were vacuous — caught before the commit, worth recording

`TestValidator.equals` is asymmetric. `json_equal_to` walks `Object.keys(x).filter((key) => x[key] !== undefined)`, so **keys holding `undefined` in the first argument are skipped entirely**. My first draft of the 13-keyword test passed the erased leaf as `x` and compared `{minimum: undefined, ...}` against `{minimum: 0, ...}` — zero keys walked, vacuous pass **against the unfixed code**. It reported green while the defect was fully present.

Restructured to one scalar assertion per keyword with the expectation first (`equals<X, Y extends X>` makes that type-check, and a top-level `typeof x !== typeof y` comparison is not subject to the key skip). The prose case deliberately keeps the leaf first — there the invented keyword is the *present* value, so first-argument position is what makes it visible. All three files now state the reason in their docblock.

### 2. The registrar wiring is untested — real gap, closing it

`tests/test-mcp` and `tests/test-vercel` contain **zero** `strict` coverage. Reverting `controller.application.config` in either registrar therefore breaks nothing, even though dropping it would silently stop enforcing every strict output constraint. Adding `test_mcp_tool_output_constraint_enforcement`: the same documented, constrained controller served twice — non-strict and strict — asserting a conforming result is accepted and that `Minimum` and `Format<"email">` violations are tool errors in **both** modes. That is the end-to-end witness of the issue's own repro and it pins the wiring.

### 3. Formatting

Five files of mine deviated from Prettier; fixed with a targeted `prettier --write --end-of-line auto` on **only those files**. This is not the repository-wide `pnpm format` the campaign defers to Post-Campaign Cleanup.

Worth recording for whoever runs that cleanup: **plain `prettier --check` flags the entire repository here.** `core.autocrlf=true` gives CRLF working files and Prettier defaults to `endOfLine: "lf"`, so all 22 files in my first check were flagged on line endings alone, including files no one has touched. With `--end-of-line auto` the real signal was 5.

### 4. Website snippet used a type foreign to the page

`json.mdx` speaks `IMember` throughout; my new section introduced `IBbsArticleController` out of nowhere. Rewritten in the page's own vocabulary, with the `app.config` form shown separately.

### 5. #2147's ASI lesson — checked, clean

#2147 recorded a silent downlevel miscompilation from a comment placed in an expression body. Both registrar edits have that shape. Verified the emitted `lib/*.js` and `lib/*.mjs`: the ternary alternative follows its comment correctly in all four outputs, and the ESM bundles carry `props.config` / `controller.application.config`.

Applying 2, 3, and 4, then re-running `pnpm test:packages` and starting a fresh round.

samchon:
## Self-Review round 1 — outcome, and a finding that changes the strict half

All four items applied. One of them turned up something that contradicts part of the issue's Approach, so recording it in full.

### `ILlmApplication.config.strict` is always `false` — the callers cannot supply a truthful config

The issue's Approach says the fix "reaches its signature and its callers", and the obvious caller-side source of truth is `controller.application.config`. **It is not a source of truth.** Probed on this branch:

```
typia.llm.application<T, { strict: true }>().config   →  {"strict":false,"validate":null}
  age description                                     →  "How old the member is.\n\n\n@minimum 0"

typia.llm.application<T>().config                     →  {"strict":false,"validate":null}
  age description                                     →  "How old the member is."

typia.llm.controller<T, { strict: true }>(...).application.config
                                                      →  {"strict":false,"validate":null}
```

The schema **was** generated in strict mode — the constraints are shifted into the description — while the config emitted beside it says `strict: false`. `strict` is read in exactly two places in the whole Go tree, both in `LlmSchemaProgrammer` (`:139`, `:431`), and nothing ever writes it into the emitted config. `ILlmApplication.config` is declared to hold "the settings that were applied during schema generation, **including strict mode**", so the runtime contradicts its own declaration — the runtime is the defect. Not filed as far as I can find.

**Consequence, stated plainly:** with the read gated on `config.strict`, a `strict` MCP/Vercel controller stops enforcing its declared output constraints, because the config it hands over is false. That is a regression this PR introduces, and it is the reason my first draft of the Vercel case failed on the strict half — the test was right and the config was lying.

I am shipping it anyway, and this is the reasoning rather than a shrug:

- The defect being fixed is **unsound** behaviour on the default path — invalid data reported valid for every documented property, with nothing else checking it.
- What regresses is a **server-side self-check** on a mode the lead has established is nearly dead, where the shape is already enforced at compile time and only tag constraints can slip.
- The alternatives are worse. Sniffing the schema is the heuristic this issue exists to remove. Passing `{ strict: true }` is a lie in the other direction. Fixing the config emission is native Go, which this issue excludes and which needs a version bump I must not make.
- Threading `application.config` is **future-correct**: it is inert today and starts working the moment the config emission is fixed, with no further change here. Not threading it would leave strict silently broken even after that fix.

**This needs a follow-up issue the lead owns**, and it is a prerequisite for #2150's own strict acceptance row at the adapter layer.

### The registrar wiring is pinned, and each mutation is caught

`tests/test-mcp` and `tests/test-vercel` had zero `strict` coverage. Both new cases correct `config.strict` on the strict fixture — pinning the plumbing against the config the contract promises, not the one typia emits — and each is independently caught:

| mutation | caught by |
| --- | --- |
| `LlmJson.validate(func.output, true)` in `McpControllerRegistrar` | `strict minimum violation is a tool error` |
| `LlmJson.validate(func.output, true)` in `VercelToolsRegistrar` | `strict minimum violation uses the failure branch` |

test-mcp 20 cases, test-vercel 23 cases, both Success, 0 failures.

### #2157 coordination — its fix is not sufficient on its own

`llmSchemaProgrammer_shift_numeric` does `delete(output, "default")` with no tag, which is #2157. But `shift_string` **does** write `@default`, and the round trip still loses it: the inverter restores no `default` for either type, and `describe()` does not filter `@default`, so the tag survives as literal prose. Measured:

```
typia.llm.schema<string & Default<"abc"> & MinLength<2>, { strict: true }>
  →  { type: "string", description: "@minLength 2\n@default abc" }
  →  invert  →  { description: "@default abc", type: "string", minLength: 2 }
```

`minLength` restored; `default` lost; the tag leaked into the description. Fixing #2157 alone would give numeric exactly this outcome. It needs a read-side counterpart — restore `default`, and add it to `describe()`'s filter list. Out of #2150's declared 13-keyword scope and overlapping #2157's owner, so not done here; **no test locks the current behaviour**, since that would snapshot the bug. My strict round-trip case deliberately covers the 13 and does not touch `default`.

Starting a fresh full round.

samchon:
## Rebase + Self-Review conclusion

Paused mid-verification during a resource crunch, resumed on an idle box. Master had advanced twice while paused, so the branch was rebased forward in two steps and re-verified each time.

### Rebases
- Onto #2165 (`2227ff47ec`) — the provider-behaviour-claim removal.
- Onto current master `df10f34bae` — #2163 (native accessor escaping) + #2167 (langchain/vercel reachability).

Both conflict-free. **#2167 overlap resolved cleanly:** it edits `VercelToolsRegistrar.ts` on the *arguments* path (removes a double ```` ```json ```` fence); my edit is on the *output-validation* path. Different hunks, both present after rebase.

### #2165 rule applied
Scanned the whole branch surface. Two comments explained the shifter by what a schema "may not carry" — provider-rooted. Reworded both to state typia's own mechanic (the shifter moves constraints into the description under `strict`) in `LlmDescriptionInverter.ts` and the website guide. Tests were already provider-neutral.

### Round-2 finding, corrected
The config doc/comments/website first recommended `application.config` unconditionally. That is wrong for class controllers: `typia.llm.application`/`controller` emit `config.strict: false` regardless of the `Config` generic (the transform defect above), while `HttpLlm.application` reports it truthfully. The guidance now distinguishes the two sources, and the registrar comments explain why threading the declared config is correct even while it is inert for class controllers today.

### Install staleness, not a defect
The session's `node_modules` predated #2167's new `@standard-schema/spec` dependency, so `test-langchain` failed to compile until `pnpm install --frozen-lockfile` re-synced it. Reinstalled, then green.

### Fail-then-pass, re-verified on the rebased tree
Reverting the whole fix unit makes the three documented tests Error, and a dump shows the documented leaves invert to bare `{description, type}` — all 13 keywords stripped (age −3, score −2, thumbnail −5, hobbies −3). Restoring passes.

### Final verification on `df10f34bae` (`26cd3c7762`), executed case counts, 0 bugs
test-utils 301 · test-mcp 20 · test-vercel 24 · test-langchain 17 · test-typia-schema 160 · test-utils-automated 163 — all Success.

Config threads through every recursive `invert` path (properties, items, additionalProperties, refs) and all three leaf readers; both mutation tests catch a dropped registrar config. Self-Review terminates clean. Not merging — lead-owned.

Issue #2159, opened 2026-07-17
Type<"int64">/<"uint64"> ranges are not enforced: the bigint path emits literally true, and the number path admits 2^63 because its boundary is an unrepresentable literal every owner copies identically


samchon:
## Problem

**`Type<"int64">` and `Type<"uint64">` do not enforce their ranges.** Two independent mechanisms in one file:

1. **`bigint & Type<"int64">` emits literally `true`** — no range check at all — on the spelling typia's own documentation recommends for 64-bit integers. `2n ** 200n` validates. `bigint & Type<"uint64">` gets a lower bound only, no upper bound.
2. **`number & Type<"int64">` admits one out-of-range value**, because its upper bound is a decimal literal JavaScript cannot represent. **All four owners write the same literal and all four are wrong together**, so every parity check between them passes.

In both cases typia certifies the value, and `protobuf.encode` then flips its sign or truncates it to `0`.

## Evidence

Verified on `master`. Owner: `packages/typia/native/core/factories/NumericRangeFactory.go`.

### 1 — the bigint path has no check

Lines 31-48, the entire function:

```go
func (numericRangeFactoryNamespace) Bigint(t string, input *shimast.Expression, emit ...*shimprinter.EmitContext) *shimast.Node {
  f := nativecontext.EmitFactoryOf(numericRangeFactory_factory, emit...)
  if t == "uint64" {
    return f.NewBinaryExpression(/* BigInt(0) <= input */)
  }
  return f.NewKeywordExpression(shimast.KindTrueKeyword)
}
```

`int64` takes the fall-through and emits `true`. `uint64` gets `0n <= input` and **no upper bound**. Compare the `Number` path twenty lines below, which does it properly:

```go
func numericRangeFactory_number_int64(input *shimast.Expression, emit ...) *shimast.Node {
  return numericRangeFactory_logical_and(
    numericRangeFactory_integer(input, emit...),
    numericRangeFactory_between("-9223372036854775808", "9223372036854775807", emit...)(input),
    emit...,
  )
}
```

So `number & Type<"int64">` is range-checked and `bigint & Type<"int64">` is not — on the type that is *supposed* to be the correct way to express a 64-bit integer.

`2n ** 200n` validates as `int64`, then `protobuf` decodes it as `0`.

### 2 — the number path's boundary does not exist

[...]

Issue #1179, opened 2024-07-25
Invalid JavaScript is generated when using Index Signatures


colinaaa:
### 📝 Summary
Write a short summary of the bug in here.

  - **Typia Version**: 6.5.3
  - **Expected behavior**:  Create a valid JavaScript
  - **Actual behavior**: Create an invalid JavaScript with SyntaxError

When using `typia` with [Index Signatures](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures), invalid JavaScript is generated:

```js
                        "string" === typeof value[0] || $report(_exceptionable, {
                            path: _path + $join(key[0]",
                            expected: "string",
                            value: value[0]
                        })
```

### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDOAzKERwERIqp4DcAUGcAHYwCmUGqAxrXAGIQRwDeZccYVDAAWAZwD8ALh59+cANpYc00TCjUA5gF1p81eqrby-AL7kTFJhCqq4AN1QAbYABMhtaQApqYAK4xpXyoAayoIAHcqAEo4AF4APgRkNAA6AEkANSdXIWBrAB4OCETY2UJUpihadyznNzoAUQBHXydRQs54zyjyMisbCEdaFMcIDU9efgc69wAaMhMesiA

### 💻 Code occuring the bug
```typescript
import typia from "typia";

interface Foo {
  paths?: {
    [from: string]: [string];
  };
}

const validate: (input: unknown) => typia.IValidation<Foo> =
  typia.createValidateEquals<Foo>();

console.log({
  validate,
});
```

samchon:
Uprade to v6.5.5, then be fixed.

colinaaa:
Thank you @samchon. It works like a charm.

PR #2163, opened 2026-07-17
fix(native): escape emitted accessor paths through one owner

### LINKED ISSUE #2158: A double quote in a property key makes typia emit invalid JavaScript while ttsc exits 0 and publishes it; the same escaping defect makes ~standard.validate() throw
## Problem

**A `"` in a property key makes typia emit invalid JavaScript, and `ttsc` exits `0` and publishes the file anyway.** `node --check` then rejects it. The build reports success; the artifact does not parse.

The same escaping defect makes a control character in a key cause `~standard.validate()` to **throw** instead of returning issues. **Fixing only the quote case leaves that one live** — they share one root cause, which is why they are filed together.

## Evidence

Verified on `master`.

`packages/typia/native/core/factories/IdentifierFactory.go:74-80`:

```go
func (identifierFactoryNamespace) Postfix(str string) string {
  if identifierFactory_variable(str) {
    return "\"." + str + "\""
  }
  quoted := strings.ReplaceAll(strconv.Quote(str), "\"", "\\\"")
  return "\"[" + quoted + "]\""
}
```

Running that function verbatim:

| key | `Postfix` output |
| --- | --- |
| `a-b` | `"[\"a-b\"]"` — correct |
| **`a"b`** | **`"[\"a\\"b\"]"`** — broken |
| `a\b` | `"[\"a\\b\"]"` — correct |

**The mechanism, exactly:** `strconv.Quote` already escapes the inner `"` to `\"`. The `ReplaceAll` pass then matches **that same `"`** and prefixes another backslash, producing `\\"`. In the emitted JavaScript `\\` is a literal backslash and the following `"` **terminates the string**. The quote is escaped twice, and the second escape lands on the wrong character.

### A correction to the reported root cause

The discovery round reported this as "escapes quotes twice **and backslashes never**." The second half is **false** — the table above shows `strconv.Quote` escapes `a\b` correctly and the output is valid. Only the double-quote path is broken. The narrower claim is the true one, and any fix should be measured against it rather than against the wider one.

### Blast radius — three call sites

```
programmers/helpers/UnionExplorer.go:242
programmers/iterate/feature_object_entries.go:60
programmers/protobuf/ProtobufEncodeProgrammer.go:926
```

**protobuf is not affected** — it rejects such a type with a diagnostic before reaching the emit. That narrows the claim honestly and was verified.

### Why nothing caught it

`ttsc` exits `0`. The emit is not parsed by anything in the pipeline, so a syntactically invalid artifact is indistinguishable from a valid one at every gate the build has. This is the same family as #2117 (a host that published without consulting diagnostics) — a build step whose success is not evidence that it produced anything usable.

**#1179**, closed in 2024, is the same class and **stays fixed** — verified, not assumed.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/typia/native/core/factories/IdentifierFactory.go:78` |
| **Affected** | any type with a property key containing `"` — the emitted file does not parse, and the build says it succeeded; and any key containing a control character, through `~standard.validate()` |
| **Not affected** | keys that are valid identifiers (the early-return branch); keys with backslashes; the protobuf path, which diagnoses first |

A build that emits unparseable output and reports success is worse than one that fails: the failure surfaces at the consumer, far from the cause.

| Case | Expected |
| --- | --- |
| key containing `"` | valid JavaScript that `node --check` accepts |
| key containing a control character | `~standard.validate()` returns issues, never throws |
| key containing `\` | unchanged — already correct |
| key that is a valid identifier | unchanged — the `.key` fast path |
| protobuf with such a key | unchanged — diagnosed, not emitted |
| any emit | parses, or the build fails |

## Approach

The invariant: **an emitted accessor is a correct JavaScript string literal for every possible key, and an emit that is not valid JavaScript is a build failure, not a published file.**

Escape once, through one owner. The defect is two escaping passes disagreeing about who already ran — a shape this campaign has found repeatedly. `strconv.Quote` is already a complete JavaScript-compatible string escaper for this purpose; the `ReplaceAll` after it is the bug. But **verify that claim across the full byte range** rather than trusting it: Go's `Quote` and JavaScript's string grammar are not the same specification, and the two differ on at least line separators (U+2028/U+2029).

Fix the shared root cause, not the quote symptom — the control-character case is the proof that a quote-only fix is incomplete.

**Then ask the harder question:** should the emit be parsed before it is published? Nothing in the pipeline reads the artifact it just wrote. That is why a 21-month-old escaping bug ships silently, and it is the reason this issue exists rather than the escaping itself. Whether to close that gap is a maintainer decision; state what you decided.

- **Positive:** keys containing `"`, `\`, `'`, backtick, `${`, newline, tab, control characters, lone surrogates, U+2028/U+2029, and non-ASCII all emit output that `node --check` accepts and that reads the right property.
- **Negative — must fail against the current code:** a type with a `"` in a key must fail a test today; a control character in a key must make `~standard.validate()` return issues rather than throw.
- **Boundary:** the identifier fast path is unchanged; a key that is `__proto__`, `constructor`, or `toString` still resolves to the property and not the prototype.
- **Regression:** #1179's case stays fixed. The protobuf diagnostic path is unchanged. Emitted output for ordinary keys is byte-identical — prove it, since this touches every emitted accessor.
- **Structural:** if the emit is not parsed as part of the fix, say why the next invalid emit will be caught.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/factories/...`; `go -C packages/typia/test test -p 2 ../native/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start`; `pnpm --filter ./tests/test-typia-cli start`.

## Note for the implementer — a real hazard

**Git Bash heredocs here collapse backslashes even with a quoted delimiter, and so does `node -e`.** Three probes were invalidated by this during discovery. Any reproduction touching escaping must have its bytes **verified on disk**, not assumed from the command you typed. Write fixtures with a file-writing tool, then read them back and check the bytes.

## Coordination

- Same family as **#2117** (a host publishing without consulting diagnostics) and **#1179** (closed 2024, same escaping class, still fixed).
- Independent of #2150 through #2157, #2145, #2146, and #1887.
- Found by discovery round D3-B. The lead reproduced the mechanism and **narrowed it**: the reported "backslashes are never escaped" half is false.

samchon:
Closes #2158.

## Intent

`IdentifierFactory.Postfix` builds the emitted accessor path (`$input["a-b"]`) as **raw JavaScript source text**, spliced into the emit through `f.NewIdentifier("_path + " + explore.Postfix)`. That text needs **two** levels of escaping — the key becomes a JSON literal, and that literal is embedded in a JavaScript string literal in the emitted file. The current code applies `strconv.Quote` once and then a `strings.ReplaceAll(…, "\"", "\\\"")` pass that only knows about `"`.

That outer pass is not an escaper. It double-escapes the `"` that `Quote` already handled, and it ignores every other character that needs escaping at the outer level.

## What measurement changed about the above

Kept as written because it is the intent this branch started from, but two parts of it did not survive verification, and the thread records the evidence:

- **`strconv.Quote` is not a usable inner escaper either.** It escapes for Go. It writes `\a` for U+0007 and `\U0001d173` for a non-printable astral rune; JavaScript has neither escape and decodes them as the identity escapes `a` and `U`, silently corrupting the key.
- **The two levels escape for different readers.** The key becomes JSON text that a user sees and `JSON.parse` reads; that text is then embedded in JavaScript source. U+2028/U+2029 are the entire difference between the grammars, so they are escaped in the source level only.
- **19 of 39 keys in the character matrix were already broken on `master`, not 2**, including `\` (emits valid JavaScript, then decodes as a backspace — the wrong key) and every control character (makes `~standard.validate()` throw instead of returning issues).

The invariant that settles it: the value of what `Postfix` returns must equal what typia's own runtime helper `typia/lib/internal/_accessExpressionAsString` computes for the same key, since dynamic keys take that runtime path while sole-literal keys are folded at compile time. That helper is `variable(str) ? ".{str}" : "[{JSON.stringify(str)}]"` — the oracle was already in the repo.

## Scope

Native Go, one owner: `packages/typia/native/core/factories/IdentifierFactory.go`, plus three tests and one feature test.

## The emit is still not parsed by the pipeline — deliberate, and here is the residual risk

typia's transform hands ttsc an AST and never sees the emitted text, so "parse what you wrote" belongs to ttsc's emitter and to #2117, not here. It would also treat the symptom: invalid text is expressible because `f.NewIdentifier("<raw source text>")` is a deliberate raw-text hatch used across the native tree for path expressions, regex literals from `metadata_to_pattern`, IIFEs in the compare programmers, and WeakSet/WeakMap recursion slots.

This bug did not survive 21 months because nothing parses the emit. **42 of the 57 `*_transform_test.go` cases already execute their emitted module through node, which parses it.** The gap was fixture coverage: no fixture had a `"` in a property key. This branch adds one, plus an explicit `node --check` on an untouched emit.

**Residual risk, stated plainly:** this change seals the escaping class at its owner — `Postfix` is now total over every scalar value, every surrogate code unit, and every single byte, proven exhaustively rather than by sampling. It does **not** make an invalid emit impossible in general. **A different raw-text splice with no fixture covering it can still ship JavaScript that does not parse behind a green build; the regex-pattern splices are the most plausible next instance.** Closing that properly means removing the raw-text hatch and building real AST nodes — a refactor across the path-building subsystem, since `FeatureProgrammer.Index` and `postfix_of_tuple` splice these strings by stripping a trailing quote. That is out of scope here and is the same shape as #2117.

## Deferred

**No `packages/typia/package.json` version bump.** Native Go source normally owes one under the development skill, but the maintainer owns releases and version numbers, and this campaign's issues state that explicitly.

The Go reserved-word list carries five words the runtime helper lacks (`module`, `package`, `private`, `protected`, `public`), so a static key `module` reports `$input["module"]` where a dynamic one reports `$input.module`. Both parse back to the same key. Cosmetic, different cause, lead is filing it separately — not folded in here.

Campaign CI is suspended per `.agents/skills/issue-campaign/development.md`; every run for each pushed SHA is cancelled at the exact-SHA gate and verification is local. Verification, the character matrix, and the outstanding gates are recorded in the thread.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/escaped_property_key_path_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/escaped_property_key_path_transform_test.go
@@ -1,0 +1,193 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "testing"
+)
+
+// TestEscapedPropertyKeyPathTransform verifies emits with hostile keys parse.
+//
+// Sole-literal property keys are folded into the emit as accessor path text, and
+// no build step reads the file back, so a mis-escaped key shipped as a broken
+// artifact while ttsc still exited 0. A `"` in a key terminated the path string
+// early and the emitted module stopped being JavaScript; a control character
+// stayed raw and the reported path became unparseable. Only running node against
+// a real emit can catch that, which is why this asserts on the transform output
+// rather than on the factory in isolation.
+//
+//  1. Transform a fixture whose keys carry quotes, backslashes, and control
+//     characters.
+//  2. Assert node --check accepts the emitted module as JavaScript.
+//  3. Execute it and assert each reported path is the accessor JSON.stringify
+//     defines for that key, which is what the runtime helper produces for
+//     dynamic keys.
+func TestEscapedPropertyKeyPathTransform(t *testing.T) {
+  project := escapedPropertyKeyPathProject(t)
+  js := escapedPropertyKeyPathTransform(t, project)
+  escapedPropertyKeyPathCheckSyntax(t, project, js)
+  escapedPropertyKeyPathRunRuntimeCases(t, project, js)
+}
+
+func escapedPropertyKeyPathProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "escaped-property-key-path-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(escapedPropertyKeyPathTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(escapedPropertyKeyPathSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func escapedPropertyKeyPathTransform(t *testing.T, project string) string {
+  t.Helper()
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("escaped property key transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  return out
+}
+
+// escapedPropertyKeyPathCheckSyntax runs node against the emit exactly as it
+// would ship. node --check only parses, so the unresolved typia imports in the
+// untouched output do not matter, and that keeps this a check of the real
+// artifact rather than of a rewritten copy.
+func escapedPropertyKeyPathCheckSyntax(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  emitted := filepath.Join(project, "emitted.cjs")
+  if err := os.WriteFile(emitted, []byte(js), 0o644); err != nil {
+    t.Fatalf("write emitted module: %v", err)
+  }
+  if output, err := exec.Command(node, "--check", emitted).CombinedOutput(); err != nil {
+    t.Fatalf("emitted module is not valid JavaScript: %v\n%s\n--- emit ---\n%s", err, output, js)
+  }
+}
+
+func escapedPropertyKeyPathRunRuntimeCases(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  runtimeJS := ttscTypiaTestRewriteCommonJS(t, js)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(runtimeJS), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
+  if err := os.WriteFile(runner, []byte(escapedPropertyKeyPathRuntimeRunner), 0o644); err != nil {
+    t.Fatalf("write runtime runner: %v", err)
+  }
+  cmd := exec.Command(node, runner)
+  cmd.Dir = project
+  if output, err := cmd.CombinedOutput(); err != nil {
+    t.Fatalf("escaped property key runtime cases failed: %v\n%s", err, output)
+  }
+}
+
+const escapedPropertyKeyPathTSConfig = `{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "commonjs",
+    "moduleResolution": "bundler",
+    "ignoreDeprecations": "6.0",
+    "types": ["*"],
+    "esModuleInterop": true,
+    "strict": true,
+    "skipLibCheck": true
+  },
+  "include": ["src"]
+}
+`
+
+const escapedPropertyKeyPathSource = `import typia from "typia";
+
+interface IHostileKeys {
+  "quote\"key": number;
+  "back\\slash": number;
+  "line\nbreak": number;
+  "tab\tkey": number;
+  "bell\u0007key": number;
+  "nul\u0000key": number;
+  "sep\u2028key": number;
+  "plain-key": number;
+  plainIdentifier: number;
+}
+
+export const KEYS: string[] = [
+  "quote\"key",
+  "back\\slash",
+  "line\nbreak",
+  "tab\tkey",
+  "bell\u0007key",
+  "nul\u0000key",
+  "sep\u2028key",
+  "plain-key",
+];
+
+const validateHostile = typia.createValidate<IHostileKeys>();
+
+export const run = () => {
+  const input: Record<string, unknown> = { plainIdentifier: 1 };
+  for (const key of KEYS) input[key] = "not a number";
+  return validateHostile(input);
+};
+`
+
+// The expected path is rebuilt with JSON.stringify rather than written out
+// literally, because that is the contract _accessExpressionAsString implements
+// for dynamic keys and the compile-time fold must agree with it.
+const escapedPropertyKeyPathRuntimeRunner = `const mod = require("./main.cjs");
+const result = mod.run();
+
+if (result.success !== false) {
+  throw new Error("validate accepted non-number values");
+}
+for (const key of mod.KEYS) {
+  const expected = "$input[" + JSON.stringify(key) + "]";
+  const found = result.errors.find((error) => error.path === expected);
+  if (!found) {
+    throw new Error(
+      "missing path for key " + JSON.stringify(key) +
+        ": expected " + JSON.stringify(expected) +
+        ", got " + JSON.stringify(result.errors.map((error) => error.path)),
+    );
+  }
+}
+if (result.errors.some((error) => error.path === "$input.plainIdentifier")) {
+  throw new Error("identifier fast path reported an error for a valid value");
+}
+`
--- a/packages/typia/native/cmd/ttsc-typia/escaped_property_key_path_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/escaped_property_key_path_transform_test.go
@@ -10,12 +10,12 @@ import (
 // TestEscapedPropertyKeyPathTransform verifies emits with hostile keys parse.
 //
 // Sole-literal property keys are folded into the emit as accessor path text, and
-// no build step reads the file back, so a mis-escaped key shipped as a broken
-// artifact while ttsc still exited 0. A `"` in a key terminated the path string
-// early and the emitted module stopped being JavaScript; a control character
-// stayed raw and the reported path became unparseable. Only running node against
-// a real emit can catch that, which is why this asserts on the transform output
-// rather than on the factory in isolation.
+// no build step reads the file back, so a wrongly escaped key shipped as a
+// broken artifact while ttsc still exited 0. A `"` in a key terminated the path
+// string early and the emitted module stopped being JavaScript; a control
+// character stayed raw and the reported path could not be parsed. Only running
+// node against a real emit can catch that, which is why this asserts on the
+// transform output rather than on the factory in isolation.
 //
 //  1. Transform a fixture whose keys carry quotes, backslashes, and control
 //     characters.
--- a/packages/typia/native/core/factories/IdentifierFactory.go
+++ b/packages/typia/native/core/factories/IdentifierFactory.go
@@ -1,8 +1,9 @@
 package factories
 
 import (
-  "strconv"
+  "fmt"
   "strings"
+  "unicode/utf8"
 
   shimast "github.com/microsoft/typescript-go/shim/ast"
   shimprinter "github.com/microsoft/typescript-go/shim/printer"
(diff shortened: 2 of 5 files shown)


samchon:

Full-surface round over the base-to-head diff: `IdentifierFactory.go`, the two factory tests, the `ttsc-typia` transform test, and the `test-typia-schema` feature test.

### 1. The issue's case table is wrong on the backslash, and I changed the fix because of it — ACCEPTED

The issue states `a\b` → `"[\"a\\b\"]"` is **correct** and must stay **unchanged**, and narrows the discovery claim ("backslashes never escaped") as false.

Measured: the emit for `a\b` does parse, so the narrowing is right that it is not a *syntax* failure. But the path it produces is `["a\b"]`, and `_createStandardSchema` hands that to `JSON.parse`, which reads `\b` as **backspace**. The reported key is silently `a<BS>b`, not `a\b`. A key that is only `\` makes the path parser **throw** (`Failed to parse path: "$input[\"\\\"]"`).

So `a\b` is a real defect and its output **must** change. The discovery round's half-claim was closer to the truth than the correction: `strconv.Quote` does escape the backslash — but the accessor needs **two** nesting levels, and only one was escaped, so the backslash is unescaped *at the outer level*.

### 2. `strconv.Quote` is not a complete escaper for this purpose — ACCEPTED

The issue proposes it is ("`strconv.Quote` is already a complete JavaScript-compatible string escaper for this purpose; the `ReplaceAll` after it is the bug") while asking that it be verified. It is not:

| key | master emits | JavaScript decodes | why |
| --- | --- | --- | --- |
| `ab` | `\a` | `aab` | Go has `\a`; JavaScript reads it as the identity escape `a` |
| `a\u{1D173}b` | `\U0001d173` | `aU0001d173b` | JavaScript has no `\U` escape at all |

Both are silent key corruption behind a build that parses. Remediation: the escaper is JSON.stringify's set, written out, with `strconv` dropped entirely.

### 3. U+2028 — the issue's one specific worry, and I got it wrong first — ACCEPTED

The issue warns Go's `Quote` and JavaScript's grammar "differ on at least U+2028/U+2029". Verified as the **only** difference: JSON forbids raw < U+0020 (covering the CR/LF JavaScript also forbids), so the line separators are the entire delta.

My first implementation escaped them at the **inner (JSON) level**. `TestEscapedPropertyKeyPathTransform` caught it: the emitted path became `$input["sep key"]` while `_accessExpressionAsString` produces `$input["sep<raw U+2028>key"]` for the same key as a dynamic key. I introduced a static-vs-dynamic path divergence in a public string.

Remediation applied: the two levels escape for **different readers**. `identifierFactory_json` matches JSON.stringify exactly; `identifierFactory_source` adds the line separators because only the JavaScript parser cares. This also restored `u2028`/`u2029` to **byte-identical** with master (15 → 17 byte-identical cases).

### 4. Byte-identity — VERIFIED, with three representation-only changes disclosed

17 of 39 keys are byte-identical, and they are exactly the ones that were already correct: the identifier fast path, `__proto__`/`constructor`/`toString`, `a-b`, `a'b`, backtick, `${`, non-ASCII, emoji, dotted, bracketed, spaced, empty, digit-first, and both line separators.

19 changed because they were broken. **Three changed that were not broken** — `\x7f`, U+FEFF, U+200D. Master escaped them via `Quote`'s `IsPrint` rule; mine leaves them raw, as JSON.stringify does. The decoded path value and recovered key are **identical**; only the source representation differs, and `node --check` accepts it. Disclosed rather than hidden.

### 5. Downstream splicing — VERIFIED safe

`FeatureProgrammer.Index` and `postfix_of_tuple` strip a trailing `"` and splice. That is safe only if `Postfix`'s last byte is the literal's *unescaped* closing quote. `identifierFactory_quote` writes that byte unconditionally, so the invariant is structural rather than incidental.

### 6. Out of scope, recorded not chased

The Go reserved-word list carries five words the runtime helper lacks (`module`, `package`, `private`, `protected`, `public`), so a static key `module` reports `$input["module"]` where a dynamic one reports `$input.module`. Both parse back to the same key, so it is a cosmetic divergence with a different cause. Not touched here.

Also related but separate, per the issue: `NumericRangeFactory` (#2159).

### Verification

Recorded in a follow-up comment once the suites finish.

samchon:
## Decision: the pipeline does not parse its emit, and this change does not add that

The issue asks the harder question directly, so here is the answer and the evidence behind it.

### What I found

The claim "nothing in the pipeline parses the artifact it just wrote" is true of the **pipeline**. It is not true of the **test suite**: 42 of the 57 `*_transform_test.go` cases in `cmd/ttsc-typia` execute their emitted module through node, which necessarily parses it. A syntactically broken emit fails those tests today.

So this bug did not survive 21 months because nothing parses the emit. It survived because **no fixture had a `"` in a property key**. The gap was fixture coverage, not the absence of parsing.

### Why the emit is not parsed in the pipeline, and why I am leaving it that way

1. **typia does not own the write.** The transform hands ttsc an AST; ttsc prints and writes the file. typia never sees the final text. A post-emit re-parse would have to live in ttsc's emitter as a whole-file cost on every build — a compiler product decision, not a typia bug fix. #2117 is the sibling issue that owns "a build step's success must mean it produced something usable," and that is the right home.

2. **It would treat the symptom.** Invalid text is expressible at all because `f.NewIdentifier("<raw source text>")` is used as a deliberate raw-text escape hatch. It is not a one-off: the native tree uses it for path expressions, regex literals from `metadata_to_pattern`, IIFEs in the compare programmers, and WeakSet/WeakMap recursion slots. The printer trusts every one of them. Re-parsing the output would catch the *next* escaping bug late and noisily; the structural fix is to build real AST nodes so invalid text cannot be expressed. That is a large refactor across the path-building subsystem — `FeatureProgrammer.Index` and `postfix_of_tuple` splice these strings by stripping a trailing quote — and it is far outside this issue's goal.

3. **The class is sealed at the owner instead.** `Postfix` is now total: `TestIdentifierFactoryPostfixSurvivesEveryCodePoint` sweeps every scalar value, every surrogate code unit, and every single byte, and decodes each result with `encoding/json`. No key can make it emit invalid text, so for this cause there is nothing left to catch downstream.

### How the next invalid emit gets caught

- **For this class:** it cannot recur. The owner is total over the entire input space, proven exhaustively rather than by sampling.
- **For the general class:** `TestEscapedPropertyKeyPathTransform` now runs `node --check` against an untouched emit, and the 42 emit-executing transform tests remain the broader net.
- **Honest limitation:** a *different* raw-text splice with no fixture covering it can still ship invalid JavaScript behind a green build. The regex-pattern splices are the most plausible next instance. That residual risk is real, it is the same shape as #2117, and closing it properly means removing the raw-text hatch — not bolting a parse onto the end.

samchon:
## Verification

Campaign CI is suspended, so every gate below is local. Case counts are quoted rather than exit codes.

### Fail-then-pass, at every level

| gate | against `origin/master` | with the fix |
| --- | --- | --- |
| `TestIdentifierFactoryPostfixEscapesEveryKey` | **19 of 37 subtests FAIL** | 37/37 pass |
| `TestIdentifierFactoryPostfixSurvivesEveryCodePoint` | **FAIL** — `Postfix("a\x00b")` is `"[\"a\x00b\"]"`, `invalid character 'x' in string escape code` | pass |
| `TestEscapedPropertyKeyPathTransform` | **FAIL** — see below | pass |

The end-to-end case is the issue's claim reproduced on a real emit that `ttsc` exited `0` for:

```
emitted module is not valid JavaScript: exit status 1
    path: _path + "[\"quote\\"key\"]",
                              ^^^^
SyntaxError: Invalid or unexpected token
```

### The `~standard.validate()` half

Driving the **verbatim** `typiaPathToStandardSchemaPath` parser from `packages/typia/src/internal/_createStandardSchema.ts` against real `Postfix` output:

| key | before | after |
| --- | --- | --- |
| `\n` `\t` `\r` `\v` `\b` `\f` `\0` `\x01` `\x1f` | **THREW** `Bad control character in string literal` | returns the key |
| `\` alone | **THREW** `Failed to parse path: "$input[\"\\\"]"` | returns the key |

10 throwing keys to 0. A quote-only fix would have left every one of these live, exactly as the issue predicted.

### Whole-matrix result

39 keys, each checked for `node --check` acceptance and for round-tripping through the real path parser: **19 broken before, 0 after.**

### Byte-identity for ordinary keys — proven

**17 of 39 keys emit byte-identical output**, and they are exactly the keys that were already correct: the identifier fast path (`field`, `_private`, `$ref`), `__proto__`, `constructor`, `toString`, `a-b`, `a'b`, `` a`b ``, `a${b}c`, `한글`, emoji, `a.b`, `a[0]`, `not valid`, `""`, `0abc`, and both line separators.

The 22 that changed are the 19 that were broken plus three that changed **representation only**, with an identical decoded path and recovered key: `\x7f`, U+FEFF, U+200D. `strconv.Quote` escaped those under its `IsPrint` rule; JSON.stringify leaves them raw, and so does this. `node --check` accepts them.

### Boundaries

- Identifier fast path unchanged, byte for byte.
- `__proto__`, `constructor`, `toString` still emit `.__proto__` etc., reading the property and never the prototype.
- **#1179 stays fixed:** its class is the dotted-vs-bracketed decision in `identifierFactory_variable`, which this change does not touch; the fast path's output is byte-identical.
- Protobuf path untouched — it diagnoses such a key before the emit.
- `FeatureProgrammer.Index` and `postfix_of_tuple` splice by stripping a trailing `"`. `identifierFactory_quote` always writes that closing quote unescaped, so the splice invariant is structural.

### Commands

| command | result |
| --- | --- |
| `go test -count=1 -p 2 ../native/core/factories/...` | **ok** (all packages) |
| `go test -p 2 ../native/core/...` | **exit 0**, 11 test-bearing packages ok — every programmer that consumes `Postfix` |
| `go test -run TestEscapedPropertyKeyPathTransform ../native/cmd/ttsc-typia/` | **ok**, fail-then-pass above |
| `go test -list` on both packages | all three new tests listed — no `_js` filename trap |
| `pnpm --filter ./tests/test-typia-schema start` | **160 cases executed, `Success`**, including `test_standard_schema_escaped_key_paths` |

### Not run, and why

`go test ../native/...` (full tree) died with `fatal error: out of memory allocating heap arena map`, and re-running it serialized with `-p 1` did not complete. The box is shared with the rest of this campaign's agents; RAM dropped to ~8 GB free with `link` alone at 1.6 GB, and D: fell from 45 GB to 22 GB of free space while GOCACHE grew past 31 GB. Environmental, not this change: `../native/core/...` is where every line of this change lives and it passes, and `cmd/ttsc-typia` is covered by the transform test above.

`test-typia-automated`, `test-typia-cli`, and root `pnpm test` are still outstanding for the same reason; results will follow in a comment.

samchon:
## Self-Review round 2 — complete fresh round, no changes applied

Full round over the whole surface at `330850f4e8`: `IdentifierFactory.go`, the two factory tests, the transform test, the schema feature test. Merge base is `976bc31f78`; the change is 5 files, +626/-4, and touches nothing else.

Nothing found that survives verification. Details of what was re-checked:

- **Escaper totality.** `identifierFactory_quote` writes the opening and closing `"` unconditionally and advances `i` on every branch, so it terminates and always returns a closed literal. The four escape paths (`\uXXXX`) only ever receive a value below U+10000 — a control character, a surrogate, or U+FFFD — which is what `\uXXXX` can spell; astral runes take the verbatim path. Documented on the helper, since a future caller passing an astral rune would silently emit a 5-hex-digit escape.
- **WTF-8 detection.** `identifierFactory_surrogate` requires `0xED` with a second byte in `0xA0..0xBF`. Valid three-byte sequences with an `0xED` lead top out at U+D7FF (`ED 80..9F`), so the range cannot catch a legitimate rune. A malformed tail falls through to the `RuneError`/`size == 1` branch and becomes `�`, one byte at a time.
- **Splice invariant.** `FeatureProgrammer.Index` and `postfix_of_tuple` strip `Postfix`'s trailing `"` and continue the literal. That is sound only because the final byte is the unescaped closing quote — structural here, since the builder appends it after the loop. Checked the escaped-quote-at-end case (`a"`) by hand: `"[\"a\\\"\"]"` strips to a valid prefix.
- **Consequence surface.** All three `Postfix` call sites re-read (`UnionExplorer`, `feature_object_entries`, `ProtobufEncodeProgrammer`). The path text flows through assert, validate, is, `json.stringify`, and the plain programmers via `feature_object_entries`, and `go test ../native/core/...` covers every one of those packages green.
- **Test independence.** The expectation helper detects WTF-8 itself rather than calling the factory's helper, so the tests do not restate the code under test. `encoding/json` is the decoder in both factory tests and `node` is the decoder in the transform test — three implementations typia does not own.
- **Reachability.** `go test -list` names all three new Go tests. No `_js` suffix; the trailing segments are `key`, `point`, and `transform`.

### Rounds

Round 1 applied one real fix — the U+2028 level error my own transform test caught — and disclosed the three representation-only byte changes. This round is complete and clean, so review stops here per the review skill's stop rule.

samchon:
## Outstanding gates — `test-typia-automated` could not be run, and I am not claiming it passes

Four attempts, none of which produced a test result. Recording the raw outcomes rather than a conclusion:

| attempt | phase reached | outcome |
| --- | --- | --- |
| 1 | `prestart` printed `Success`, then `start` | exit **1 with zero output** — no case named, no error text |
| 2 | — | exit **127** (`command not found`) on `prestart` |
| 3 | `prestart` | never completed (~40 min) |
| 4 | `prestart` | never completed (~40 min) |

None of these is a test failure. A failing case in this suite names itself; exit 1 with no output and exit 127 are a killed and an unspawnable process. Directly observed in the same window:

- `go test ../native/...` died with `fatal error: out of memory allocating heap arena map`.
- Free RAM fell to ~5 GB of 31 GB with `link.exe` alone at 1.6 GB; 26 node processes from the rest of this campaign's agents were live.
- D: fell from 45 GB to 18 GB free while GOCACHE grew past 31 GB.

**I am not asserting the suite is green.** What I can assert is that the suite's failures never named a case, and that the surface it would cover is covered elsewhere:

- `go test -p 2 ../native/core/...` — exit 0, 11 test-bearing packages, including `programmers`, `programmers/helpers`, `programmers/iterate`, `programmers/json`, and `programmers/plain`. Every consumer of `Postfix` compiles and passes there.
- `pnpm --filter ./tests/test-typia-schema start` — 160 cases executed, `Success`, through the same plugin binary and the same emit path.
- `TestEscapedPropertyKeyPathTransform` — a real emit, `node --check`ed and executed.

`test-typia-cli` and root `pnpm test` are outstanding for the same reason. If the lead wants the automated matrix green before merge, it needs a quieter box; it is the one gate this batch did not clear.

samchon:
## The three outstanding gates now ran — counts, not exit codes

The box was freed of the campaign's other agents and all three completed. Superseding my earlier "could not run" comment.

| gate | executed | result |
| --- | --- | --- |
| `pnpm --filter ./tests/test-typia-automated start` | **3,947 cases** | `# fail 0`, exit **0** |
| `pnpm --filter ./tests/test-typia-cli start` | **3 cases** | 0 failures, exit **0** |
| `pnpm --filter ./tests/test-typia-generate start` | **10 tests: `# pass 9 / # fail 0 / # skipped 1`** | exit **0** |
| `pnpm test` | see below | exit **1** — one contended timeout |

### `pnpm test` = exit 1, and the one failure is a timeout, not this change

Ten workspaces reported `Success` (`test-feature-identity`, `test-mcp`, `test-langchain`, `test-typia-exact-optional`, `test-typia-cli`, `test-typia-schema`, `test-typia-bundler-cache`, `test-utils-automated`, `test-vercel`, `test-utils`). `test-error` rejected every fixture — 50 files, 845 typia diagnostics. `test-interface` completed.

The single failure:

```
not ok 5 - physical output aliases are excluded
  duration_ms: 180042.9533
  error: spawnSync ... ETIMEDOUT
# tests 10 / # pass 8 / # fail 1 / # skipped 1
```

`180042 ms` against a `180000 ms` hard cap — the process was killed at the limit, not failed by an assertion. The test is `output-directory-alias.test.cjs`, which builds symlinked output directories and never reaches an accessor path.

**Re-run in isolation, it passes, and the timing is the proof:**

| | under `pnpm test` | isolated |
| --- | --- | --- |
| that subtest | `not ok`, 180,042 ms (hard cap) | **`ok`, 37,671 ms** |
| suite total | `# fail 1`, 207,789 ms | **`# fail 0`, 56,101 ms** |

The same work is ~5x faster on that test and 3.7x faster suite-wide once nothing else is competing. That is contention, measured rather than asserted.

### Two facts about these runners that this exposed

- **`pnpm test` fail-fast aborts the rest.** `test-typia-automated` printed no outcome under `pnpm test` because `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL` cancelled it when `test-typia-generate` timed out. Silence there means *cancelled*, not *green*. The standalone 3,947-case run above is what actually covers it.
- **`test-typia-automated` is unsafe to run concurrently with itself, and says nothing about it.** An earlier run of mine overlapped another and exited **0** reporting `Success` over **3,825** cases — 122 short of the clean 3,947 — because the two runs were deleting and regenerating the same `src/features` tree. It did not warn, crash, or fail; it silently ran a subset and called it green. That result was discarded and the suite re-run alone against a freshly regenerated tree. Only the count distinguished the two.

PR #2165, opened 2026-07-17
docs(interface,mcp): remove provider behaviour claims from doc comments and fix the mcp README setup

### LINKED ISSUE #2152: @typia/mcp README tells users to run a command that was removed in v13, and omits the devDependencies both sibling adapters list
## Problem

`packages/mcp/README.md`'s setup instructions do not work. They tell the user to run **`npx typia setup`**, a command that **no longer exists** — the typia 13 CLI has exactly one command, `generate`, and anything else prints a usage error and exits `-1`. They also omit the dev dependencies both sibling adapter READMEs list, so following the page leaves a project that cannot compile.

This is the **only** setup instruction the package's README offers.

## Evidence

Verified on `master`.

`packages/mcp/README.md:24-27`:

```bash
npm install @typia/mcp @modelcontextprotocol/sdk typia
npx typia setup
```

`packages/typia/src/executable/typia.ts` — the whole command surface:

```ts
const type: string | undefined = process.argv[2];
if (type === "generate") {
  // ...
  await TypiaGenerateWizard.generate();
} else halt(USAGE);
```

`halt` prints `Wrong command has been detected.` and calls `process.exit(-1)`. `setup` takes the `else`.

Both siblings get it right — `packages/vercel/README.md:20-21` and `packages/langchain/README.md:20-21`:

```bash
npm install @typia/vercel ai typia
npm install -D ttsc typescript@rc
```

`@typia/mcp` lists neither `ttsc` nor `typescript`, and substitutes a wizard that was removed.

**`website/src/content/docs/setup/legacy.mdx` is not affected and must not be changed.** It also runs `npx typia setup`, but deliberately: it documents typia 12 and installs `typia@12` explicitly first, under an "Legacy setup (typia 12 / TypeScript 6)" heading that tells new projects to use typia 13 instead. The wizard existed in v12. That page is correct.

So the defect is exactly one file.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/mcp/README.md` |
| **Affected** | every new `@typia/mcp` user who follows the README — the published npm page shows this text |
| **Not affected** | the code; the other adapter READMEs; `website/src/content/docs/setup/legacy.mdx`, which is correct for what it documents |

Documentation only. Nothing a user installs changes — but the first thing a user *does* fails.

| Case | Expected |
| --- | --- |
| following `@typia/mcp`'s README from an empty project | a project that compiles and runs |
| the dev dependencies it lists | the same set its siblings list, unless `@typia/mcp` genuinely needs a different one |
| `legacy.mdx` | unchanged — it documents v12 and pins `typia@12` |

## Approach

The invariant: **a README's setup instructions produce a working project when followed literally.**

Align with the siblings rather than inventing a third form — three adapter READMEs describing the same toolchain three ways is how this drift started. Check the package's own website guide for what it should say, and check whether `@typia/mcp` really needs anything its siblings do not before adding it.

While here: the reason this went unnoticed is that no check verifies a README's commands exist. Whether that gap is worth closing is a maintainer decision, but it is why a removed command still ships on the npm page.

- **Positive:** the README's setup steps, followed literally in an empty project, produce a working `@typia/mcp` project.
- **Negative:** no instruction references a CLI command the current binary does not implement.
- **Boundary:** `legacy.mdx` still says `npx typia setup` and still pins `typia@12` — do not sweep it in.
- **Regression:** the vercel and langchain READMEs are unchanged; they are the reference.
- **Commands:** `pnpm --filter ./tests/test-mcp start`; the repository spell check — a spell-check failure is real and has caught genuine typos three times this campaign.

## Coordination

- Documentation only. Read `.agents/skills/documentation/SKILL.md` first. No code change, and **no package version bump** — releases and version numbers belong to the maintainer.
- Independent of #2150 and #2151, which touch the same packages' code but not this file.
- Found by discovery round D3-C. The lead reproduced it and **narrowed it**: D3-C reported the command as broken generally; it is broken in exactly one file, and the `legacy.mdx` occurrences are correct.

### LINKED ISSUE #2160: Provider behaviour claims do not belong in @typia/interface doc comments: one is already false, contradicts a sibling comment 18 lines away, and misled two readers in opposite directions

samchon:
Documentation batch: two independent documentation defects, one commit each.

### #2160 — provider behaviour claims leave `@typia/interface` doc comments

A provider's behaviour is not typia's contract: typia cannot guarantee it, cannot test it, and cannot update a comment when a provider changes. The claims were true when written — OpenAI and Gemini genuinely ran non-standard schema subsets that rejected constraint keywords, until roughly last year — and nobody edited them as the world moved. `ILlmSchema.ts:378` is now false and sits eighteen lines below the three-way `additionalProperties` contract it contradicts.

The behavioural claims are removed rather than corrected. A corrected claim is true today exactly as `:378` was true when written, and inherits the same fate while looking more trustworthy. References that name an option's origin stay: `strict` **is** OpenAI's structured output mode, so saying so describes typia, not OpenAI.

Comments only. No declaration in `packages/interface/src` is touched — the `IObject.additionalProperties` three-way contract that #2147 enforces must survive, and `@typia/interface`'s emitted `.d.ts` is byte-identical.

### #2152 — the `@typia/mcp` README's only setup instruction works

`packages/mcp/README.md` told users to run `npx typia setup`, a command the typia 13 CLI does not implement (`generate` is the whole surface; anything else prints a usage error and exits `-1`). It also omitted the devDependencies both sibling adapters list. Aligned with `packages/vercel/README.md` and `packages/langchain/README.md` rather than inventing a third form.

`website/src/content/docs/setup/legacy.mdx` is deliberately **not** touched: it documents typia 12, pins `typia@12`, and the wizard existed in v12.

## Verification

Recorded per commit; campaign CI is suspended, so verification is local. `pnpm format` is not run on this branch per the campaign no-format rule.

Closes #2160
Closes #2152


samchon:
## Self-Review — Round 1

Complete round over the full base-to-head surface (`976bc31f78..75e6590ca9`): 5 files, both commits, plus the consequence surface (emitted declarations, converter behaviour, website guides, sibling READMEs, CI).

### Findings

**1. `ILlmFunction.ts:27-28` prose wrapping deviates from `prettier-plugin-jsdoc`.** My line is 81 columns; the formatter would wrap `and` onto the next line. 3 of the 4 changed interface files match prettier's output exactly — this one does not. Remediation: rewrap that one line to prettier's exact output. This is not the repository-wide formatter (the campaign no-format rule stands); it is one line I authored, kept formatter-stable so Post-Campaign Cleanup produces no diff from this branch.

### Verified sound, no action

- **`prettier --check` fails on this branch — and that is a pre-existing CRLF artifact, not a defect.** Files I never touched (`ILlmApplication.ts`, `IJsonSchemaAttribute.ts`, `IValidation.ts`) fail identically. The Windows checkout is CRLF and prettier defaults to `endOfLine: lf`. My first comparison against pristine copies was invalid — `git show > file` writes LF and the copies sat outside the repo, so they resolved no `prettier.config.js` and skipped the jsdoc plugin. Re-run with line endings normalized and config resolution inside the repo, the result above is the real one.
- **No declaration changed.** 82 of 86 emitted `.d.ts` files are byte-identical; the 4 that differ are exactly the 4 edited, and with JSDoc stripped all 4 declaration sets are identical. `IObject.additionalProperties`' three-way contract survives verbatim, as #2147 requires.
- **`{@link ILlmSchema.IConfig.strict}` in `IJsonSchemaTransformError.ts` is unimported** — consistent with that file's existing `{@link ILlmSchema}` at line 9 and `ILlmFunction`'s existing cross-package `{@link LlmJson.stringify}`. Not a regression.
- **`website/` is entirely untouched**, so `legacy.mdx` is safe by construction.
- **Spell check passed on CI** for the exact head SHA (run 29581862257, `success`), independent of the local `typos` 1.48.0 run.

### Correction to the issue's premise, recorded for the lead

#2160's acceptance asks that the emitted `.d.ts` be *unchanged*. That cannot hold: TypeScript emits doc comments into declaration files, so editing a comment necessarily changes the `.d.ts` bytes. The meaningful invariant — declarations identical, only comments moved — is proven above and is what this branch satisfies.

samchon:
## Self-Review — Rounds 2-5

### Round 2 (after round 1's rewrap)
Clean re-verification. All four changed interface files match `prettier-plugin-jsdoc` output exactly; declarations invariant; spell check clean; `test-mcp` and `test-interface` green.

**Corrected two of my own measurements rather than trusting them:**
- I reported `test-mcp` as **16 cases**. The real count is **19**, in both runs. The 16 came from counting an 18-line `tail` excerpt of the log instead of the log — the exact "never trust a truncated view" failure. Re-counted against the full log.
- A `grep -icE "error|fail"` hit turned out to be the test *name* `test_mcp_class_controller_error_handling`. Zero real failures; `Success` is the authoritative line.

Also confirmed the briefed zero-match trap on this worktree: `pnpm --filter=./tests/nope-does-not-exist -r exec echo hi` prints "No projects matched" and exits **0**. Every suite result here is backed by the package banner and 19 named cases, not an exit code.

### Round 3 — finding, applied
**The guide pointer violated the repository's linking convention.** I had written it as inline prose carrying a bare URL — the only `typia.io/docs` URL in all of `packages/interface/src`. The repository links docs from a doc comment through `@see` (`packages/typia/src/protobuf.ts` does it five times; this file's sibling `ILlmFunction` already uses `@see`). Rewritten as `@see https://typia.io/docs/llm/application`.

### Round 4
Full re-verification after that change: prettier-exact on all four files, `BUILD_EXIT=0`, declarations invariant (82/86 `.d.ts` byte-identical, the other 4 identical with JSDoc stripped), spell clean, `test-interface` green, `test-mcp` 19/19 `Success`.

### Round 5 — complete round, no change to this branch
Verified the README's build line empirically: `ttsx` ships in `.bin` from `npm i -D ttsc`, so "Build with `ttsc`, or run TypeScript directly with `ttsx`" is accurate.

**Two observations outside this batch's declared surface (`packages/interface/src`), reported rather than fixed:**

1. **`packages/typia/src/json.ts:107` may be stale in the same way #2160 describes.** It says `llm.application` "provides provider-specific schemas (ChatGPT, Claude, Gemini, etc.)". `ILlmSchema.IConfig` carries only `strict` — there is no per-provider schema model. This looks like a survivor of the removed per-provider `ILlmSchema<Model>` era: a sentence describing typia's own API that the code moved past. Same rot, different cause (typia changed, not the world). Lead's call.
2. **`packages/typia/src/llm.ts:37,102` and `packages/utils/src/converters/LlmSchemaConverter.ts:36`** name providers as examples and name `strict`'s origin — the legitimate pattern this issue preserves. No action.

**Stop condition met:** round 5 produced no sound improvement to this branch's surface.

samchon:
## Round 6 — lead ruling on `Exclude.ts:12`, applied

The lead reversed the keep-list entry and supplied the sharper test: **not whether a provider is named, but whether the sentence states typia's behaviour or predicts someone else's server.** Split accordingly.

**Removed** — "OpenAI strict mode (and therefore `typia.llm.*` with a strict model configuration) cannot express `not`": a claim about a third party's schema dialect.

**Kept, now standing on its own** — "Strict LLM schemas reject this tag": typia's contract. Verified rather than assumed, per this batch's own lesson that two removed bullets misdescribed our own code:
- `packages/typia/native/core/programmers/llm/LlmSchemaProgrammer.go:175-177` appends `"Strict mode does not support exclude type tag."` when an exclude-kind tag is scanned under strict.
- `llm_schema_programmer_rejects_exclude_tag_in_strict_test.go` pins strict rejecting and non-strict accepting. Executed: `=== RUN` → `--- PASS`. Control: a bogus `-run` filter returns `ok … [no tests to run]`, so the pass is real.

### Verification
- All **5** changed interface files match `prettier-plugin-jsdoc` exactly. The formatter caught my reflow on this file too — same class as round 1, fixed before commit.
- `Exclude.d.ts`: rebuilt from pristine master vs mine — the **only** delta is the comment; declarations **identical** with JSDoc stripped. `TagBase` shape untouched.
- `test-mcp` **19/19 Success**; `test-interface` green; spell clean; `packages/interface` builds.

### Related claim left in place, deliberately
`LlmSchemaProgrammer.go:157-158` carries the same sentence — *"OpenAI strict structured outputs cannot express the `not` keyword"* — as **implementation rationale** in native Go source, not a published doc comment. Two reasons to leave it: it explains why the branch exists at the site of the branch, and editing `packages/typia/native/**` would require a `typia` version bump in the same change, which this batch is forbidden to make. Flagging it as the lead's call.

Provider references remaining in `packages/interface/src` are now exclusively option-origin naming: `ILlmSchema.ts:15,38,42`, `IHttpLlmApplication.ts:23`, `ILlmApplication.ts:15`, `ILlmStructuredOutput.ts:34`.

PR #2166, opened 2026-07-17
fix(native): enforce Type<"int64">/Type<"uint64"> ranges on both the bigint and number paths

samchon:
Implements #2159.

## Intent

`Type<"int64">` and `Type<"uint64">` do not enforce their ranges, through two independent mechanisms:

1. **The bigint path emits literally `true`.** `bigint & Type<"int64">` — the spelling typia's own documentation recommends for 64-bit integers — falls through to a `true` keyword and is entirely unchecked. `bigint & Type<"uint64">` gets `0n <= input` and no upper bound.
2. **The number path's boundary does not exist.** `9223372036854775807` (`2^63 - 1`) is not representable as a JavaScript number; it rounds to `2^63`. So the emitted `input <= 9223372036854775807` is really `input <= 2^63`, and `2^63` — outside int64 — passes.

The invariant: **typia never certifies a value its own encoder will corrupt.** Today typia certifies these values and `protobuf.encode` then flips their sign or truncates them to `0`.

## Scope

Native Go only. Batch scope is `#2159` alone.

## Notes

- **No `packages/typia/package.json` version bump.** The maintainer owns releases and version numbers; the development skill's native-change bump rule is deliberately deferred to them here.
- `pnpm format` is not run on this branch, per the issue campaign's no-format rule. Post-Campaign Cleanup owns the repository-wide formatter result.
- Campaign CI is suspended; every pushed SHA passes the exact-SHA cancellation gate and verification is local.

## Verification

Pending — this is a claim pull request opened before implementation to reserve the batch.

### DIFF
--- a/packages/interface/src/tags/Type.ts
+++ b/packages/interface/src/tags/Type.ts
@@ -69,12 +69,12 @@ export type Type<
               : Value extends "int64"
                 ? {
                     number: `$importInternal("isTypeInt64")($input)`;
-                    bigint: `true`;
+                    bigint: `$importInternal("isTypeInt64Bigint")($input)`;
                   }
                 : Value extends "uint64"
                   ? {
                       number: `$importInternal("isTypeUint64")($input)`;
-                      bigint: `BigInt(0) <= $input`;
+                      bigint: `$importInternal("isTypeUint64Bigint")($input)`;
                     }
                   : Value extends "float"
                     ? `$importInternal("isTypeFloat")($input)`
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -342,9 +342,13 @@ func metadataCommentTagFactory_parse_type(props struct {
   } else if value == "uint32" {
     validate = "Math.floor($input) === $input && 0 <= $input && $input <= 4294967295"
   } else if value == "int64" {
-    validate = "Math.floor($input) === $input && -9223372036854775808 <= $input && $input <= 9223372036854775807"
+    // Delegate to the runtime helper instead of restating the bound. The 64-bit
+    // maxima are not representable as JavaScript numbers, so an inline literal
+    // here silently rounds up and admits an out-of-range value; the helper owns
+    // the one correct spelling and documents why it is exclusive.
+    validate = "$importInternal(\"isTypeInt64\")($input)"
   } else if value == "uint64" {
-    validate = "Math.floor($input) === $input && 0 <= $input && $input <= 18446744073709551615"
+    validate = "$importInternal(\"isTypeUint64\")($input)"
   } else if value == "float" {
     validate = "-3.4028235e38 <= $input && $input <= 3.4028235e38"
   }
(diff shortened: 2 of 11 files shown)


samchon:
## Implemented and verified

Rebased onto `master` `a651378939`; no collision with #2158 (`IdentifierFactory.go`) — separate file in the same directory. Final SHA `c4737c73acb09c3f801e64fa18d3c1ecb8755081`.

### The four owners of the bound literal, fixed together

| Owner | Path | Before | After |
| --- | --- | --- | --- |
| number runtime helper | `_isTypeInt64.ts` / `_isTypeUint64.ts` | `value <= 2 ** 63 - 1` (rounds to `<= 2 ** 63`) | `value < 2 ** 63` exclusive |
| intersection-type emitter directive | `interface/src/tags/Type.ts` `validate` | bigint int64 = `` `true` ``, uint64 = `BigInt(0) <= $input` | delegate to bigint helpers |
| `@type` JSDoc path | `MetadataCommentTagFactory.go` | inline `... <= 9223372036854775807` | delegate to the runtime helper |
| protobuf union discriminator | `NumericRangeFactory.go` | `Bigint` fell through to a bare `true`; number path inclusive-rounding | `Bigint` explicit per type; number path exclusive `< 2 ** 63` |

Two further owners were brought along: the new `_isTypeInt64Bigint.ts` / `_isTypeUint64Bigint.ts` give the bigint path a single exact source of truth, and `RandomProgrammer.go`'s typed-array table had `BigInt64Array` mislabelled `uint64` (a signed array tagged unsigned) — corrected to `int64`. Its range strings are unchanged, so generated data is byte-identical; the table is now documented as the one deliberately-inexact copy (a generation hint, not a validation gate).

### The unrepresentable-boundary decision

`2 ** 63 - 1` and `9223372036854775807` are the **same double** — both round to `2 ** 63`, so no `number` literal can carry the true int64 maximum, and a decimal rewrite rounds identically. **Decision: bound the `number` path exclusively at the power of two (`input < 2 ** 63`, `input < 2 ** 64`).** This is exact, not approximate: IEEE-754 doubles near `2 ** 63` are 1024 apart (2048 near `2 ** 64`), so none exists between the true maximum and the power of two, and the largest representable in-range value (`2 ** 63 - 1024`) is still accepted. The minimum stays inclusive because `-(2 ** 63)` is a power of two, hence exact. The `bigint` path takes the **true inclusive** bound, which a bigint represents exactly, via `BigInt("...")` — a string, because `BigInt(9223372036854775807)` would round its number literal before `BigInt` ever parsed it. Every emitted check is written so the rounding is visible, with the reasoning at each site.

### The oracle

Expectations come from **BigInt comparison** against `-(2n ** 63n)` / `2n ** 63n - 1n` (and the uint64 pair), plus a **protobuf `encode` → `decode` byte-identity** round trip. Both can *represent* the boundary a `number` literal cannot; a spec-derived `number` literal would reproduce the defect inside the test.

### Fail-then-pass (the three negative cases)

Reverting only the defect sources to `master` and running `test-typia-schema`: **161 cases ran**, both new tests error — e.g. `Bug on _isTypeInt64(9223372036854776000) === false` (that value is `2 ** 63`). With the fix restored: **162 cases, Success.** Native Go: `TestNumericRangeFactoryEnforces64BitBounds` fails on `master` (`bigint range int64 emits a bare true keyword`) and passes with the fix. `go test -list` confirms the file is not `_js`-excluded.

Post-fix behaviour (debug probe): `bigint & Type<"int64">` rejects `2n ** 200n` and `2n ** 63n`, accepts `-(2n ** 63n)` and `2n ** 63n - 1n`; `bigint & Type<"uint64">` rejects `2n ** 64n` and `-1n`; `number & Type<"int64">` rejects `2 ** 63`, accepts `2 ** 63 - 1024`. Round trip: `-(2n ** 63n)`, `2n ** 63n - 1n`, `0n`, `±1n`, `2n ** 62n` all decode unchanged with **byte-identical** re-encode.

### The uint32 upper-bound question (#2146 / coordinator)

Not the same root, and not a validation defect. Measured: `typia.is<{ value: number & Type<"uint32"> }>({ value: 4294967296 })` returns **`false`** — validation enforces the uint32 upper bound correctly. The observation is that the emitted **JSON Schema** for `uint32` is `{ "type": "integer", "minimum": 0 }` with no `maximum` keyword (and `int64` is `{ "type": "integer" }`). That is a property of the `schema` field in the off-limits `interface/src/tags/Type.ts` declaration, a JSON-Schema-emission concern in a different layer from my part-1 fall-through (a Go validation emitter). It does not belong in this batch; flagged for the lead as a possible separate JSON-Schema-strictness issue, scope not widened.

### Verification (executed case counts)

| Command | Result |
| --- | --- |
| `GOMAXPROCS=4 go test ... ../native/core/factories/...` | ok, 2 packages |
| `GOMAXPROCS=4 go test -p 2 ../native/...` | ok, 14 packages, 0 failures |
| `test-typia-schema` | 162 cases, Success |
| `test-typia-automated` (solo) | 3947 cases, Success |

Regression: `int32` / `uint32` bounds stay inclusive (exactly representable), `float` unchanged; the #2084 / #2090 / #2119 reader contracts are exercised by the unchanged 64-bit-varint decode test in `test-typia-schema` and did not shift.

No `packages/typia/package.json` version bump — the maintainer owns releases. No `pnpm format`. Not merging — lead-owned.

Issue #2151, opened 2026-07-17
@typia/langchain never runs typia: its validation and coercion are both unreachable, and the arguments path double-fences the feedback the model reads


samchon:
## Problem

**`@typia/langchain`'s two advertised features are both unreachable.** The package promises *"Every tool call is validated by typia. If LLM provides invalid arguments, returns validation error formatted by `LlmJson.stringify` so that LLM can correct them automatically."* Neither the validation nor the coercion ever runs: the tool is registered with a raw JSON Schema, LangChain validates the arguments itself with `@cfworker/json-schema`, and throws before typia's code is reached.

**And typia's LLM feedback is double-fenced on the arguments path** in both `@typia/vercel` and `@typia/langchain`. `LlmJson.stringify` already emits its own ```` ```json ```` fence; the arguments path wraps it in a second one. The model receives ```` ```json\n```json ````.

The two are filed together because **the LangChain half of the second defect is masked by the first** — fixing the unreachability alone surfaces a double fence that nothing tests.

## Evidence

Verified on `master`.

### 1 — LangChain never reaches typia

`packages/langchain/src/internal/LangChainToolsRegistrar.ts`, `createTool`:

```ts
tool(
  async (args: unknown): Promise<unknown> => {
    const valid: IValidation<unknown> = LlmJson.validateArguments(entry.function, args ?? {});
    if (valid.success === false) throw new ToolInputParsingException(/* ... */);
    const result: unknown = await entry.execute(valid.data);
    // ...
  },
  {
    name: entry.name,
    description: entry.function.description ?? "",
    schema: entry.function.parameters,   // <-- a raw JSON Schema
  },
)
```

LangChain's `tool()` validates against a supplied JSON Schema **before** invoking the body. So `LlmJson.validateArguments` never sees invalid arguments, and `valid.data` — the coerced arguments — is never produced.

Byte-identical inputs through all three adapters, as a single-variable control:

| Adapter | `"42"` where a number is expected | Invalid arguments |
| --- | --- | --- |
| `@typia/mcp` | coerced to `42`, returns `47` | typia's feedback |
| `@typia/vercel` | coerced to `42`, returns `47` | typia's feedback |
| **`@typia/langchain`** | **rejected** — `"Received tool input did not match expected schema"` | **LangChain's message, not typia's** |

[...]

PR #2167, opened 2026-07-17
fix(langchain,vercel): make typia's validation reachable in LangChain and fence the LLM feedback once

samchon:
Implements #2151.

## Scope

Two defects that must land together, because the second one's LangChain instance is masked by the first.

1. **`@typia/langchain` never reaches typia.** `LangChainToolsRegistrar.createTool` registers with `schema: entry.function.parameters` — a raw JSON Schema. LangChain's `StructuredTool.call` runs `@cfworker/json-schema` against that schema *before* invoking the tool body, so `LlmJson.validateArguments` never sees invalid arguments and `valid.data` — the coerced arguments — is never produced. Both advertised features of the package are unreachable.
2. **The arguments path double-fences typia's feedback.** `LlmJson.stringify` already emits its own ` ```json ` fence; `VercelToolsRegistrar` and `LangChainToolsRegistrar` each wrap it in a second one on the arguments path. The output path in both is already correct.

## Verification

Local only — campaign CI is suspended and cancelled per exact-SHA gate.

Pending; recorded on this thread as it completes.

### DIFF
--- a/packages/langchain/package.json
+++ b/packages/langchain/package.json
@@ -23,6 +23,7 @@
   },
   "homepage": "https://typia.io",
   "dependencies": {
+    "@standard-schema/spec": "^1.1.0",
     "@typia/interface": "workspace:^",
     "@typia/utils": "workspace:^"
   },
--- a/packages/langchain/package.json
+++ b/packages/langchain/package.json
@@ -27,7 +27,7 @@
     "@typia/utils": "workspace:^"
   },
   "peerDependencies": {
-    "@langchain/core": ">=1.0.0"
+    "@langchain/core": ">=1.1.30"
   },
   "devDependencies": {
     "@langchain/core": "^1.1.31",
--- a/packages/langchain/src/internal/LangChainParameterConverter.ts
+++ b/packages/langchain/src/internal/LangChainParameterConverter.ts
@@ -1,3 +1,4 @@
+import { StandardJSONSchemaV1 } from "@standard-schema/spec";
 import { IHttpLlmFunction, ILlmFunction } from "@typia/interface";
 
 /**
@@ -22,17 +23,18 @@ import { IHttpLlmFunction, ILlmFunction } from "@typia/interface";
  * whose `validate` is likewise left undefined so that `safeValidateTypes`
  * short-circuits its own validation.
  *
- * `toJsonSchema` only reads `~standard.jsonSchema` from `@langchain/core@1.1.30`
- * onwards; older versions return the wrapper itself and would advertise it to the
- * model in place of the parameters. That is what the package's `@langchain/core`
- * peer range is pinned to.
+ * Two version floors hold this together, and both are pinned in the package
+ * manifest. `toJsonSchema` only reads `~standard.jsonSchema` from
+ * `@langchain/core@1.1.30` onwards — older versions return the wrapper itself and
+ * would advertise *that* to the model in place of the parameters — and
+ * `StandardJSONSchemaV1` only exists from `@standard-schema/spec@1.1.0`.
  *
  * @see https://github.com/standard-schema/standard-schema
  */
 export namespace LangChainParameterConverter {
   export const convert = (
     func: ILlmFunction | IHttpLlmFunction,
-  ): IStandardJsonSchema => {
+  ): StandardJSONSchemaV1 => {
     // Coercion never changes the shape a value must end up in, so the input and
     // output schemas are the same document.
     const jsonSchema = (): Record<string, unknown> =>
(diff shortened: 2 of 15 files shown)


samchon:
## Self-Review — Round 1 (opening)

Complete surface under review: the 12 files in `4b9a5bf..13a3a11` (2 adapter sources + 1 new converter, 7 test files, 1 website guide), plus their consequence surface — the emitted `lib/`, the published `.d.ts`, the three adapters' agreement, and the LangChain APIs the fix leans on.

### Verification so far

| Suite | Executed cases | Result |
| --- | --- | --- |
| `tests/test-langchain` | 16 | Success, exit 0 |
| `tests/test-vercel` | 23 | Success, exit 0 |
| `tests/test-mcp` | 19 | Success, exit 0 |
| `tests/test-utils` | 297 | Success, exit 0 |
| `tests/test-feature-identity` | 7 | Success, exit 0 |

`pnpm --filter ./packages/langchain build` and `./packages/vercel build` both exit 0.

### Fail-then-pass

Against `origin/master`'s registrar, the new/replaced tests fail with messages naming the invariant (`feedback comes from typia, not LangChain's JSON Schema validation`; `argument feedback opens exactly one json fence`). Deleting the registrar's typia validation outright also fails 5 tests. On `master`, that same deletion left the suite **green at 13/13** — confirming the issue's "oracles that cannot fail", and finding a third instance beyond the two it named.

### Findings to check in this round

1. Whether `test_langchain_tool_description`'s `typeof schema === "object"` assertion is now weaker than it reads.
2. Whether `@typia/langchain`'s README or `src/index.ts` still describes behavior that changed.
3. Whether the `as unknown as JSONSchema` cast can be narrowed, and whether `@langchain/core/utils/json_schema` is a safe import for a package that declares `@langchain/core` only as a peer dependency.
4. Whether any other repo code hands a raw JSON Schema to a framework that validates it.
5. Whether the HTTP controller path is covered for coercion, not only for validation.

Remediation for anything that survives verification will be pushed as its own commit and recorded here.

samchon:
## Self-Review — closing (rebased onto `2227ff47ec`)

Rebased cleanly onto current master (picked up #2147 and #2165); no conflicts, no changes to `packages/utils`, `packages/interface`, or `packages/mcp`. Final head: **`30a7fccaf5`**.

### Rounds since the draft opened

Five complete rounds ran; each that changed anything was followed by a fresh full pass, ending on one that found nothing:

- **R1** — found and fixed the real shipping risk: `toJsonSchema` only reads a Standard JSON Schema from `@langchain/core@1.1.30` (verified against real installs of 1.0.0/1.1.29/1.1.30/1.1.31 — older return the wrapper object, which would be advertised to the model). Raised the peer floor from `>=1.0.0`. Added the HTTP-controller coercion case.
- **R2** — narrowed a double cast to a single one; corrected a guide sentence that claimed `@typia/mcp` states a no-validator contract (it doesn't — the low-level MCP server just never validates `arguments`).
- **R3** — guarded the langchain fence test's error read; dropped a brittle "fifteen lines apart" detail. **Closed an evidence gap**: proved the *vercel* half of defect 2 fails on master (fence test found 2 fences, expected 1) while the old oracles stayed green.
- **R4** — restored the referenced-numeric negative in the JSON-pointer test that the oracle swap had silently dropped (`count: "42"` is now correctly coerced, so it needed an uncoercible `"not a number"`).
- **R5** — replaced the hand-rolled `IStandardJsonSchema` with `@standard-schema/spec`'s own `StandardJSONSchemaV1`, matching what `packages/typia` already depends on and imports.
- **Post-rebase** — reworded "the provider is sent" → "the model is shown" to keep clear of the absolute no-provider-language rule (#2165). No provider behaviour is asserted anywhere in this change.

### Re-verification on the rebased tree (executed case counts)

| Suite | Cases | Terminal | Exit |
| --- | --- | --- | --- |
| `tests/test-langchain` | 17 | Success | 0 |
| `tests/test-vercel` | 23 | Success | 0 |
| `tests/test-mcp` | 19 | Success | 0 |
| `tests/test-utils` | 297 | Success | 0 |
| `tests/test-feature-identity` | 7 | Success | 0 |

`pnpm --filter ./packages/langchain build` and `./packages/vercel build` both exit 0. Public `lib/index.d.ts` leaks no `@standard-schema` type; the internal converter emit produces `~standard`.

### Byte-identical agreement (re-run post-rebase)

`{x:"42", y:5}` → `{value:47}` and invalid args → typia's annotated feedback, on all three adapters; `@typia/vercel` and `@typia/langchain` outputs are byte-identical (`diff` empty), all three fence exactly once. `@typia/mcp` differs only in its protocol envelope (`structuredContent`, no title), which is its own pre-existing spec contract.

READY for merge (lead-owned).

Issue #563, opened 2023-03-22
Support `@format date` and `@format datetime` tags


samchon:
```typescript
export interface TagFormat {
    /**
     * Universal Unique Identifier.
     * 
     * @format uuid
     */
    uuid: string;

    /**
     * Email address
     *
     * @format email
     */
    email: string;

    /**
     * URL address.
     *
     * @format url
     */
    url: string;

    /**
     * IPv4 address.
     *
     * @format ipv4
     */
    ipv4: string;

    /**
     * IPv6 address.
     *
     * @format ipv6
     */
    ipv6: string;

    /**
     * Date only.
     *
     * @format date
     */
    date: string;

    /**
     * Date and time.
     *
     * @format datetime
     */
    datetime: string;
}
```

PR #2168, opened 2026-07-17
fix(native): align the date-time format aliases and enforce array-form tag exclusivity

### LINKED ISSUE #2153: @format datetime emits format: "date-time" but accepts February 30th: a fourth, unaligned owner of the date-time contract
## Problem

`@format datetime` and `@format dateTime` emit `format: "date-time"` while generating a **different, far weaker validator** than `@format date-time` does. typia's own emitted schema contradicts typia's own generated validator, for a type typia itself accepted.

`@format datetime` **accepts `"2020-02-30T00:00:00Z"`** — February 30th — and **rejects a legal RFC 3339 leap second**. It diverges from `@format date-time` in both directions.

## Evidence

Verified on `master` through a real `ttsx` compile.

`packages/typia/native/core/factories/MetadataCommentTagFactory.go:429-436` builds its format table by copying every `FormatCheatSheet` entry, then explicitly overrides the two aliases:

```go
var metadataCommentTagFactory_FORMATS = func() map[string][2]string {
  output := map[string][2]string{}
  for key, value := range FormatCheatSheet {
    output[key] = [2]string{key, value}
  }
  output["datetime"] = [2]string{"date-time", "!isNaN(new Date($input).getTime())"}
  output["dateTime"] = [2]string{"date-time", "!isNaN(new Date($input).getTime())"}
  return output
}()
```

The first element is the emitted format; the second is the validator. So both aliases emit `format: "date-time"` and validate with `!isNaN(new Date($input).getTime())`.

What `FormatCheatSheet.go:24` actually specifies for `date-time` is a full RFC 3339 validator: an anchored grammar, per-month day limits, leap-year handling for February, and explicit leap-second logic for `:60`.

| Input | `@format date-time` | `@format datetime` | RFC 3339 |
| --- | --- | --- | --- |
| `"2020-02-30T00:00:00Z"` | rejected | **accepted** | invalid — February has no 30th |
| a legal `:60` leap second | accepted | **rejected** | valid |

`new Date("2020-02-30T00:00:00Z")` rolls over to March 1 and is not `NaN`, so the alias accepts it. The cheat sheet's validator checks `day > [31, 28, ...][month - 1] + leap` and rejects it. In the other direction, `Date` cannot parse `:60`, so the alias rejects what RFC 3339 permits and what the cheat sheet explicitly handles.

**This is a fourth owner of the `date-time` contract**, and it is the only one out of alignment.

### Why nothing caught it

D1-A swept for three-owner disagreements on formats and correctly reported zero — it enumerated the 22 `FormatCheatSheet` entries, and **this owner is not one of them**. It is added *after* the copy loop, so it is invisible to any sweep that starts from the cheat sheet. A clean sweep is only as wide as its enumeration.

**#2075's declared purpose was format parity. It edited this exact file and widened the gap.**

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/typia/native/core/factories/MetadataCommentTagFactory.go:433-434` |
| **Affected** | any type using `@format datetime` or `@format dateTime` — every generated validator, and every consumer of the emitted schema, which are now inconsistent with each other |
| **Not affected** | `@format date-time`, the `Format<"date-time">` type tag, and the other 21 cheat-sheet formats |

| Case | Expected |
| --- | --- |
| `@format datetime`, `"2020-02-30T00:00:00Z"` | rejected — matching the `format: "date-time"` it emits |
| `@format datetime`, a legal RFC 3339 leap second | accepted — matching `@format date-time` |
| `@format datetime`, an ordinary valid timestamp | accepted — unchanged |
| `@format date-time` | unchanged in every case — it is the correct owner |
| the emitted schema for either spelling | `format: "date-time"` — unchanged |

## Approach

The invariant: **a format alias is an alias. If it emits `format: "date-time"`, it validates exactly as `date-time` does.**

**Align, do not remove.** `@format datetime` is a *supported* feature: **#563** requested it, was implemented, and was closed and labeled `documentation`. Deleting it would break users who were told it works.

The mechanism is the tell — the aliases are added *after* a loop that copies the cheat sheet. An alias that carries its own validator string is not an alias; it is a second implementation that happens to share a name. Prefer making the alias resolve to the cheat sheet's entry over copying a corrected string into a third place, so a future fix to `date-time` cannot leave the aliases behind again — which is exactly what has already happened once.

- **Positive:** `@format datetime` and `@format dateTime` accept and reject exactly what `@format date-time` accepts and rejects; the emitted schema still says `format: "date-time"`.
- **Negative — these must fail against the current code:** February 30th, and a legal leap second.
- **Boundary:** the full RFC 3339 surface the cheat sheet already covers — month lengths, leap years, offsets, fractional seconds.
- **Regression:** `@format date-time` and the other 21 formats are unchanged. #2075's parity work is not reverted.
- **Structural:** a sweep that enumerates format owners must now find this one. If the fix leaves the aliases outside `FormatCheatSheet`, say why the next sweep will see them.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/factories/...`; `go -C packages/typia/test test -p 2 ../native/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start`.

## Coordination

- Related to **#563** (the alias is supported) and **#2075** (format parity, same file).
- Independent of #2150, #2151, #2152, #2145, #2146, and #1887.
- Found by discovery round D3-E and reproduced by the lead.

### LINKED ISSUE #2154: Every array-form tag exclusivity is inert: the check compares a list of kinds against a tag name, behind a guard that excludes the cross-kind pairs the list exists for
## Problem

**Every array-form tag exclusivity declaration is inert.** `MetadataTypeTagFactory.Validate`'s `[]string` branch can never fire, so all six tags that declare exclusivity as a list — `Minimum`, `Maximum`, `ExclusiveMinimum`, `ExclusiveMaximum`, `Format`, `Pattern` — accept combinations their own published contract forbids.

The result is that typia's emitted schema contradicts typia's generated validator, for a type typia itself accepted. `number & Minimum<5> & Minimum<0>` compiles, emits `{"minimum": 0}`, and `typia.is(3)` returns `false`.

## Evidence

Verified on `master`.

### The contract

`packages/interface/src/tags/Minimum.ts` — the declaration and its doc comment are the specification:

```ts
export type Minimum<Value extends number | bigint> = TagBase<{
  kind: "minimum";
  exclusive: ["minimum", "exclusiveMinimum"];
  // ...
}>;
```

> "This constraint is **mutually exclusive** with {@link ExclusiveMinimum} — you cannot use both on the same property."

`exclusive` holds **kinds**: `"minimum"` (itself, so it cannot be duplicated) and `"exclusiveMinimum"` (the cross-kind partner).

### The check

`packages/typia/native/core/factories/MetadataTypeTagFactory.go:248-256`:

```go
case []string:
  var some *schemametadata.IMetadataTypeTag
  for j, opposite := range props.Tags {
    if i != j && opposite.Kind == tag.Kind && metadataTypeTagFactory_includes(exclusive, opposite.Name) {
```

**Two errors compound, and each alone would be fatal:**

1. **`opposite.Kind == tag.Kind`** restricts the scan to same-kind tags — excluding exactly the cross-kind pairs the list exists to catch. `Minimum` has kind `"minimum"`, `ExclusiveMinimum` has kind `"exclusiveMinimum"`; they never meet.
2. **`metadataTypeTagFactory_includes(exclusive, opposite.Name)`** compares a list of **kinds** against a tag's **Name**. `exclusive` holds `"minimum"`; `opposite.Name` is `"Minimum"`. They never match.

So the same-kind case survives guard 1 and dies on guard 2; the cross-kind case dies on guard 1. **The branch cannot fire for any input.**

The `bool` branch immediately above it works and is unaffected.

### Observed

| Type | Emitted schema | `typia.is` | Contract |
| --- | --- | --- | --- |
| `number & Minimum<5> & Minimum<0>` | `{"minimum": 0}` | `is(3)` → `false` | should not compile |
| `number & Minimum<0> & ExclusiveMinimum<5>` | accepted | — | "you cannot use both" |

The schema says the minimum is 0. The validator requires 5. Both came from the same type.

### The decisive experiment

A one-token change — `opposite.Name` → `opposite.Kind`, and drop the same-kind guard — **rejects exactly the declared pairs and leaves every legal combination compiling.** That separates "the check is inert" from "the tags are mis-declared", which reading alone cannot: it proves the declarations are right and the check is wrong. It is offered as evidence, not as the proposed patch.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/typia/native/core/factories/MetadataTypeTagFactory.go:248-256` |
| **Affected** | all six array-form tags: `Minimum`, `Maximum`, `ExclusiveMinimum`, `ExclusiveMaximum`, `Format`, `Pattern` — a forbidden combination compiles and produces a schema and a validator that disagree |
| **Not affected** | the `bool` branch and every tag using it; any type that does not combine exclusive tags |

Fixing this **turns previously-compiling types into compile errors**. That is the intended effect — the contract always said so — but it is user-visible and belongs in the release notes.

| Case | Expected |
| --- | --- |
| `Minimum<5> & Minimum<0>` | rejected at compile time |
| `Minimum<0> & ExclusiveMinimum<5>` | rejected — the doc comment's exact case |
| `Minimum<0>` alone | compiles — unchanged |
| `Minimum<0> & Maximum<10>` | compiles — not an exclusive pair |
| every legal combination across all six tags | compiles — unchanged |
| the `bool`-form tags | unchanged |

## Approach

The invariant: **a tag's `exclusive` declaration is enforced. It names kinds, and the check compares kinds.**

Both errors must go; fixing either alone leaves the branch inert. And a check that cannot fire is indistinguishable from a check that passes — which is why this survived. Whatever the fix, it needs a test that fails when the branch is reverted to comparing `Name`, and another that fails when the same-kind guard is restored.

Enumerate all six tags' declarations against what the fixed check enforces before landing. The experiment above shows the declarations are consistent today, but that is a fact about today.

- **Positive:** every legal tag combination across all six still compiles; the `bool` branch is untouched.
- **Negative — these must fail against the current code:** `Minimum<5> & Minimum<0>` and `Minimum<0> & ExclusiveMinimum<5>` must both become compile errors, with the message naming the conflict.
- **Boundary:** all six array-form tags, in both the same-kind and cross-kind directions; `bigint` as well as `number`.
- **Mutation:** reverting `Kind` to `Name` must turn a test red; restoring the same-kind guard must turn a different test red. Prove the branch can fail, since its inability to fail is the defect.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/factories/...`; `go -C packages/typia/test test -p 2 ../native/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start`; `pnpm --filter ./tests/test-error start` — this defect's fix produces new compile errors, and that workspace is where deliberately-rejected call sites live.

## Coordination

- Native Go, plus new rejection fixtures. **No `packages/typia/package.json` version bump** — releases and version numbers belong to the maintainer.
- Related to **#2153** (`@format datetime`), which is the same "typia's schema contradicts typia's validator" shape in the sibling factory, but an independent cause and fix.
- Independent of #2150, #2151, #2152, #2145, #2146, and #1887.
- Found by discovery round D3-E and reproduced by the lead.

samchon:
Fixes #2153. Fixes #2154.

Two adjacent but independently caused defects in `packages/typia/native/core/factories/`. They share one sentence — **typia's own emitted schema contradicts typia's own generated validator, for a type typia itself accepted** — and no code. They land as two separate commits so either can be reverted alone.

## #2153 — `@format datetime` accepts February 30th

`MetadataCommentTagFactory` copies every `FormatCheatSheet` entry into its format table, then overrides the `datetime` / `dateTime` aliases with their own validator string. Both aliases emit `format: "date-time"` while validating with `!isNaN(new Date($input).getTime())`, so `"2020-02-30T00:00:00Z"` is accepted (`Date` rolls it to March 1) and a legal RFC 3339 leap second is rejected (`Date` cannot parse `:60`) — the opposite of what `@format date-time` does in both directions.

An alias that carries its own validator string is not an alias; it is a second implementation sharing a name. The fix makes the aliases **resolve to** the cheat sheet rather than copy a corrected string into a third place.

`@format datetime` is a supported feature (#563) and is **aligned, not removed**. #2075's format-parity work is not reverted.

## #2154 — every array-form tag exclusivity is inert

`MetadataTypeTagFactory.Validate`'s `[]string` branch compares a list of **kinds** against a tag's **Name**, behind a same-kind guard that excludes exactly the cross-kind pairs the list exists to catch. No real tag declaration can satisfy both. All six array-form tags — `Minimum`, `Maximum`, `ExclusiveMinimum`, `ExclusiveMaximum`, `Format`, `Pattern` — accept combinations their own published contract forbids.

The `exclusive` declarations in `packages/interface/src/tags` are correct; the Go check is the defect. Fixing it turns previously-compiling types into compile errors — intended, and covered in `tests/test-error`.

## Deferred

No `packages/typia/package.json` version bump. Native Go normally owes one, but releases and version numbers belong to the maintainer.

## Verification

Pending — recorded on this thread before the review-ready hand-off. Campaign CI is suspended per `.agents/skills/issue-campaign/development.md`; verification is local.

### DIFF
--- a/packages/typia/native/core/factories/FormatCheatSheet.go
+++ b/packages/typia/native/core/factories/FormatCheatSheet.go
@@ -4,6 +4,39 @@ func formatCheatSheet_RegexCall(text string) string {
   return text + ".test($input)"
 }
 
+// formatCheatSheet_ALIASES maps an alternative `@format` spelling onto the
+// canonical FormatCheatSheet key it means.
+//
+// An alias owns no validator. It resolves to the cheat sheet entry through
+// formatCheatSheet_resolve, so FormatCheatSheet stays the single owner of every
+// format's validator and a correction to one reaches every spelling of it. The
+// aliases previously carried validator strings of their own, which made each of
+// them a second implementation sharing a name and free to drift: `datetime`
+// emitted `format: "date-time"` while validating with `new Date`, so it accepted
+// February 30th and rejected the RFC 3339 leap second the cheat sheet handles.
+//
+// Aliases live beside the cheat sheet so that enumerating this file enumerates
+// every owner of the `@format` contract.
+var formatCheatSheet_ALIASES = map[string]string{
+  "datetime": "date-time",
+  "dateTime": "date-time",
+}
+
+// formatCheatSheet_resolve resolves a written `@format` value to the canonical
+// format name and the validator FormatCheatSheet owns for it, reporting whether
+// the value names a supported format at all.
+func formatCheatSheet_resolve(value string) (string, string, bool) {
+  name := value
+  if canonical, ok := formatCheatSheet_ALIASES[value]; ok {
+    name = canonical
+  }
+  validate, ok := FormatCheatSheet[name]
+  if ok == false {
+    return "", "", false
+  }
+  return name, validate, true
+}
+
 var FormatCheatSheet = map[string]string{
   "byte":                  `/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test($input)`,
   "password":              "true",
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -274,11 +274,11 @@ var metadataCommentTagFactory_PARSER = map[string]metadataCommentTagFactory_pars
     Report func(msg string) any
     Value  string
   }) metadataCommentTagFactory_TagRecord {
-    matched, ok := metadataCommentTagFactory_FORMATS[props.Value]
+    name, validate, ok := formatCheatSheet_resolve(props.Value)
     if ok == false {
       return metadataCommentTagFactory_TagRecord{}
     }
-    return metadataCommentTagFactory_TagRecord{"string": {{Name: "Format<" + strconv.Quote(matched[0]) + ">", Target: "string", Kind: "format", Value: matched[0], Validate: matched[1], Exclusive: true, Schema: map[string]any{"format": matched[0]}}}}
+    return metadataCommentTagFactory_TagRecord{"string": {{Name: "Format<" + strconv.Quote(name) + ">", Target: "string", Kind: "format", Value: name, Validate: validate, Exclusive: true, Schema: map[string]any{"format": name}}}}
   },
   "pattern": func(props struct {
     Report func(msg string) any
@@ -436,16 +436,6 @@ func metadataCommentTagFactory_parse_integer(props struct {
   value := int(parsed)
   return &value
 }
-
-var metadataCommentTagFactory_FORMATS = func() map[string][2]string {
-  output := map[string][2]string{}
-  for key, value := range FormatCheatSheet {
-    output[key] = [2]string{key, value}
-  }
-  output["datetime"] = [2]string{"date-time", "!isNaN(new Date($input).getTime())"}
-  output["dateTime"] = [2]string{"date-time", "!isNaN(new Date($input).getTime())"}
-  return output
-}()
 
 func metadataCommentTagFactory_includes(values []string, target string) bool {
   for _, value := range values {
--- a/packages/typia/native/core/factories/MetadataTypeTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataTypeTagFactory.go
@@ -246,9 +246,21 @@ func (metadataTypeTagFactoryNamespace) Validate(props struct {
         }{Property: nil, Message: "kind '" + tag.Kind + "' can't be duplicated"})
       }
     case []string:
+      // The `exclusive` list names every kind this tag may not share a property
+      // with, itself included, and it is the authoritative spec. Reject when any
+      // other tag's kind appears in the list, matching kinds against kinds.
+      //
+      // Two bugs made this inert: it compared the list of kinds against
+      // `opposite.Name` ("minimum" against "Minimum<0>"), which never matched,
+      // and it also required `opposite.Kind == tag.Kind`, which excluded the
+      // cross-kind partners the list exists to catch. Both must go, or a
+      // forbidden combination keeps compiling: `Minimum<5> & Minimum<0>` emits
+      // `{"minimum":0}` yet its validator requires 5, and `Minimum<0> &
+      // ExclusiveMinimum<5>` or `Format<"uuid"> & Pattern<...>` coexist though
+      // the declarations forbid it.
       var some *schemametadata.IMetadataTypeTag
       for j, opposite := range props.Tags {
-        if i != j && opposite.Kind == tag.Kind && metadataTypeTagFactory_includes(exclusive, opposite.Name) {
+        if i != j && metadataTypeTagFactory_includes(exclusive, opposite.Kind) {
           copied := opposite
           some = &copied
           break
--- a/packages/typia/native/core/factories/factory_utility_coverage_test.go
+++ b/packages/typia/native/core/factories/factory_utility_coverage_test.go
@@ -499,8 +499,14 @@ export let inner!: Outer.Inner;
 		Report: typeTagReport,
 		Type:   "string",
 		Tags: []schemametadata.IMetadataTypeTag{
-			{Target: "string", Kind: "range", Name: "minimum", Exclusive: []string{"maximum"}},
-			{Target: "string", Kind: "range", Name: "maximum", Exclusive: []string{"minimum"}},
+			// Format and Pattern as they are actually declared: both list the
+			// kinds `["format", "pattern"]`, so intersecting them on one property
+			// is forbidden. The cross-kind conflict fires on the opposite's kind,
+			// not its name, and without a same-kind guard. The previous fixture
+			// named the opposite in `exclusive` by tag name, the only shape the
+			// doubly-broken check could fire on and no real declaration produces.
+			{Target: "string", Kind: "format", Name: "Format<\"uuid\">", Exclusive: []string{"format", "pattern"}},
+			{Target: "string", Kind: "pattern", Name: "Pattern<\"^x$\">", Exclusive: []string{"format", "pattern"}},
 		},
 	})
 	invalidExclusiveReports := []string{}
--- a/packages/typia/native/core/factories/format_alias_resolves_to_cheat_sheet_test.go
+++ b/packages/typia/native/core/factories/format_alias_resolves_to_cheat_sheet_test.go
@@ -1,0 +1,57 @@
+package factories
+
+import (
+  "testing"
+
+  schemametadata "github.com/samchon/typia/packages/typia/native/core/schemas/metadata"
+)
+
+// TestFormatAliasResolvesToCheatSheet verifies every `@format` alias generates the cheat sheet validator it emits.
+//
+// An alias that carries its own validator string is not an alias; it is a
+// second implementation sharing a name, free to drift from the format it
+// claims. `@format datetime` did exactly that: it emitted `format: "date-time"`
+// while validating with `!isNaN(new Date($input).getTime())`, so it accepted
+// February 30th and rejected the RFC 3339 leap second the cheat sheet handles.
+// This pins the alias to the cheat sheet as the single owner of the validator.
+//
+// 1. Parse each alias spelling of the `date-time` format as a JSDoc comment tag.
+// 2. Require the emitted format name, schema, and tag name to be the canonical one.
+// 3. Require the generated validator to be the cheat sheet entry verbatim.
+func TestFormatAliasResolvesToCheatSheet(t *testing.T) {
+  canonical := "date-time"
+  expected, ok := FormatCheatSheet[canonical]
+  if ok == false {
+    t.Fatalf("the cheat sheet must own the %s format", canonical)
+  }
+  for _, alias := range []string{"datetime", "dateTime", "date-time"} {
+    record := metadataCommentTagFactory_parse(struct {
+      Report func(msg string) any
+      Tag    schemametadata.IJsDocTagInfo
+    }{
+      Report: func(msg string) any { return nil },
+      Tag: schemametadata.IJsDocTagInfo{
+        Name: "format",
+        Text: []schemametadata.IJsDocTagInfo_IText{{Text: alias}},
+      },
+    })
+    tags := record["string"]
+    if len(tags) != 1 {
+      t.Fatalf("@format %s should generate exactly one string tag, generated %d", alias, len(tags))
+    }
+    tag := tags[0]
+    if tag.Value != canonical {
+      t.Fatalf("@format %s should carry the canonical %s value, carried %v", alias, canonical, tag.Value)
+    }
+    if tag.Name != "Format<\""+canonical+"\">" {
+      t.Fatalf("@format %s should name the canonical format tag, named %s", alias, tag.Name)
+    }
+    schema, ok := tag.Schema.(map[string]any)
+    if ok == false || schema["format"] != canonical {
+      t.Fatalf("@format %s should emit format %s, emitted %v", alias, canonical, tag.Schema)
+    }
+    if tag.Validate != expected {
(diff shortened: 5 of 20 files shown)


samchon:
## Self-Review round 1 — scope correction on #2154 (Format & Pattern must compose)

**Finding.** Issue #2154's stated fix — drop the same-kind guard so cross-kind "mutually exclusive" pairs (Format & Pattern, Minimum & ExclusiveMinimum) are rejected — does not survive verification. Building it turned four committed tests into compile errors:

- `tests/test-typia-schema/src/features/llm.schema/test_llm_schema_spec_strict_string.ts` (since #1830)
- `tests/test-typia-schema/src/features/llm.schema/test_llm_schema_parity_invert.ts`
- `tests/test-utils/src/features/llm/invert/test_llm_invert_documented_constraints_survive.ts` (added in **#2162**, the commit this branch is rebased onto)
- `tests/test-utils/src/features/llm/invert/test_llm_invert_strict_constraints_restored.ts` (added in **#2162**)

Each intersects `Format<...> & Pattern<...>` (and #2162's use `Format<"uri"> & Pattern<"^https://">`) on one property and asserts **both** `format` and `pattern` survive into the schema. Format+Pattern coexistence is intended, tested behavior, reaffirmed in a PR merged this session.

**Why the runtime, not the declaration.** The declarations in `packages/interface/src/tags` are untouched and correct. The batch's invariant is "emitted schema contradicts generated validator," and that contradiction is exactly the **same-kind** case: two tags of one kind write one schema slot, the schema keeps the last value, the validator ANDs all, so they disagree (`Minimum<5> & Minimum<0>` emits `{"minimum":0}` yet rejects 3). Two **different** kinds write different slots (`minimum` beside `exclusiveMinimum`, `format` beside `pattern`) and compose into one consistent schema and validator — no contradiction, so they stay legal.

Master's `[]string` branch was inert (it compared the kind list against `opposite.Name`), so every combination compiled. The corrected fix changes exactly one behavior: same-kind array-form duplication now errors. Cross-kind pairs — including Minimum+ExclusiveMinimum — compiled on master and still compile, unchanged.

**Fix.** Compare kinds (`opposite.Name` → `opposite.Kind`) and keep the same-kind guard; the guard is load-bearing. This rejects the same-kind duplicates that produce the contradiction and leaves cross-kind composition legal.

**Mutation proof (both directions, different tests).**
- Revert the kind comparison to `Name` → `TestMetadataTypeTagFactoryRejectsDuplicatedArrayFormTags` red (the kind comparison is load-bearing for the duplication rejection).
- Drop the same-kind guard — the issue's prescribed experiment → `TestMetadataTypeTagFactoryAllowsCrossKindExclusiveTags` red (the guard is load-bearing for composition). This is the direct demonstration that the issue's "decisive experiment" over-rejects.

**Test changes.** Replaced the cross-kind-rejection unit test with a cross-kind-**allow** test; kept the duplication rejection test; in `tests/test-error` removed the two cross-kind fixtures (now legal) and added a `Format & Format` same-kind fixture beside `Minimum & Minimum`; added `test_tag_format_pattern_compose` to `test-typia-schema` locking the plain-json.schema Format+Pattern path.

**Recommendation for the lead:** narrow issue #2154 to same-kind duplication and drop its cross-kind acceptance rows. Marking this batch **not** ready-to-merge pending lead validation of the narrowed scope.

### Local verification (executed counts)
| suite | result |
| --- | --- |
| `go test ../native/core/factories/...` | ok, 11 tests |
| `go test ../native/...` | 14 ok, 0 fail |
| `go test -tags typia_native_internal ../native/...` | 28 ok, 0 fail |
| `test-typia-schema` | 162 tests, Success (both formerly-failing tests green) |
| `test-utils` | 301 tests, Success (#2162 invert tests green) |
| `test-typia-automated` (solo) | 3491 tests, Success |
| `test-error` | 52 fixtures rejected, 847 diagnostics; same-kind fixtures name the conflict |
| `test-feature-identity` | Success |

#2153 (format alias) verified independently: February 30th rejected, legal leap second accepted, all 22 formats unchanged, `format: "date-time"` still emitted. Campaign CI suspended per `.agents/skills/issue-campaign/development.md`; verification is local. No `packages/typia/package.json` bump — releases belong to the maintainer.

samchon:
## Rework of #2154 per maintainer ruling — cross-kind exclusive pairs are compile errors

The maintainer confirmed the schema is authoritative: `Format & Pattern` and `Minimum & ExclusiveMinimum` on one property MUST be compile errors. My earlier "same-kind only" narrowing was wrong; the tests combining those pairs encode forbidden types and are the defect. This reverts to issue #2154's original prescription.

### The fix
`MetadataTypeTagFactory.Validate`'s `[]string` branch now compares kinds against kinds and drops the same-kind guard, so a tag whose kind appears in another tag's `exclusive` list is rejected regardless of whether the kinds match. Both bugs the issue named are fixed together.

Six-declaration enumeration (each honoured in both directions):

| tag | kind | exclusive | rejects |
| --- | --- | --- | --- |
| Minimum | minimum | ["minimum","exclusiveMinimum"] | Minimum, ExclusiveMinimum |
| Maximum | maximum | ["maximum","exclusiveMaximum"] | Maximum, ExclusiveMaximum |
| ExclusiveMinimum | exclusiveMinimum | ["exclusiveMinimum","minimum"] | ExclusiveMinimum, Minimum |
| ExclusiveMaximum | exclusiveMaximum | ["exclusiveMaximum","maximum"] | ExclusiveMaximum, Maximum |
| Format | format | ["format","pattern"] | Format, Pattern |
| Pattern | pattern | ["format","pattern"] | Pattern, Format |

### Mutation proof (three-way)
- Revert the kind comparison to `Name` → **both** `RejectsDuplicatedArrayFormTags` and `RejectsCrossKindExclusiveTags` red (the kind comparison is load-bearing for the whole branch).
- Restore the same-kind guard → **only** `RejectsCrossKindExclusiveTags` red (dropping the guard is load-bearing for cross-kind; same-kind duplication is still caught).
- `tests/test-error` fail-then-pass for the cross-kind half: `Format & Pattern` and `Minimum & ExclusiveMinimum` compiled before the fix and are now rejected, each naming the conflict.

### Blast radius — I re-swept the whole repo
My scan found exactly **4** files with a compiled forbidden intersection, all Format & Pattern; split each so `format` and `pattern` live on separate leaves, preserving every assertion:
- `test_llm_schema_spec_strict_string.ts` — two strict strings (format-side, pattern-side).
- `test_llm_schema_parity_invert.ts` — two strict round-trips.
- `test_llm_invert_documented_constraints_survive.ts` — pattern moved to a `homepage` leaf; all 13 keyword assertions kept.
- `test_llm_invert_strict_constraints_restored.ts` — same split, plus its description assertion.

Two files the loose grep flagged are **false positives** and correctly need no change — the tag names appear only inside string literals, never as compiled types:
- `test_llm_stringify_multiple_errors_same_path.ts` — `Format`/`Pattern` are `expected` strings in an `IValidation.IFailure`, on separate cases.
- `test_openapi_naming_numeric_constraints.ts` (#2169) — `Minimum`/`ExclusiveMinimum` are `expected` string literals of a naming function's output; the `schema` field is raw `{minimum, exclusiveMinimum}`.

### Inherited failure, NOT from this batch
`test-utils` has 2 failures — `test_http_migrate_v30` and `test_document_upgrade_v30` — both a `Type<"int64">` range rejection of `examples/v3.0/openai.json`'s `seed.maximum: 9223372036854776000` (int64-max rounded up by float64). This is #2166's new int64 enforcement rejecting a committed fixture value; the fixture is byte-identical to master and untouched here, my changes never touch the int64 path, and the same suite passed all 301 tests on my pre-#2166 base. It is present on master and belongs to the NumericRangeFactory / #2166 line of work, not this batch. Flagging for the maintainer.

### Local verification (executed counts, GOMAXPROCS=4)
- `go test ../native/core/factories/...`: ok, 11 tests · `../native/...`: 14 ok, 0 fail · `-tags typia_native_internal`: green
- `test-typia-schema`: 163 tests, Success (both split tests green)
- `test-utils`: 300/302 pass; the 2 failures are the inherited #2166 int64 ones above
- `test-typia-automated` (solo): _recorded below_
- `test-error`: 54 fixtures rejected; 4 tag fixtures name the conflict

No `packages/typia/package.json` bump. Campaign CI suspended; verification is local.

Issue #2156, opened 2026-07-17
json.schema/schemas/application with "3.0" emit a 3.1 document with the version string rewritten: the Go port dropped the downgrader


samchon:
## Problem

`typia.json.schema/schemas/application<T, "3.0">()` emit a **3.1 document labelled `"3.0"`**. The `"3.0"` path builds the 3.1 document and rewrites the version string. Nothing else happens.

This is a **regression from the TypeScript implementation**, which downgraded. The Go port kept `writeV3_0`'s skeleton and dropped the downgrade.

## Evidence

Verified on `master`.

`packages/typia/native/core/programmers/json/JsonSchemasProgrammer.go:105-112` — the whole function:

```go
func jsonSchemasProgrammer_writeV3_0(metadataList []*nativemetadata.MetadataSchema) JsonSchemasProgrammer_IJsonSchemaCollection {
  collection := jsonSchemasProgrammer_writeV3_1(metadataList)
  return JsonSchemasProgrammer_IJsonSchemaCollection{
    Version:    "3.0",
    Components: collection.Components,
    Schemas:    collection.Schemas,
  }
}
```

It calls `writeV3_1` and relabels. There is **no downgrader anywhere in the native tree** — `OpenApiV3Downgrader.ts` exists only under `packages/utils/src/converters/internal/`, on the TypeScript side that the port replaced.

The result violates all three differences typia's own `OpenApiV3.ts` documentation lists between the dialects: `nullable`, the absence of `const`, and the absence of `prefixItems`.

The proof used is typia's **own `validateEquals` rejecting typia's own output**, with a passing 3.1 control — not an external validator's opinion.

### Why nothing caught it

There is **zero 3.0-dialect coverage in the repository.** The single file that touches `"3.0"` asserts only dialect-independent facts, so it passes against a 3.1 document. The dialect is a documented option with no test that knows what the dialect means.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/typia/native/core/programmers/json/JsonSchemasProgrammer.go:105` and its `schema`/`application` siblings |
| **Affected** | every consumer passing `"3.0"` — they receive a document that declares 3.0 and is not 3.0, and any conformant 3.0 tool will reject or misread it |
| **Not affected** | the `"3.1"` path, which is correct and is the default |

| Case | Expected |
| --- | --- |
| `"3.0"`, a nullable type | `nullable: true` — not a 3.1 type union |
| `"3.0"`, a literal type | no `const` |
[...]

PR #2172, opened 2026-07-17
docs(typia): remove removed-API and provider-behaviour claims from packages/typia/src doc comments

### LINKED ISSUE #2171: json.application points readers at "provider-specific schemas" that llm.application has never had: a comment describing an API typia itself removed
## Problem

`packages/typia/src/json.ts:107` tells the reader that `llm.application` **"provides provider-specific schemas (ChatGPT, Claude, Gemini, etc.)"**. It does not. There is no per-provider model — `llm.application`'s only schema knob is `ILlmSchema.IConfig.strict`.

This is the same rot as #2160, arriving by the **opposite mechanism**, and that is why it is worth its own issue. #2160's comments went stale because **the world moved**. This one went stale because **typia moved**: it reads as a survivor of the removed `ILlmSchema<Model>` era, and it now points users at a capability that no longer exists.

## Evidence

Verified on `master`.

`packages/typia/src/json.ts:100-107`:

```
 * Use cases:
 *
 * - Custom LLM function calling schema formats
 * - API documentation or code generation tools
 * - Alternative LLM integrations beyond built-in providers
 *
 * For standard LLM function calling, use {@link llm.application} instead, which
 * provides provider-specific schemas (ChatGPT, Claude, Gemini, etc.).
```

What `llm.application` actually takes (`packages/typia/src/llm.ts:117-120`):

```ts
export function application<
  Class extends Record<string, any>,
  Config extends Partial<
    ILlmSchema.IConfig & { ... }
```

`ILlmSchema.IConfig` declares exactly one field — `strict: boolean`. `ILlmApplication.IConfig` adds `validate`. **No provider dimension exists anywhere in the signature or the config.**

So the sentence promises a selector the API does not have, and the phrase "beyond built-in providers" in the bullet above implies "built-in providers" are a thing here. They are not.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/typia/src/json.ts:100-107` |
| **Affected** | anyone reading `json.application`'s doc — this ships on the published package. A reader is told to use `llm.application` for a feature it does not have, and may go looking for the knob |
| **Not affected** | runtime behaviour; `llm.application` itself, which is fine — only the sentence describing it is wrong |

Documentation only. Recorded because this campaign has been misled by its own records five times, and a comment that describes a removed API is a record actively pointing the reader at nothing.

| Case | Expected |
| --- | --- |
| `json.application`'s doc | states what `llm.application` actually offers, or simply points to it without inventing a capability |
| "beyond built-in providers" | either the built-in providers are real, or the phrase goes |
| `llm.application`'s own docs | unchanged — they do not make this claim |

## Approach

The invariant: **a doc comment describes the API that exists.**

Say what `llm.application` actually is — the LLM function-calling application builder, configured by `strict` — and drop the provider dimension. Do not replace it with a different provider claim; #2160 established that provider behaviour does not belong in a doc comment at all, and that a *corrected* claim rots exactly as the original did.

**Sweep for siblings while here.** #2160's cleanup found that its own inventory was incomplete because one instance — `ILlmStructuredOutput.ts:38`, "Most LLM providers accept this directly" — **named no provider and so escaped the grep**. This one is in `packages/typia/src`, which #2160's scope did not cover at all. The honest question is how many more comments describe an API typia has since removed, and a grep for provider names will not find them.

## A related, deliberately separate question for the maintainer

`packages/typia/native/core/programmers/llm/LlmSchemaProgrammer.go:157-158` carries the sentence *"OpenAI strict structured outputs cannot express the `not` keyword"* — the same claim #2165 just deleted from `Exclude.ts`.

It was left in place deliberately, and the reasoning is worth recording rather than acting on unilaterally: it is **implementation rationale sitting beside the branch it explains**, not a doc comment that ships on a public type. It explains why *typia's* code does what it does. That is arguably on the right side of #2160's rule.

But it is also where the prediction survives after the doc comments are cleaned — and it is the *reason* the `Exclude` rejection exists, so if that dialect claim ever goes stale, what needs revisiting is **the code it justifies**, not the comment. Whether rationale comments are in scope for the no-provider-claims rule is the maintainer's call, not the campaign's. **This issue does not propose touching it.**

- **Positive:** `json.application`'s doc describes the real API; a reader following it finds what it promises.
- **Negative:** no comment in `packages/typia/src` claims a capability the signature does not have.
- **Regression:** no code change; the emitted `.d.ts` differs only in the edited comments — and note that it **will** differ, because TypeScript emits doc comments into `.d.ts`. The proof to use is #2165's: every unedited declaration file byte-identical, and the edited ones identical with JSDoc stripped.
- **Commands:** `pnpm --filter ./packages/typia build`; the repository spell check.

## Coordination

- Documentation only. Read `.agents/skills/documentation/SKILL.md` first. **No package version bump** — releases and version numbers belong to the maintainer.
- **Sibling of #2160** (fixed by PR #2165), which removed provider behaviour claims from `@typia/interface`. That scope stopped at `packages/interface/src`; this is `packages/typia/src`.
- **The schema spec is off-limits.** The type declarations in `packages/interface/src` have zero errors and their non-standard properties are deliberate. Nothing here touches a declaration.
- Found by the #2160 batch, which correctly declined to sweep it in, and confirmed by the lead against `llm.application`'s actual signature.

samchon:
Closes #2171.

## Scope

Documentation-only sweep of `packages/typia/src` doc comments for two classes the maintainer's ruling on #2171 condemns (latest spec only):

- **Provider-behaviour claims** — sentences predicting what OpenAI/Gemini/Anthropic do or don't support. Deleted, not "corrected" (a corrected claim rots identically). #2160/#2165 established this for `packages/interface/src`; this batch covers `packages/typia/src`, which that scope did not touch.
- **Removed-API descriptions** — comments describing a capability, option, or type that no longer exists. `json.ts:107`'s "provider-specific schemas (ChatGPT, Claude, Gemini, etc.)" is the seed: a survivor of the removed `ILlmSchema<Model>` era. `llm.application`'s only schema knob is `ILlmSchema.IConfig.strict`; there is no per-provider model.

The invariant: a doc comment describes the API that exists and states typia's contract, not a third party's behaviour. Naming a provider to explain an option's *origin* survives (e.g. `strict` is OpenAI's structured-output mode).

## Constraints honoured

- **No package version bump.** `packages/typia/package.json` is untouched; releases and version numbers belong to the maintainer.
- **`packages/typia/native/**` untouched.** `LlmSchemaProgrammer.go:157-158` ("OpenAI strict cannot express `not`") is owned by the #2156/#2157 batch, which removes it. Not touched here.
- **`packages/interface/src` declarations off-limits.** Already swept by #2160/#2165.
- No `pnpm format`; campaign CI suspended per the issue-campaign rules.

## Verification

- `pnpm --filter ./packages/typia build`; the emitted `.d.ts` differs only in the edited comments (byte-identical with JSDoc stripped, per #2165's method).
- Repository spell check.

### DIFF
--- a/packages/typia/src/json.ts
+++ b/packages/typia/src/json.ts
@@ -101,10 +101,10 @@ export function application(): never;
  *
  * - Custom LLM function calling schema formats
  * - API documentation or code generation tools
- * - Alternative LLM integrations beyond built-in providers
+ * - Alternative LLM integrations that need full JSON Schema
  *
  * For standard LLM function calling, use {@link llm.application} instead, which
- * provides provider-specific schemas (ChatGPT, Claude, Gemini, etc.).
+ * builds the LLM-optimized function schema directly.
  *
  * @template Class Target class or interface type
  * @template Version OpenAPI version (`"3.0"` | `"3.1"`). Default `"3.1"`


samchon:
## Implementation complete — rebased, verified, Self-Reviewed

Rebased onto `ae2173e36a` (clean; #2173 already removed the `LlmSchemaProgrammer.go` provider sentence, which was out of scope here anyway). Head is `b2b79a5957`. Campaign CI cancelled per the exact-SHA gate after both the force-push and this update.

### The sweep result (`packages/typia/src`, read-and-classify, not grep)

One defect, in `json.ts`'s `json.application` doc:

- **Removed-API claim** — "use `llm.application` ..., which **provides provider-specific schemas (ChatGPT, Claude, Gemini, etc.)**". `llm.application`'s only schema knob is `ILlmSchema.IConfig.strict` (a single boolean; verified against `llm.ts`'s signature and `ILlmSchema.IConfig`). No per-provider model exists — this described the removed `ILlmSchema<Model>` API. Replaced with "builds the LLM-optimized function schema directly."
- **False implication** — the bullet "Alternative LLM integrations **beyond built-in providers**" implied typia has built-in LLM providers here; it does not. Replaced with "that need full JSON Schema."

The rest of the scope (all public files + ~120 internal helpers, `protobuf`/`functional`/`notations`/`transformers`/etc.) was read; no other provider-behaviour claim or removed-API description was found.

### Deliberately kept (the one judgment call)

`llm.ts:37` (`llm.controller`) and `llm.ts:102` (`llm.application`): "When passed to LLM providers (ChatGPT, Claude, Gemini, etc.), the LLM automatically selects functions and fills arguments from conversation." **Kept unchanged.** This is the keep-side of the #2165 precedent: it makes no schema accept/reject/support prediction (the class that rots), only describes the function-calling usage workflow and names providers as the *destination* of typia's output — exactly like the sentence #2165 kept in `ILlmStructuredOutput.ts` ("Pass this schema to LLM providers (OpenAI, Anthropic, Google, etc.) ... to constrain the output format"). Issue #2171's own case matrix also states "`llm.application`'s own docs — unchanged." Flagging it here so the lead can overrule; if a reframe is wanted, the model is #2165's "pass this to a provider to do X" phrasing.

### Verification

- `pnpm --filter ./packages/typia build` — green (both edited HEAD and pristine `ae2173e36a`).
- **`.d.ts` proof (#2165's method), rebuilt on `ae2173e36a`:** only `lib/json.d.ts` differs among declarations, and it is **identical with JSDoc stripped**; every other declaration file and all emitted `.js`/`.mjs` are byte-identical. (`lib/json.mjs.map` also differs — a sourcemap embeds the source text, so a comment change shifts it; not a declaration and not a code change.)
- **Spell check (`typos` v1.48.0, `./typos.toml`):** `json.ts` is clean. The repository run reports 6 pre-existing `labelled/unlabelled/labelling` (en-GB) errors, all in files outside this diff (`website/.../protobuf/message.mdx`, `packages/utils/.../OpenApiSchemaNamingRule.ts`, two `packages/typia/test/contract/native/*_test.go`, one `tests/test-typia-schema/...`) that landed on `ae2173e36a` via sibling batches — reported to the lead, not in scope here.

### Constraints honoured

`packages/typia/native/**` untouched (diff is `json.ts` only). No `packages/typia/package.json` bump. `packages/interface/src` untouched. No `pnpm format`.

### Noted for the lead (out of scope, not edited)

`packages/interface/src/schema/IJsonSchemaApplication.ts:13-14` still carries "features that **some LLM providers don't support** (tuples, `additionalProperties`, etc.)" — a brand-less "don't support" prediction (same class as the `ILlmStructuredOutput.ts:38` one #2160 missed) that survived the #2160/#2165 interface sweep. It's in `packages/interface/src` (off-limits to this batch), so left untouched; recording it since the premise "#2160 already swept interface, expect little there" holds only approximately — one genuine instance remains.

Solo Self-Review: one complete round, nothing to change.

PR #2173, opened 2026-07-17
fix(native): OpenAPI 3.0 dialect downgrade and strict numeric default (#2156, #2157)

### LINKED ISSUE #2156: json.schema/schemas/application with "3.0" emit a 3.1 document with the version string rewritten: the Go port dropped the downgrader

### LINKED ISSUE #2157: llm.* with strict silently drops a numeric default: #2056 was fixed in @typia/utils and left in the Go emitter that typia.llm.* actually calls

samchon:
Campaign batch: two independent defects in `packages/typia/native/core/programmers/`, each where the Go port dropped a step its TypeScript predecessor performed. Two separate commits on one branch so each can be reverted alone.

## #2156 — `"3.0"` emitted a 3.1 document with the version string rewritten

`json.schema`/`json.schemas`/`json.application` under `"3.0"` built the emended 3.1 document and rewrote only its version string, so the output declared 3.0 while carrying `{"type":"null"}` union members (not the `nullable` flag), `const`, and `prefixItems` — none of which OpenAPI 3.0 defines. The TypeScript implementation called `@typia/utils`' `OpenApiConverter.downgrade*`; the Go port kept `writeV3_0`'s skeleton and dropped the downgrade. There was zero 3.0-dialect coverage to catch it.

**Decision — reimplement the downgrade in Go.** The Go transform cannot call into TypeScript, so `"3.0"` cannot be honoured by delegating to `@typia/utils`; dropping the option is a public-contract change the maintainer owns. `OpenApiV3Downgrader` (new, in the iterate package) is a faithful, deliberately literal port of `@typia/utils`' `OpenApiV3Downgrader`. That makes `@typia/utils` a **second owner** of one contract; they are kept aligned by `test_json_schemas_v3_0_parity_converter`, which runs both owners on the same emended input over the full construct set and fails whichever drifts.

Covers all three shared entry points (`schema`/`schemas`/`application`) with dialect-aware tests plus a colocated Go unit test. The `"3.1"` default path is unchanged.

## #2157 — `llm.*` with `strict` silently dropped a numeric `default`

`shift_numeric` did `delete(output, "default")` with no tag, while `shift_string` shifts it to `@default <value>`. PR #2062 fixed the `@typia/utils` copy and left the Go emitter that `typia.llm.*` calls. `"default"` now joins the numeric key list.

**Read-side counterpart (per the #2157 lead comment, verified against merged #2162):** even the string `@default` that was written leaked back as prose on inversion — `LlmDescriptionInverter` restored no `default` and `describe()` filtered none. Both readers now restore `default` and drop its tag, so the value round-trips symmetrically for numeric and string.

The durable finding is the blind fixture: `test_llm_schema_parity_converter_strict` now exercises the full numeric/string/array shift sets rather than a sample, and `test_llm_schema_invert_default_round_trip` pins the round trip. Severity is low — reachable only under `strict`. Also removes a provider-behaviour comment at the exclude-tag gate, reworded to typia's own mechanic per the campaign rule (relates to #2171).

## Release / scope

**No `packages/typia/package.json` version bump.** Native Go normally owes one, but the maintainer owns releases and this campaign defers version numbers to them. No `pnpm format`. No schema declarations touched — the runtime was the defect in both cases.

## Local verification

Campaign CI is suspended per the exact-SHA cancellation gate; verification is local. See PR comments for executed case counts and the fail-then-pass observation for both issues.

Closes #2156
Closes #2157

### DIFF
--- a/packages/typia/native/core/programmers/iterate/openapi_v3_downgrader.go
+++ b/packages/typia/native/core/programmers/iterate/openapi_v3_downgrader.go
@@ -1,0 +1,637 @@
+package iterate
+
+import (
+  "reflect"
+  "strings"
+
+  nativefactories "github.com/samchon/typia/packages/typia/native/core/factories"
+)
+
+// OpenApiV3Downgrader rewrites an emended (OpenAPI 3.1) schema collection into
+// the OpenAPI 3.0 dialect. `json.schema`, `json.schemas`, and `json.application`
+// all build their document through json_schema_station, which speaks 3.1 only:
+// it emits `const`, `prefixItems`, and a `{"type": "null"}` union member for a
+// nullable type. None of those exist in 3.0, which spells the same three things
+// as `enum`, `items` + `minItems`/`maxItems`, and the `nullable` flag.
+//
+// This is a port of `@typia/utils`' OpenApiV3Downgrader, which the TypeScript
+// implementation of JsonSchemasProgrammer called through
+// `OpenApiConverter.downgradeComponents` / `downgradeSchema` before the Go port
+// replaced it. The Go transform cannot call into TypeScript, so the two are
+// necessarily separate implementations of one contract; the port is kept
+// deliberately literal so the correspondence stays reviewable, and
+// `tests/test-typia-schema/src/features/json.schemas/test_json_schemas_v3_0_parity_converter.ts`
+// pins the two owners against each other on every emitted construct.
+//
+// Only the components-and-schemas surface is ported. The TypeScript owner also
+// downgrades paths, operations, and security schemes, none of which typia emits.
+type OpenApiV3Downgrader_IComponentsCollection struct {
+  // Original is the emended 3.1 collection, read to resolve `$ref` targets
+  // while deciding nullability. It is never mutated.
+  Original *OpenApi_IComponents
+  // Downgraded accumulates the 3.0 result, including the `X.Nullable`
+  // companion schemas that a nullable reference forces into existence.
+  Downgraded *OpenApi_IComponents
+}
+
+// OpenApiV3Downgrader_downgrade_components downgrades every schema of the
+// collection, preserving the discovery order recorded in Order. A schema is
+// downgraded before its own key is registered, so the `X.Nullable` companions
+// that its references create land ahead of it exactly as the TypeScript owner's
+// argument-evaluation order places them.
+func OpenApiV3Downgrader_downgrade_components(input *OpenApi_IComponents) *OpenApiV3Downgrader_IComponentsCollection {
+  collection := &OpenApiV3Downgrader_IComponentsCollection{
+    Original:   input,
+    Downgraded: &OpenApi_IComponents{},
+  }
+  if input == nil || input.Schemas == nil {
+    return collection
+  }
+  collection.Downgraded.Schemas = map[string]JsonSchema{}
+  for _, key := range input.Order {
+    value, ok := input.Schemas[key]
+    if ok == false {
+      continue
+    }
+    downgraded := OpenApiV3Downgrader_downgrade_schema(collection, value)
+    collection.Downgraded.emplaceSchemaKey(key)
+    collection.Downgraded.Schemas[key] = downgraded
+  }
+  return collection
+}
+
+// OpenApiV3Downgrader_downgrade_schema rewrites one emended schema into 3.0.
+//
+// The emended dialect expresses a union as `oneOf` and nullability as a
+// `{"type": "null"}` member of it. 3.0 has neither, so the schema is flattened
+// into a union of concrete members, the null member is dropped, and the
+// surviving members carry `nullable: true` instead. Constants collapse into the
+// `enum` of a member of the same primitive type.
+func OpenApiV3Downgrader_downgrade_schema(collection *OpenApiV3Downgrader_IComponentsCollection, input JsonSchema) JsonSchema {
+  nullable := openApiV3Downgrader_is_nullable(map[string]bool{}, collection.Original, input)
+  union := []JsonSchema{}
+
+  var discriminator any
+  if openApiV3Downgrader_is_one_of(input) {
+    if value, ok := input["discriminator"]; ok && openApiV3Downgrader_is_nil(value) == false {
+      discriminator = openApiV3Downgrader_clone_value(value)
+    }
+  }
+  // A nullable union cannot keep its discriminator: 3.0 requires every mapped
+  // branch to stay a distinct `oneOf` member, and the null member is gone.
+  preserveDiscriminator := discriminator != nil && nullable == false
+
+  var visit func(schema JsonSchema)
+  visit = func(schema JsonSchema) {
+    if openApiV3Downgrader_is_boolean(schema) {
+      union = append(union, JsonSchema{"type": "boolean"})
+    } else if openApiV3Downgrader_is_integer(schema) ||
+      openApiV3Downgrader_is_number(schema) ||
+      openApiV3Downgrader_is_string(schema) ||
+      openApiV3Downgrader_is_reference(schema) {
+      union = append(union, openApiV3Downgrader_omit_examples(schema))
+    } else if openApiV3Downgrader_is_array(schema) {
+      next := openApiV3Downgrader_omit_examples(schema)
+      next["items"] = OpenApiV3Downgrader_downgrade_schema(
+        collection,
+        openApiV3Downgrader_schema(schema["items"]),
+      )
+      union = append(union, next)
+    } else if openApiV3Downgrader_is_tuple(schema) {
+      union = append(union, openApiV3Downgrader_downgrade_tuple(collection, schema))
+    } else if openApiV3Downgrader_is_object(schema) {
+      union = append(union, openApiV3Downgrader_downgrade_object(collection, schema))
+    } else if openApiV3Downgrader_is_one_of(schema) {
+      tracked := openApiV3Downgrader_same(schema, input) && discriminator != nil
+      for _, branch := range openApiV3Downgrader_schema_array(schema["oneOf"]) {
+        previous := len(union)
+        visit(branch)
+        // A branch that did not contribute exactly one member (a constant,
+        // or a nested union) breaks the one-branch-per-mapping rule the
+        // discriminator depends on.
+        if tracked && len(union) != previous+1 {
+          preserveDiscriminator = false
+        }
+      }
+    }
+  }
+
+  insertConstant := func(value any) {
+    typeName := openApiV3Downgrader_typeof(value)
+    for _, member := range union {
+      if member["type"] == typeName {
+        enum, _ := member["enum"].([]any)
+        member["enum"] = append(enum, value)
+        return
+      }
+    }
+    union = append(union, JsonSchema{"type": typeName, "enum": []any{value}})
+  }
+  visitConstant := func(schema JsonSchema) {
+    if openApiV3Downgrader_is_constant(schema) {
+      insertConstant(schema["const"])
+    } else if openApiV3Downgrader_is_one_of(schema) {
+      for _, branch := range openApiV3Downgrader_schema_array(schema["oneOf"]) {
+        if openApiV3Downgrader_is_constant(branch) {
+          insertConstant(branch["const"])
+        }
+      }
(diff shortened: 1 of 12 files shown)


samchon:
## Local verification (campaign CI suspended per exact-SHA gate)

Final SHA `491aa12758`. Executed case counts, `GOMAXPROCS=4`, Go at `-p 2`:

| Suite | Result |
| --- | --- |
| test-typia-schema | 165 cases, all Success |
| test-utils | 300 cases, all Success (exercises the inverter change) |
| test-mcp | 20 cases, all Success |
| test-typia-automated (solo) | 3945 cases, all Success |
| Go native `../native/core/programmers/...` | ok (iterate, json, llm, notations, plain) |
| Go native `../native/...` | ok (full tree) |
| Go public `./...` | ok |

### Fail-then-pass (reconstructed through the real emit path)

Reverting the two Go source fixes + the read-side inverter to `master` and rebuilding the plugin made exactly six tests Error, and only those six — no collateral:

- `test_json_schemas_v3_0_dialect`, `test_json_schema_v3_0_dialect`, `test_json_application_v3_0_dialect`, `test_json_schemas_v3_0_parity_converter` (#2156)
- `test_llm_schema_parity_converter_strict`, `test_llm_schema_invert_default_round_trip` (#2157)

Restoring the fixes turned all six green.

### 3.0 downgrade — decision and owner alignment

Reimplemented the downgrade in Go (`OpenApiV3Downgrader`, iterate package), a faithful port of `@typia/utils`' `OpenApiV3Downgrader`. The Go transform cannot call into TypeScript, so the two are necessarily separate owners of one contract. They are kept aligned by `test_json_schemas_v3_0_parity_converter`, which runs both owners on the same emended input across the full construct set — nullable atomics and references, tuples, rest tuples, literal and mixed literal unions, recursion, dictionaries, and both a preserved and a dropped discriminator — and fails whichever drifts. All three shared entry points (`json.schema`/`schemas`/`application`) are covered by dialect tests.

### #2157 round-trip coordination with #2150 (merged as #2162)

#2162 rewrote `LlmDescriptionInverter` to gate reads on `strict` and to drop keys whose lookup found nothing, but it restored no `default` and filtered none — so a repaired numeric write-side default would still leak as prose on inversion, exactly as string already did. This change adds `default` restoration + `describe()` filtering to both the numeric and string readers, and `test_llm_schema_invert_default_round_trip` pins the write-then-invert round trip for both. `test_llm_invert_strict_constraints_restored` (test-utils) has no `default` tag and is unaffected.

### Self-Review

Solo Self-Review found and fixed, each re-verified:

1. `openApiV3Downgrader_typeof` classified a numeric literal `object`, mislabelling the numeric branch of a mixed literal union — a numeric literal arrives as the compiler's named `jsnum.Number` (`type Number float64`), which an explicit `case float64` does not match. Reclassified by reflected kind; a Go unit-test case locks it.
2. `test_llm_schema_spec_strict_number` had snapshotted the dropped `@default` — corrected to include `@default 3` (called out as a spec-locking test).
3. The `preserveDiscriminator` branch was untested — added a preserved and a dropped discriminator to the parity fixture.

Final complete round found nothing further.

### Notes

No `packages/typia/package.json` version bump — releases and version numbers belong to the maintainer. No `pnpm format` (deferred to Post-Campaign Cleanup); the two new Go files were hand-normalized to the repo's 2-space Go indentation to match their directory. No schema declarations touched — the runtime was the defect in both cases. The provider-behaviour comment at `LlmSchemaProgrammer.go:157` is removed, reworded to typia's own mechanic (relates to #2171).

PR #2174, opened 2026-07-17
fix(ci): make the artifact gate, the website lint config, and the wasm build honest

### LINKED ISSUE #2170: Three gates that are not gates: the artifact CI job omits config/**, the website lint config cannot execute, and build/wasm.js returns success with a stale playground
## Problem

Three things in the website and CI surface look like gates and are not. Each is independently verified; together they are why a three-year-old `ReferenceError` and a twice-broken `.mjs` build could ship behind green.

1. **`experiments.yml` — the only CI gate that consumes a published artifact — omits `config/**` from its path filter**, while `build.yml` and `test.yml` both include it. `config/` owns both files that define the `.mjs` contract.
2. **`website/eslint.config.mjs` is dead configuration.** eslint is not a dependency, there is no lint script, and no CI step runs it.
3. **`website/build/wasm.js` returns success while leaving whatever wasm is on disk**, so a "successful" build can ship a stale playground.

## Evidence

Verified on `master`.

### 1 — the artifact gate does not watch the artifact's definition

```yaml
# .github/workflows/experiments.yml
paths:
  - '.github/workflows/experiments.yml'
  - 'packages/**'
  - 'experiments/**'
  - 'package.json'
  - 'pnpm-lock.yaml'
  - 'pnpm-workspace.yaml'
```

No `config/**`. Its siblings have it — `build.yml:6` and `test.yml:9`.

This matters because of what `experiments/` **is**: the harness that packs a tgz and imports it from a real ESM consumer. It is the only thing in the repository that consumes the **published** artifact rather than the workspace source. The in-repo `ttsx` suites cannot see this class of defect at all.

And `config/` owns both files that define the `.mjs` contract: `config/rolldown.config.mjs` and `config/tsconfig.json`'s `isolatedModules`.

D3-D reproduced the **exact v13.0.0-class defect** by editing `config/rolldown.config.mjs` alone — two deletions:

- `pnpm run build` → **exit 0**
- the tarball ships **0 of 115** `internal/*.mjs`; the CJS twins are all intact
- a CJS consumer works perfectly; an ESM top-level `import { is }` **also** works
- an ESM **deep** import → `ERR_MODULE_NOT_FOUND`
- the real `experiments/esm` harness → `ttsc` compiles green, then `TEST_EXIT=1`

**The gate is a working oracle. It just is not listening to the file it guards.** `git log -S` shows `build.yml` gained `config/**` in #1751 — the PR that created the concern — and `experiments.yml` never had it. **#2016**, which both fixed the twice-broken `.mjs` and created `config/rolldown.config.mjs`, edited `experiments.yml` while leaving its `paths` byte-identical.

**Severity: latent, and stated that way deliberately.** D3-D swept all **59** commits touching `config/` and **0** would have hit the gap — it caught itself nearly reporting "all 30 commits" from a `-30`-capped log.

### 2 — the lint config nothing runs

`website/eslint.config.mjs` exists (409 bytes, dated 2025-10-27) and extends `next/core-web-vitals` + `next/typescript`. Measured against `website/package.json`:

```
eslint deps  : NONE
lint script  : NONE
scripts referencing eslint : NONE
CI steps running lint : NONE
```

**This is why `clear()` survived three years.** `website/build/deploy.js`'s success path called an undefined `clear` — a typo for gh-pages' exported `deploy.clean()` — from 2023 until PR #2161 fixed it. `no-undef` catches that on day one.

A config that looks like a gate but cannot execute is worse than no config: the next reader believes it does something.

### 3 — a successful build that ships a stale playground

`website/build/wasm.js:138`:

```js
if (!hasNativeDir || !wasmExecSrc) {
  if (fs.existsSync(wasmOut) && fs.existsSync(wasmExecOut)) {
    console.warn(
      "build/wasm.js: Go toolchain or typia native dir missing; reusing existing prebuilt artifacts.",
    );
    return;
  }
```

When the Go toolchain or the ttsc tree is unavailable it logs a notice and **returns successfully**, leaving whatever wasm is on disk. Line 119 additionally skips the rebuild on an mtime comparison (`wasmMtime >= sourceMtime`), and mtimes are unreliable after a clone or a cache restore, which stamps them at restore time.

**CI is insulated only incidentally:** `public/compiler/` is gitignored and not cached, so the artifact is always absent and the hard-fail path wins. That insulation is an accident of caching policy. Anyone who caches `public/compiler/`, or builds locally with a populated one, gets a silently stale playground from a green build.

## Consequence surface

| # | Owner | Affected | Not affected |
| --- | --- | --- | --- |
| 1 | `.github/workflows/experiments.yml` | a `config/`-only change to the `.mjs` contract merges with the one gate that could catch it never running | today's history — 0 of 59 commits would have hit it |
| 2 | `website/eslint.config.mjs` | anyone who believes the website is linted; every undefined reference in `website/` | runtime — nothing executes it |
| 3 | `website/build/wasm.js` | anyone caching `public/compiler/` or building locally with one populated | CI today, incidentally |

None is a product defect. All three are reasons a product defect would not be caught, which is the same family as #2117 (a host that published without consulting diagnostics) and #2141 (a tagged suite no workflow runs).

| Case | Expected |
| --- | --- |
| a `config/rolldown.config.mjs` change that breaks `.mjs` | `experiments` runs and fails |
| a `config/`-only change generally | the artifact gate is not skipped |
| an undefined reference in `website/` | caught before merge, or the config that claims to catch it is gone |
| `build/wasm.js` with no Go toolchain | fails, or says loudly that the artifact is unverified |
| `build/wasm.js` with a stale cached wasm | rebuilt, or refused |
| CI today | unchanged — all three already pass |

## Approach

The invariant: **a job watches every input it consumes, and a check that cannot fail is removed rather than kept.**

**1** is the same soundness argument PR #2161 just landed for `website.yml` and `release.yml`: removing a path filter does not relax a guard, it makes the guard **pass stale**. Add `config/**`, then ask whether any other workflow's `paths` omits an input it reads — the three disagree today and nothing forces them to agree.

**2** has three options and only two are acceptable: wire eslint up, or delete the config. **Leaving a config that looks like a gate but is not is the worst of the three.** If it is wired up, expect existing violations — `clear()` was one for three years, and there may be more.

**3**: decide deliberately whether a missing toolchain is a build failure or a documented "unverified artifact" state, and do not rely on `public/compiler/` staying uncached to enforce it. The mtime skip is a second, independent instance of the same leniency; both should move together.

- **1:** reproduce D3-D's control — delete the two lines from `config/rolldown.config.mjs`, confirm `experiments` now triggers and fails, then restore. That is the negative case, and it must fail against the current filter.
- **2:** if wired, a deliberately undefined reference in `website/` fails the check; if deleted, nothing claims to lint.
- **3:** a build with the Go toolchain removed does not silently succeed with a stale artifact; a normal build is unchanged.
- **Regression:** CI's green stays green — all three pass today. `website.yml`'s provenance test and its `release.yml` filter (PR #2161) are untouched.
- **Commands:** `pnpm --filter ./website run build`; the `experiments` harness; the repository spell check.

## Coordination

- CI configuration and website build scripts only. No product change, and **no package version bump** — releases and version numbers belong to the maintainer.
- **Related to #2161** (merged), which fixed the mirror-image defect in `website.yml` — a deleted path filter that made a provenance guard pass stale.
- **Do not touch `website/package.json`'s duplicate `typescript` key.** It is **deliberate**, confirmed by the maintainer: Next.js cannot use TypeScript 7 because 7.0 ships no `lib/typescript.js` (the JS Compiler API, deferred to 7.1) and `next build` requires it. It is not a defect and is out of scope here.
- Items 1 and 3 found by discovery round D3-D; item 2 by the website deploy batch. All three confirmed by the lead on merged master.

samchon:
## Scope

Fixes #2170 — three CI/website constructs that look like gates but cannot gate. Landed as three independent commits so each reverts alone.

1. **`.github/workflows/experiments.yml` omits `config/**`.** `experiments/` is the only CI job that consumes the *published* tarball (`pnpm run package:tgz` → `pnpm pack` → `prepack` build), and every package's `rolldown.config.mjs` re-exports `config/rolldown.config.mjs` while every package `tsconfig.json` extends `config/tsconfig.json`. A `config/`-only change to the `.mjs` contract could merge without the one gate that catches it ever running. `build.yml:6` and `test.yml:9` already list `config/**`. Add it here so the three agree.
2. **`website/eslint.config.mjs` is dead config** — eslint is not a dependency, no lint script, no CI step runs it. A config that looks like a gate but cannot execute is worse than none. Deleted.
3. **`website/build/wasm.js` returns success while leaving a possibly-stale wasm** when the Go toolchain is unavailable, and skips rebuild on an unreliable mtime comparison. Made a missing toolchain a build failure by default, with an explicit opt-in (`WEBSITE_WASM_ALLOW_PREBUILT`) for the documented "reuse unverified prebuilt" state. CI is no longer insulated only by `public/compiler/` being gitignored.

## Constraints honored

- CI configuration and website build scripts only. No product change, no version bump, no `packages/typia/package.json` edit.
- No `pnpm format` (campaign rule).
- `website/package.json`'s duplicate `typescript` key left untouched (deliberate, per maintainer).
- `website.yml`'s provenance test and `release.yml`'s filter (#2161) untouched.

## Verification

Recorded per commit as it lands. Draft until verification completes.

Related to #2161 (merged), which fixed the mirror-image defect in `website.yml`.


samchon:
## Self-Review round 1 (solo) — clean, no changes to the diff

Three isolated commits on `a651378939`, each reverting alone:

| commit | file | item |
| --- | --- | --- |
| `e214e79cdf` | `.github/workflows/experiments.yml` (+1) | 1 |
| `11a9cfdb9e` | `website/eslint.config.mjs` (−16, deleted) | 2 |
| `efde7637ab` | `website/build/wasm.js` (+46 −15) | 3 |

### Item 1 — load-bearing reproduction reproduced

Deleted the two `preserveModules` lines from `config/rolldown.config.mjs` alone, then a full `package:tgz`:

- `pnpm pack` for `typia` → **exit 0**; the tarball ships **0 of 115** `lib/internal/*.mjs` (the modules flatten to `lib/*.mjs`), while the **115** CJS `lib/internal/*.js` twins stay intact.
- `experiments/esm`: `ttsc` compiles **green (exit 0)**, then `node lib/index.js` dies with `Error [ERR_MODULE_NOT_FOUND]: Cannot find module '…/typia/lib/internal/_isTypeInt32.mjs' imported from …/lib/test_langchain.js` → **`TEST_EXIT=1`**.
- Control: with the two lines present, the same harness passes **all** checks (exit 0). The only variable is `config/rolldown.config.mjs`.
- Path filter: `minimatch` confirms `config/rolldown.config.mjs` and `config/tsconfig.json` are **SKIPPED** by the pre-fix filter and **TRIGGER** after adding `config/**`. The gate is a working oracle that was not listening to the file it guards. Restored `config/rolldown.config.mjs` after.

### Item 2 — dead on every axis

`eslint` is not a `website/` dependency, no lint script, no CI step, and `next.config.ts` sets `eslint.ignoreDuringBuilds: true` so `next build` skips linting. The file even imports `@eslint/eslintrc`, which is not installed, so it throws before evaluating. No other eslint config exists and nothing references it. Verified `no-undef` reports `'clear' is not defined` (exit 1) on the exact pre-#2161 `deploy.js` success-path pattern — a live gate would have caught it on day one; the dead one caught nothing for three years. Removed.

### Item 3 — deliberate failure, opt-in reuse, normal build unchanged

Observed the negative case failing first, then fixed:

| scenario | before | after |
| --- | --- | --- |
| toolchain missing, **stale** prebuilt present, no opt-in (`[B]` reuse) | exit **0** "reusing existing prebuilt artifacts" | exit **1** |
| toolchain missing, **fresh-mtime** prebuilt present, no opt-in (`[A]` skip) | exit **0** "skipping rebuild" | exit **1** |
| toolchain missing, prebuilt present, `WEBSITE_WASM_ALLOW_PREBUILT=1` | — | exit **0** + explicit "unverified/may be stale" warning |
| normal: Go present, real native dir, fresh wasm | exit 0 mtime skip | exit 0 mtime skip (identical — guard conds true) |
| normal: Go present, no prebuilt → real `buildWasm()` | exit 0, 40 MB wasm | exit 0, 40 MB wasm (real build run, exit 0) |

The `[A]` guard adds `hasNativeDir && wasmExecSrc`, which are both true whenever the toolchain is present, so the Go-present path is byte-for-byte the original. CI no longer depends on `public/compiler/` being gitignored to enforce the failure.

### Regression / scope

Spell check (`typos --config ./typos.toml`, whole tree): **exit 0**. `website.yml` provenance test and `release.yml` filter (#2161) untouched. No product change, no version bump, `packages/typia/package.json` untouched, `website/package.json`'s duplicate `typescript` key untouched, no `pnpm format`.

### Additional finding for the lead (out of #2170 scope) — `nestia.yml`

`nestia.yml` runs `pnpm test` (= `test:packages` + `test:go`) but its `paths` **omit `config/**`, `tests/**`, and `examples/**`** while listing `experiments/**` and `scripts/**` that `pnpm test` does not run. It is the same class as item 1: created in #1896 to "verify typia tarballs against nestia next", its `paths` are stale relative to the current `pnpm test` body. Reported, not fixed — publication is lead-owned.

Issue #2169, opened 2026-07-17
OpenApiSchemaNamingRule reads exclusiveMinimum/Maximum as booleans while typia declares them as numbers, so an exclusive bound names itself just "number"


samchon:
## Problem

`OpenApiSchemaNamingRule.getNameOfInteger` reads `exclusiveMinimum`/`exclusiveMaximum` as **booleans**, while typia's own type declares them as **numbers**. Two consequences:

1. **`{ type: "integer", exclusiveMinimum: 5 }` names itself just `number`** — the constraint is silently dropped, because the whole branch is gated on `minimum !== undefined` and `minimum` is absent.
2. When both are present, the emitted tag carries **`minimum`'s value** under an `ExclusiveMinimum` name.

Separately, the function **ignores a declared `format`**, so `{ type: "integer", format: "int32" }` also names itself just `number`. That is the mirror image of #2146's third defect, which invented a `Type<"int32">` the schema never declared — this one drops the one it did.

## Evidence

Verified on `master`. **The specification here is typia's own declaration**, not any external standard.

`packages/interface/src/openapi/OpenApi.ts:402-419` — `IJsonSchema.IInteger`:

```ts
export interface IInteger extends IJsonSchemaAttribute.IInteger {
  minimum?: number & tags.Type<"int64">;
  maximum?: number & tags.Type<"int64">;
  exclusiveMinimum?: number & tags.Type<"int64">;
  exclusiveMaximum?: number & tags.Type<"int64">;
  multipleOf?: number & tags.ExclusiveMinimum<0>;
}
```

`exclusiveMinimum` is a **number**. `packages/utils/src/validators/internal/OpenApiSchemaNamingRule.ts:35-54` treats it as a flag on `minimum`:

```ts
const getNameOfInteger = (schema: OpenApi.IJsonSchema.IInteger): string[] => [
  "number",
  ...(schema.minimum !== undefined
    ? [
        schema.exclusiveMinimum
          ? `tags.ExclusiveMinimum<${schema.minimum}>`
          : `tags.Minimum<${schema.minimum}>`,
      ]
    : []),
  ...(schema.maximum !== undefined
    ? [
        schema.exclusiveMaximum
          ? `tags.ExclusiveMaximum<${schema.maximum}>`
          : `tags.Maximum<${schema.maximum}>`,
      ]
    : []),
  ...(schema.multipleOf !== undefined
    ? [`tags.MultipleOf<${schema.multipleOf}>`]
    : []),
];
```

[...]

PR #2175, opened 2026-07-17
fix(utils): name numeric exclusive bounds and declared integer format from the schema

samchon:
Implements #2169 — `OpenApiSchemaNamingRule` reads `exclusiveMinimum`/`exclusiveMaximum` as booleans while typia's own `OpenApi.IJsonSchema.IInteger`/`INumber` declare them as numbers, so an exclusive bound names itself just `number` (or, when paired with `minimum`, carries the wrong value under an exclusive name). Claim commit only at open; implementation follows.

## Scope (`@typia/utils` only)

`packages/utils/src/validators/internal/OpenApiSchemaNamingRule.ts`:

- **`getNameOfInteger` / `getNameOfNumber`** — read `exclusiveMinimum`/`exclusiveMaximum` as the independent numbers the declaration says they are, each with its own value, rather than as a boolean modifier gated on `minimum`/`maximum`. The decisive case is `exclusiveMinimum: 0`, falsy as a number, which the current `schema.exclusiveMinimum ?` test drops entirely.
- **`getNameOfInteger`** — echo a declared `format` as `tags.Type<...>`, defensively (the normalized `IInteger` does not model `format`, but `OpenApiConverter` passes an external document's `format` through at runtime). This mirrors `OpenApiIntegerValidator.describeType`, repaired in #2148.

## Invariant

The name states exactly the constraints the schema declares, each with its own value. Cross-checked so each naming function agrees with its paired validator on what a schema declares:

| naming function | paired validator | format echoed? |
| --- | --- | --- |
| `getNameOfInteger` | `OpenApiIntegerValidator` (`describeType`) | yes — `tags.Type<...>` |
| `getNameOfNumber` | `OpenApiNumberValidator` | no (validator does not echo format either) |
| `getNameOfString` | `OpenApiStringValidator` | already `tags.Format<...>`, unchanged |

## Regression

#2146's contract holds: no `int32` (or any format) is invented where the schema declares none.

## Coordination

No native Go. No package version bump — releases belong to the maintainer. No `pnpm format` on this campaign branch.

## Verification

Campaign CI is suspended; verification is local and recorded on this thread:

- `./tests/test-utils-automated` (`pnpm --filter ./tests/test-utils-automated start`)
- `pnpm test:packages`

### DIFF
--- a/packages/utils/src/validators/internal/OpenApiSchemaNamingRule.ts
+++ b/packages/utils/src/validators/internal/OpenApiSchemaNamingRule.ts
@@ -32,42 +32,59 @@ export namespace OpenApiSchemaNamingRule {
     return "unknown";
   };
 
-  const getNameOfInteger = (schema: OpenApi.IJsonSchema.IInteger): string[] => [
-    "number",
-    ...(schema.minimum !== undefined
-      ? [
-          schema.exclusiveMinimum
-            ? `tags.ExclusiveMinimum<${schema.minimum}>`
-            : `tags.Minimum<${schema.minimum}>`,
-        ]
-      : []),
-    ...(schema.maximum !== undefined
-      ? [
-          schema.exclusiveMaximum
-            ? `tags.ExclusiveMaximum<${schema.maximum}>`
-            : `tags.Maximum<${schema.maximum}>`,
-        ]
-      : []),
-    ...(schema.multipleOf !== undefined
-      ? [`tags.MultipleOf<${schema.multipleOf}>`]
-      : []),
-  ];
+  const getNameOfInteger = (schema: OpenApi.IJsonSchema.IInteger): string[] => {
+    // `minimum`, `maximum`, `exclusiveMinimum`, and `exclusiveMaximum` are all
+    // numbers in the declaration, each an independent bound rather than a
+    // boolean modifier on another. Read every one on its own value, so an
+    // exclusive bound is never dropped (including the falsy `0`) nor labelled
+    // with a sibling's value.
+    //
+    // The normalized integer schema does not model `format`, but
+    // `OpenApiConverter` passes an external document's `format` (`"int32"` /
+    // `"int64"`) straight through, so read it defensively and echo only what is
+    // there as `tags.Type<...>` — mirroring `OpenApiIntegerValidator.describeType`,
+    // which names the same declared width and invents none.
+    const format: unknown = (schema as { format?: unknown }).format;
+    return [
+      "number",
+      ...(typeof format === "string"
+        ? [`tags.Type<${JSON.stringify(format)}>`]
+        : []),
+      ...(schema.minimum !== undefined
+        ? [`tags.Minimum<${schema.minimum}>`]
+        : []),
+      ...(schema.exclusiveMinimum !== undefined
+        ? [`tags.ExclusiveMinimum<${schema.exclusiveMinimum}>`]
+        : []),
+      ...(schema.maximum !== undefined
+        ? [`tags.Maximum<${schema.maximum}>`]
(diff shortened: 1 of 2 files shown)


samchon:
## Implementation + Self-Review (round 1, clean)

Campaign CI is suspended; verification is local and recorded here.

### Change

`packages/utils/src/validators/internal/OpenApiSchemaNamingRule.ts`:

- `getNameOfInteger` / `getNameOfNumber` now read `exclusiveMinimum` / `exclusiveMaximum` as the independent numbers `OpenApi.IJsonSchema` declares, each named with its own value — not as a boolean modifier gated on `minimum` / `maximum`.
- `getNameOfInteger` echoes a declared `format` as `tags.Type<...>`, read defensively (`(schema as { format?: unknown }).format`, `typeof … === "string"`), since the normalized `IInteger` does not model `format` but `OpenApiConverter` passes an external document's `format` through. This mirrors `OpenApiIntegerValidator.describeType` (#2148).
- `getNameOfNumber` echoes **no** format: `OpenApiNumberValidator` reports none either, so echoing one would make the naming rule and the validator disagree on what the schema declares. `getNameOfString` was already correct (format handled; no exclusive bounds).

### Fail → pass (observed through `OpenApiValidator.validate`, reading the report's `expected`)

| Schema | Before (buggy) | After (fixed) |
| --- | --- | --- |
| `{ integer, exclusiveMinimum: 5 }` | `number` | `number & tags.ExclusiveMinimum<5>` |
| `{ integer, minimum: 0, exclusiveMinimum: 5 }` | `number & tags.ExclusiveMinimum<0>` | `number & tags.Minimum<0> & tags.ExclusiveMinimum<5>` |
| `{ integer, exclusiveMinimum: 0 }` (boundary) | `number` | `number & tags.ExclusiveMinimum<0>` |
| `{ integer, maximum: 10, exclusiveMaximum: 7 }` | `number & tags.ExclusiveMaximum<10>` | `number & tags.Maximum<10> & tags.ExclusiveMaximum<7>` |
| `{ integer, format: "int32" }` | `number` | `number & tags.Type<"int32">` |
| `{ number, exclusiveMinimum: 0 }` (boundary) | `number` | `number & tags.ExclusiveMinimum<0>` |
| `{ number, minimum: -1.5, exclusiveMaximum: 2.5 }` | `number & tags.Minimum<-1.5>` | `number & tags.Minimum<-1.5> & tags.ExclusiveMaximum<2.5>` |
| `{ integer }` / `{ number, format: "double" }` | `number` | `number` (no format invented; #2146's contract holds) |

The `exclusiveMinimum: 0` boundary is the decisive case: falsy as a number, the old `schema.exclusiveMinimum ?` test dropped a declared constraint.

### Cross-check with `OpenApiIntegerValidator`

`describeType` reads `format` with the same `typeof format === "string"` guard and names the same width; the naming rule differs only by the `tags.` prefix (as `getNameOfString` differs from `OpenApiStringValidator`). They agree on what a schema declares.

### Verification (executed case counts)

- `pnpm --filter ./tests/test-utils start` → **302 cases, Success** (includes the new `test_openapi_naming_numeric_constraints`).
- `pnpm --filter ./tests/test-utils-automated start` (path form; its package name is `@typia/test-openapiautomated`, so the name filter would silently match zero) → **163 cases, Success**.
- `pnpm test:packages` → running as the broad gate; result to follow.

### Self-Review

One complete solo round over the change surface (source, test, the sole consumer `OpenApiStationValidator`, cross-check with all three atomic validators, boundaries, backward-compatibility of the error-message text). Found nothing to change: the round is clean.

No native Go, no version bump, no `pnpm format` on this campaign branch.

PR #2184, opened 2026-07-18
fix(native): guard native generic detection against WeakMap*/WeakSet*-named user types

### LINKED ISSUE #2181: WeakMap*/WeakSet*-named user types are misdetected as native WeakMap/WeakSet: native generic match uses bare HasPrefix while Map/Set are <-guarded
## Problem

A user type whose name begins with `WeakMap` or `WeakSet` — e.g. `interface WeakMapLike { ... }` — is misdetected as the native `WeakMap`/`WeakSet`, because native detection matches those generics with a bare `strings.HasPrefix(name, generic.Name)` while every sibling (`Map`, `Set`, `ReadonlyMap`, `ReadonlySet`) requires a `<` after the name. So `is`/`assert`/`validate`/`json.schema`/`llm`/`random`/`clone`/`classify`/`compare`/`protobuf` all treat the user type as an unsupported native.

## Evidence

Verified on `master` (`5260e78433`). The same file guards `Map`/`Set` and does not guard `WeakMap`/`WeakSet`.

`packages/typia/native/core/factories/internal/metadata/iterate_metadata_native.go` (the primary detector):

```go
for _, generic := range iterate_metadata_native_generics {
  if strings.HasPrefix(name, generic.Name) {          // bare prefix — no `<`
    iterate_metadata_native_take(props.Metadata, generic.Name)
    return true
  }
}
```

`iterate_metadata_intersection.go:493-498` — the asymmetry in one place:

```go
for _, generic := range iterate_metadata_native_generics {
  if strings.HasPrefix(name, generic.Name) { return false }        // WeakMap/WeakSet: bare
}
if strings.HasPrefix(name, "Map<") || strings.HasPrefix(name, "ReadonlyMap<") ||
   strings.HasPrefix(name, "Set<") || strings.HasPrefix(name, "ReadonlySet<") { return false }   // Map/Set: `<`-guarded
```

`HasPrefix("WeakMapLike", "WeakMap")` is `true`, so `WeakMapLike` is taken as a native `WeakMap`. `HasPrefix("MapLike", "Map<")` is `false`, so `MapLike` is correctly a plain object — the two are handled inconsistently.

Reproduced end to end with three identically-shaped interfaces:

```
is<WeakMapLike>({ id, count })   → false     (should be true)
is<WeakSetLike>({ id, count })   → false     (should be true)
is<NormalLike>({ id, count })    → true       (control — no Weak prefix)
json.schema<WeakMapLike>()       → errors "does not support WeakMap type"
```

**Provenance:** the pre-Go detector matched with `name === c || HasPrefix(name, c + "<")`; the Go port (`e15205745f`) dropped the `<` guard for the generics list while keeping it for `Map`/`Set`.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `iterate_metadata_native.go` (the `generics` loop) and the four `HasPrefix(name, generic.Name)` sites in `iterate_metadata_intersection.go` (`:493, :566, :681, :990`) |
| **Affected** | any user type whose name starts with `WeakMap` or `WeakSet` — misclassified as an unsupported native across every feature |
| **Not affected** | `Map`/`Set`/`ReadonlyMap`/`ReadonlySet` user-alike names (already `<`-guarded); the genuine native `WeakMap<K,V>`/`WeakSet<T>`, which always carry `<` |

This is a soundness defect: a valid user type is rejected (or mis-emitted) for every feature.

| Case | Expected |
| --- | --- |
| `is<WeakMapLike>(...)` where `WeakMapLike` is a user interface | validates as the interface |
| a genuine `WeakMap<K, V>` | detected as the native — unchanged |
| `MapLike` / `SetLike` user types | plain objects — unchanged |
| a user type `WeakMapEntry`, `WeakSetView`, etc. | its own shape, not a native |

## Approach

The invariant: **a name is a native generic only when it is that generic — exact, or followed by `<` — never merely prefixed by it.** Bring the `generics` match onto the same `<`-guard the `Map`/`Set` siblings already use (or `name == generic.Name || HasPrefix(name, generic.Name + "<")`, the pre-Go form). Apply it at every one of the five sites — a partial fix leaves the misdetection on whichever path is missed.

Genuine `WeakMap`/`WeakSet` are always generic, so requiring `<` rejects nothing real; confirm against a bare `WeakMap` reference if the codebase can produce one.

- **Positive:** `WeakMapLike`/`WeakSetLike`/`WeakMapEntry` user types validate, serialize, and schema-generate as their real shape; a genuine `WeakMap<K,V>`/`WeakSet<T>` is still the native.
- **Negative — must fail against current code:** `is<WeakMapLike>({...})` returns `false` today; reproduce it, then fix.
- **Boundary:** all five call sites; both `WeakMap` and `WeakSet`; names that are an exact native vs prefixed vs `<`-followed; the same across is/assert/validate/json.schema/llm/random/clone/classify/compare/protobuf.
- **Regression:** `Map`/`Set` detection unchanged; genuine natives unchanged.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/factories/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start` (solo).

## Coordination

- Found by discovery round D4-C and confirmed by the lead against the guarded/unguarded asymmetry in the same file and the reproduction above.

samchon:
Fixes #2181.

## Intent

A user type whose name begins with `WeakMap`/`WeakSet` (e.g. `interface WeakMapLike {}`) is misdetected as the native `WeakMap`/`WeakSet`, because native generic detection matches with a bare `strings.HasPrefix(name, generic.Name)` while every sibling (`Map`/`Set`/`ReadonlyMap`/`ReadonlySet`) requires the name to be exactly that generic. So `is<WeakMapLike>({...})` returns `false` and `json.schema<WeakMapLike>()` errors "does not support WeakMap type", across every feature.

## Scope

Bring the five `iterate_metadata_native_generics` match sites onto the invariant **a name is a native generic only when it is that generic — exact, or followed by `<` — never merely prefixed by it**, matching the pre-Go form `name == generic.Name || HasPrefix(name, generic.Name + "<")`:

- `iterate_metadata_native.go` — the primary detector's `generics` loop.
- `iterate_metadata_intersection.go` — the four sibling sites (`is_removable_object_constraint`, `identify`, `is_plain_object_only`, `category`).

Because the name is pre-stripped of `<...>` by `iterate_metadata_native_getNativeName`, a genuine `WeakMap<K,V>`/`WeakSet<T>` reduces to exactly `"WeakMap"`/`"WeakSet"` and the `name == generic.Name` clause still detects it as the native.

## Verification

Pending — recorded on the thread as the campaign local-verification gate. Regression test lands first and is observed failing before the fix.

## CI

Campaign CI is deliberately suspended; runs for each pushed SHA are cancelled by the exact-SHA gate and local verification is recorded on the thread.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/native_generic_name_guard_is_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/native_generic_name_guard_is_transform_test.go
@@ -1,0 +1,194 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestNativeGenericNameGuardIsTransform verifies that a user type whose name
+// begins with WeakMap/WeakSet is not misdetected as the native generic.
+//
+// Native generic detection matched WeakMap/WeakSet with a bare
+// strings.HasPrefix(name, generic.Name) while the Map/Set siblings require the
+// name to be exactly the generic. A user interface named WeakMapLike therefore
+// took the native path and emitted `input instanceof WeakMap`, so
+// typia.is<WeakMapLike>({...}) returned false (issue #2181). The guard now
+// admits a native only when the stripped name equals the generic, so a genuine
+// WeakMap<K, V>/WeakSet<T> stays native while WeakMapLike validates its shape.
+//
+//  1. Transform typia.is calls for WeakMapLike, WeakSetLike, WeakMapEntry, a
+//     MapLike control, and genuine WeakMap<object, number>/WeakSet<object>.
+//  2. Assert the emitted TypeScript keeps the native `instanceof` path for the
+//     genuine generics.
+//  3. Execute the emitted validators so the user-named objects validate as
+//     their shape while the genuine natives still accept only their instance.
+func TestNativeGenericNameGuardIsTransform(t *testing.T) {
+  project := nativeGenericNameGuardProject(t)
+  ts := nativeGenericNameGuardTransform(t, project, "ts")
+  for _, needle := range []string{"instanceof WeakMap", "instanceof WeakSet"} {
+    if !strings.Contains(ts, needle) {
+      t.Fatalf("expected genuine native path %q in transform output:\n%s", needle, ts)
+    }
+  }
+  nativeGenericNameGuardRunRuntimeCases(t, project, nativeGenericNameGuardTransform(t, project, "js"))
+}
+
+func nativeGenericNameGuardProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "native-generic-name-guard-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(nativeGenericNameGuardTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(nativeGenericNameGuardSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func nativeGenericNameGuardTransform(t *testing.T, project string, output string) string {
+  t.Helper()
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", output,
+    })
+  })
+  if code != 0 {
+    t.Fatalf("native generic name guard transform failed: output=%s code=%d stderr=\n%s", output, code, errText)
+  }
+  return out
+}
+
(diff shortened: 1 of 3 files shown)


samchon:
## Implementation + local verification (campaign CI suspended)

Fix at all five `iterate_metadata_native_generics` match sites, plus a command-level transform regression test. `typia` package version intentionally not bumped per issue #2181 (releases belong to the maintainer).

### Fail-then-pass (regression test)

`TestNativeGenericNameGuardIsTransform` (command-level transform + node runtime):

- **Before the fix:** `WeakMapLike accepts its shape: expected true, got false` — reproduces the issue exactly.
- **After the fix:** PASS.

### Root-cause confirmation (both issue symptoms), via the debug runner

```
json.schema<WeakMapLike>() -> {"type":"object","properties":{"id":{"type":"number"},"count":{"type":"number"}},"required":["id","count"],"additionalProperties":false}
is<WeakMapLike>({id:1,count:2})  -> true
is<WeakMapLike>(new WeakMap())   -> false
```

No more "does not support WeakMap type"; the user type validates and schema-generates as its real shape. Genuine `WeakMap<object, number>`/`WeakSet<object>` still detected as the native (instanceof path), confirmed at runtime.

### Local verification (GOMAXPROCS=12)

| Command | Result |
| --- | --- |
| `go test ../native/core/factories/...` (`-p 2`) | ok (factories + internal/metadata) |
| `go test ../native/...` (`-p 2`) | ok — full native tree |
| `go test ./...` (public contract tree, `-p 2`) | ok — all packages |
| `pnpm --filter ./tests/test-typia-schema start` | 168 cases, Success |
| `pnpm --filter ./tests/test-typia-automated start` (solo) | 3947 cases, Success |
| `pnpm --filter ./tests/test-error start` | Every fixture rejected: 54 files, 849 diagnostics — genuine WeakMap/WeakSet still rejected by random/clone/protobuf |
| `go vet` (changed packages) | clean |

### Self-Review (solo)

One complete round over the whole change surface (5 production sites + test). Findings: none requiring change. The `HasPrefix(name, generic.Name+"<")` clause is post-strip-dead but matches the pre-Go form and the Map/Set sibling `<`-guard; the `name == generic.Name` clause is what preserves genuine natives (proven at runtime). Swept the whole repo — no sixth prefix-detection site. SPEC-over-reality gate: a bare `WeakMap`/`WeakSet` type reference does not typecheck (the generic has no key default), so genuine natives always carry `<`; the guard rejects nothing real.

READY.

PR #2185, opened 2026-07-18
fix(native): single-source JSDoc tag exclusivity from the interface authority (#2182)

### LINKED ISSUE #2182: JSDoc @minimum/@maximum/@format/@pattern exclusivity is unenforced: the comment factory omits or under-specifies every exclusive list #2154 relies on
## Problem

The tag-exclusivity contract that #2154 enforced for **type tags** is not enforced for the same constraints written as **JSDoc comments**, because the JSDoc comment factory's hand-written `Exclusive` lists are almost entirely missing or incomplete. So `@minimum 0` + `@exclusiveMinimum 5`, or `@pattern ^a` + `@pattern ^b`, compile through JSDoc where the equivalent type tags are a compile error — and emit a schema that contradicts the generated validator, with no diagnostic.

## Evidence

Verified on `master` (`5260e78433`). The interface tag declarations are the authority (`packages/interface/src/tags/*.ts`); the comment factory's `Exclusive` fields (`packages/typia/native/core/factories/MetadataCommentTagFactory.go`) diverge from them:

| JSDoc tag | factory `Exclusive` | interface `exclusive` (authority) |
| --- | --- | --- |
| `@minimum` | *(field absent)* | `["minimum", "exclusiveMinimum"]` |
| `@maximum` | *(field absent)* | `["maximum", "exclusiveMaximum"]` |
| `@exclusiveMinimum` | *(field absent)* | `["exclusiveMinimum", "minimum"]` |
| `@exclusiveMaximum` | *(field absent)* | `["exclusiveMaximum", "maximum"]` |
| `@format` | *(field absent)* | `["format", "pattern"]` |
| `@pattern` | `["format"]` | `["format", "pattern"]` — missing `pattern` |

The exclusivity check (`MetadataTypeTagFactory.Validate`, fixed in #2154) reads a tag's `Exclusive` list. A missing list means nothing to check against; an incomplete list misses the omitted partner. So through JSDoc:

- `@pattern ^a` + `@pattern ^b` on one property → **compiles**, emits `{"pattern": "^b"}`, but `is({value:"b"})` → **`false`** (the validator ANDs both patterns) — schema contradicts validator, no diagnostic. (Reproduced.)
- `@minimum 0` + `@exclusiveMinimum 5` → **compiles** (no `Exclusive` at all), while `number & Minimum<0> & ExclusiveMinimum<5>` is a compile error (#2154).

This is the campaign's signature shape: **one contract (tag exclusivity), two owners (the interface tag declarations and the comment factory's `Exclusive` lists), silently diverging.** #2154 fixed the check; the JSDoc factory's data was never aligned to it. It is a mirror of #2116/#2117 — a valid-looking input producing an inconsistent artifact with no error.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `MetadataCommentTagFactory.go` — the `Exclusive` field of each numeric/format/pattern JSDoc tag |
| **Affected** | any property whose constraints are written as JSDoc `@minimum`/`@maximum`/`@exclusiveMinimum`/`@exclusiveMaximum`/`@format`/`@pattern` in a forbidden combination — the emitted schema and the generated validator disagree, silently |
| **Not affected** | the type-tag path (`Minimum<0> & ...`), which #2154 enforces correctly; a single JSDoc tag per property |

| Case | Expected |
| --- | --- |
| `@pattern ^a` + `@pattern ^b` | compile error naming the conflict — as the type tags are |
| `@minimum 0` + `@exclusiveMinimum 5` | compile error |
| `@format email` + `@pattern ^x` | compile error (the interface makes format and pattern mutually exclusive) |
| a single `@minimum`, `@format`, `@pattern` | unchanged |
| the equivalent type-tag combinations | unchanged (#2154) |

## Approach

The invariant: **the JSDoc path and the type-tag path enforce the same exclusivity, because both derive it from the same authority — the interface `exclusive` declarations.** The durable fix is to source the comment factory's `Exclusive` lists from the tag declarations rather than re-typing them by hand, so they cannot drift again; #2154 already established that the interface `exclusive` is the authority. If a hand-written list is kept, complete every one to match the table above — but a hand-copied list is exactly what produced this divergence, so prefer single-sourcing.

Do not touch the interface `exclusive` declarations — they are correct and are the authority this enforces.

- **Positive:** every forbidden JSDoc constraint combination is a compile error naming the conflict, matching the type-tag path; a single tag per property is unchanged.
- **Negative — must fail against current code:** `@pattern ^a @pattern ^b` and `@minimum 0 @exclusiveMinimum 5` compile today and must become compile errors. Add both to `tests/test-error`.
- **Boundary:** every tag pair the interface declares mutually exclusive, in both orders, through JSDoc; and same-kind duplication of each.
- **Regression:** the type-tag path (#2154) is unchanged; a single JSDoc tag still works.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/factories/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-error start`.

## Coordination

- **Direct follow-up to #2154** (merged), which enforced this for type tags; this is the second owner it exposed but did not fix. Broader than a single tag: the sweep found the omission across all six numeric/format/pattern JSDoc tags, not just `@pattern`.
- Found by discovery round D4-C; the lead broadened it from the reported `@pattern` case to the full set after sweeping every JSDoc tag's `Exclusive` list against the interface authority.

samchon:
## Summary

Fixes #2182.

The tag-exclusivity contract that #2168 enforced for **type tags** (`Minimum<0> & ExclusiveMinimum<5>` is a compile error) was under-enforced for the same constraints written as **JSDoc comment tags**, because the comment factory's hand-written `Exclusive` lists had drifted from the `@typia/interface` tag declarations that are the authority.

At `origin/master` the only divergent lists are:

| JSDoc tag | factory `Exclusive` (before) | interface authority |
| --- | --- | --- |
| `@format` | `true` | `["format", "pattern"]` |
| `@pattern` | `["format"]` | `["format", "pattern"]` |

`@minimum`/`@maximum`/`@exclusiveMinimum`/`@exclusiveMaximum` already carried correct lists (the issue's "field absent" claim was against an earlier state). The consequence of the two divergent lists: `@pattern ^a` + `@pattern ^b` and `@format email` + `@pattern ^x` compiled through JSDoc while the type-tag equivalents are compile errors, emitting a schema that contradicts the generated validator with no diagnostic.

## Approach

Single-source every JSDoc tag's `Exclusive` value from one authority table in the comment factory, keyed by kind and mirroring `packages/interface/src/tags/*.ts`. Both the numeric parsers and the string `@format`/`@pattern` parsers now read their list from the one table, so a hand-copied list can no longer drift per-parser. The interface declarations are unchanged (they are the authority).

## Scope

- **Native Go only.** `packages/typia/native/core/factories/MetadataCommentTagFactory.go`.
- **No `packages/interface/src` change** — the `exclusive` declarations are the authority.
- **No `packages/typia/package.json` version bump** — release belongs to the maintainer.
- Regression fixtures added to `tests/test-error` for every interface-declared mutually-exclusive JSDoc pair (both orders) plus same-kind duplication.

## Verification

Local verification recorded in the campaign knowledge base; CI intentionally suspended per the issue-campaign flow (exact-SHA runs cancelled after each push).

### DIFF
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -207,8 +207,8 @@ var metadataCommentTagFactory_PARSER = map[string]metadataCommentTagFactory_pars
       Value    string
     }{Report: props.Report, Value: props.Value, Unsigned: true})
     return metadataCommentTagFactory_TagRecord{"array": {
-      {Name: "MinItems<" + props.Value + ">", Target: "array", Kind: "minItems", Value: value, Validate: props.Value + " <= $input.length", Exclusive: true, Schema: map[string]any{"minItems": value}},
-      {Name: "MaxItems<" + props.Value + ">", Target: "array", Kind: "maxItems", Value: value, Validate: "$input.length <= " + props.Value, Exclusive: true, Schema: map[string]any{"maxItems": value}},
+      {Name: "MinItems<" + props.Value + ">", Target: "array", Kind: "minItems", Value: value, Validate: props.Value + " <= $input.length", Exclusive: metadataCommentTagFactory_exclusive("minItems"), Schema: map[string]any{"minItems": value}},
+      {Name: "MaxItems<" + props.Value + ">", Target: "array", Kind: "maxItems", Value: value, Validate: "$input.length <= " + props.Value, Exclusive: metadataCommentTagFactory_exclusive("maxItems"), Schema: map[string]any{"maxItems": value}},
     }}
   },
   "minItems": func(props struct {
@@ -220,7 +220,7 @@ var metadataCommentTagFactory_PARSER = map[string]metadataCommentTagFactory_pars
       Unsigned bool
       Value    string
     }{Report: props.Report, Value: props.Value, Unsigned: true})
-    return metadataCommentTagFactory_TagRecord{"array": {{Name: "MinItems<" + props.Value + ">", Target: "array", Kind: "minItems", Value: value, Validate: props.Value + " <= $input.length", Exclusive: true, Schema: map[string]any{"minItems": value}}}}
+    return metadataCommentTagFactory_TagRecord{"array": {{Name: "MinItems<" + props.Value + ">", Target: "array", Kind: "minItems", Value: value, Validate: props.Value + " <= $input.length", Exclusive: metadataCommentTagFactory_exclusive("minItems"), Schema: map[string]any{"minItems": value}}}}
   },
   "maxItems": func(props struct {
     Report func(msg string) any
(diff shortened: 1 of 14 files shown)


samchon:
### Verification (rebased onto `origin/master` `f682316125`, head `8cb2c6eff4`)

CI intentionally suspended per the issue-campaign flow; exact-SHA runs cancelled after each push. Local verification (GOMAXPROCS=12; Go temp/cache redirected to a scratch dir because the D: drive was full mid-run):

| Suite | Result |
| --- | --- |
| `go test ../native/core/factories/...` | ok — includes new pin-test |
| `go test ../native/...` | ok — 14 packages, 0 fail |
| `go test -tags typia_native_internal ../native/core/factories/...` | ok |
| `tests/test-error` | Every fixture rejected: **66 files, 861 typia diagnostics** |
| `tests/test-typia-schema` | Success — **169 cases, 0 fail** |
| `tests/test-typia-automated` (solo) | Success — **3947 cases (442 CommentTag), 0 fail** |

**Fail-then-pass (test-error).** With the pre-fix lists restored (`@format` = `true`, `@pattern` = `["format"]`), 11 of the 12 new JSDoc fixtures already rejected; the one that compiled through was `error_tag_jsdoc_pattern_duplicated.ts` — `@pattern ^a @pattern ^b` emitting `{"pattern":"^b"}` while the validator ANDs both regexes. After single-sourcing, all 12 reject with the conflict named, e.g. `kind 'pattern' can't be used with 'Pattern<"^b">'` and `kind 'format' can't be used with 'Pattern<"^x">'` — the same wording the type-tag path produces.

Note: the issue's premise that `@minimum 0 @exclusiveMinimum 5` and `@format email @pattern ^x` "compile today" did not hold at `origin/master` — those were already enforced (their lists already contained the partner kind). The genuine defects were `@pattern` (missing its own kind → duplicate `@pattern` compiled) and `@format` (bare `true`, a masked divergence). The fixtures lock all 12 combinations regardless.

**Self-Review:** one complete solo round, no issues found. Every parser routes its `Exclusive` through one authority map keyed by kind; the shared `[]string` values are read-only in `Validate`; an unknown kind panics rather than emitting an unchecked tag.

PR #2187, opened 2026-07-18
fix(compare,plain): stop the object branch over-matching arrays and native instances

### LINKED ISSUE #2180: plain.prune crashes/corrupts and compare.equals reports two different natives equal, both from an object branch that over-matches arrays and native instances
## Problem

Two `plain.*`/`compare.*` programmers over-match with an object branch that also catches arrays and native instances, because the branch composition is unguarded — independent `if`s in one, an `OR` of all branches in the other. `clone` and `classify` handle the identical unions correctly with an `else`-if ladder, so the repository already contains the right shape twice over.

1. **`typia.plain.prune<T>` crashes or corrupts** on a union that mixes an object type with an array, tuple, or native instance.
2. **`typia.compare.equals`/`cover<native | object>` returns `true` for two different native instances.**

## Evidence

Verified on `master` (`5260e78433`), each with a control.

### 1 — `plain.prune` over a mixed union

`PlainPruneProgrammer.go` emits one **independent** `if` per union member (`:316-323`):

```go
for _, u := range unions {
  statements = append(statements, f.NewIfStatement(u.Is(), plainPruneProgrammer_to_statement(u.Value(), ...), nil))
}
```

The object member's guard is `IsObject` with **`CheckArray: false`** (`:295-297`) — it does not exclude arrays, and it matches native instances too. Because the `if`s are independent (no `else`), an array or native input matches both its own branch *and* the object branch, and both run. Observed:

- `prune<Uint8Array | { x: number }>` on a `Uint8Array` → **throws** `Cannot delete property '0'`.
- `prune<{ y }[] | { x }>` on `[{y},{y}]` → array elements **deleted** → `[null, null]`.
- `prune<Date | { x }>` on a `Date` → the Date's own props **stripped**.

Control: `plain.clone` over the identical unions is correct — it uses a conditional ladder (`PlainClassifyProgrammer.go:1088` `} else {`), so exactly one branch runs.

### 2 — `compare.equals` over `native | object`

`CompareEqualProgrammer.go`'s `schema` builds a branch per member — natives, tuples, arrays, **objects** (`g.objectCall`) — and returns `g.or(same(x,y), branches...)` (`:169`). For `Date | { timestamp: number }` it ORs `g.native("Date", …)` with `g.objectCall({timestamp}, …)`. The object comparator matches a `Date` (a Date *is* an object) and compares it by the declared keys of `{ timestamp }`, which a Date does not have, so both `x.timestamp` and `y.timestamp` are `undefined` and compare equal. The `OR` then accepts. Observed:

- `equals<Date | { timestamp: number }>(new Date(0), new Date(5000))` → **`true`**.
- Control: `equals<Date>(new Date(0), new Date(5000))` → `false` (correct).

This is a soundness defect: two genuinely different values are reported equal. `cover` over the same union has the same shape.

## Why nothing caught it, and the spec gate

- **`clone`/`classify` are the second producer that answers correctly** — both resolve these unions with an `else`-if ladder, so the divergence is invisible to any test that only exercises the correct owners.
- **Both fixes are pure tightening in the accept-wrong direction, so they pass the int64/spec gate.** Excluding arrays/natives from the object branch rejects no legitimate input: prune stops double-processing a value one branch already handled, and equals stops calling two different natives equal. No value that *should* be pruned or *should* compare equal is lost — verified by the `clone`/`classify` controls, which accept exactly the legitimate cases.

## Consequence surface

| | Affected | Not affected |
| --- | --- | --- |
| **1 `prune`** | any `prune<T>` where `T` unions an object with an array, tuple, or native — crash or silent corruption on valid input | pure-object types, pure-array types, and `clone`/`classify`, which ladder correctly |
| **2 `equals`/`cover`** | any `equals`/`cover<native | object>` — two different natives reported equal | `equals<Native>` alone; `equals<object>` alone; `clone`/`classify` |

| Case | Expected |
| --- | --- |
| `prune<Uint8Array \| {x}>` on a `Uint8Array` | the native is left intact, not deleted-into |
| `prune<{y}[] \| {x}>` on an array | array elements pruned as objects, not the array deleted |
| `prune<Date \| {x}>` on a `Date` | the Date left intact |
| `equals<Date \| {timestamp}>(Date(0), Date(5000))` | `false` |
| `equals<Date>` / `equals<{timestamp}>` alone | unchanged |
| every legitimate prune target / equal pair | unchanged |

## Approach

The invariant: **the object branch matches only plain objects — not arrays, not native instances — and each input is handled by exactly one branch.** The correct shape already exists in `clone`/`classify`: a conditional ladder where the array/tuple/native branches precede the object branch and are mutually exclusive with it. Bring `prune` and `compare.equals`/`cover` onto that shape, or give the object branch a guard that excludes what the earlier branches claim (`CheckArray` and a native check), so no input reaches two branches.

Do not fix this by special-casing `Date`/`Uint8Array` — the defect is structural (an object branch that swallows opaque instances), and the fix must cover every native the earlier branches recognize.

- **Positive:** the mixed unions above prune and compare correctly; pure-object and pure-array types are unchanged; `clone`/`classify` are untouched.
- **Negative — must fail against current code:** the `Uint8Array` prune throw, the array-element deletion, and the two-Dates-equal result must each be reproduced red, then fixed.
- **Boundary:** every native kind (Date, Uint8Array, Map, Set, RegExp, …) unioned with an object, in both directions; a union of an object with a *tuple*; nested unions.
- **Regression:** `clone` and `classify` over these unions stay correct (they are the control); no legitimate prune or equal case is lost.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/programmers/plain/... ../native/core/programmers/compare/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start` (solo).

## Coordination

- Two independent commits recommended (`prune` and `compare` are separate owners), each revertable alone.
- `plain.classify`/`plain.clone` are the reference shape and must not regress.
- Found by discovery round D4-A and confirmed by the lead against the branch-composition code and the `clone`/`classify` controls.

samchon:
## Intent

Fixes #2180. Two independent programmers over-match with an object branch that also catches arrays and native instances, because the branch composition is unguarded. `clone`/`classify` handle the identical unions correctly with an `else`-if ladder — that is the control and the fix shape.

This is a draft claim reservation. Implementation follows as **two independent commits**, each revertable alone.

## Scope

1. **`plain.prune<T>`** over a union mixing an object with an array/tuple/native crashes or corrupts. `PlainPruneProgrammer.go` emits independent `if`s (no `else`) per union member; the object branch's guard is `IsObject` with `CheckArray: false`, so an array/native matches both its own branch and the object branch, and both run.
   - `prune<Uint8Array | {x}>` on a `Uint8Array` throws `Cannot delete property '0'`.
   - `prune<{y}[] | {x}>` on `[{y},{y}]` → `[null, null]`.
   - `prune<Date | {x}>` strips the Date's props.

2. **`compare.equals`/`cover<native | object>`** returns `true` for two different natives. `CompareEqualProgrammer.go` ORs all member branches; the object branch matches a `Date` and compares it by the absent declared keys of `{timestamp}` (both `undefined` → equal), so `equals<Date | {timestamp}>(Date(0), Date(5000))` → `true`.

## Invariant

The object branch matches only plain objects — not arrays, not native instances — and each input is handled by exactly one branch. `prune` adopts the `clone`/`classify` else-if ladder; `compare.equals`/`cover` guards its object branch against the sibling natives the union declares. Pure tightening in the accept-wrong direction: no legitimate prune target or equal pair is lost.

## Verification (planned)

- Runtime reproductions in `packages/typia/native/cmd/ttsc-typia` for all four failures, red before the fix and green after.
- `go -C packages/typia/test test -count=1 -p 2 ../native/core/programmers/plain/... ../native/core/programmers/compare/...`; `../native/...`.
- `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start` (solo).
- Regression: `clone`/`classify` over these unions stay correct (the control).

## Coordination

Native Go. No `packages/typia/package.json` version bump (releases belong to the maintainer). Campaign CI is deliberately suspended; verification is local per the issue-campaign development skill.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/compare_native_object_overmatch_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/compare_native_object_overmatch_transform_test.go
@@ -1,0 +1,161 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "testing"
+)
+
+// TestCompareNativeObjectOvermatchTransform pins equals/cover over native|object.
+//
+// CompareEqualProgrammer.schema ORs a branch per union member and the object
+// branch (`_eo`) guards only with `typeof === "object" && !Array.isArray`,
+// which a native instance also passes. For `Date | { timestamp: number }` the
+// object branch compares two Dates by the absent `timestamp` key — both
+// `undefined`, so it reports them equal, and the OR accepts even though the
+// Date branch (comparing getTime) correctly said not-equal. The object branch
+// must never claim a value a sibling native branch owns. This test drives Date,
+// Uint8Array, and RegExp unioned with an object, in both orders, and requires
+// two different natives to compare not-equal while every legitimate pair and
+// the native-alone / object-alone controls are unchanged.
+//
+//  1. Transform equals/cover factories over native|object and object|native.
+//  2. Require two different natives to compare not-equal (equals and cover).
+//  3. Require same natives, plain-object pairs, and the native-alone control
+//     to keep their correct equal / not-equal results.
+func TestCompareNativeObjectOvermatchTransform(t *testing.T) {
+  project := compareNativeObjectOvermatchProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("compare native/object over-match transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  compareNativeObjectOvermatchRunRuntimeCases(t, project, out)
+}
+
+func compareNativeObjectOvermatchProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "compare-native-object-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(compareNativeObjectOvermatchTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(compareNativeObjectOvermatchSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func compareNativeObjectOvermatchRunRuntimeCases(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  runtimeJS := ttscTypiaTestRewriteCommonJS(t, js)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(runtimeJS), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
+  if err := os.WriteFile(runner, []byte(compareNativeObjectOvermatchRuntimeRunner), 0o644); err != nil {
+    t.Fatalf("write runtime runner: %v", err)
+  }
+  cmd := exec.Command(node, runner)
+  cmd.Dir = runtimeDir
+  output, err := cmd.CombinedOutput()
+  if err != nil {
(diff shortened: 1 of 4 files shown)


samchon:
## Self-Review (solo, round 1 — clean)

Complete fresh pass over the whole change surface (both source hunks, both new tests, emitted output). No finding required a change; the round is clean.

**Reproductions observed red on pre-fix code, green after:**
- `prune<Uint8Array | {x}>` → `TypeError: Cannot delete property '0' of [object Uint8Array]` (throw)
- `prune<{y}[] | {x}>` → `[null, null]` (array slots deleted)
- `prune<IBox | [IElem, IElem]>` → `[null, null]` (tuple slots deleted)
- `prune<Date | {x}>`, `Map`, `Set`, `RegExp` → native own-state stripped
- `equals<Date | {timestamp}>(new Date(0), new Date(5000))` → `true` (and `cover` likewise)

Each corruption was reproduced independently (a catch-and-continue harness against the reverted prune source printed RED for all seven prune cases, including the exact `holes: [null,null]`), then all pass post-fix.

**Correctness of the fix shape:**
- `plain.prune` now emits a mutually exclusive `if / else if` ladder with the object arm last, so each runtime value is pruned by exactly one arm — the shape `clone`/`classify` already use. Is()/Value() are evaluated in reverse per union, matching the clone ladder; helper-name derivation is index-based, so order is immaterial.
- `compare.equals`/`cover` guards the object arm with `!(x instanceof N) && !(y instanceof N)` for every sibling native `N` the union declares. Emitted shape confirmed: `... || ((!(x instanceof Date) && !(y instanceof Date)) && _eo0(x, y))`.

**No legitimate case lost (pure tightening):**
- Non-native unions emit byte-identical output (`notNatives` returns empty; the existing `TestCompareEqualCoverTransform` object/union/recursive fixtures are unchanged and green).
- The only results that change are accept-wrong ones (a native treated as the object member). Analyzed the cover cross-member edge (`cover(new Date(5), {})`): pre-fix `true` came from the object arm mis-treating the Date as the object member; post-fix `false` is correct, and a legitimate cover (both operands the object member) is unaffected.

**Controls untouched:** `PlainCloneProgrammer.go` / `PlainClassifyProgrammer.go` are not in the diff. `test-typia-schema` (native clone/classify union tests) and `test-typia-automated` (full prune/clone/classify/equals matrix) both pass.

**Integrity:** two disjoint-file commits, each revertable alone. No `packages/typia/package.json` bump (release-owned). `go vet` clean; 2-space Go style matches the repo. Campaign CI suspended; local verification is the gate.

PR #2188, opened 2026-07-18
fix(typia): decode absent required protobuf scalar to its proto3 default

### LINKED ISSUE #2179: protobuf.decode seeds an absent required number/boolean/bigint to undefined instead of its proto3 default, so assertDecode throws on valid peer-encoded messages
## Problem

`typia.protobuf.decode` / `assertDecode` / `validateDecode` seed an **absent required `number`, `boolean`, or `bigint`** field to `undefined` instead of its proto3 typed default (`0` / `false` / `0n`), while an absent required **`string`** correctly defaults to `""`. So decoding a valid, all-defaults proto3 message throws on data that every conformant peer produces.

## Evidence

Verified on `master` (`5260e78433`). Owner: `protobufDecodeProgrammer_write_property_default_value`, `packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go:274-301`:

```go
} else if protobufDecodeProgrammer_hasStringDefault(value) {
  expression = f.NewStringLiteral("", shimast.TokenFlagsNone)   // string → ""
} else if protobufDecodeProgrammer_hasDynamicObject(value) {
  expression = f.NewObjectLiteralExpression(nil, false)
} else {
  expression = f.NewIdentifier("undefined")                    // number / boolean / bigint → undefined
}
```

`hasStringDefault` returns true for an atomic `string` (line 605), giving `""`. There is **no matching branch for number/boolean/bigint**, so they fall to the final `else` → `undefined`.

In proto3, a singular scalar equal to its default is **not written on the wire**, and an absent scalar decodes to its type default: `number → 0`, `boolean → false`, `bigint → 0n`, `string → ""`. So decoding the empty message `{}` for a type `{ n: number; b: boolean; s: string }` must yield `{ n: 0, b: false, s: "" }`. typia yields `{ n: undefined, b: undefined, s: "" }`, and `assertDecode` then **throws on valid data** (`invalid type on $input.n`).

This is the mirror of #2116 (a valid message wrongly rejected).

## Why nothing caught it, and why the fix is safe

- **The oracle is a copy checked against its other copy.** `test_protobuf_decode.ts` decodes only typia's *own* encode output. typia's encoder writes every required non-nullable field **unconditionally, even at the scalar default** — confirmed at `ProtobufEncodeProgrammer.go:454-456`, where the required-non-nullable wrapper is the identity `return block`. So the field is always present in typia↔typia traffic and the absent case never arises in the test.
- **The fix is a pure loosening, and it passes the spec gate.** Only a *peer* encoder (Go/Java/Python/C++), which omits default-valued fields per standard proto3, produces the absent field. Changing `undefined` → the typed default aligns typia with proto3 and rejects nothing it should accept; typia↔typia is byte-unchanged because typia never omits the field.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `ProtobufDecodeProgrammer.go:297-298` (the `else → undefined` for required number/boolean/bigint) |
| **Affected** | any `decode`/`assertDecode`/`validateDecode` of a message a conformant non-typia peer encoded with a default-valued required scalar — thrown/rejected as invalid |
| **Not affected** | typia↔typia round-trips (the encoder always writes the field); absent required `string` (already `""`); optional fields (correctly `undefined`); nullable fields (correctly `null`) |

| Case | Expected |
| --- | --- |
| decode `{}` for required `number` | `0` |
| decode `{}` for required `boolean` | `false` |
| decode `{}` for required `bigint` | `0n` |
| decode `{}` for required `string` | `""` — unchanged |
| optional scalar absent | `undefined` — unchanged |
| nullable scalar absent | `null` — unchanged |
| a peer message with all scalars at default | `assertDecode` passes |

## Approach

The invariant: **an absent required scalar decodes to its proto3 typed default, for every scalar kind, not just `string`.** Add the numeric/boolean/bigint defaults beside the existing `string → ""` branch, reading the kind the same way `hasStringDefault` does. Keep the optional (`undefined`), nullable (`null`), array (`[]`), map (`new Map`), and native (`Uint8Array`) cases exactly as they are.

`bigint` needs `0n` (a `BigInt` literal), not `0`. Check whether a constrained scalar (e.g. `number & Minimum<1>`) at default `0` should then fail validation with a *minimum* error rather than a *type* error — that is the correct, more precise diagnostic, and it is the peer's out-of-spec message, correctly reported.

- **Positive:** decoding `{}` yields `0`/`false`/`0n`/`""` per kind; a peer message with all-default required scalars passes `assertDecode`.
- **Negative — must fail against current code:** the same decode currently produces `undefined` and `assertDecode` throws; reproduce that, then fix.
- **Boundary:** optional stays `undefined`, nullable stays `null`, `string` stays `""`; a constrained scalar at default is judged by its constraint, not by "invalid type".
- **Regression:** the reader contracts from #2084, #2090, #2116, #2119, #2135 are unchanged — this touches only the default-value seed, not the reader. typia↔typia encode/decode is byte-identical.
- **Replace the oracle:** the test must decode a message with the field **absent** (a hand-built or peer-built payload), not only typia's own encode output — otherwise the case that fails today stays untested.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/programmers/protobuf/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start` (solo).

## Coordination

- Same file family as #2116/#2119/#2135; the reader contracts there must not shift.
- Found by discovery round D4-B and reproduced by the lead (the code path and the encoder's unconditional-write are both confirmed on `5260e78433`).

samchon:
Fixes #2179.

## Problem

`typia.protobuf.decode` / `assertDecode` / `validateDecode` seed an absent **required** `number` / `boolean` / `bigint` field to `undefined` instead of its proto3 typed default (`0` / `false` / `0n`), while an absent required `string` already defaults to `""`. Decoding a valid, all-defaults proto3 message (what any conformant peer produces by omitting default-valued fields) then throws `invalid type`.

Owner: `protobufDecodeProgrammer_write_property_default_value` in `packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go`. The `string -> ""` branch is `hasStringDefault`; there was no numeric / boolean / bigint equivalent, so those kinds fell through to the final `else -> undefined`.

## Fix

Add a number -> `0`, boolean -> `false`, bigint -> `0n` (a `BigInt` literal, not `0`) branch beside the existing `string -> ""`, reading the atomic kind the same way `hasStringDefault` reads it. Optional stays `undefined`, nullable stays `null`, array stays `[]`, map stays `new Map`, native stays `Uint8Array`.

## Why it is safe

typia's encoder writes every required non-nullable field unconditionally, even at the scalar default, so typia<->typia traffic never omits the field: the seed is always overwritten and the decoded object is unchanged, and the encoder is untouched so the wire bytes are byte-identical. Only a peer encoder (Go/Java/Python/C++) omits a default-valued field, and for those `undefined -> typed default` is a pure loosening that aligns typia with proto3. A constrained scalar (e.g. `number & Minimum<1>`) at default `0` now fails validation with the correct, more precise `minimum` error instead of an `invalid type` error.

## Scope

- Native Go only. No `packages/typia/package.json` version bump (releases belong to the maintainer).
- Same file family as #2084 / #2090 / #2119 / #2135; the reader contracts there are unchanged (this touches only the default-value seed, not the reader).

Local verification and the fail-then-pass regression evidence are recorded as a comment after implementation.

Campaign implementation pull request: automatic CI is deliberately suspended and exact-SHA runs are cancelled per push under the campaign's CI-suspension policy.

### DIFF
--- a/packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go
+++ b/packages/typia/native/core/programmers/protobuf/ProtobufDecodeProgrammer.go
@@ -292,6 +292,12 @@ func protobufDecodeProgrammer_write_property_default_value(value *schemametadata
     )
   } else if protobufDecodeProgrammer_hasStringDefault(value) {
     expression = f.NewStringLiteral("", shimast.TokenFlagsNone)
+  } else if protobufDecodeProgrammer_hasAtomicDefault(value, "number") {
+    expression = f.NewNumericLiteral("0", shimast.TokenFlagsNone)
+  } else if protobufDecodeProgrammer_hasAtomicDefault(value, "boolean") {
+    expression = f.NewKeywordExpression(shimast.KindFalseKeyword)
+  } else if protobufDecodeProgrammer_hasAtomicDefault(value, "bigint") {
+    expression = f.NewBigIntLiteral("0n", shimast.TokenFlagsNone)
   } else if protobufDecodeProgrammer_hasDynamicObject(value) {
     expression = f.NewObjectLiteralExpression(nil, false)
   } else {
@@ -620,6 +626,21 @@ func protobufDecodeProgrammer_hasStringDefault(value *schemametadata.MetadataSch
   }
   for _, template := range value.Templates {
     if len(template.Row) == 1 && template.Row[0].GetName() == "string" {
+      return true
+    }
+  }
+  return false
+}
+
+// protobufDecodeProgrammer_hasAtomicDefault reports whether an absent required
+// field of this schema decodes to the proto3 typed default of the given atomic
+// kind. proto3 does not write a singular scalar equal to its default, so an
+// absent number/boolean/bigint must seed 0/false/0n the same way an absent
+// string seeds "" through protobufDecodeProgrammer_hasStringDefault; the atomic
+// kind is read exactly as hasStringDefault reads its own atomic branch.
+func protobufDecodeProgrammer_hasAtomicDefault(value *schemametadata.MetadataSchema, typ string) bool {
+  for _, atomic := range value.Atomics {
+    if atomic.Type == typ {
       return true
     }
   }
--- a/packages/typia/native/core/programmers/protobuf/protobuf_programmer_helper_coverage_test.go
+++ b/packages/typia/native/core/programmers/protobuf/protobuf_programmer_helper_coverage_test.go
@@ -267,6 +267,32 @@ func TestProtobufProgrammerHelperCoverage(t *testing.T) {
 			t.Fatal("protobuf decode default value returned nil")
 		}
 	}
+	// #2179: an absent required number/boolean/bigint seeds its proto3 typed
+	// default (0/false/0n), just as an absent required string seeds "". Each
+	// seed is wrapped in an `as any`, so unwrap it and pin the literal so a
+	// regression to `undefined` cannot pass.
+	for _, expected := range []struct {
+		typ  string
+		kind shimast.Kind
+		text string
+	}{
+		{typ: "number", kind: shimast.KindNumericLiteral, text: "0"},
+		{typ: "boolean", kind: shimast.KindFalseKeyword, text: ""},
+		{typ: "bigint", kind: shimast.KindBigIntLiteral, text: "0n"},
+		{typ: "string", kind: shimast.KindStringLiteral, text: ""},
+	} {
+		seed := protobufDecodeProgrammer_write_property_default_value(protobufProgrammerAtomic(expected.typ), emit)
+		if seed == nil || seed.Kind != shimast.KindAsExpression {
+			t.Fatalf("protobuf %s default should be an as-expression", expected.typ)
+		}
+		inner := seed.Expression()
+		if inner == nil || inner.Kind != expected.kind {
+			t.Fatalf("protobuf %s default seeded the wrong node kind", expected.typ)
+		}
+		if expected.text != "" && inner.Text() != expected.text {
+			t.Fatalf("protobuf %s default seeded %q instead of %q", expected.typ, inner.Text(), expected.text)
+		}
+	}
 	if protobufDecodeProgrammer_write_object_function_body(protobufDecodeProgrammer_objectBodyProps{
 		Context: nativecontext.ITypiaContext{},
 		Condition: protobufDecodeProgrammer_factory.NewKeywordExpression(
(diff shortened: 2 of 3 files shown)


samchon:
## Local verification

Implementation SHA `4e8f00665ac89b37727e05a494428ad8deb8a0ac`. Native Go build temp redirected to a spare volume (`GOTMPDIR`) to work around shared-box disk pressure; the workspace `GOCACHE` and the ttsc plugin cache were untouched.

### Fail-then-pass on the new regression test

The new `test_protobuf_decode_absent_scalar_defaults` decodes an all-defaults proto3 message as the empty payload (`new Uint8Array(0)`), i.e. the absent-field case a conformant peer produces — not typia's own encode output.

- Pre-fix (`--include absent_scalar`): FAILS — `decode: absent required number seeded undefined instead of 0.`
- Post-fix: `test_protobuf_decode_absent_scalar_defaults: 2 ms — Success`

Observed decode of the empty payload:

| kind | pre-fix | post-fix (proto3 default) |
| --- | --- | --- |
| required `number` | `undefined` | `0` |
| required `boolean` | `undefined` | `false` |
| required `bigint` | `undefined` | `0n` |
| required `string` | `""` | `""` (unchanged) |
| optional scalar | `undefined` | `undefined` (unchanged) |
| nullable scalar | `null` | `null` (unchanged) |
| `number & Minimum<1>` | error on `value: undefined` (invalid type) | error on `value: 0` (minimum) |

### typia↔typia is byte-identical

The encoder is untouched. typia encodes an all-default `{ n, b, big, s }` to `[9,0,0,0,0,0,0,0,0, 16,0, 24,0, 34,0]` — every required field is written even at its default — so the field is always present on the wire and the decode seed is always overwritten; the round-trip decode is unchanged. `test-typia-automated` (the protobuf encode/decode matrix) passes.

### Commands

- `go -C packages/typia/test test -tags typia_native_internal -count=1 -p 2 ../native/core/programmers/protobuf/...` → `ok` (the `write_property_default_value` oracle, extended to pin `0`/`false`/`0n`/`""`)
- `go -C packages/typia/test test -count=1 -p 2 ../native/...` → all `ok` (incl. `cmd/ttsc-typia` command-level transform tests)
- `go -C packages/typia/test test -count=1 -p 2 ./helpers/protobuf/... ./contract/...` → all `ok`
- `pnpm --filter ./tests/test-typia-schema start` → `Success` (all reader-contract tests #2084/#2090/#2119/#2135 unchanged)
- `pnpm --filter ./tests/test-typia-automated start` → `Success` (solo)
- `pnpm --filter ./tests/test-feature-identity start` → `Success`

No `packages/typia/package.json` version bump (releases belong to the maintainer). No `pnpm format` (campaign rule). Automatic CI for this campaign commit was cancelled per the exact-SHA gate.

Issue #2045, opened 2026-07-16
Decimal MultipleOf and numeric random generation disagree


samchon:
## Problem

Decimal `multipleOf` validation uses JavaScript remainder equality, rejecting valid values, while random numeric generation both derives dimensionally wrong default ranges and can emit values that fail its own constraint.

## Evidence

The documented `MultipleOf<0.01>` currency example rejects `0.03` and `1.01` because binary floating remainder is nonzero. Type tags, JSDoc tags, `OpenApiNumberValidator`, and schema coverage use the same arithmetic assumption. `_randomNumber` rounds with `value - value % multipleOf`, which can turn `0.5` into a floating `0.49...` that the validator still rejects.

Separately, `_randomInteger` and `_randomNumber` multiply the entire fallback bound expression: `{ maximum: 1000, multipleOf: 2 }` derives minimum 1800 and throws although the domain is satisfiable. Closed PR #1864 contains only partial random-bound observations; #554 is a different decimal-type feature request. No open implementation claim exists.

JSON Schema defines validity by whether division by `multipleOf` produces an integer, independent of binary representation artifacts.

Normative reference: [JSON Schema Validation 2020-12, numeric validation](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.2.1).

## Consequence surface

Runtime/native type and JSDoc validators, OpenAPI/LLM validators, coverage, integer/number random generation, one-sided/inclusive/exclusive bounds, negative/large/small decimals, and random round-trip guarantees.

## Approach

Create one finite decimal divisibility invariant based on normalized decimal/rational arithmetic or an equivalently justified algorithm. Derive fallback bounds in value units before snapping to multiples; generated values must always revalidate.

## Acceptance and verification

- Cover 0.01/0.1 decimals, integers, negative values, scientific notation, large/small magnitudes, and near misses.
- Cover all one/two-sided and exclusive bound combinations with `multipleOf` greater/less than one.
- Run deterministic random loops and require every result to validate; impossible ranges fail promptly and clearly.
- Keep type/JSDoc/OpenAPI/coverage/random surfaces in parity and run focused native/runtime plus full tests.

## Coordination

The corrected divisibility helper should be reused by the companion OpenAPI schema-coverage issue from this campaign.

PR #2192, opened 2026-07-18
fix(native): honor minLength/maxLength on string leaves with a pattern or format in random

### LINKED ISSUE #2189: random<T> produces values is<T> rejects: the generator drops minLength/maxLength when a string leaf also has a pattern or format
## Problem

`typia.random<T>()` produces values that `typia.is<T>()` rejects whenever a `string` leaf combines a `pattern` or `format` with a `minLength`/`maxLength`. The generator's format and pattern branches drop the length information, so a satisfiable constrained string is generated at the wrong length and fails its own validator. This contradicts the documented contract that `random` honors satisfiable length tags.

## Evidence

Verified on `master` (`903f057e3d`). Owner: `randomProgrammer_compose_atomic`, `RandomProgrammer.go:615-660`.

```go
if raw, ok := schema["format"].(string); ok && raw != "" {
  // ... returns Internal: "randomFormat"+pascal(raw)   — NO Arguments: no length passed
}
if pattern, ok := schema["pattern"].(string); ok && pattern != "" {
  // ... returns Internal: "randomPattern", Arguments: [ RegExp ]   — only the regex, no length
}
// ... only the plain-string fall-through passes the full schema:
return { Internal: "random"+capitalize(atomicType), Arguments: [ LiteralFactory.Write(schema) ] }
```

The `format` branch passes **nothing** about length; the `pattern` branch passes only the `RegExp`. Only a plain string (no format, no pattern) forwards the full schema, which is what carries `minLength`/`maxLength`. So a string leaf with a format or pattern **and** a length constraint generates without knowing the length.

Reproduced end to end through real `ttsx`, 2000 draws each, each type satisfiability-proven (a hand-crafted valid value passes `is`):

| Type | `random` values that `is` rejects |
| --- | --- |
| `string & Pattern<"^[0-9]+$"> & MinLength<10>` | 177 / 2000 |
| `string & Pattern<"^[0-9]+$"> & MaxLength<3>` | 1944 / 2000 |
| `string & Format<"email"> & MinLength<40>` | 2000 / 2000 |
| `string & Pattern<"^[0-9]+$">` (control) | 0 / 2000 |
| `string & MinLength<10>` (control) | 0 / 2000 |

This directly contradicts the guide, `website/src/content/docs/random.mdx:17` — *"It honors satisfiable type tags ... `tags.MaxLength<8>` will cap a string's length"* — and `:21` — *"Explicit constraints still win."*

## Consequence surface

| | |
| --- | --- |
| **Owner** | `RandomProgrammer.go:615-660` — the `format` and `pattern` branches of `compose_atomic` |
| **Affected** | `typia.random<T>()` for any string leaf combining a format or pattern with `minLength`/`maxLength` — the generated value fails `is`/`assert`/`validate` of the same type |
| **Not affected** | format-only, pattern-only, and length-only strings (the controls pass); numbers; non-string leaves |

Same generator↔validator divergence class typia has already fixed for numbers (`#2045`, `#2159`): the generator and the validator are two owners of one contract, and only the validator was right.

| Case | Expected |
| --- | --- |
| `Pattern<"^[0-9]+$"> & MinLength<10>` | a digit string of length ≥ 10 that `is` accepts |
| `Pattern<...> & MaxLength<3>` | a matching string of length ≤ 3 |
| `Format<"email"> & MinLength<40>` | an email of length ≥ 40 |
| a genuinely unsatisfiable combination (e.g. `Format<"email"> & MaxLength<3>`) | rejected at compile time, or a documented throw — the same treatment the guide already gives unsatisfiable recursion |
| format-only / pattern-only / length-only | unchanged (already correct) |

## Approach

The invariant: **`random` produces a value its own `is` accepts, for every satisfiable combination of a string's tags — length included.**

The length information must reach the format and pattern generators. That is more than threading an argument, because a format or pattern generator cannot in general hit an arbitrary target length while staying valid — an email of length ≥ 40 is satisfiable, but the generator has to construct one, and `Format<"email"> & MaxLength<3>` is *unsatisfiable* and must be rejected the way the guide already rejects unsatisfiable recursion, not silently produced. So the fix has two parts: pass the length bounds through, and decide per-branch how to honor them (regenerate-until-fit is not acceptable if it can loop; prefer a length-aware construction, and a compile-time or throwing rejection for the unsatisfiable case).

Check the runtime `_randomFormat*`/`_randomPattern` helpers — if they do not accept length bounds today, that capability is part of the fix. And confirm whether the emitter can *prove* unsatisfiability at compile time (as it does for recursion) or must throw at runtime.

This is a generator fix, not a validator change — it rejects nothing the validator accepts, so it passes the spec gate trivially; the only new rejection is the genuinely-unsatisfiable combination, which should never have produced a value.

- **Positive:** for each satisfiable row above, a large sample of `random<T>()` values all pass `is<T>()`.
- **Negative — must fail against current code:** the three failing rows produce `is`-rejected values today; reproduce with a sample and a nonzero failure count, then fix to zero.
- **Boundary:** an exactly-`minLength` and exactly-`maxLength` value; a pattern that fixes the length (`^[0-9]{5}$`) combined with a conflicting `MinLength`; an unsatisfiable format+length combination (must be rejected, not looped or silently produced).
- **Regression:** format-only, pattern-only, and length-only strings unchanged; numbers and other leaves unchanged; `random`'s existing recursion/termination behavior unchanged.
- **Commands:** `go -C packages/typia/test test -count=1 -p 2 ../native/core/programmers/...`; `pnpm --filter ./tests/test-typia-schema start`; `pnpm --filter ./tests/test-typia-automated start` (solo). The decisive test is a `random`-then-`is` round trip over the combined-tag matrix — the existing automated suite fuzzes random-vs-is but did not cover format/pattern combined with length.

## Coordination

- Native Go, plus possibly the `_randomFormat*`/`_randomPattern` runtime helpers if they must learn length. **No `packages/typia/package.json` version bump** — releases belong to the maintainer.
- Same class as `#2045` and `#2159` (generator↔validator divergence); the validator is correct and is the reference.
- Found by discovery round D5-B, reproduced end to end, and confirmed by the lead against the branch that drops the length information.

samchon:
Fixes #2189.

## Intent

`typia.random<T>()` produces values `typia.is<T>()` rejects whenever a `string` leaf combines a `pattern`/`format` with `minLength`/`maxLength`, because `randomProgrammer_compose_atomic` routes the format branch with no length information and the pattern branch with only the `RegExp`. Only the plain-string fall-through forwards the full schema that carries the length bounds.

This PR threads the length bounds through to the pattern and format generators and makes those generators honor them: length-aware construction where the format/pattern structure allows it, and a runtime throw for a genuinely unsatisfiable combination (matching the existing `_randomInteger` convention that already throws when `minimum > maximum`).

## Scope

- Native `RandomProgrammer.go`: forward `minLength`/`maxLength` to the pattern and format branches (only when present, so format-only / pattern-only emission stays byte-identical).
- Runtime `_randomPattern` and the `_randomFormat*` helpers: accept optional length bounds and honor them.

## Invariant

`random` produces a value its own `is` accepts, for every satisfiable combination of a string's tags — length included. The only new rejection is the genuinely-unsatisfiable combination, which is thrown rather than silently produced.

This is a generator fix, not a validator change: the validator is the reference and is untouched. Same class as #2045 / #2159.

## Local verification

Pending — will be recorded on the thread as the batch completes. No `packages/typia/package.json` version bump (releases belong to the maintainer). Campaign CI is cancelled per the issue-campaign protocol; local verification is the gate.


samchon:
## Self-Review round 1

Complete solo round over the whole change surface (native transform, 23 runtime helpers, shared length module, regression test, native test, doc).

### Findings and remediation (pushed in `465138ded4`)

1. **Min-only growable formats produced a fixed length.** With only a lower bound (`MinLength<40>` and no max), the growable formats (`email`/`url`/`hostname`/`json-pointer`/`password`) collapsed to exactly `minLength` characters, whereas `_randomString` deliberately spreads a min-only string over a `+range`. Aligned the growable helpers to spread the segment across the fallback range so lengths vary. Still satisfies the lower bound.
2. **Coverage gap for the other growable helpers.** The first round-trip matrix only pinned `email` among the growable formats. Extended it with `hostname` (a `MinLength<80>` that forces the multi-label builder past the 63-char label cap), `url`, `password`, and an `email` shrink window near the format minimum, so every growable helper is pinned by `is<T>()`.

### Verified as sound (no change needed)

- Format-only, pattern-only, and length-only emission stays byte-identical: the native `randomProgrammer_length_argument` returns `nil` when no length tag is present, so no extra argument is emitted and every runtime helper called with no argument runs its original path. Confirmed by `test-typia-automated` (Success) which fuzzes random↔is over the format templates.
- Unsatisfiable combinations throw at runtime, matching `_randomInteger`'s existing impossible-range throw; the negative rows (`Format<"email"> & MaxLength<3>`, `Pattern<"^[0-9]{5}$"> & MinLength<10>`, `Format<"uuid"> & MaxLength<10>`) assert the throw.

### Documented narrow scoping (intentionally not "fixed")

Two structured formats commit to a fixed prefix the generator has always used, so a *tiny* upper bound below that prefix throws rather than switching shape:

- `Format<"url"> & MaxLength<11>` — the builder emits `https://…` (min 12 chars); an http URL of 11 is satisfiable but out of this generator's shape.
- `Format<"json-pointer"> & MaxLength<19>` — the builder keeps the `/components/schemas/` prefix (20 chars).

Both are strictly better than the pre-fix behavior (which silently produced an `is`-rejected value) and lie outside any realistic constraint. They throw honestly instead of emitting an invalid value.

### Verification

`go -C packages/typia/test test -count=1 -p 2 ../native/...` ok; `pnpm --filter ./tests/test-typia-schema start` Success; `pnpm --filter ./tests/test-typia-automated start` Success (solo); `test-feature-identity` Success. Every satisfiable row draws 1000 samples with zero `is` failures; the three unsatisfiable rows throw.

PR #2191, opened 2026-07-18
fix(native): derive notation snake/kebab/camel/pascal runtime from the *Case<T> oracle on underscore keys

samchon:
## Scope

Fixes #2190.

`typia.notations.snake`/`kebab` produce object keys that violate their declared `SnakeCase<T>`/`KebabCase<T>` type when a key combines an underscore with an internal case boundary: `snake<{ fooBar_baz }>` returns runtime `foobar_baz` while the type declares `foo_bar_baz`, so reading `.foo_bar_baz` is `undefined` at runtime — the exact soundness hole #2186 fixed for `pascal`, left on its siblings. `camel`/`pascal` share a narrower form on trailing-underscore keys.

The invariant: the runtime output equals `SnakeCase<T>`/`KebabCase<T>`/`CamelCase<T>`/`PascalCase<T>` for every key. The `*Case<T>` type declarations in `packages/interface/src` are the authority and stay untouched; the runtime helpers (`packages/typia/src/internal/_notation*`) and their byte-identical Go compile-time twins (`NotationGeneralProgrammer.go`) are the defect and are changed together.

## Plan

- Regression test asserting all three producers agree — TS runtime, Go compile-time emit, and the `*Case<T>` type — over a matrix (`fooBar_baz`, `openAI_key`, `HTTP_fooBar`, `fooBar`, `fooBar_`, `_fooBar`, `userID`, `a_b_c`, `MAX_COUNT`), observed red first.
- Fix the TS runtime helper and its Go twin for snake/kebab, and the camel/pascal trailing-underscore case, each derived from its own `*Case<T>` oracle.

## Status

Draft — implementation in progress. No `packages/typia/package.json` version bump (releases belong to the maintainer). Campaign CI is suspended per the issue-campaign flow; local verification recorded on merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Fg7AePEdpjaxHrmVffuSjy

Issue #2190, opened 2026-07-18
notations.snake/kebab violate their declared SnakeCase/KebabCase type on underscore+camelCase keys: the exact soundness hole #2186 fixed for pascal, left on its siblings


samchon:
## Problem

`typia.notations.snake` and `kebab` produce object keys that violate their own declared `SnakeCase<T>`/`KebabCase<T>` return type whenever a key combines an underscore with an internal case boundary. `snake<{ fooBar_baz: number }>(x).foo_bar_baz` type-checks and is `undefined` at runtime — the exact soundness hole #2186 just fixed for `pascal`, left unfixed on its siblings. (`camel`/`pascal` share a narrower form of it on trailing-underscore keys.)

## Evidence

Verified on `master` (`903f057e3d`), by direct runtime invocation:

```
snake("fooBar_baz") → "foobar_baz"    SnakeCase<"fooBar_baz"> → "foo_bar_baz"
snake("openAI_key") → "openai_key"    SnakeCase<"openAI_key"> → "open_ai_key"
kebab("fooBar_baz") → "foobar-baz"    KebabCase<"fooBar_baz"> → "foo-bar-baz"
snake("fooBar")     → "foo_bar"        (correct — no underscore)
snake("userID")     → "user_id"        (correct)
```

Owner: `packages/typia/src/internal/_notationSnake.ts` / `_notationKebab.ts`, and their byte-identical Go compile-time twins in `NotationGeneralProgrammer.go:1370,1423`. When the key contains an internal `_`, the runtime does `str.split("_").map(s => s.toLowerCase()).join("_")` — it lowercases each segment **atomically** and never applies the case-boundary walk it uses for underscore-free keys. So `["fooBar","baz"]` becomes `["foobar","baz"]` and the camelCase boundary inside `fooBar` is lost. The type-level `SnakeCase`/`KebabCase` walk the key character by character, inserting a separator at every case boundary and keeping existing underscores, so they produce `foo_bar_baz`.

`snake<{ fooBar_baz: number }>({ fooBar_baz: 1 })` therefore returns `{ foobar_baz: 1 }`, and reading the typed key `.foo_bar_baz` is `undefined`. This is a soundness hole: the value disagrees with its declared type.

[...]

PR #2186, opened 2026-07-18
fix(native): lowercase pascal multi-word notation emit to match type and runtime

samchon:
## Scope

Fixes #2183.

`typia.notations.pascal`'s **compile-time Go emit** does not lowercase the inner characters of a multi-word key, so the produced value disagrees with both the declared `PascalCase<T>` type and the runtime `_notationPascal` helper:

| Source | Value for `MAX_COUNT` |
| --- | --- |
| Compile-time (Go) emit | `MAXCOUNT` (buggy) |
| Declared type `PascalCase<{ MAX_COUNT }>` | `MaxCount` |
| Runtime `_notationPascal` | `MaxCount` |

So `.MaxCount` is `undefined` on the compile-time result. The compile-time emit is the sole outlier.

## Root cause

`notationGeneralProgrammer_pascal`'s multi-word callback (`packages/typia/native/core/programmers/notations/NotationGeneralProgrammer.go`) runs every word through `capitalize` only, omitting the `strings.ToLower` that the sibling `camel` multi-word callback and the runtime `__notationCapitalize` already apply. The fix brings pascal's multi-word callback to `capitalize(strings.ToLower(value))`, matching the two producers that already answer correctly.

The single-word `plain` path is untouched: single words are dispatched through `plain` (`userID -> UserID`, `ID -> ID`), which is already correct.

## Deferred / notes

- Native Go only; no `packages/typia/package.json` version bump (releases belong to the maintainer).
- Independent of the other campaign findings.

## Verification

Local verification pending; recorded on the thread after implementation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Fg7AePEdpjaxHrmVffuSjy

PR #2194, opened 2026-07-18
fix(utils): align NamingConvention case converters with the notation contract (#2193)

### LINKED ISSUE #2193: @typia/utils NamingConvention is a third divergent copy of the case-conversion contract, still carrying the #2186 and #2190 bugs in HTTP route/endpoint naming
## Problem

`@typia/utils`' `NamingConvention` is a third, independent copy of the case-conversion contract that #2186 and #2190 just fixed in the notation helpers and their Go twins — and it still carries both bugs. `NamingConvention.snake`/`kebab`/`camel`/`pascal` produce the wrong case for a key that combines an underscore with an internal case boundary, and `NamingConvention.pascal` also fails to lowercase multi-word inner characters (the #2186 bug). It is a published utility used for HTTP route and endpoint naming, so the wrong output reaches generated names.

## Evidence

Verified on `master` by direct invocation:

```
NamingConvention.snake("fooBar_baz")  → "foobar_baz"     (should be "foo_bar_baz")
NamingConvention.snake("openAI_key")  → "openai_key"     (should be "open_ai_key")
```

Owner: `packages/utils/src/utils/NamingConvention.ts`. Its `snake` uses the same segment-atomic lowercasing the notation helper used before #2190 (`unsnake` + per-segment `toLowerCase`), and its `pascal` uses the same multi-word callback the notation helper used before #2186. So all four conversions diverge from what `typia.notations.*` now produces, and `NamingConvention.pascal("MAX_COUNT")` has the pre-#2186 `MAXCOUNT` bug.

This is the campaign's signature shape: **one contract, multiple owners, silently diverging.** The notation helper, its Go compile-time twin, and `NamingConvention` are three owners; #2186 fixed one pair for pascal and #2190 fixed the same pair for all four, but `NamingConvention` was scoped out of both and never realigned. Its own test documents that it "must stay byte-identical" to the conversion — pinning the buggy values.

**Not a soundness hole.** Unlike #2186/#2190, `NamingConvention.snake` is a plain `(str: string) => string` with no `*Case<T>` return type, so there is no typed-key-returns-`undefined` failure. The defect is wrong output from a published utility.

## Consequence surface

| | |
| --- | --- |
| **Owner** | `packages/utils/src/utils/NamingConvention.ts` (all four converters) |
| **Affected** | consumers of `NamingConvention` — `EndpointUtil.pascal` (`.map(NamingConvention.pascal)`), `HttpMigrateRouteAccessor`, `HttpMigrateRouteComposer`, and any external caller (it is exported from `@typia/utils`'s `utils/index.ts`) — a name derived from a key with an underscore-plus-camel boundary, or a multi-word all-caps key, comes out wrong |
| **Not affected** | `typia.notations.*` (fixed by #2186/#2190); keys without an underscore-adjacent case boundary; the runtime soundness of `typia.is`/`assert` (this is a naming utility, not a validator) |

| Case | Expected |
| --- | --- |
| `NamingConvention.snake("fooBar_baz")` | `"foo_bar_baz"` |
| `NamingConvention.pascal("MAX_COUNT")` | `"MaxCount"` |
| `NamingConvention.kebab("openAI_key")` | `"open-ai-key"` |
| a name derived from a plain identifier | unchanged |

## Approach

The invariant: **`NamingConvention`'s four converters produce the same result as `typia.notations.*` and the `*Case<T>` types** — they are the same conversion, and there should not be a third divergent copy. The durable fix is to make `NamingConvention` delegate to (or share the implementation of) the now-correct notation helpers rather than keep a hand-maintained duplicate, so it cannot drift again — the same single-sourcing lesson as #2185. If a shared implementation is not practical across the package boundary, port the #2186/#2190 fixes exactly and add a test that asserts `NamingConvention` equals `typia.notations.*` for the same matrix, so the two can never diverge silently again.

Update the existing `NamingConvention` test that pins the buggy "byte-identical" values — it was locking the defect.

- **Positive:** `NamingConvention.{snake,kebab,camel,pascal}` equal `typia.notations.{snake,kebab,camel,pascal}` and the `*Case<T>` type across the #2190 matrix (`fooBar_baz`, `openAI_key`, `HTTP_fooBar`, `fooBar`, `fooBar_`, `_fooBar`, `userID`, `a_b_c`, `MAX_COUNT`).
- **Negative — must fail against current code:** `NamingConvention.snake("fooBar_baz")` returns `foobar_baz` and `NamingConvention.pascal("MAX_COUNT")` returns `MAXCOUNT` today.
- **Boundary:** the HTTP route naming paths (`HttpMigrateRouteAccessor`/`Composer`, `EndpointUtil`) produce the corrected names; confirm no generated route name regresses for ordinary identifiers.
- **Regression:** ordinary identifiers unchanged; `typia.notations.*` unchanged (already correct).
- **Commands:** `pnpm --filter ./tests/test-utils start`; `pnpm --filter ./tests/test-typia-schema start`; and any HTTP/route test that exercises endpoint naming.

## Coordination

- `@typia/utils` only. **No `packages/typia/package.json` version bump** — releases belong to the maintainer.
- **Direct follow-up to #2186 and #2190** (both merged), which fixed the same conversion in the notation helpers and Go twins and each explicitly scoped `NamingConvention` out for separate adjudication. This is that follow-up.
- Found by the #2190 batch, which flagged the duplicate and correctly declined to widen its scope; confirmed by the lead by direct invocation.

samchon:
Fixes #2193.

## Intent

`@typia/utils`' `NamingConvention` (`packages/utils/src/utils/NamingConvention.ts`) was a third, independent copy of the case-conversion contract that #2186 and PR #2191 (issue #2190) fixed in the notation helpers and their Go twins — and it still carried both bugs:

- `NamingConvention.snake("fooBar_baz")` → `"foobar_baz"` (should be `"foo_bar_baz"`)
- `NamingConvention.pascal("MAX_COUNT")` → `"MAXCOUNT"` (should be `"MaxCount"`)

The invariant: `NamingConvention.{snake,kebab,camel,pascal}` must produce the same result as `typia.notations.*` and the `*Case<T>` types. There should not be a third divergent copy.

## Scope

- `@typia/utils` only. Ports the #2186 / #2191 (issue #2190) case-boundary fixes into `NamingConvention` exactly, mirroring the now-correct `_notationSnake`/`_notationCamel`/`_notationPascal`/`_notationKebab` per-segment walk. Sharing the implementation across the package boundary is impractical — `@typia/utils` depends only on `@typia/interface`, and the notation runtime helpers are internal to the `typia` package (not exported) — so the issue's fallback (port exactly + assert equivalence in a test) applies.
- Updates the existing `NamingConvention` tests that pinned the buggy "byte-identical" values, and adds a regression test that asserts `NamingConvention.{snake,kebab,camel,pascal}` equals `typia.notations.*` over the #2190 matrix.
- No `packages/interface/src` change (the `*Case<T>` types are the authority). No `packages/typia/package.json` version bump.

## Verification

Local (draft; will be filled in on the implementation push):

## Campaign

Issue-campaign implementation PR. Automatic CI for campaign commits is cancelled per the exact-SHA gate; local verification is the gate. No repository Actions or workflow settings changed.

### DIFF
--- a/packages/utils/src/utils/NamingConvention.ts
+++ b/packages/utils/src/utils/NamingConvention.ts
@@ -2,9 +2,14 @@
  * String naming convention converters.
  *
  * `NamingConvention` converts between common code naming conventions:
- * camelCase, PascalCase, and snake_case. Handles edge cases like consecutive
- * uppercase letters (e.g., `XMLParser` → `xml_parser`) and leading
- * underscores.
+ * camelCase, PascalCase, snake_case, and kebab-case. The four converters share
+ * the case-conversion contract of `typia.notations.*` and the `CamelCase<T>` /
+ * `PascalCase<T>` / `SnakeCase<T>` / `KebabCase<T>` typings, and are kept in
+ * step with them: each runs the case-boundary walk within every
+ * underscore-delimited segment (`fooBar_baz` → `foo_bar_baz`), lowercases the
+ * inner characters of an all-caps run (`MAX_COUNT` → `MaxCount` under pascal),
+ * collapses an acronym run (`XMLParser` → `xmlparser` under snake), and
+ * preserves leading underscores.
  *
  * Functions:
  *
@@ -24,16 +29,18 @@ export namespace NamingConvention {
    * @param str Input string
    * @returns CamelCase string
    */
-  export function camel(str: string) {
-    return unsnake({
+  export function camel(str: string): string {
+    return renameOuter({
+      // `CamelCase<T>` lowercases an all-caps key and otherwise lowercases only
+      // the first character of an underscore-free key (`HTTP` → `http`, `userID`
+      // stays).
       plain: (str) =>
-        str.length
-          ? str === str.toUpperCase()
-            ? str.toLocaleLowerCase()
-            : `${str[0]!.toLowerCase()}${str.substring(1)}`
-          : str,
-      snake: (str, i) =>
-        i === 0 ? str.toLowerCase() : capitalize(str.toLowerCase()),
+        str === str.toUpperCase()
+          ? str.toLowerCase()
+          : `${str[0]!.toLowerCase()}${str.substring(1)}`,
+      // With an underscore present, the character after each underscore is
+      // uppercased and the rest lowercased, matching `CamelizeSnakeString`.
+      snake: camelSnake,
     })(str);
   }
 
@@ -43,11 +50,14 @@ export namespace NamingConvention {
    * @param str Input string
    * @returns PascalCase string
    */
-  export function pascal(str: string) {
-    return unsnake({
-      plain: (str) =>
-        str.length ? `${str[0]!.toUpperCase()}${str.substring(1)}` : str,
-      snake: capitalize,
+  export function pascal(str: string): string {
+    return renameOuter({
+      // `PascalCase<T>` capitalizes the first character of an underscore-free
+      // key and keeps the rest verbatim (`userID` → `UserID`).
+      plain: (str) => `${str[0]!.toUpperCase()}${str.substring(1)}`,
+      // With an underscore present, each segment is capitalized and its tail is
+      // lowercased (`MAX_COUNT` → `MaxCount`), matching `PascalizeSnakeString`.
+      snake: pascalSnake,
     })(str);
   }
 
@@ -61,7 +71,6 @@ export namespace NamingConvention {
     if (str.length === 0) return str;
 
     // PREFIX
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     let prefix: string = "";
     for (let i: number = 0; i < str.length; i++) {
       if (str[i] === "_") prefix += "_";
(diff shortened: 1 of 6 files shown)


samchon:
## Implementation complete — Self-Review clean

### What changed

`NamingConvention`'s four converters now port the #2186 / #2191 (issue #2190) per-segment case-boundary walk exactly, mirroring `_notationSnake` / `_notationCamel` / `_notationPascal` / `_notationKebab`:

- `snake` runs the case-boundary walk within each underscore-delimited segment (`fooBar_baz` → `foo_bar_baz`) instead of lowercasing each segment atomically.
- `camel` / `pascal` route on underscore *presence* (via the shared `renameOuter` outer walk, replacing the old `unsnake` that filtered empty segments and collapsed trailing underscores), uppercase the character after each underscore, and lowercase segment tails — so `MAX_COUNT` → `MaxCount`, `fooBar_baz` → pascal `FoobarBaz`, `fooBar_` → camel `foobar_` / pascal `Foobar`.
- `kebab` derives from the corrected `snake`.

### Single-sourcing decision

Sharing the implementation across the package boundary is impractical: `@typia/utils` depends only on `@typia/interface`, and the notation runtime helpers (`_notation*`) are internal to the `typia` package and not exported. So this takes the issue's port-exactly fallback, backed by a parity test (`test_naming_convention_notation_parity`) that asserts `NamingConvention.{snake,kebab,camel,pascal}` equals the live `typia.notations.*` output over the #2190 matrix — the two producers can no longer diverge silently.

### Before → after over the #2190 matrix (all now equal to `typia.notations.*` / `*Case<T>`)

| Input | snake before → after | pascal before → after |
| --- | --- | --- |
| `fooBar_baz` | `foobar_baz` → `foo_bar_baz` | `FooBarBaz` → `FoobarBaz` |
| `openAI_key` | `openai_key` → `open_ai_key` | `OpenAIKey` → `OpenaiKey` |
| `HTTP_fooBar` | `http_foobar` → `http_foo_bar` | `HTTPFooBar` → `HttpFoobar` |
| `fooBar_` | `foobar_`* → `foo_bar_` | `FooBar` → `Foobar` |
| `MAX_COUNT` | `max_count` (ok) | `MAXCOUNT` → `MaxCount` |
| `a_b_c` | `a_b_c` (ok) | `ABC` (ok) |
| `userID` | `user_id` (ok) | `UserID` (ok) |

(*`fooBar_`'s old snake also mis-collapsed the boundary; camel's old value was `fooBar`, now `foobar_`.)

### Regression-test-first

The parity test was written and observed **red** before the fix — `snake("fooBar_baz")` returned `foobar_baz` while `typia.notations.snake` gave `foo_bar_baz` — then green after. Existing `snake`/`kebab` tests gained the underscore-boundary rows they lacked; new dedicated `camel`/`pascal` per-value tests added.

### Verification (campaign CI suspended — local is the gate)

- `pnpm --filter ./tests/test-utils start` — **305 cases Success** (all `test_naming_convention_*` + the full HTTP migrate route suite: accessor, v30/v31/v32)
- `pnpm --filter ./tests/test-typia-schema start` — **171 cases Success** (`test_notation_underscore_boundary` confirms `typia.notations.*` unchanged — the oracle I aligned to)
- `pnpm --filter ./packages/utils build` — clean (`ttsc` strict typecheck + `rolldown`)
- `pnpm --filter ./tests/test-feature-identity start` — Success

### Consequence surface

The only consumer of a changed converter is `EndpointUtil.pascal` (used by `HttpMigrateRouteComposer` to build DTO type names). All other `NamingConvention` consumers (`HttpMigrateRouteAccessor`, `HttpMigrateRouteComposer`, `JsonDescriptor`, `OpenApiObjectValidator`, `stringifyValidationFailure`) use only `variable` / `reserved`, which are unchanged. The HTTP migrate route suite passes unchanged, so no generated route/DTO name regresses for ordinary identifiers.

### Note on the issue framing

The committed `snake`/`kebab` test **case lists** did not actually pin buggy values — every prior case produced identical output before and after the fix (e.g. `XMLParser` → `xmlparser`, `toHTML` → `to_html`). The defect was **uncovered**, not mis-pinned: the "must stay byte-identical" doc claim was never exercised against an underscore-plus-case-boundary key. That gap is now closed by the added rows and the parity test.

**Self-Review:** one complete solo round, no issues found. Ready for merge.

Issue #2232, opened 2026-07-18
Lead self-audit: five closed issues carry a wrong or regressive Approach section, plus three process errors


samchon:
A self-audit of the lead agent's own errors during the 2026-07 campaign. Filed because several of them are **still sitting as wrong text in closed issue bodies**, where anyone reading later — to understand a fix, or to model a similar one — will be misled. The code that shipped is fine; the prescriptions are not.

---

## 1. Five published "Approach" sections prescribe a fix that is wrong or regressive

In each case the implementing agent verified against the real code, found the prescription wrong, and did the right thing instead. The wrong text remains in the issue.

### #2196 — the "Preferred" option would have changed emit for a dozen unrelated callers
I labeled as **Preferred**: route the `is` gate through `OptionPredicator.Numeric(props.Context.Options)` at `IsProgrammer.go:104`.
**Wrong.** `Configure` at that line is shared by nil-option callers — the union-discrimination scoped checkers inside `plain.clone`/`prune`/`classify`, `notations`, the validated `json.stringify` internals, and `protobuf` encode — plus the assert/validate fast-path combiners that deliberately hardcode `Numeric: true`. Taking my preference would have silently changed all of them.
**Correct answer (shipped):** carry `Finite` on the `Decompose`-built options only, so the change is confined to top-level `is`/`assert`/`validate` and the json/protobuf is-encoders.

### #2200 — the "Preferred" mechanism would have rejected the real `File`/`Blob`
I labeled as **Preferred**: a default-lib check parallel to #2178's `metadata_symbol_from_default_lib`.
**Wrong, and it would have been a regression.** Node's `File`/`Blob` globals are declared in `@types/node/*.d.ts`, which is not a `lib.*.d.ts`, so that gate would have declassified the genuine natives. I did flag the risk in the same issue and the agent probed it, so it was caught — but the recommendation itself was unsafe.
**Correct answer (shipped):** gate on `IsDeclarationFile` (any `.d.ts`), a strict superset that cannot over-reject a real built-in.

### #2216 — the prescribed substitution would have left two of the three cases broken
I wrote: make `Validate` inspect `MetadataSchema_unalias(p.Value)` for the gate **and** the per-function checks.
**Wrong for the required/nullable checks.** `MetadataSchema_unalias` deliberately does not unwrap an optional or nullable alias — those modifiers live at the reference site — so the naive substitution leaves the optional and nullable cases still silently dropped.
[...]

PR #2241, opened 2026-07-18
docs(json): expose malformed parse exceptions

### LINKED ISSUE #1053: JSON parser that never throws
When using `json.isParse` when data on the client, there is no guarantee that the data is valid JSON, since the user may have tampered with it. In such cases, `json.isParse` throws `SyntaxError`. This makes typia cumbersome to use, since I have to handle exceptions when all I want to know is whether the data parses to the correct type. When the data is not parsable it does not parse to the correct type, so I would expect the function to just return `null` in such cases.

This would be a breaking change, so I suggest either adding a parameter to `isParse` or creating a separate `safeParse` that returns `null` when an error is thrown.

For example:
```ts
export const safeParse = <Type>(input: string) => {
  try {
    return assertParse<Type>(input);
  } catch (error) {
    return null;
  }
};
```

I tried just making this helper function, but then I get a warning: `[vite] warning: @rollup/plugin-typescript TS(typia.json.assertParse): non-specified generic argument.` Seems like I can't use typia with generics?

### LINKED ISSUE #2235: docs(json): distinguish malformed JSON from validation failures
## Problem

The JSON parse guide says `typia.json.isParse`, `assertParse`, and `validateParse` convert every failure into their corresponding validator result. The generated functions instead let native `JSON.parse` throw `SyntaxError` for syntactically malformed input. The guide conflates malformed JSON with valid JSON whose value does not match `T`.

## Evidence

The discovery reviewer and the lead reproduced the behavior independently on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6` with:

```typescript
interface Parsed { value: number; }

const isParse = typia.json.createIsParse<Parsed>();
const validateParse = typia.json.createValidateParse<Parsed>();
const assertParse = typia.json.createAssertParse<Parsed>();
```

The native transformer emits the parse before or inside the validator call without a catch boundary:

```javascript
input => {
  input = JSON.parse(input);
  return __is(input) ? input : null;
}

input => __validate(JSON.parse(input))
(input, errorFactory) => __assert(JSON.parse(input), errorFactory)
```

The lead's runtime matrix produced:

```json
{
  "isValid": { "returned": { "value": 1 } },
  "isMismatch": { "returned": null },
  "isMalformed": { "threw": "SyntaxError" },
  "validateValid": { "returned": { "success": true, "data": { "value": 1 } } },
  "validateMismatch": { "returned": { "success": false } },
  "validateMalformed": { "threw": "SyntaxError" },
  "assertValid": { "returned": { "value": 1 } },
  "assertMismatch": { "threw": "TypeGuardError" },
  "assertMalformed": { "threw": "SyntaxError" }
}
```

`website/src/content/docs/json/parse.mdx` currently presents one “If anything is wrong” column:

| Function | Current guide claim |
| --- | --- |
| `json.isParse<T>` | returns `null` |
| `json.assertParse<T>` | throws `TypeGuardError` |
| `json.validateParse<T>` | returns `IValidation.IFailure` |

The same page later says `assertParse` throws if either parsing or validation fails, without distinguishing the error classes. Those claims are accurate for valid JSON with a type mismatch but false or incomplete for malformed JSON.

This behavior predates the current guide. Closed issue #1053 reported `json.isParse` throwing `SyntaxError` for malformed input in 2024; the maintainer recommended an explicit caller-side safe-parse wrapper around a generated parser rather than changing the parse function. The conflicting guide table was introduced by `7c4f72e5f17` / PR #1852 in 2026. No later issue or pull request owns the mismatch.

## Consequence surface

| Input class | `isParse` | `validateParse` | `assertParse` |
| --- | --- | --- | --- |
| Valid JSON matching `T` | parsed value | success | parsed value |
| Valid JSON not matching `T` | `null` | failure | `TypeGuardError` |
| Malformed JSON | `SyntaxError` | `SyntaxError` | `SyntaxError` |

Readers following the current table can omit an exception boundary around untrusted malformed text, causing an unexpected throw in all three APIs. The direct and factory forms share the same native programmers. Ordinary `JSON.parse` supplies the same syntax error and is the control.

## Approach

Enforce one public-contract invariant: the JSON parse guide and the generated runtime behavior must agree for syntactically malformed input and for valid JSON that fails type validation, with the distinction stated separately for all three parse variants. The implementation mechanism is deliberately left to the owner to validate against the historical contract and current call graph.

- The guide presents separate outcomes for malformed JSON and valid-but-type-invalid JSON for `isParse`, `validateParse`, and `assertParse`.
- A reader can tell which error type is thrown by each failing path and cannot infer that `isParse` or `validateParse` is exception-free on malformed text.
- Direct and factory forms are covered by the stated contract.
- Valid-success, valid-mismatch, and malformed-input runtime probes agree with the final prose.
- If the guide continues to recommend a never-throw wrapper, its example supplies the concrete generic parser through a generated factory and does not claim generic runtime transformation is possible.
- Keep each prose paragraph on one source line and verify the JSON guide page under the documentation workflow. No unrelated guide or runtime behavior changes.

## Coordination

This documentation-contract mismatch is independent of the two native D10 defects. Historical #1053 is evidence about the intended behavior, not an implementation prescription; per #2232, the final mechanism must follow measured current behavior and the maintained contract. No package version change is part of this issue.

samchon:
## Intent

The JSON parse guide grouped malformed JSON and post-parse type mismatches under one failure outcome. That implied `json.isParse` and `json.validateParse` never throw, although all generated parse variants preserve native `JSON.parse` syntax errors.

## Scope

- Split the guide's contract table into matching JSON, valid-but-type-invalid JSON, and malformed JSON.
- State the distinct `null`, `IValidation.IFailure`, `TypeGuardError`, and `SyntaxError` outcomes for all three parse variants.
- State that the direct and factory forms share the same contract.
- Clarify the `assertParse` example without changing runtime behavior or adding a generic transform pattern.

No runtime, package version, interface declaration, workflow, or unrelated guide changed.

## Contract evidence

The following transformed runtime matrix was captured before the edit and repeated after the final prose on exact base `5fce10b93dfe3807423312b447d55f7b2265f5c6`:

| Input class | `isParse` direct/factory | `validateParse` direct/factory | `assertParse` direct/factory |
| --- | --- | --- | --- |
| Valid JSON matching `T` | parsed value | `IValidation.ISuccess` with data | parsed value |
| Valid JSON not matching `T` | `null` | `IValidation.IFailure` with errors and data | `TypeGuardError` |
| Malformed JSON | `SyntaxError` | `SyntaxError` | `SyntaxError` |

Both malformed paths were `instanceof SyntaxError` and not `TypeGuardError`. The mismatch assertion method was `typia.json.assertParse` for the direct call and `typia.json.createAssertParse` for the factory. Native `JSON.parse` was the syntax-error control.

History was also checked: closed #1053 records the intentional caller-side exception boundary, while PR #1852 / merge `7c4f72e5f17` introduced the conflicting guide table. Per #2232, that history was treated as evidence rather than an implementation prescription; current direct and factory behavior was measured independently.

## Pre-PR chronology and Self-Review

- The first cold transformed probe reached its 15-minute timeout; the warmed retry completed in 11.8 seconds and produced the red matrix above.
- Root packaging first reached a 20-minute timeout during the cold `typia` prepack. A complete retry passed in 22 minutes 53 seconds and produced all six expected tarballs.
- The first website build reached the playground Wasm step and failed with `fatal error: out of memory allocating heap arena map` during concurrent Go linking. After the resource contention cleared, the same Wasm build passed in 24 minutes 43 seconds and the complete website build passed.
- Solo Self-Review inspected the complete one-file diff and consequence surface. The first full round was clean, so no remediation round was required.

## Verification

- `pnpm install` — passed (1 minute 49.6 seconds).
- `pnpm build` — passed (96.5 seconds).
- `pnpm package:tgz` — passed (22 minutes 53.4 seconds); `interface`, `langchain`, `mcp`, `typia`, `utils`, and `vercel` tarballs confirmed and left ignored.
- `node build/wasm.js` in `website/` — passed after resource contention cleared (24 minutes 43.4 seconds; 39.27 MiB).
- `pnpm build` in `website/` — passed (171.1 seconds); Next compiled, type-checked, generated 74/74 static pages including `/docs/json/parse`, exported, indexed, and generated the sitemap.
- Rendered `/docs/json/parse` HTML — confirmed the three input columns, all `SyntaxError` cells, factory parity sentence, and assertion distinction.
- Final transformed direct/factory runtime matrix — passed (5.1 seconds; all 18 paths matched the table).
- `typos v1.48.0 --config typos.toml website/src/content/docs/json/parse.mdx` — passed with en-US configuration.
- `git diff --check` — passed.

Skipped by campaign contract: repository-wide `pnpm format`, the full `pnpm test` suite for this documentation-only change, and the already-complete D9 combined checks. Campaign CI is cancelled per exact SHA; the push gate observed no runs after two polls.

Fixes #2235

### DIFF
--- a/website/src/content/docs/json/parse.mdx
+++ b/website/src/content/docs/json/parse.mdx
@@ -17,13 +17,15 @@ namespace json {
 }
 ```
 
-Each function maps to one of the [runtime validators](/docs/validators/is):
+Each function maps to one of the [runtime validators](/docs/validators/is), but validation starts only after native `JSON.parse` succeeds:
 
-| Function | If JSON is valid AND matches `T` | If anything is wrong |
-|----------|----------------------------------|----------------------|
-| `json.isParse<T>` | returns the parsed value | returns `null` |
-| `json.assertParse<T>` | returns the parsed value | throws `TypeGuardError` |
-| `json.validateParse<T>` | returns `IValidation.ISuccess<T>` | returns `IValidation.IFailure` |
+| Function | Valid JSON matching `T` | Valid JSON not matching `T` | Malformed JSON |
+|----------|-------------------------|-----------------------------|----------------|
+| `json.isParse<T>` | returns the parsed value | returns `null` | throws `SyntaxError` |
+| `json.assertParse<T>` | returns the parsed value | throws `TypeGuardError` | throws `SyntaxError` |
+| `json.validateParse<T>` | returns `IValidation.ISuccess<T>` | returns `IValidation.IFailure` | throws `SyntaxError` |
+
+The factory functions `createIsParse`, `createAssertParse`, and `createValidateParse` produce parsers with the same outcomes.
 
 ## First example
 
@@ -37,7 +39,8 @@ interface User {
 
 const text: string = await fetch("/me").then((r) => r.text());
 const user = typia.json.assertParse<User>(text);
-// → parsed User, validated. Throws if either parse or validation fails.
+// → parsed User when text is valid JSON matching User.
+// Throws SyntaxError for malformed JSON or TypeGuardError for a type mismatch.
 ```
 
 <Tabs items={['TypeScript Source', 'Compiled JavaScript']}>


samchon:
Post-creation campaign CI gate passed for `c26bb4fd11aef89ab7df8ddfb33defe2b1c3e536` on `docs/json-parse-malformed-contract` after four polls.

- [Run 29659313230](https://github.com/samchon/typia/actions/runs/29659313230): `completed` / `cancelled`
- [Run 29659313220](https://github.com/samchon/typia/actions/runs/29659313220): `completed` / `cancelled`

The earlier push gate observed no runs after two polls. Repository Actions settings and workflows were not changed.

samchon:
Upstream advanced after the initial verified push: `origin/master` now includes #2236 and the release commit at `4cba3fdee3`, including package-version and website/MCP documentation changes.

Before updating this PR, I will rebase onto that exact master, require the final base-to-head diff to remain limited to `website/src/content/docs/json/parse.mdx`, preserve upstream version `13.1.19`, and repeat the focused direct/factory parse runtime probe plus the JSON parse route/build verification. The rebased push will then pass a new exact-SHA cancellation gate.

samchon:
Rebase remediation is complete.

- New head: `669000995005dd2e66e9f24d724edeaa9d49eb65`
- Parent/base: `4cba3fdee3fec343db9dd10d988ab68cbaa1cf42`
- Final base-to-head diff: only `website/src/content/docs/json/parse.mdx`
- Preserved upstream release: root, website, and typia package versions are all `13.1.19`; the version manifests have no PR diff
- Focused transformed direct/factory parse matrix: passed in 3.9 seconds; all 18 matching, mismatch, and malformed paths still match the documented contract
- `pnpm build` in `website/`: passed in 194.6 seconds as `@typia/website@13.1.19`; Next compiled and type-checked, generated 74/74 pages including `/docs/json/parse`, exported, indexed, and generated the sitemap
- Rendered parse route: confirmed the three input columns, malformed `SyntaxError` contract, factory parity, and assertion error distinction
- Fresh post-rebase Self-Review: clean

The lease-protected rebased push passed its exact-SHA gate after four polls:

- [Run 29659558210](https://github.com/samchon/typia/actions/runs/29659558210): `completed` / `cancelled`
- [Run 29659558211](https://github.com/samchon/typia/actions/runs/29659558211): `completed` / `cancelled`

Issue #2225, opened 2026-07-18
compare.equals still false-equals distinct values of a non-discriminable object union (the still-open half of #2214): equals({a:1},{}) is true


samchon:
## Problem

`typia.compare.equals` still reports two distinct values as equal for a **non-discriminable** object union — a union of ≥2 object members with no unique required distinguishing key (all members' distinguishing keys are optional), e.g. `{ a?: number } | { b?: string }`. #2214 fixed the discriminable case but left this fallback on the pre-#2214 flat-OR, which is unsound. A populated object even compares equal to `{}`.

## Evidence — bytes + runtime on `master` (`dbf147b4bb`, the #2214 fix commit)

`typia.compare.createEquals<{ a?: number } | { b?: string }>()` emits the flat OR (no discrimination):

```js
_eo0 = (x, y) => x === y || (<both objects> && (x.a === y.a || (typeof x.a==="number" && typeof y.a==="number" && x.a === y.a)));
_eo1 = (x, y) => x === y || (<both objects> && (x.b === y.b || (typeof x.b==="string" && typeof y.b==="string" && x.b === y.b)));
eq   = (x, y) => (x === y || _eo0(x, y) || _eo1(x, y));
```

Runtime (distinct references, both operands `is`-valid):

```
eq({a:1}, {a:2})  →  true   (expected false)
eq({a:1}, {})     →  true   (expected false — a populated object equals {})
eq({a:1}, {b:"p"})→  false  (correct)
eq({a:1}, {a:1})  →  true   (correct)
```

`eq({a:1}, {})`: `_eo0` (member `{a?}`) inspects `a` and returns false (`1 !== undefined`); `_eo1` (member `{b?}`) never inspects `a`, and `x.b === y.b` is `undefined === undefined` → true → `eq` is `false || false || true` = **true**. The wrong member's comparator passes because the discriminant it needs is optional (so `undefined === undefined`) and the *declared* property `a` is treated as "extra" under that member.

## Root cause

[...]

Issue #2214, opened 2026-07-18
compare.equals reports two distinct object-union values as equal: it flat-ORs member comparators with no discrimination


samchon:
## Problem

`typia.compare.equals` / `createEquals` over a union of object types reports two **distinct** values as equal. It assembles the union as a flat OR of member comparators with no discrimination, so a sibling member's comparator matches spuriously — its discriminant passes through a generic fallthrough and its payload keys, absent in both operands, compare `undefined === undefined`. The operands' real difference is never examined.

## Evidence — bytes + runtime on `master` (`e4a4eb5fd7`)

`typia.compare.createEquals<{ type: "a"; x: number } | { type: "b"; y: string }>()` emits:

```js
_eo0 = (x, y) => x === y || (<both objects> && ((x.type === y.type || (x.type === "a" && y.type === "a")) && (x.x === y.x || <number>)));
_eo1 = (x, y) => x === y || (<both objects> && ((x.type === y.type || (x.type === "b" && y.type === "b")) && (x.y === y.y || <string>)));
eqDU = (x, y) => (x === y || _eo0(x, y) || _eo1(x, y));   // flat OR — no discrimination
```

Runtime (distinct references, both operands `is`-valid):

```
eqDU({type:"a",x:1}, {type:"a",x:2})  →  true   (expected false)
eqDU({type:"b",y:"p"}, {type:"b",y:"q"})  →  true   (expected false)
eqDU({type:"a",x:1}, {type:"a",x:1})  →  true   (correct)
```

Trace for the first: `_eo0` (member "a") correctly returns `false` (`1 !== 2`). Then `_eo1` (member "b") returns `true` — its discriminant passes via `x.type === y.type` (`"a" === "a"`, even though neither is `"b"`), and its payload compares `x.y === y.y` = `undefined === undefined`. So `eqDU = false || false || true = true`.

## Root cause

`packages/typia/native/core/programmers/compare/CompareEqualProgrammer.go` — `schema()` assembles an object union as a flat `_eo0(x,y) || _eo1(x,y) || …` (the union-assembly at ~`:152-171`/`:181-184`), and each member comparator's discriminant is weakened by the `x[key] === y[key]` first branch that `schema()` always prepends (~`:137-142`/`:184`): the discriminant is satisfied whenever x and y merely agree, not when they actually are that member.

[...]

PR #2243, opened 2026-07-18
fix(native): stabilize dynamic-key schema emit

samchon:
## Intent

Make JSON schema emission byte-deterministic for objects with multiple numeric or template index signatures, without changing schema meaning or unrelated stable ordering.

Fixes #2233

## Root cause and scope

The Go migration retained dynamic pattern/value pairs in a map and ranged that map before `MetadataSchema_merge`. The legacy TypeScript producer used `Object.values`, whose string-key enumeration preserves first insertion order, so the migration made the resulting `oneOf` order depend on Go map iteration.

This change records each pattern's first insertion position beside the map and merges values in that order. Reassigning an existing pattern keeps its original slot and replaces its value, matching the earlier plain-object contract. The producer boundary feeds `json.schema`, `json.schemas`, `json.application`, the OpenAPI 3.0 downgrade, and the LLM schema/parameters/application consumers.

No package version, `packages/interface/src` declaration, workflow, dependency, or provider-specific behavior changes are included.

## Regression and emit evidence

On exact base `5fce10b93dfe3807423312b447d55f7b2265f5c6`, the new real-transform regression failed on the second unchanged TypeScript emit:

- baseline SHA-256: `b5d99f1f0c464b44482539e17eab0146698829e111053ee1e6c25c890265edd2`
- current SHA-256: `154e56197d98f94b5d260293643476a3c03db718cd8d05465ded335468ec8c52`
- first differing byte: `1374`

After the fix and rebase onto `643446af098866eba95ba672551d59d7f3890177`, 32 TypeScript and 32 JavaScript transforms were byte-identical:

- TypeScript SHA-256: `a87c4778bf525478e2bdc1e786b80f0c2da5a7cc346d49d520f3af6b5b51d6d9`
- JavaScript SHA-256: `8b49c22648563ab643862c509bf3c2ba28e28e4e1bfa614cf1f6bfa9fc40001b`

The runtime oracle pins the intended declaration-order stabilization and branch completeness:

- `DynamicComposite`: `[number, string, boolean]`
- `DynamicTemplate`: `[string, number, boolean]`
- `DynamicUnion`: `[string, number]`

It also preserves literal property order, component/schema order, a single dynamic signature, and an ordinary union. A separate unit negative control verifies that a repeated pattern keeps one ordered slot while retaining the replacement value.

## Local verification

- `go -C packages/typia/native test ./cmd/ttsc-typia -run '^TestJsonSchemaDynamicKeyDeterminismTransform$' -count=1 -v`
[...]

PR #2244, opened 2026-07-18
fix(native): preserve template-literal union membership

### LINKED ISSUE #2234: fix(native): preserve template-literal union membership in compare.equals
## Problem

`typia.compare.equals` can report two distinct, valid object-union values as equal when the union members are distinguished by template-literal string domains. The comparison fallback weakens each template-literal member check to a plain string check, resolves a later-member value to the first member, and then ignores a property declared only by the actual later member.

## Evidence

The discovery reviewer and the lead reproduced this independently on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6` with this tag-free type:

```typescript
type TemplateUnion =
  | { kind: `a_${string}`; common: string }
  | { kind: `b_${string}`; common: string; b?: string };

const is = typia.createIs<TemplateUnion>();
const clone = typia.plain.createClone<TemplateUnion>();
const equals = typia.compare.createEquals<TemplateUnion>();
```

The native transformer emitted full template checks for `is` and clone membership:

```javascript
const _io0 = input =>
  typeof input.kind === "string" && /^a_(.*)/.test(input.kind) && /* common */;
const _io1 = input =>
  typeof input.kind === "string" && /^b_(.*)/.test(input.kind) &&
  /* common and optional b */;
```

The compare member matchers discarded the template domains:

```javascript
const _ui0 = v => typeof v.kind === "string" && typeof v.common === "string";
const _ui1 = v =>
  typeof v.kind === "string" && typeof v.common === "string" &&
  (v.b === undefined || typeof v.b === "string");
```

The fallback selects the first matching `_ui` member. Running the emitted JavaScript on distinct references produced these exact bytes in both reproductions:

```json
{"isX":true,"isY":true,"cloneX":{"kind":"b_x","common":"same","b":"left"},"cloneY":{"kind":"b_x","common":"same","b":"right"},"equals":true}
```

Both values satisfy the declared union, and clone resolves both to the `b_*` member and preserves the differing `b` values. `equals` should therefore be `false`; a Node negative assertion exited `7` on current master.

The owner is `packages/typia/native/core/programmers/compare/CompareEqualProgrammer.go`. `objectUnion` sends a non-specialized object union through `matchLadder`; `matchLadder` chooses the first matcher that returns true; and `match` represents a template-literal metadata bucket only as `typeof value === "string"`. The authoritative template checker in `packages/typia/native/core/programmers/iterate/check_template.go` retains the pattern. This weak matcher arrived in the #2225 fallback at current HEAD.

This is a residual of #2225, not a reason to follow or revert that issue's stale Approach text. Per #2232, #2225's factual claim about clone was wrong; the evidence here comes from current emitted bytes and runtime behavior.

GitHub issue and pull-request searches for `compare template literal union` and `compare.equals template` found no duplicate. #2214 and #2225 cover other object-union resolution failures but not constraint-sensitive template discriminants.

## Why nothing caught it

Before #2225 there was no TypeScript-suite coverage of `typia.compare.*`. Current master has one `compare.equals` TypeScript test file for non-discriminable all-optional unions, but no template-literal discriminator case. Native compare tests cover literals, unique keys, ordinary discriminants, native/object unions, recursion, arrays, and tuples, but no object-union member whose discriminator is a template-literal domain.

## Consequence surface

| Surface | Consequence |
| --- | --- |
| `typia.compare.equals<T>(x, y)` | Direct calls can false-equal distinct valid values. |
| `typia.compare.createEquals<T>()` | Factory calls use the same programmer and have the same defect. |
| Nested object unions | The same member matcher can ignore later-member properties below a containing object. |
| `typia.plain.clone` / `is` | Not defective for the witness; they are authoritative positive controls and select the `b_*` member. |
| `typia.compare.cover` | Returned `false` for the witness and is excluded. |
| `typia.compare.less` | Rejects multi-object unions and is excluded. |

The false positive requires an earlier member that the weakened matcher over-accepts and a later-member property whose difference matters. Ordinary literal discriminants, non-string discriminants, a one-member object, and equal later-member values are negative controls.

## Approach

Enforce one invariant at object-union comparison: for is-valid operands, `compare.equals` must resolve each operand to the same declared member as the authoritative membership path used by `is` and `plain.clone`, including template-literal constraints, and must compare every property declared by that member. The implementation mechanism is deliberately left to the implementer to test against the full call graph.

- A red regression on current `5fce10b93d` proves two valid `b_*` values with different `b` properties are not equal after the fix; it must also prove `is` accepts both and clone preserves both properties.
- Equal `b_*` values compare equal, distinct `a_*` values still compare by their declared properties, and an `a_*` value does not equal a `b_*` value.
- Invalid discriminator strings remain rejected rather than being routed to an arbitrary member.
- The invariant holds for direct and factory `equals`, and for the same union nested in an object.
- Existing #2214/#2225 discriminated and non-discriminable union regressions remain green; no branch is hardcoded to this fixture.
- Add focused native transform/runtime coverage and a matching TypeScript-suite case with the repository's one-case-per-file doc comment contract.
- Run the focused native regressions, the native Go suite, `go vet`, and `tests/test-typia-schema`; record emitted-byte evidence that the intended member-resolution behavior changed while unrelated compare output did not.

## Coordination

This issue and the JSON-schema determinism defect touch different owners and can be implemented independently from the same master. Do not bump a package version during the campaign; release numbering belongs to the maintainer. Do not rerun the already-passing D9 combined build/vet/fix-test check. Per #2232, any implementation idea remains a hypothesis until the red test, runtime controls, full call graph, and emitted bytes verify it.

samchon:
## Summary

Preserve structural template-literal domains when `compare.equals` resolves a fallback object-union member. Both operands now reach the same declared member as the `is`/`plain.clone` structural template path before equality compares that member's fields.

Fixes #2234

## Root cause and impact

The first-match fallback added for non-discriminable object unions represented every template bucket as only `typeof value === "string"`. For

```ts
{ kind: `a_${string}`; common: string }
  | { kind: `b_${string}`; common: string; b?: string }
```

both `_ui` matchers accepted any string `kind`. A valid `b_*` value therefore resolved to the earlier `a_*` member, whose comparator does not declare `b`; distinct `b` values compared equal. The same weakening also routed two identical invalid `c_*` discriminators to an arbitrary member.

## Implementation

- Expose the structural pattern already owned by the runtime template checker without exposing its capture/tag machinery.
- Retain that pattern only in the fallback member matcher that had weakened template metadata.
- Leave union specialization, dynamic-key equality, and compare's established `typia.tags.*` policy unchanged.
- Add one tag-free native transform/runtime regression and one matching function-per-file TypeScript regression.

No package version, `packages/interface/src`, workflow, dependency, or provider claim changes are included. The rebased branch preserves upstream `typia` version `13.1.19`.

## Emitted evidence

Before, both member matchers accepted every string discriminator:

```js
const _ui0 = (v) => typeof v.kind === "string" && typeof v.common === "string";
const _ui1 = (v) => typeof v.kind === "string" && typeof v.common === "string" && (v.b === undefined || typeof v.b === "string");
```

After, the same matchers preserve their declared structural domains:

```js
const _ui0 = (v) => typeof v.kind === "string" && RegExp("^a_(.*)").test(v.kind) && typeof v.common === "string";
const _ui1 = (v) => typeof v.kind === "string" && RegExp("^b_(.*)").test(v.kind) && typeof v.common === "string" && (v.b === undefined || typeof v.b === "string");
```

An unrelated non-union `compare.equals<{ kind: string; common: string; b?: string }>` export was byte-identical before and after: SHA-256 `f2a311c5577a6393c03a521d2d2024a3987dd83ce4e0474d6381833534db0a19`.

## Regression coverage

The native and TypeScript cases prove:

- both `b_*` operands are `is`-valid;
- clone resolves both to the later member and preserves different `b` values;
- direct and factory equals reject different `b` values and accept equal ones;
- `a_*` values compare their declared fields, while `a_*` and `b_*` differ;
- invalid discriminators are rejected instead of routed arbitrarily;
- direct/factory behavior holds for the same union nested in an object; and
- literal-discriminated (#2214), non-discriminable (#2225), and non-union controls remain unchanged.

## Local verification

Final rebased HEAD `2494e19cb477243cd0e195e0df8dbae29d52567f` is based directly on `643446af098866eba95ba672551d59d7f3890177`.

- PASS red-to-green `TestCompareTemplateLiteralUnionTransform` (current master produced 7 intended runtime assertion failures before the fix).
- PASS `go -C packages/typia/native test -p=1 ./cmd/ttsc-typia -run '^TestCompare(TemplateLiteralUnion|NonDiscriminableUnion|UnionDiscrimination)Transform$' -count=1` after Self-Review tightening and again after rebase.
- PASS `go -C packages/typia/test test -p=1 ../native/...` (full native suite; 360.7 seconds).
- PASS `go -C packages/typia/test vet -p=1 ../native/...`.
- PASS `pnpm --filter ./tests/test-typia-schema start -- --include compare_equals_object_union_template_literal` after rebase, including a fresh source-plugin build for the `13.1.19` cache key.
- PASS `pnpm --filter ./tests/test-typia-schema start` after rebase.
- PASS `pnpm --filter ./tests/test-feature-identity start`.
- PASS `git diff --check`, targeted Prettier check, filename/export identity, and en-US spelling sweep.

The full native suite and vet preceded Self-Review's removal of two unrelated emit expansions; the final fallback-only implementation was rerun through the focused native/adjacent tests and both schema commands. The already-passing D9 combined gate and repository-wide `pnpm build` were not rerun. Per campaign policy, `pnpm format` was not run.

## Pre-PR Self-Review chronology

The PR follows implementation, so these findings were resolved locally before the PR existed:

1. Round 1 removed unverified template-pattern changes from specialization and dynamic-key equality. `MetadataSchema_intersects` deliberately keeps template-vs-template/literal/string properties out of specialization, while dynamic keys were outside #2234's evidence.
2. Round 2 removed a diagnostic-only emit environment hook and narrowed comments to the structural template invariant actually implemented.
3. Round 3 reviewed the complete source/test/consequence surface from scratch and was clean.

Campaign CI is intentionally not used as a verification substitute. The push exact-SHA gate polled twice and observed no runs; the same gate will run again immediately after draft PR creation.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/compare_template_literal_union_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/compare_template_literal_union_transform_test.go
@@ -1,0 +1,241 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestCompareTemplateLiteralUnionTransform verifies template-literal object
+// union members retain their runtime domains during equality dispatch.
+//
+// The #2225 first-match fallback must use the same structural template
+// membership decision as is and clone. Weakening both template domains to
+// `typeof string` routes valid `b_*` values through the `a_*` comparator and
+// drops `b`.
+//
+//  1. Transform direct and factory is, clone, and equals calls for a tag-free
+//     template-literal object union and the same union nested in an object.
+//  2. Prove is accepts both witnesses and clone selects the `b_*` member while
+//     preserving their different `b` values.
+//  3. Exercise equal, cross-member, invalid-discriminator, nested, literal-
+//     discriminated, non-discriminable, and non-union controls at runtime.
+func TestCompareTemplateLiteralUnionTransform(t *testing.T) {
+  project := compareTemplateLiteralUnionProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("compare template-literal union transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  compareTemplateLiteralUnionRunRuntimeCases(t, project, out)
+}
+
+func compareTemplateLiteralUnionProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "compare-template-literal-union-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(compareTemplateLiteralUnionTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(compareTemplateLiteralUnionSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func compareTemplateLiteralUnionRunRuntimeCases(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "throw-type-guard-error-stub.cjs"), []byte(compareTemplateLiteralUnionThrowStub), 0o644); err != nil {
+    t.Fatalf("write throw stub: %v", err)
+  }
+  js = strings.ReplaceAll(js, `require("typia/lib/internal/_throwTypeGuardError")`, `require("./throw-type-guard-error-stub.cjs")`)
+  runtimeJS := ttscTypiaTestRewriteCommonJS(t, js)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(runtimeJS), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
+  if err := os.WriteFile(runner, []byte(compareTemplateLiteralUnionRuntimeRunner), 0o644); err != nil {
+    t.Fatalf("write runtime runner: %v", err)
+  }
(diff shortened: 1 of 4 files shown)


samchon:
Lead rebase verification on current master (after #2243): focused native tests TestCompareTemplateLiteralUnionTransform, TestCompareNonDiscriminableUnionTransform, and TestCompareUnionDiscriminationTransform all pass. Rebased head: e03e1c1eef4950eaeb33027de341bbc245a18f34. Exact-SHA Actions cancellation completed.

Issue #2239, opened 2026-07-18
fix(native): distinguish structural package declarations from colliding runtime natives


samchon:
## Problem

The native transformer classifies ordinary package-exported declarations as JavaScript runtime natives when their public `.d.ts` names happen to be `Date`, `Map`, `Set`, or `WeakMap`. A structural dependency type then rejects its own plain values and accepts unrelated native instances.

## Evidence

The D10 reviewer and the lead independently reproduced this on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6`. The lead's real-transform runtime fixture reported all eight package identity mismatches in one run: every plain package value was `false` and every unrelated native instance was `true`. The fixture imported these declarations from an ordinary `models.d.ts`:

```typescript
export interface Date { stamp: number; }
export interface Map<K, V> { brandMap: K; valueMap: V; }
export interface Set<T> { brandSet: T; }
export interface WeakMap<K extends object, V> {
  brandWeakMap: K;
  valueWeakMap: V;
}
```

Compile-time bidirectional `Same<X, Y>` assertions proved each imported declaration equivalent to a local structural control. The transformed validators nevertheless produced the opposite runtime identity for every package type:

| Declaration | Own structural value | Unrelated native instance |
| --- | ---: | ---: |
| package `Date` | `false` | `true` |
| package `Map` | `false` | `true` |
| package `Set` | `false` | `true` |
| package `WeakMap` | `false` | `true` |
| equivalent local structural controls | `true` | n/a |

The shared classification input is the declaration name plus whether its source file is a declaration file. That distinguishes `.ts` from `.d.ts`, but ordinary dependency declarations are normally shipped as `.d.ts` too. The executed witnesses cover the simple-native, `Map`, `Set`, and generic-native classifier families.

## Verified owner

`packages/typia/native/core/factories/internal/metadata/MetadataHelper.go:318` defines the shared declaration-file predicate, which returns the first locatable declaration's `IsDeclarationFile` flag. `iterate_metadata_native.go:25`, `iterate_metadata_map.go:19`, and `iterate_metadata_set.go:19` combine that predicate with the matching name for the executed simple, generic, `Map`, and `Set` paths. Those bytes explain the `.ts`/`.d.ts` split without establishing a safe replacement identity mechanism.

[...]

Issue #2212, opened 2026-07-18
A user type named Map/Set/WeakMap/WeakSet is classified as the built-in native by name, so its own members are dropped (the #2200 collection residual)


samchon:
## Problem

A user type whose exact name is `Map`, `Set`, `WeakMap`, or `WeakSet`, declared at module scope, is classified as the built-in native collection by **name (and arity) alone** — with no check that the type actually is the built-in. Its own members are dropped and it is validated with `input instanceof Map` (etc.). This is the collection half of #2200 (which fixed the `simples` name table via a declaration-file identity gate but left the collection and generic paths ungated).

## Evidence — bytes on `master` (`e4a4eb5fd7`)

```ts
interface Map<K, V> { brandMap: string; keyMap: K; valMap: V; }
interface WeakMap<K extends object, V> { brandWeak: string; }
```
```js
isUserMap  = input => input instanceof Map && (() => [...input].every(elem => Array.isArray(elem) && (elem.length === 2 && "string"===typeof elem[0] && "number"===typeof elem[1])))();
isUserWeak = input => input instanceof WeakMap;
```

The user's `brandMap`/`keyMap`/`valMap` (and `brandWeak`) are ignored; the emit is **byte-identical** to the real-lib `Map`/`WeakMap` output — typia cannot tell the user type from the global. `isUserMap({ brandMap: "x", ... })` is `false` (a valid value rejected).

## Root cause

The exact-name `simples` path got the identity gate in #2200 (`metadata_symbol_from_declaration_file`), but the collection and generic paths match by name only:

- `iterate_metadata_map.go:10` — `if metadata_type_symbol_base_name(typ) != "Map" { return false }` then classifies as native Map.
- `iterate_metadata_set.go:10` — the same for `Set`.
- `iterate_metadata_native.go:37-42` — the WeakMap/WeakSet generics loop matches `name == generic.Name || HasPrefix(name, generic.Name+"<")` (the `+"<"` guard from #2181 stops the `WeakMapLike` prefix over-match, but nothing checks declaration-file identity, so an exact `interface WeakMap<K,V>` still matches).

None of these calls the `metadata_symbol_from_declaration_file` gate #2200 added, so a user type of the same name — declared in a `.ts` module, not a `.d.ts` — is taken as the built-in.

## Why nothing caught it

[...]

Issue #2200, opened 2026-07-18
A user type is classified as a built-in native by name alone: is<File>/<Blob>/<Date>… emit instanceof and reject valid values (or throw)


samchon:
## Problem

typia classifies a user type as a built-in native type **by name alone**, with no check that the type is actually the built-in. A user interface named `File`, `Blob`, `Date`, `RegExp`, `DataView`, `Number`, `String`, `Boolean`, `BigInt`, `ArrayBuffer`, or a typed-array name is validated with `input instanceof <Native>` instead of against its own members — so `is`/`assert`/`validate` reject every valid value of that user type (and, where the global does not exist at runtime, throw `ReferenceError`). This is the same class as #2178 (name matching that ignores symbol identity).

## Evidence — bytes on `master` (`9ba43d73b4`)

`is<T>()` emitted for user interfaces whose names collide with native names:

```ts
interface File { name: string; size: number; }
interface Blob { size: number; }
interface WeakMapCache { size: number; }
```
```js
isFile         = input => input instanceof File;                                   // MISCLASSIFIED — user fields ignored
isBlob         = input => input instanceof Blob;                                    // MISCLASSIFIED
isWeakMapCache = input => "object"===typeof input && null!==input && _io0(input);   // correct — structural (_io0 checks size)
```

`isFile({ name: "a", size: 1 })` is **false** (a valid `File` object rejected); in a project without the DOM lib, `File` is not defined at runtime and `isFile(...)` throws `ReferenceError: File is not defined`.

The `WeakMapCache` line shows the generics prefix is already correct — #2181's `name == generic.Name || HasPrefix(name, generic.Name+"<")` guard rejects `WeakMapCache`. The defect is only the **exact-name** path.

## Root cause

`Iterate_metadata_native` (`packages/typia/native/core/factories/internal/metadata/iterate_metadata_native.go:18`):

```go
name = iterate_metadata_native_getNativeName(name)     // the type's symbol base name, e.g. "File"
if _, ok := iterate_metadata_native_simples[name]; ok { // <-- pure name lookup, no symbol-identity / structural check
  iterate_metadata_native_take(props.Metadata, name)
  return true
}
```

`name` is the symbol's base name, so a user `interface File` yields `"File"`, matches `iterate_metadata_native_simples["File"]`, and is taken as native. Native is tried **before** object in the iterator (`iterate_metadata.go`: native at the array/native step, object last), so the structural object path never runs.

[...]

PR #2247, opened 2026-07-18
fix(native): preserve native-named structural intersections in unions

### LINKED ISSUE #2237: fix(native): preserve native-named structural intersections in unions
## Problem

The native transformer can remove an inhabited structural intersection from a union solely because a local user declaration shares the name `Date`. The same type validates correctly by itself, but becomes unreachable after union composition.

## Evidence

The D10 reviewer and the lead independently reproduced this on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6`. The lead's real-transform runtime fixture repeated the exact standalone/intersection/union matrix below and reported `localDateUnionDate` as the only valid arm mismatch:

```typescript
interface Date { stamp: number; }
type LocalDateIntersection = Date & { label: string };
type LocalDateUnion = LocalDateIntersection | { ok: boolean };
```

The runtime matrix was:

| Validator/input | Result |
| --- | ---: |
| `createIs<Date>()({ stamp: 1 })` | `true` |
| `createIs<LocalDateIntersection>()({ stamp: 1, label: "x" })` | `true` |
| `createIs<LocalDateUnion>()({ stamp: 1, label: "x" })` | `false` |
| `createIs<LocalDateUnion>()({ ok: true })` | `true` |
| `createIs<LocalDateUnion>()({ stamp: 1 })` | `false` |

The emitted union validator contains only the `{ ok: boolean }` arm. The direct classifier already treats the local `.ts` declaration structurally; the divergence appears only when the intersection is categorized for composition and pruning. GitHub searches for `Date intersection` and native intersection classification found no issue or pull request owning this witness.

## Verified owner

`packages/typia/native/core/factories/internal/metadata/iterate_metadata_intersection.go:701` decides whether a composed intersection is impossible, while the category path around line 957 derives native identity from a symbol base name. That independent category can mark the local structural `Date` native and conflict it with the real object member. This locates the divergence without prescribing how composition should recover the correct identity.

## Consequence surface

The proven boundary is an inhabited intersection used in a union/prunable context. Standalone structural `Date` and standalone `Date & { label }` are correct. Operations consuming the composed metadata can lose the valid arm before their programmer runs; `createIs` is the executed witness.

## Approach

Enforce one composition invariant: composing a structural type into an intersection or union cannot change its runtime identity category. An inhabited structural intersection cannot be removed as `never` solely because one user declaration shares a built-in name. The implementation mechanism is deliberately left to the implementer to validate against the full intersection and pruning call graph.

- A regression covers standalone local `Date`, standalone `Date & { label }`, the same intersection as a union arm, the other union arm, and an invalid near-miss; it fails on current `5fce10b93d`.
- The valid structural intersection arm remains accepted after union composition.
- The `{ ok: boolean }` arm and invalid near-miss retain their current results.
- Genuine native intersections and truly impossible intersections keep their existing pruning behavior.
- The test executes emitted JavaScript and inspects enough emit to prove the structural arm is retained.
- Run the focused native regression, related intersection/native suites, the full native Go suite, `go vet`, and the relevant TypeScript validator workspace.

## Coordination

This issue is separate from package `.d.ts` native-name classification: its witness uses a correctly classified local `.ts` declaration and fails only during composition. It requires no campaign version bump. Per #2232, proposed mechanisms are hypotheses until positive and impossible-intersection controls both pass.

samchon:
Fixes #2237.

## Intent

Preserve an inhabited structural intersection when a local declaration shares a simple native name and the intersection is composed into a union.

## Scope

- Pair each intersection classifier's normalized type name with the symbol that supplied it.
- Apply the checker-resolved global type identity from #2245 at the four simple-native intersection paths: removable constraints, union identity, plain-object detection, and impossible-intersection categories.
- Add a tag-free real-transform regression covering local `Date`, `String`, and `Number` structural intersections; union alternatives and near-misses; genuine native and wrapper pruning; impossible intersections; and emitted-JavaScript proof.
- Keep generic, collection, and atomic-like name branches unchanged. Provisioned controls show local `Map`, `Set`, `WeakMap`, and `WeakSet` intersections already retain their structural union arms. The local `String`/`Number` controls pass on this PR without an additional atomic-like gate, while genuine global wrapper controls retain their existing pruning behavior.

No package version, `packages/interface/src`, dependency, workflow, provider claim, or public declaration changes are included.

## Red / green evidence

With the production file restored byte-for-byte to the verified base, the focused fixture emitted only the alternative object arms for all three colliding structural intersections:

```text
Date:   missing input.stamp and input.label
String: missing input.stringValue and input.label
Number: missing input.numberValue and input.label
```

The baseline runner executed all 44 runtime cases and reported exactly three mismatches: the valid local `Date & { label }`, `String & { label }`, and `Number & { label }` union arms. Collection-name controls and genuine global `String`/`Number` wrapper-pruning controls produced no emit or runtime mismatch.

On the current PR production tree, the same emit oracle retains every local structural field and alternative arm, does not emit the local declarations as genuine natives, and all 44 runtime cases pass. This also disproves the reviewer hypothesis that the atomic-like name branch requires a separate production gate.

## Verification

- `pnpm install --frozen-lockfile` (no lockfile or dependency diff)
- verified-base RED and current-head green: `go -C packages/typia/native test -p=1 -count=1 ./cmd/ttsc-typia -run '^TestNativeNamedIntersectionUnionTransform$' -v`
- focused and adjacent 10-test native matrix covering #2200, #2212, #2239, generic native-name guards, brand/tag intersections, all-never pruning, every nonsensible-intersection subtest, union validation, and atomic schema
- `go -C packages/typia/native test -p=1 -count=1 ./...` (116.7s, success on the same production blob before the test-only wrapper oracle extension)
- `go -C packages/typia/native vet -p=1 ./...` (8.9s, success on the same production blob before the test-only wrapper oracle extension)
- `pnpm --filter @typia/test-typia-automated start -- --include is` (412.7s, success)
- `git diff --check`
- solo Self-Review: final complete round clean

Full native and vet were not repeated after the test-only wrapper oracle extension; the changed fixture was instead exercised by the focused 44-case run and the adjacent 10-test matrix. Campaign policy intentionally skipped repository-wide `pnpm format`, the full repository `pnpm test`, and the already-completed D9 composition check. Automatic Actions runs for this implementation SHA are cancelled by the exact-SHA campaign gate; the local verification above is the delivery gate.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/native_named_intersection_union_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/native_named_intersection_union_transform_test.go
@@ -1,0 +1,390 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestNativeNamedIntersectionUnionTransform verifies a structural intersection
+// keeps its object identity when a user declaration shares a native name.
+//
+// The standalone local Date and Date intersection already use structural
+// metadata. Union composition must retain that classification instead of
+// pruning the inhabited arm as a native/object conflict. Native-named local
+// String and Number declarations must follow the same structural path, while
+// genuine native wrappers and impossible intersections retain their existing
+// pruning paths.
+//
+//  1. Transform standalone, union, genuine-native, impossible, and colliding
+//     native-name controls.
+//  2. Inspect each union factory to prove local structural arms are retained
+//     while genuine wrappers and existing collection arms keep their emit.
+//  3. Execute both valid arms, invalid near-misses, and pruning controls.
+func TestNativeNamedIntersectionUnionTransform(t *testing.T) {
+  project := nativeNamedIntersectionUnionProject(t)
+  js := nativeNamedIntersectionUnionTransform(t, project)
+
+  localUnion := nativeNamedIntersectionUnionExport(t, js, "isLocalUnion", "isLocalMapUnion")
+  for _, retained := range []string{"input.stamp", "input.label", "input.ok"} {
+    if !strings.Contains(localUnion, retained) {
+      t.Errorf("local structural union lost %q in emitted factory:\n%s", retained, localUnion)
+    }
+  }
+  if strings.Contains(localUnion, "instanceof Date") {
+    t.Errorf("local structural union was emitted as a genuine Date:\n%s", localUnion)
+  }
+
+  localWrapperUnions := []struct {
+    name    string
+    next    string
+    members []string
+    native  string
+  }{
+    {"isLocalStringUnion", "isLocalNumberUnion", []string{"input.stringValue", "input.label", "input.stringOk"}, "String"},
+    {"isLocalNumberUnion", "isLocalMapUnion", []string{"input.numberValue", "input.label", "input.numberOk"}, "Number"},
+  }
+  for _, control := range localWrapperUnions {
+    segment := nativeNamedIntersectionUnionExport(t, js, control.name, control.next)
+    for _, retained := range control.members {
+      if !strings.Contains(segment, retained) {
+        t.Errorf("%s lost structural member %q in emitted factory:\n%s", control.name, retained, segment)
+      }
+    }
+    if strings.Contains(segment, "instanceof "+control.native) {
+      t.Errorf("%s was emitted as a genuine %s wrapper:\n%s", control.name, control.native, segment)
+    }
+  }
+
+  collectionUnions := []struct {
+    name    string
+    next    string
+    members []string
+  }{
+    {"isLocalMapUnion", "isLocalSetUnion", []string{"input.brandMap", "input.valueMap", "input.label", "input.mapOk"}},
+    {"isLocalSetUnion", "isLocalWeakMapUnion", []string{"input.brandSet", "input.label", "input.setOk"}},
+    {"isLocalWeakMapUnion", "isLocalWeakSetUnion", []string{"input.brandWeakMap", "input.valueWeakMap", "input.label", "input.weakMapOk"}},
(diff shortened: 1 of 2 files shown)


samchon:
Commit-integrity follow-up:

- Amended the existing commit message only: c670207d970c9db2d66702e0d385083478662851 -> 383cc697a843aea7f23e3acef3547954033800e0.
- The code tree is byte-identical: tree 26aa4b64a080d10256af33efd8567be8844b8d2f before and after; git diff between the two trees is empty.
- git interpret-trailers --parse now recognizes all eight Lore trailers from Constraint through Related.
- No tests were rerun because the committed tree did not change; the PR's existing verification evidence therefore remains exact.
- The exact-SHA cancellation gate polled three times and confirmed all six workflow runs for the new SHA completed as cancelled.

samchon:
Reviewer-hypothesis follow-up:

- Added real-transform emit/runtime controls for local structural String and Number intersections in unions, plus genuine global String and Number wrapper controls.
- On the base production file, the fixture was RED: local Date/String/Number structural arms were absent from emit and all three valid runtime cases returned false.
- On this PR's unchanged production implementation, all 44 runtime cases and emit oracles pass; genuine wrapper intersections retain their existing pruning behavior. The proposed separate atomic-like name gate was therefore rejected as unnecessary.
- The 10-test focused/adjacent matrix passed, including all IntersectionNonsensibleBackstop subtests and the #2200/#2212/#2239 identity controls.
- The production blob remains exactly 336fd1e5b1a78a411fda513a9a6565e988d99fbb; compared with 383cc697a843aea7f23e3acef3547954033800e0, only the regression-test file changed.
- Commit amended to 53ae517dfc17b5c6d0f6951590645f163905983d. All nine Lore trailers parse contiguously.
- Full native and vet were not repeated because production is byte-identical; the new cases were exercised by focused and adjacent runs.
- Exact-SHA cancellation polled five times and confirmed all six workflow runs completed as cancelled.

Issue #2226, opened 2026-07-18
A symbol-keyed data member leaks a mangled key: is<T> is always false for a real instance and json.schema emits an uninhabitable required property (the #2213 residual for symbol keys)


samchon:
## Problem

A `symbol`-keyed data property leaks into the validated shape and the JSON schema under the TypeScript-internal escaped name (`"�@<name>@<id>"`) that no runtime object carries. `is`/`assert`/`validate` are then unconditionally `false` for a real instance, and `json.schema` emits an uninhabitable required property. This is the exact #2213 pattern (`#private` leaking a mangled key), for symbol keys instead of private fields.

## Evidence — bytes on `master` (`dbf147b4bb`)

```ts
declare const sym: unique symbol;
interface Foo { name: string; [sym]: number; [Symbol.toStringTag]: string; }
typia.createIs<Foo>();
typia.json.schemas<[Foo]>();
```
```js
// is<Foo>
_io0 = input => "string"===typeof input.name
             && "number"===typeof input["�@sym@5"]
             && "string"===typeof input["�@toStringTag@6"];
// json.schemas<[Foo]> — the symbol keys become required properties under mangled names
{ properties: { name:{…}, "�@sym@5":{…}, "�@toStringTag@6":{…} },
  required: ["name", "�@sym@5", "�@toStringTag@6"], additionalProperties:false }
```

`input["�@sym@5"]` is a mangled string key no object carries (the real member is under a `symbol`, unreachable by that string), so `isFoo({ name:"x", [sym]:1, [Symbol.toStringTag]:"Foo" })` is **`false`** — a real object with every declared member fails its own guard. The emitted schema declares an uninhabitable required property under the mangled name. An intersection (`A & B`) leaks identically, and `Promise<T>` / boxed `Symbol` inherit it through their `[Symbol.toStringTag]` member.

## Root cause

`packages/typia/native/core/factories/internal/metadata/emplace_metadata_object.go` — the `pred` member filter (`~:93-111`) excludes `private`/`protected` keyword members and (post-#2213) ES `#private`, but not symbol-keyed members; the key is then built at `~:140` (`MetadataHelper.Literal_to_metadata(symbol.Name)`) from the internal escaped symbol name, and the intersection path (`~:562`) has the same gap (confirmed leaking).

[...]

PR #2248, opened 2026-07-18
fix(native): keep symbol index signatures out of string-key metadata

### LINKED ISSUE #2240: fix(native): keep symbol index signatures out of string-key metadata
## Problem

A TypeScript symbol index signature is converted into a string-key dynamic property by the native transformer. It incorrectly constrains unrelated `Object.keys` values and becomes JSON Schema `additionalProperties`, even though symbol keys are outside typia's structural JSON shape.

## Evidence

The D10 reviewer and the lead independently reproduced this on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6`. The lead's real-transform runtime fixture repeated the validator mismatch and emitted these exact schema differences: `SymbolIndex.additionalProperties` was `{ "type": "number" }`, while the plain control's was `false`.

```typescript
interface SymbolIndex {
  name: string;
  [key: symbol]: number;
}

type _NoStringIndex = Assert<string extends keyof SymbolIndex ? false : true>;
type _HasSymbolIndex = Assert<symbol extends keyof SymbolIndex ? true : false>;
```

The compile-time assertions prove the checker retains distinct key domains. The emitted validator nevertheless applies the `number` value type to ordinary string extras:

| Input | Result |
| --- | ---: |
| `{ name: "x" }` | `true` |
| `{ name: "x", extra: 1 }` | `true` |
| `{ name: "x", extra: "not-a-number" }` | `false` |
| `{ name: "x", [Symbol()]: 1 }` | `true` |
| `{ name: "x", [Symbol()]: "ignored" }` | `true` |

The emitted `SymbolIndex` schema contains `additionalProperties: { type: "number" }`; an otherwise identical `PlainObject` schema contains `additionalProperties: false`. This is distinct from #2226: explicit symbol-keyed members use a different metadata path and their regression already passes. Open issue and pull-request searches for `symbol index` and `symbol-keyed index` found no owner.

## Verified owner

`packages/typia/native/core/factories/internal/metadata/emplace_metadata_object.go:184` iterates every checker `IndexInfo`, analyzes its key and value, and inserts it as a dynamic object property. The path does not distinguish the executed symbol key domain before downstream validators and schema programmers consume it. This identifies the domain conversion boundary without prescribing a filter or representation.

## Consequence surface

The executed consequences cover non-strict validation and JSON Schema. Every consumer of object dynamic-property metadata can see the symbol index as a JSON string index, with operation-specific effects. Explicit symbol values are correctly ignored by the runtime witness; the defect is the accidental constraint on unrelated string keys.

## Approach

Enforce one domain invariant: symbol and string index domains remain distinct throughout metadata analysis. The established exclusion of symbol-keyed members from typia's structural JSON shape applies consistently to symbol index signatures; a symbol index cannot constrain `Object.keys` values or become JSON Schema `additionalProperties`. The implementation mechanism is deliberately left to the implementer to validate against checker index metadata and downstream consumers.

- A regression with `[key: symbol]: number` and compile-time `keyof` controls fails on current `5fce10b93d`.
- In non-strict validation, unrelated string extras behave exactly as they do for the plain-object control.
- Symbol-valued properties remain outside the structural JSON validation surface.
- `SymbolIndex` and its `PlainObject` control emit exactly equivalent string-key JSON Schema constraints.
- A normal `[key: string]` control continues to validate and emit as a string-key dynamic property.
- Existing explicit symbol-member regressions remain green.
- Run the focused native regression, symbol-member/schema suites, the full native Go suite, `go vet`, and relevant TypeScript validator/schema workspaces.

## Coordination

This is a separate metadata ingestion path from #2226 and requires no campaign version bump. Per #2232, any metadata-filtering mechanism is a hypothesis until both symbol-index and ordinary string-index controls pass.

samchon:
## Intent

Keep TypeScript symbol index signatures outside typia's string-keyed structural JSON shape. A symbol index, including a branded/intersection symbol domain, must not constrain `Object.keys` values or become JSON Schema `additionalProperties`.

## Scope

- Exclude checker index entries whose key domain is symbol-like directly or through nested intersections before dynamic object metadata is created.
- Preserve string and number index entries on the existing dynamic-property path; TypeScript-Go already exposes union index domains as separate `IndexInfo` entries.
- Add a tag-free real-transform regression with bare and branded symbol indexes, compile-time `keyof` controls, validator runtime cases, exact JSON Schema controls, a normal string index, and the explicit symbol-member behavior from #2226.
- Leave the separate explicit symbol-member filter unchanged.

## Evidence

- Initial RED: a bare symbol index rejected an unrelated string extra and emitted `additionalProperties: { type: number }` instead of matching the plain-object schema.
- Reviewer hypothesis RED on the initial PR implementation: `symbol & { readonly __brand: unique symbol }` was accepted as an index key by the checker but failed both `typia.createIs` and `typia.json.schemas` transforms as unsupported.
- Temporary diagnostics identified the relevant checker representation: bare symbol key flags were `512` (`ESSymbolLike`), while the branded key flags were `268435456` (`Intersection`). This disproved the initial direct-flags-only mechanism.
- GREEN: recursive intersection inspection keeps both bare and branded symbol domains outside typia's structural JSON shape while string-index and explicit-symbol controls remain unchanged.

## Verification

- `go test ./cmd/ttsc-typia -run '^TestSymbolIndexDomainTransform$' -count=1 -v`
- `go test ./cmd/ttsc-typia -run '^(TestSymbolIndexDomainTransform|TestSymbolKeyMemberFilterTransform|TestIntersectionBrandTransform|TestJsonSchemaDynamicKeyDeterminismTransform)$' -count=1 -v`
- `go test -p=1 -count=1 ./...`
- `go vet -p=1 ./...`
- repository Go formatter and `git diff --check`
- Solo Self-Review completed with a clean full round after the branded-symbol remediation.

The full native suite passed in 98.7 seconds and the full schema workspace passed in 69.9 seconds. `pnpm install --frozen-lockfile` provisioned the worktree without changing the lockfile. The repository-wide `pnpm format` and full `pnpm test` were intentionally not run under the active issue-campaign policy; package-scoped native and schema verification is recorded above. No version bump is included because campaign releases remain maintainer-owned.

## Campaign chronology

This PR is implementation-first: the initial implementation commit and draft PR existed before the campaign's claim-first reservation and knowledge-base record. No empty claim commit was created later to manufacture the preferred chronology. The exact chronology, prior gates, amended head, and one corrected gate-script aggregation error are recorded in the campaign knowledge base.

Automatic Actions for this campaign commit are cancelled by exact SHA after every push and relevant pull-request mutation.

Fixes #2240

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/symbol_index_domain_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/symbol_index_domain_transform_test.go
@@ -1,0 +1,316 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "testing"
+)
+
+// TestSymbolIndexDomainTransform verifies a symbol index signature stays out
+// of typia's string-keyed validation and JSON Schema shape (#2240).
+//
+// TypeScript keeps string and symbol index domains distinct, including when a
+// symbol index key is intersected with a nominal brand. This regression pins
+// that boundary through the real transform because treating every checker
+// IndexInfo as a JSON string index makes unrelated Object.keys values inherit a
+// symbol index's value constraint. An ordinary string index and the explicit
+// symbol-member behavior fixed by #2226 are adjacent controls.
+//
+//  1. Compile checker-level keyof assertions for symbol and string indexes.
+//  2. Execute validators over ordinary extras and symbol-valued properties.
+//  3. Compare the symbol-index schema exactly with a plain-object control while
+//     requiring a normal string index to retain additionalProperties.
+func TestSymbolIndexDomainTransform(t *testing.T) {
+  project := symbolIndexDomainProject(t)
+  js := symbolIndexDomainTransform(t, project)
+  symbolIndexDomainRunRuntimeCases(t, project, js)
+}
+
+func symbolIndexDomainProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "symbol-index-domain-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(symbolIndexDomainTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(symbolIndexDomainSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func symbolIndexDomainTransform(t *testing.T, project string) string {
+  t.Helper()
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("symbol index domain transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  return out
+}
+
+func symbolIndexDomainRunRuntimeCases(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  runtimeJS := ttscTypiaTestRewriteCommonJS(t, js)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(runtimeJS), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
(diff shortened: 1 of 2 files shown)


samchon:
Solo Self-Review completed with a clean full round.

- Reviewed the entire two-file base-to-head diff and the symbol-index consequence surface from checker `IndexInfo` through validator `Object.keys` emit and JSON Schema `additionalProperties`.
- Confirmed the production condition is limited to `TypeFlagsESSymbolLike`; string and number index entries remain on the existing path, including the normal string-index runtime/schema control.
- Confirmed the explicit symbol-member path from #2226 is unchanged and its real-transform regression passes.
- Checked test isolation, Windows/POSIX path handling, cache cleanup, exact schema assertions, public compatibility, package boundaries, version/interface/workflow exclusions, and formatting.
- No sound improvement remained after the full round.

The first adjacent run before `pnpm install` reproduced the known unprovisioned tag symptom in `TestIntersectionBrandTransform`. After `pnpm install --frozen-lockfile` (no lockfile diff), the same test and the complete adjacent bundle passed; the full native suite, vet, and schema workspace then passed.

Exact-SHA campaign gate after PR creation observed five cancelled runs and one run that was already terminal with success; the gate completed successfully after four polls.

samchon:
Final Solo Self-Review completed with a clean full round after the branded-symbol remediation. This supplements the initial review comment rather than erasing its chronology.

- Treated the branded-symbol report as a hypothesis and reproduced it through the real transform before changing production code: both validator and schema transforms rejected the branded index key as unsupported.
- Confirmed the actual checker representation with temporary diagnostics: bare `symbol` used flags `512` (`ESSymbolLike`), while `symbol & brand` used flags `268435456` (`Intersection`). The temporary diagnostics are absent from the final diff.
- Reviewed the complete base-to-head diff and the consequence path from checker `IndexInfo` through metadata ingestion, `Object.keys` validation, and JSON Schema `additionalProperties`.
- Confirmed the helper recurses only through intersection constituents; direct symbol-like keys are excluded, while TypeScript-Go's separate union `IndexInfo` entries keep string and number siblings on the existing dynamic-property path.
- Confirmed the #2226 explicit symbol-member path is unchanged and that bare symbol, branded symbol, normal string index, explicit symbol member, intersection-brand, and schema-determinism controls all pass.
- Full evidence is green: focused and adjacent real-transform tests; `go test -p=1 -count=1 ./...`; `go vet -p=1 ./...`; full `test-typia-schema`; repository Go formatter; `git diff --check`.
- Checked test isolation, Windows/POSIX path handling, exact schema assertions, public compatibility, package boundaries, version/interface/dependency/provider/workflow exclusions, Lore trailer continuity, and worktree cleanliness.
- No sound improvement remained after the full round.

Amended implementation SHA: `85bca47ebe472bd95527deed1b4ccf11535742ed`. Its push gate cancelled five exact-SHA runs; spell-check was already terminal success. The initial gate-script invocation had a nested-array aggregation error and failed without cancelling unrelated runs; the corrected five-poll gate ended with two quiet terminal polls. The campaign knowledge base records both the implementation-first chronology deviation and this gate correction.

PR #2254, opened 2026-07-19
fix(json): keep contextual any and unknown serialization valid

### LINKED ISSUE #2253: fix(json): keep contextual any and unknown serialization valid
## Problem

`typia.json.*Stringify` can return success-shaped strings that are not JSON when a statically accepted `any` or `unknown` value contains a function or symbol. Dynamic object properties concatenate the token `undefined`, array slots become empty elements, and validation wrappers can report success for the malformed text.

## Evidence

The D11 reviewer and lead independently executed this on `c7f7939d2bb367bf5a0571b1becf889b67c3a472`; current master `ce88a887c9b9fac0ebbbdff088ae38df325200d3` changes only skill documents after that SHA.

For `Record<string, any>` values, all 16 direct/factory `stringify`, `isStringify`, `assertStringify`, and `validateStringify` function/symbol cells produced malformed text. Representative results were:

```json
{
  "rawFunction":"{\"keep\":1,\"omit\":undefined,\"nil\":null}",
  "rawSymbol":"{\"keep\":1,\"omit\":undefined,\"nil\":null}",
  "guardedFunction":"{\"keep\":1,\"omit\":undefined,\"nil\":null}",
  "validatedSymbol":{"success":true,"data":"{\"keep\":1,\"omit\":undefined,\"nil\":null}"},
  "arrayFunction":"[,null,null]"
}
```

Every shown string fails `JSON.parse`. Native `JSON.stringify` omits function, symbol, and `undefined` object-property values and writes `null` for the same values in array slots. The same defect was executed for `Record<string, unknown>`, `any[]`, `unknown[]`, tuples, nested containers, and static `any`/`unknown` symbol properties. Arrays emitted `[,null,null,1]`, tuples emitted `[undefined,null,null,1]`, and a static symbol property emitted `{"value":undefined}`.

Static function omission, all tested `undefined` container cases, scalars, explicit function metadata, and `toJSON` matched native JSON. Top-level `any` function, symbol, and `undefined` matched native `JSON.stringify` across the executed direct/factory and guarded families by returning JavaScript `undefined`; `validate<any>` returned `success: true, data: undefined`. The executed raw `unknown` factory also returned JavaScript `undefined`. Those top-level controls do not promise JSON text and are outside the malformed-child defect.

The executed owner is `packages/typia/native/core/programmers/helpers/StringifyJoinder.go` together with its callers in `JsonStringifyProgrammer.go`. Dynamic composition filters exact `undefined` before concatenating an expression whose evaluated result can still be `undefined`, while array composition can join an empty serialized expression. This is ownership evidence, not an implementation prescription.

Open and closed searches found no owner. #1673 concerns non-finite numeric tokens, #719 optional `any` property disappearance, #342 `Array<any>` performance/delegation, and #2201 static-key escaping. None owns contextual function or symbol child serialization.

## Invariant

Whenever ECMAScript `JSON.stringify` produces text for a statically accepted input, every typia success payload must be parseable and context-equivalent. A child serialization of `undefined` is omitted from objects and becomes `null` in arrays and tuples. Top-level JavaScript `undefined` remains an explicit control unless separately changed. Direct and factory, raw and guarded, and assertion and validation forms must agree on the resulting text or failure.

- Differentially compare typia output with native `JSON.stringify`; whenever the oracle produces text, require every typia success result to pass `JSON.parse` and match the oracle's contextual semantics.
- Cross `any` and `unknown` with function, symbol, and `undefined` values in static properties, dynamic/index-signature properties, arrays, tuples, nested containers, and mixed valid neighbors.
- Cross direct/factory and stringify/is/assert/validate families; validation cannot report success for malformed text.
- Retain already-correct static-function omission, `undefined` container behavior, top-level `undefined` results, ordinary scalar/object/array behavior, key escaping, `toJSON`, and explicit function-metadata controls.
- Enumerate every caller of the shared joinder and every context that can evaluate an accepted expression to `undefined`; each positive cell needs a one-axis valid or invalid twin.
- Inspect generated bytes and execute them against the independent native/parse oracle.
- Retain the regression while the independent verifier mutates only the product fix and proves the canonical command fails while adjacent controls pass.
- Run `go -C packages/typia/test test ../native/cmd/ttsc-typia -count=1`, `go -C packages/typia/test test ../native/...`, `go -C packages/typia/native vet -p=1 ./...`, `go -C packages/typia/test vet -p=1 ./...`, `pnpm --filter ./tests/test-typia-automated start`, `pnpm build`, and `pnpm test` from fresh worktrees at both the prospective merge and exact merged SHA. Allow at least 20 minutes for the first cold native build.

## Coordination

No package version, tag, release, or publication change belongs in this issue. The issue states contextual output invariants and owner constraints only; the implementer validates the mechanism.

samchon:
Closes #2253.

This draft reserves the independently reproduced JSON contextual-serialization defect before implementation. The product invariant is that every child value accepted through `any` or `unknown` follows ECMAScript JSON context: object children that serialize to JavaScript `undefined` are omitted, while array and tuple children become `null`; successful text must parse as JSON.

Implementation has not started. The batch will add real-transform regressions across direct/factory and raw/is/assert/validate families, enumerate every shared joinder consumer, inspect emitted JavaScript, and mutation-prove the permanent oracle before the prospective and exact-merge integration gates.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/json_stringify_contextual_undefined_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/json_stringify_contextual_undefined_transform_test.go
@@ -1,0 +1,556 @@
+package main
+
+import (
+  "fmt"
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestJsonStringifyContextualUndefinedTransform pins every json stringify family
+// against ECMAScript JSON.stringify wherever a child serialization evaluates to
+// JavaScript undefined (#2253).
+//
+// A child can serialize to undefined while the value itself is present: `any`
+// and `unknown` delegate to JSON.stringify, which answers undefined for a
+// function, a symbol, and a toJSON that returns undefined; a property typed
+// `undefined` has nothing to serialize; and an array hole reads as undefined at
+// any element type. The object joiner decided omission from a test on the input
+// and the array joiner joined the mapped results, so those children concatenated
+// the token `undefined` into a static property, produced `{"k":undefined}` for a
+// dynamic one, and left an empty array slot - text that fails JSON.parse, which
+// the is/assert/validate families then reported as success.
+//
+//  1. Transform one fixture that exports the direct and factory forms of
+//     stringify, isStringify, assertStringify, and validateStringify for `any`
+//     and `unknown` in static, optional, dynamic, array, tuple, rest, and nested
+//     positions, plus the toJSON, sparse-array, and typed controls beside them.
+//  2. Execute the emitted serializers in Node and compare every success payload
+//     with native JSON.stringify of the same input, parsing both so the
+//     comparison is contextual equivalence rather than the current emit.
+//  3. Require the guarded families to reject each one-axis invalid twin, run the
+//     whole matrix again under `undefined: false`, and report the exact executed
+//     case count so a runner that silently skipped cells cannot pass as green.
+func TestJsonStringifyContextualUndefinedTransform(t *testing.T) {
+  for _, row := range jsonStringifyContextualUndefinedRows() {
+    t.Run(row.name, func(t *testing.T) {
+      project := jsonStringifyContextualUndefinedProject(t)
+      js := jsonStringifyContextualUndefinedTransform(t, project, row.payload())
+      jsonStringifyContextualUndefinedRunRuntime(t, project, js, row)
+    })
+  }
+}
+
+type jsonStringifyContextualUndefinedRow struct {
+  name string
+  // config is the JSON fragment appended to typia's resolved plugin entry.
+  config string
+  // strict marks the row where `undefined: false` makes the checkers reject an
+  // explicitly undefined optional property. The serialized text of everything
+  // the checkers still accept must not move.
+  strict bool
+}
+
+func (row jsonStringifyContextualUndefinedRow) payload() string {
+  return `[{"config":{"transform":"typia/lib/transform"` + row.config + `},"name":"typia","stage":"transform"}]`
+}
+
+func jsonStringifyContextualUndefinedRows() []jsonStringifyContextualUndefinedRow {
+  return []jsonStringifyContextualUndefinedRow{
+    {name: "default options", config: ``, strict: false},
+    {name: "undefined false", config: `,"undefined":false`, strict: true},
+  }
+}
+
+// jsonStringifyContextualUndefinedTypes names every fixture type the runner
+// drives. The Go side owns the list so the exported symbol names and the
+// runner's lookup keys cannot drift apart.
+var jsonStringifyContextualUndefinedTypes = []struct{ Key, TS string }{
+  {"anyProperty", "IAnyProperty"},
+  {"unknownProperty", "IUnknownProperty"},
+  {"undefinedProperty", "IUndefinedProperty"},
+  {"optionalAny", "IOptionalAny"},
+  {"recordAny", "IRecordAny"},
+  {"recordUnknown", "IRecordUnknown"},
+  {"mixedRecord", "IMixedRecord"},
+  {"anyArray", "IAnyArray"},
+  {"unknownArray", "IUnknownArray"},
+  {"numberArray", "INumberArray"},
+  {"tuple", "ITuple"},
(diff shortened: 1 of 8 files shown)


samchon:
State of this pull request after the campaign consolidation.

#2253 is **not** included in the consolidated #2266 and stays open here. The branch now carries the analysis and the partial work as an explicitly incomplete commit, so nothing is lost, but it must not be merged as-is.

**Root cause, re-derived from source.** `jsonStringifyProgrammer_decode` emits `JSON.stringify(input)` for `Any` metadata, and `JSON.stringify` answers JavaScript `undefined` for a function, a symbol, and a `toJSON` that returns undefined while the value is still present. The object joiner decides omission from a test on the input (`stringifyJoiner_regular_condition`: `undefined === input || "function" === typeof input`), which cannot see a symbol or a `toJSON` result. The array joiner uses `.map(...).join(",")`, where a mapped `undefined` and a hole both render as empty text.

**The class is wider than the reported witness.** A sparse `number[]` emits `[,3]`, and a required property whose `toJSON` returns undefined emits `{"k":undefined}` — text that is not valid JSON at all. A hole exists at runtime whatever the element type declares, so this is not only an `any`/`unknown` concern.

**What is missing, and why this was excluded rather than shipped.**

1. The `_jsonStringifyArray` runtime helper is orphaned. `StringifyJoiner.Array` still emits `input.map(arrow).join(",")` and nothing imports or calls the helper.
2. The object-property half has no implementation at all, while the tests on this branch assert it.
3. The `undefined: false` oracle question is unresolved. Under that option typia may legitimately skip undefined checks, so native `JSON.stringify` is not an unqualified oracle there, yet the tests assert against it in both modes. Resolving that contract must come before the implementation, or the fix will be built to satisfy a wrong assertion.

**Nothing here has been compiled or executed.** It was written under a maintainer stop on all local building and testing; no cell of either test has ever run.

samchon:
Implemented. This pull request now carries the fix, not the analysis.

**Root cause.** Omission was decided from a test on the *input* (`undefined === input || "function" === typeof input`), which cannot see a symbol or a `toJSON` result, and array elements were joined with `.map(...).join(",")`, where a mapped `undefined` and a hole both render as empty text. The decision is now made from the serialized result, evaluated once — a `toJSON` may answer differently on a second call, so testing the input and then serializing is a different program.

**Three positions, three answers**, matching ECMAScript rather than one rule everywhere: `_jsonStringifyProperty` drops a member together with its separator; `_jsonStringifyArray` and `_jsonStringifyElement` write `null`, because an array position cannot be dropped without shifting what follows.

**Fixed as a class, not as the reported witness:**

| symptom | position |
| --- | --- |
| `{"k":undefined}` | static member, index-signature member |
| `[,1]` | sparse array |
| `[undefined,1]` | array or tuple element with no serialization |
| `{"a":1,}` | trailing separator when the omitted member was last |

That last one was reachable without any of this issue's inputs: the tail helper was skipped whenever the last member was declared *required*, but a required `any` member is still omitted when its value has no text. A declared `toJSON` whose return type admits `undefined` reaches the same state with no `any` anywhere, and a hole exists at runtime whatever the element type declares — so neither is confined to `any`/`unknown`.

**Verification.** The transform regression executes 354 runtime cases across direct and factory forms and all four stringify entry points, in default and `undefined: false` modes, comparing every result against native `JSON.stringify` as the oracle. The count is asserted, not the exit code, so a row that stops running cannot pass as a row that agreed. Fail-before was observed on this same regression — it reported malformed text for records, dynamic members, arrays, and tuples before the change, and the failures disappeared class by class as each position was fixed. The whole `native/cmd/ttsc-typia` package passes locally (106 s). The TypeScript suite added here exercises the real shipped helpers rather than the stubs, and CI owns it.

**Change integrity.** `transform_helpers_test.go` gains runtime stubs for the three new helpers. Its shared rewriter fails loudly on any unmapped `typia/lib/internal/` import, so without them every transform test that serializes an array would stop resolving.

**Deferred, not dropped.** A non-callable `toJSON` twin is rejected by the declared type, but the checker admits it and the serializer then throws `input.value.toJSON is not a function` — so `isStringify` and `validateStringify` throw where their contract is to answer, never to throw. That is a callability rule rather than this contextual one, and mixing them would put two root causes in one diff. It is filed separately and keeps its own twin there.

Issue #2238, opened 2026-07-18
fix(native): preserve function semantics for call-only and construct-only interfaces


samchon:
## Problem

Call-only and construct-only interfaces are treated as ordinary object shapes, while mutually assignable function and constructor aliases receive typia's documented function metadata semantics. Real functions and classes are therefore rejected solely because the type was declared with interface syntax.

## Evidence

The D10 reviewer and the lead independently reproduced this on exact `origin/master` `5fce10b93dfe3807423312b447d55f7b2265f5c6`. The lead's real-transform fixture repeated the interface failures beside the passing alias controls under both default and `functional: true` options.

```typescript
interface Callable { (value: number): string; }
interface Constructable { new (value: number): { value: number }; }
type CallableAlias = (value: number) => string;
type ConstructableAlias = new (value: number) => { value: number };
```

Bidirectional `Same<X, Y>` assertions compile for each interface/alias pair. For holders with these nested properties, the alias matrix follows the validator option contract:

| Spelling/mode | Real function + class | Object placeholders |
| --- | ---: | ---: |
| aliases, default | `true` | `true` |
| aliases, `functional: true` | `true` | `false` |
| interfaces, default | `false` | `false` |
| interfaces, `functional: true` | `false` | `false` |

The interface emit expands apparent `Function` properties and first requires `typeof value === "object"`, which rejects actual functions and classes. This is a residual distinct from #2178's global `Function` interface fix. Open issue and pull-request searches for `callable interface`, `call signature`, and `construct signature` found no owner.

## Verified owner

`packages/typia/native/core/factories/internal/metadata/MetadataHelper.go:142` recognizes function-like declaration nodes and a narrow global `Function` interface case. A user call-only or construct-only interface has an `InterfaceDeclaration` node and falls through to structural object metadata despite its checker signatures. This locates the syntax-dependent split without prescribing how signatures or hybrid interfaces should be represented.

After the independent runtime runs, `origin/master` advanced to `4cba3fdee3` through an MCP feature and maintainer release commit. A lead `git diff --quiet 5fce10b93d..4cba3fdee3 -- packages/typia/native` check passed: every owning and emitting native byte above is unchanged on current master.

## Consequence surface

The executed witness covers nested call-only and construct-only properties under default and `functional: true` transform options. Top-level occurrences pass through related metadata recognition, but they were not executed and are not claimed here. Hybrid callable objects with declared data properties are also outside the reproduced scope and must be treated as a negative boundary rather than silently generalized.

[...]

PR #2255, opened 2026-07-19
fix(compare): preserve functional union membership during equality routing

### LINKED ISSUE #2252: fix(compare): preserve functional union membership during equality routing
## Problem

With `functional: true`, `typia.compare.equals` can route two valid values of a later object-union member through an earlier function-bearing member and ignore data that belongs to the actual member. The validator and plain transforms select the later member, but equality can return `true` for values whose later-only fields differ. Union declaration order therefore changes equality.

## Evidence

The D11 reviewer and lead independently executed this on `c7f7939d2bb367bf5a0571b1becf889b67c3a472`. The only later change through current master `ce88a887c9b9fac0ebbbdff088ae38df325200d3` is the four skill documents in #2251; `packages/**` and `tests/**` are byte-identical.

```typescript
type FunctionUnion =
  | { handler: (value: number) => string; common: string }
  | { handler: string; common: string; label: string };

const left = { handler: "text", common: "same", label: "left" };
const right = { handler: "text", common: "same", label: "right" };
```

The real transform and runtime produced:

```json
{"isLeft":true,"isRight":true,"directEquals":true,"factoryEquals":true,"nestedEquals":true,"cloneLeft":{"handler":"text","common":"same","label":"left"},"cloneRight":{"handler":"text","common":"same","label":"right"}}
```

The authoritative `is` and plain membership emit distinguishes `typeof input.handler === "function"` from `"string"`. Equality's first-member fallback omits the function-valued property while matching the earlier member, so both string-arm values enter that member and their distinct `label` fields are never compared. Reversing the union order returns the correct `false`; equal string-arm values remain equal; two actual function-arm values retain the existing method-identity-ignored result; `compare.cover` returns the correct `false` for this witness.

The independent critic also executed two distinct `{ handler: 1, common: "same", label: ... }` objects. Under `functional: true`, `is` rejects both values but equality still returns `true`: the fallback assigns both operands to the first member and discards the differing labels. The same reference remains equal through the established `x === y` fast path and is a separate control. Under default mode the function-valued property is deliberately ignored, so the number-valued objects satisfy the first member and their equality result is not evidence for this defect.

## Verified ownership

The executed owner is `packages/typia/native/core/programmers/compare/CompareEqualProgrammer.go` and its shared method classifier in `CompareLessProgrammer.go`. The fallback and method omission were last changed in #2225, while #2234 changed template membership without changing this function-property path. This identifies the owner and regression history but does not prescribe a replacement algorithm.

Open and closed issue and pull-request searches found no owner. #2214, #2225, and #2234 cover adjacent union-routing cases, while #2238 covers callable metadata; none states or tests this transform-option-dependent membership divergence.

## Consequence surface

The executed failures cover direct, factory, and nested `compare.equals` under `functional: true`. Direct and factory `is`, clone, classify, and prune are positive membership controls and preserve the later member's data. Reversed order proves the result is routing-dependent. Distinct function-property values rejected by `is` still acquire the fallback member and compare equal. Default mode, `compare.cover`, and the same-reference fast path are executed controls; `compare.less` rejects multi-object unions with its existing diagnostic. Pure callable and constructable spellings, hybrid callable objects, aliases, intersections, and every compare caller sharing the owner remain mandatory implementation boundaries rather than inferred failures.

## Invariant

Under the active transform options, object-union equality must resolve is-valid operands consistently with authoritative validator and plain membership paths, and distinct operands rejected by that membership cannot acquire a fallback member and become equal. Function-valued properties may retain their established comparison treatment after a member is selected. The existing same-reference fast path is a separate boundary.

- Preserve the current failure as a real emitted-JavaScript regression for direct, factory, and nested equality.
- Cover both union orders, equal and unequal later-member values, two distinct rejected values, the same rejected reference as a fast-path control, and actual function-member values.
- Cross default and `functional: true`; each applicable positive cell has a one-axis negative twin.
- Keep direct and factory `is`, clone, classify, prune, cover, and less boundaries explicit, including the operations that correctly reject or handle this union.
- Enumerate every caller of the owning membership helper and exercise every equivalent function spelling and declaration provenance that reaches the same control flow, including the #2238 interface and type-literal boundaries once integrated.
- Inspect generated bytes and execute them; metadata-only assertions are insufficient.
- The independent verifier retains the regression while reverting or mutating only the product fix and proves the expected assertion fails while controls pass.
- Run `go -C packages/typia/test test ../native/cmd/ttsc-typia -count=1`, `go -C packages/typia/test test -tags typia_native_internal ../native/...`, `go -C packages/typia/native vet -p=1 ./...`, `go -C packages/typia/test vet -p=1 ./...`, `pnpm --filter ./tests/test-typia-automated start`, `pnpm build`, and `pnpm test` from fresh worktrees at both the prospective merge and exact merged SHA. Allow at least 20 minutes for the first cold native build.

## Coordination

No package version, tag, release, or publication change belongs in this issue. Closed issue mechanisms are hypotheses; the implementer owns executable consequence analysis and the solution.

samchon:
Closes #2252.

This draft reserves the independently reproduced functional union-membership defect before implementation. Equality must use the same active-option member distinction as authoritative validation and plain routing; two distinct operands rejected by that membership must not acquire a fallback member and become equal. Existing same-reference and post-selection function-property comparison behavior remain explicit controls.

Implementation has not started. The batch will add direct/factory/nested real-transform regressions, both union orders, valid and rejected distinct-reference controls, enumerate every shared classifier caller, inspect emitted JavaScript, and mutation-prove the permanent oracle before prospective and exact-merge integration gates. The #2238 declaration-spelling residual is a cross-batch integration dependency, not a mechanism prescription.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/compare_functional_union_membership_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/compare_functional_union_membership_transform_test.go
@@ -1,0 +1,343 @@
+package main
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestCompareFunctionalUnionMembershipTransform verifies that compare.equals
+// selects an object-union member with the same functional-property semantics as
+// typia.is and the plain member routers (samchon/typia#2252).
+//
+// Object comparison deliberately ignores function-valued properties after it
+// has selected a member. That must not make those properties invisible while
+// selecting a member when the active `functional: true` option makes them part
+// of membership. Otherwise a string-valued later member enters an earlier
+// function-bearing member, and equality or cover can ignore the later member's
+// distinguishing data. The default mode remains intentionally lenient for a
+// sole function-valued property.
+//
+//  1. Transform functional and default-option variants of direct/factory is,
+//     equals, cover, clone, prune, and nested equality calls over both union
+//     orders.
+//  2. Execute distinct valid, invalid, same-reference, reversed-order, nested,
+//     and actual-function-member controls against the emitted JavaScript.
+//  3. Require compare.less to retain its existing multi-object-union diagnostic.
+func TestCompareFunctionalUnionMembershipTransform(t *testing.T) {
+	project := compareFunctionalUnionMembershipProject(t)
+	functional := compareFunctionalUnionMembershipTransform(t, project, true)
+	defaultMode := compareFunctionalUnionMembershipTransform(t, project, false)
+	compareFunctionalUnionMembershipRunRuntimeCases(t, project, functional, defaultMode)
+	if strings.Contains(functional, `typeof v.handler === "function"`) == false {
+		t.Fatalf("functional equality emit did not retain the function membership guard:\n%s", functional)
+	}
+	compareFunctionalUnionMembershipRejectsLess(t, project)
+}
+
+func compareFunctionalUnionMembershipProject(t *testing.T) string {
+	t.Helper()
+	base := os.Getenv("TYPIA_TEST_TMPDIR")
+	if base == "" {
+		root := ttscTypiaTestRepoRoot(t)
+		base = filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+	}
+	if err := os.MkdirAll(base, 0o755); err != nil {
+		t.Fatalf("mkdir temp base: %v", err)
+	}
+	dir, err := os.MkdirTemp(base, "compare-functional-union-")
+	if err != nil {
+		t.Fatalf("create temp fixture: %v", err)
+	}
+	t.Cleanup(func() {
+		_ = os.RemoveAll(dir)
+	})
+	src := filepath.Join(dir, "src")
+	if err := os.MkdirAll(src, 0o755); err != nil {
+		t.Fatalf("mkdir fixture src: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "tsconfig.main.json"), []byte(compareFunctionalUnionMembershipTSConfig("src/main.ts")), 0o644); err != nil {
+		t.Fatalf("write main tsconfig: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


samchon:
## Self-review opened

The reservation now has an implementation candidate at `bdd52cf079d18b58a70ebdffd6bfba0685bcc80a`.

Scope is limited to #2252: `compare.equals` must use the active `functional` option while routing object-union members, without changing post-selection function-property comparison behavior. The regression fixture covers direct/factory APIs, both declaration orders, nesting, `cover`, cross-member rejection, partial cover, extra keys, function identity, and declared `equals` delegation. No package, release, publication, or schema-interface change is in scope.

I will re-read the exact diff and transformed 52-case proof, including the canonical one-line `Functional: false` mutation. Independent verification has passed, but the prospective and exact-merged integration gates remain pending; this is not a completion or merge recommendation.

samchon:
## Self-review complete

Reviewed `bdd52cf079d18b58a70ebdffd6bfba0685bcc80a` against the #2252 reservation invariant.

- Scope remains limited to functional object-union membership in `compare.equals` plus its transform regression; no package, release, publication, or schema-interface change.
- `go -C packages/typia/native/cmd/ttsc-typia test -run '^TestCompareFunctionalUnionMembershipTransform$' -count=1 .` passed with all 52 transformed Node runtime cases.
- The fixture proves direct/factory `is`, `equals`, and `cover`, nested equality, both union orders, valid cross-member rejection, post-selection compatibility controls, and the default-mode boundary.
- Canonical mutation evidence: changing only `CompareEqualProgrammer.go`'s `Functional: nativehelpers.OptionPredicator.Functional(...)` to `Functional: false` makes the named second-function invalid equal-label direct/factory equality assertions fail with `expected false, got true`. The mutation worktree was restored cleanly.

Independent verification also passed. The pull request stays draft because #2257 must integrate first; prospective and exact-merged integration gates remain mandatory, so this self-review is not a merge recommendation.

samchon:
Superseded by #2266.

PR #2257, opened 2026-07-19
fix(native): preserve user-global native identity

### LINKED ISSUE #2200: A user type is classified as a built-in native by name alone: is<File>/<Blob>/<Date>… emit instanceof and reject valid values (or throw)

samchon:
Closes #2200.

This draft now reserves only the user-global runtime-native identity residual. A user-authored global `File` or `Blob` must remain structural when it merges with a built-in or Node `buffer` declaration; global-table resolution or a same-name declaration alone cannot promote that user type to a runtime native.

The campaign lead re-applied the five-row admission table and split #2238 into its own claim. Callable and constructable declaration-shape semantics have a distinct root cause, focused matrix, rollback unit, and downstream dependency on #2252; they are not part of this pull request.

This replan does not push the local #2200 implementation candidate. The remote branch remains the implementation-free reservation until the provenance matrix, independent verification, and integration gates are complete.


samchon:
Superseded by #2266.

PR #2258, opened 2026-07-19
fix(protobuf): reject malformed varints across values and framing

### LINKED ISSUE #2256: fix(protobuf): reject malformed varints across values and framing
## Problem

The shared Protocol Buffer varint readers accept malformed ten-byte payloads as ordinary values. On a declared integer field, a ten-byte sequence with no terminator decodes as `0n`; a terminating tenth byte carrying payload outside the 64-bit wire domain decodes as `9223372036854775807n`. Raw, assertion, predicate, and validation decode families return success-shaped data. The same `uint32` reader also owns field tags and every length prefix, so the full defect class includes those framing consumers rather than only declared integer values.

## Evidence

The D11 reviewer and lead independently executed this on `c7f7939d2bb367bf5a0571b1becf889b67c3a472`; current master `ce88a887c9b9fac0ebbbdff088ae38df325200d3` changes only skill documents after that SHA.

For `{ value: bigint & tags.Type<"uint64"> }`, field tag `0x08` followed by ten `0x80` bytes returned `{ value: 0n }`. The tag followed by nine `0xff` bytes and `0x02` returned `{ value: 9223372036854775807n }`. Direct/factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` all accepted the malformed inputs; validation returned `success: true`. The canonical ten-byte maximum, nine `0xff` bytes followed by `0x01`, decoded correctly as `18446744073709551615n`. The valid non-canonical encoding `0x80 0x00` also remained accepted. Direct `_ProtobufReader` calls reproduced the same malformed acceptance through `uint32`, `int32`, `sint32`, `bool`, `uint64`, `int64`, and `sint64`.

The repository's existing official Go dependency supplied the independent oracle. `google.golang.org/protobuf/encoding/protowire.ConsumeVarint` returned `n=-3` for each malformed payload and `n=10` with `math.MaxUint64` for the canonical maximum.

The expanded critic executed the shared framing paths on current master. A malformed tag encoding field 1 / VARINT as `0x88` followed by nine `0x80` bytes, and its excess-tenth-payload twin `0x88` followed by eight `0x80` bytes and `0x02`, were both accepted by direct/factory decode and validate forms; each returned `{ value: 1 }`, and validation reported success. The direct reader returned tag `8` at index 10. Go `protowire.ConsumeTag` rejected both with `n=-3`.

A known string length encoded as tag `0x0a` plus ten `0x80` bytes was accepted as `""`; the equivalent bytes field became an empty byte array, packed `uint32` became an empty array, a map became `{ "": 0 }`, and a nested message became `{ value: 0 }`. An unknown length-delimited field was silently ignored, as was the same malformed length inside a group. Direct `bytes`, `fork`/`close`, and `skipType(LEN)` calls also accepted it. Representative direct/factory decode and validate forms all returned success-shaped data, while Go `protowire.ConsumeField` or `ConsumeBytes` rejected every corresponding payload with `n=-3`.

The executed runtime owner is `packages/typia/src/internal/_ProtobufReader.ts`. Its value-reading paths return after byte ten without rejecting a continuing tenth byte or discarded high payload bits. The adjacent unknown-field `skipVarint` path already rejects ten continuation bytes. This locates the boundary without prescribing a loop, mask, or error implementation.

Closed #2135/#2139 concern unknown-field skipping and are not owners for malformed known-field values. Their explanation claimed value readers reject beyond the limit, but current execution disproves that premise. The #2139 regression `test_protobuf_reader_unknown_field_faults` currently asserts that `uint32` may consume ten continuation bytes and return zero; that assertion freezes the defect and must be corrected. Open searches found no current owner.

## Consequence surface

The executed value path reaches `_ProtobufReader.varint32` and `varint64`, all seven public integer/boolean reader methods, and all eight direct/factory `decode`, `assertDecode`, `isDecode`, and `validateDecode` wrappers. Generated scalar fields and repeated, packed, map, and nested consumers share those methods. `_ProtobufReader.uint32` additionally consumes every generated field tag, `bytes` and `string` lengths, `fork` boundaries, unknown length-delimited fields, and nested group tags; malformed tag and length-prefix cells therefore belong to the same consequence matrix even where a particular byte sequence still reaches a later framing error.

The documented Wire errors contract in `website/src/content/docs/protobuf/decode.mdx` says malformed bytes raise an `Error` before validation and propagate from every wrapper. The correction must preserve #2051 valid high-bit behavior, #2083 truncated length-delimited and bounded-subrecord behavior, #2088 invalid UTF-8 rejection, #2116 zero-length advancement, #2135/#2139 truncated-before-limit and overlong unknown-varint faults, canonical maximum values, valid non-canonical in-range encodings, and encoder round trips.

## Invariant

Every Protocol Buffer varint used as a value, field tag, or length prefix must terminate within its legal wire width and cannot carry payload outside the 64-bit wire domain before any decoded value is returned or validated. Malformed varints are wire errors: direct and factory raw, assert, is, and validate decode forms all throw before validation rather than returning data, `null`, or an unsuccessful `IValidation`. Canonical maximum values and valid shorter or non-canonical-but-in-range encodings remain accepted.

- Preserve exact malformed no-terminator and excess-tenth-payload witnesses beside the canonical ten-byte maximum.
- Cross direct/factory raw, assert, is, and validate decode families and the direct `_ProtobufReader` value methods.
- Enumerate every known-field scalar, enum-like, boolean, signed/unsigned, 32/64-bit, repeated, packed, map, and nested path that reaches the shared readers; record non-applicable cells with evidence.
- Exercise malformed field tags and every `uint32` length-prefix consumer: bytes, strings, packed fields, maps, nested messages, unknown length-delimited fields, and groups. A later error is evidence only when the reader rejects the malformed varint itself rather than accidentally reinterpreting its remainder.
- Cross zero, one-byte, width boundaries, valid non-canonical in-range encodings, truncation at every byte, ten continuation bytes, excess tenth payload, trailing fields, unknown-field skipping, and length-delimited framing controls.
- Use the official Protocol Buffer oracle on the same bytes and keep encoder round trips green.
- Retain the regression while the independent verifier mutates only the product fix and proves the canonical command fails while controls pass.
- Add stable `test_protobuf_reader_varint_bounds` and `test_protobuf_decode_varint_bounds` regressions and a permanent Go test pinning the official oracle's verdicts. Run `pnpm --filter @typia/test-typia-schema start -- --include protobuf_reader_varint_bounds`, `pnpm --filter @typia/test-typia-schema start -- --include protobuf_decode_varint_bounds`, and `go -C packages/typia/test test -run '^TestProtobufVarintCorpusMatchesProtowire$' -count=1 .` as focused gates.

  **Correction, 2026-07-19.** This bullet originally named a `TestProtobufVarintBoundsOracle` Go oracle, and the implementation delivered exactly that. Independent verification then found that the delivered test **passes with the product fix fully reverted**: it exercises `protowire` alone, never calls typia, and its agreement with the TypeScript regressions was hand transcription rather than any mechanical coupling. A permanent test that cannot fail when the work regresses is a false green, so the batch was returned for a test-asset remediation: the byte strings and oracle verdicts now live in one shared `packages/typia/test/protobuf_varint_corpus.json` that both the Go test and both TypeScript regressions read, and the Go test is renamed to state that it pins the oracle only. Anyone running the originally published command against the current tree gets `ok ... [no tests to run]`, exit 0 — a false green in the opposite direction. Use the command above.

  The general lesson is the one #2232 records: an issue should state the invariant and leave the mechanism to whoever can run it. Naming a test artifact by shape is a mechanism, and this one specified an artifact that could not carry the guarantee its name implied.
- Run `pnpm --filter ./tests/test-typia-schema start`, `pnpm build`, `pnpm test`, `go -C packages/typia/test test -tags typia_native_internal ../native/...`, `go -C packages/typia/native vet -p=1 ./...`, and `go -C packages/typia/test vet -p=1 ./...` from fresh worktrees at both the prospective merge and exact merged SHA. Allow at least 20 minutes for the first cold native build.

## Coordination

No package version, tag, release, or publication change belongs in this issue. The issue defines framing invariants and the shared owner only; the implementer validates the mechanism.

samchon:
Closes #2256.

A malformed Protocol Buffer varint could become success-shaped decoded output. All value and framing consumers share the tenth wire byte, so the fix enforces its single legal payload bit in one reader boundary in `_ProtobufReader.ts`, and corrects the #2139 reader regression that had frozen ten-continuation acceptance.

## Verification

Independently verified by an agent that did not write the implementation and built its own oracle rather than reusing the batch's test material.

| Measure | Result |
| --- | --- |
| Differential vs official `protowire` v1.34.2 — 885 byte strings x 8 reader entry points | **7080/7080 cells agree, 0 disagreements** |
| Same differential on the base commit | 3971 disagreeing cells over 530 inputs, every one *accepts-illegal* |
| Regressions / loosened cells introduced | **0 / 0** |
| Oracle-accepted inputs decoding to the exact value at the exact consumed length | 208 / 208 |
| Exhaustive tenth-byte partition, both prefix families | accepts exactly `{0x00, 0x01}`, bit-identical to the oracle; base accepted all 256 |
| Cross-reader consistency among the 8 consumers | 0 disagreements, against 269 on base |
| Public wrapper matrix (direct/factory x raw/assert/is/validate) | 152/152 threw a prefixed `Error` — never `null`, never `success: false` |

Mutation proof: reverting only `_ProtobufReader.ts` fails exactly the three coupled regressions while all fifteen adjacent protobuf controls pass. Gates at the head: `pnpm build`, 18 protobuf cases, 178-case `test-typia-schema`, feature identity, `go vet` in both trees, and tagged native `../native/...` (28 packages, 0 failures).

## The changed #2139 expectation is a correction, not a laundered regression

A previously merged test asserted that typia must *accept* `0x80 x10, 0x01` as `0`. `protowire` rejects that input with `n = -3`. The old expectation was wrong; the new one matches both the wire specification and the oracle. Adjudicated independently against the specification rather than against this pull request's narrative.

## Second commit: the oracle test could not fail

Independent verification found the first commit's permanent `TestProtobufVarintBoundsOracle` **passed with the product fix fully reverted** — it exercised `protowire` alone, never called typia, and its agreement with the TypeScript regressions was hand transcription. A permanent test that cannot fail when the work regresses is a false green, so the batch was returned for a test-asset remediation.

Every byte string and verdict now lives in one `packages/typia/test/protobuf_varint_corpus.json` that the Go test and both TypeScript regressions read. The Go test is renamed to state that it pins the official parser's verdicts only. Verified by mutating the corpus seven ways — value, consumed length, fault relabelling, emptied entries, unknown class, a fabricated contradictory row, and an accepted/rejected flip — with **both halves reacting to every mutation**.

The same commit splits one error message that had described two different faults: a continuing tenth byte keeps `varint exceeds 10 bytes.`, while a terminating tenth byte carrying payload above bit 0 now reports `varint exceeds 64 bits.` **Acceptance is byte-identical** — proven by re-running the full 7080-cell differential across both heads: 0 acceptance differences, 0 value differences, 2088 message-only differences. No website page, documentation, example, or tracked test outside this batch pinned the old string.

## Out of scope, recorded for the next discovery round

Two verified, user-visible defects present identically on the base commit and on this fix, outside this issue's published invariant: length prefixes and field tags are silently truncated to 32 bits, so a field declaring 4,294,967,301 bytes reads five, and field 536870913 aliases onto field 1. A third consumer of the same truncation, `bool()`, decodes from the low 32 bits only. None is locked in by the new tests, so the next round can decide the contract rather than inherit a snapshot.

## Campaign constraints

No package version, tag, release, publication, or `packages/interface/src` change. `pnpm format` was not run.


samchon:
## Self-Review — round 1

Solo Self-Review under `.agents/skills/review/SKILL.md`. Snapshot `2d3b4181c28038a4b8429c86a6818c182d8dfbba`, complete surface `3f59f641a9..2d3b4181c2` (eight files). Findings and remediation plan recorded here before any of them is applied.

### Verified, no change needed

- **The product fix matches its oracle byte for byte.** Read the pinned `google.golang.org/protobuf@v1.34.2` `encoding/protowire/wire.go` source directly: `ConsumeVarint` requires the tenth byte to be `< 2` and returns `errCodeTruncated = -1` / `errCodeOverflow = -3`. `lastVarintByte` accepts exactly `0x00` and `0x01` and rejects everything else, splitting the verdict into a length fault and a domain fault. Same accept/reject partition, finer error naming.
- **Both fault classes reach every consumer.** `varint32` reads five bytes, then four, then `lastVarintByte`; `varint64` reads nine, then `lastVarintByte`; `skipVarint` loops bytes one through nine, then `lastVarintByte`. Ten bytes maximum on all three paths, and the skip path now rejects the tenth-byte-payload class it previously accepted.
- **The corpus is genuinely consumed, and a corrupted row fails a real assertion.** Recomputed every accepted row's value independently from its own bytes (seven-bit little-endian groups, `asUintN(64)`); all 15 match the recorded value, every accepted row carries continuation bits on all but its last byte, and no accepted row exceeds ten bytes or carries more than bit 63 in its tenth. The Go side re-derives the same verdicts from `protowire` plus an internal-consistency gate; the TypeScript side reads the same rows through `typia.assert<IDocument>` and fails loudly on a lost field or an unknown fault class. Neither side transcribes a byte string.
- **Repository format pipeline.** `scripts/format-go.cjs` runs `gofmt -w` and then converts leading tabs to two spaces; the new Go test is byte-identical to that pipeline's output (the only diff is CRLF vs LF from the Windows checkout), so it matches every other Go file in the repository. Prettier reports the same style drift on untouched neighbours as on the changed files, so it is a repository-wide state that Post-Campaign Cleanup owns, not a defect of this branch.
- **`packages/typia/test` is not in the published `files` list**, so the corpus does not ship to npm consumers.
- **Correcting `test_protobuf_reader_unknown_field_faults`.** The `#2139` assertion that `uint32` may consume ten continuation bytes and return zero froze the defect; #2256 explicitly requires correcting it. It is replaced by a rejection plus a positive control that a legal ten-byte `0xff`×9 `0x01` still reads as `0xffffffff` at index 10. Flagged as a deliberate specification change under the development skill's Change Integrity rule.

### Findings to act on

1. **An inaccurate claim I introduced one commit ago.** `test_protobuf_reader_varint_bounds` now says the corpus carries "the canonical maximum that terminates at each width from one byte through ten". That is false for width two: the only two-byte rows are `ac02` (300) and `8000` (non-canonical zero), neither of which is the two-byte maximum. **Plan:** add the `ff7f` / `16383` row rather than soften the sentence, so the family is uniform and the claim is exactly true.
2. **`skipVarint`'s doc comment no longer describes what `skipVarint` does.** It still explains only the ten-byte length limit, but the method now also rejects a tenth byte carrying payload above bit 63. **Plan:** extend the comment to name both faults, matching `lastVarintByte`.

### Deliberately not changed

- **`website/src/content/docs/protobuf/decode.mdx`.** Its "Wire errors" section already states that a malformed payload raises an `Error` before validation and propagates out of `isDecode` and `validateDecode`. This change makes the runtime match that documented contract rather than altering it, and #2256 names the page as the contract to preserve. Adding a varint sentence would be an enhancement that also drags in the website skill's full fresh-build validation lane, which is outside this unit's gate set.
- **Renaming the local `OVERFLOW` constant** in the protobuf tests, even though the corpus now uses "overflow" for the 64-bit-domain fault while that constant means `buffer overflow.`. Seven files across the suite already use that name for the reader's own message text; a one-file rename would break an established convention rather than fix an inconsistency.
- **Recording the runtime divergence in `_ProtobufReader.ts` itself.** The evidence belongs where a test pins it; a claim about another implementation's behaviour rots in shipped source.

### Evidence run so far

Independent pre-flight against the real reader source (byte-identical copy with the enum and `Singleton` stubbed, run under `node --experimental-strip-types`), crossing all 32 corpus rows with `uint32`, `int32`, `sint32`, `uint64`, `int64`, `sint64`, `bool`, and `skipVarint`:

| tree | assertions passed | assertions failed |
| --- | --- | --- |
| head `2d3b4181c2` | 256 | 0 |
| `_ProtobufReader.ts` reverted to `3f59f641a9` | 211 | 45 |

The 45 base failures are exactly the three over-long rows through the seven value readers (21) and the three overflow rows through the seven value readers plus the skip path (24); every accepted row still passes on the base, so the controls are not what moves. The base returned `9223372036854775807` from `uint64` for `ff`×9 `02` — the issue's own witness, reproduced from scratch on current master. This is a prediction that de-risks the long suite run, not a substitute for it; `pnpm --filter ./tests/test-typia-schema start`, `pnpm test:go:public`, `pnpm test:go:native`, `pnpm --filter ./packages/typia build`, and the disposable-worktree mutation proof are still required and are running or queued.

samchon:
Superseded by #2266.

PR #2264, opened 2026-07-20
fix(native): preserve callable declaration identity

### LINKED ISSUE #2238: fix(native): preserve function semantics for call-only and construct-only interfaces

samchon:
Closes #2238.

This draft reserves the remaining function-shaped declaration identity defect. Call-only and construct-only interfaces and type literals must keep the same function metadata behavior as mutually assignable aliases, while required members and index signatures remain structural boundaries.

The campaign lead split this issue from #2200 under the five-row admission table. Runtime-native File and Blob provenance has a different root cause, focused oracle, rollback unit, and acceptance matrix; it remains in re-scoped draft #2257. This claim owns only the callable declaration-shape invariant and its downstream #2252 integration dependency.

This is an implementation-free reservation at `9cc3ade21ae516e624929650adb7424ce470c1e9` from `master` `3f59f641a92e6a730072a65fda2ed7f2ec90c9ad`. Product implementation, local verification, solo self-review, independent verification, and prospective and exact-merged integration gates are pending. The exact-SHA campaign Actions record is maintained separately and is not acceptance evidence.


samchon:
Superseded by #2266.

Issue #2276, opened 2026-07-20
Intersection spellings of callable shapes are half accepted and half rejected


samchon:
## Problem

A member-carrying callable type has two equivalent spellings: a declaration body, and an intersection of a call signature with the rest. typia treats the intersection spelling inconsistently — it accepts six of these shapes and rejects six structurally analogous ones as a nonsensible intersection.

Where it rejects, the intersection also disagrees with the declaration spelling of the same type, which is accepted. So the same type is supported or refused depending on how it was written, and which way it goes is not predictable from any stated rule.

## Evidence

Measured on `master` at `ab978b0a27`, transforming one single-shape project per spelling and recording the exit code and diagnostic counts. Each row is the same type written two ways.

| Shape | Interface spelling | Intersection spelling |
| --- | --- | --- |
| `Method` | `code=0` | `code=0` |
| `OptionalMethod` | `code=0` | `code=0` |
| `OptionalMethodProperty` | `code=0` | `code=0` |
| `OptionalMember` | `code=0` | `code=0` |
| `SymbolMember` | `code=0` | `code=0` |
| `OptionalSymbolMember` | `code=0` | `code=0` |
| `Overload` | `code=0` | `code=3`, 2 nonsensible |
| `ConstructOverload` | `code=0` | `code=3`, 2 nonsensible |
| `CallAndConstruct` | `code=0` | `code=3`, 2 nonsensible |
| `MethodProperty` | `code=0` | `code=3`, 2 nonsensible |
| `RequiredMember` | `code=0` | `code=3`, 2 nonsensible |
| `IndexSignature` | `code=0` | `code=3`, 2 nonsensible |

Two diagnostics per rejected row because the fixture exercises both the direct and the factory form.

The split is not random, which is what makes it worth stating precisely rather than as "sometimes":

- `Method` is accepted while `MethodProperty` is rejected — `cancel(): void` and `cancel: () => void` are the same type.
- `OptionalMember` is accepted while `RequiredMember` is rejected — the only difference is `?`.
- Every shape whose intersection combines two call signatures (`Overload`, `ConstructOverload`, `CallAndConstruct`) is rejected.

Owner: `packages/typia/native/core/factories/internal/metadata/iterate_metadata_intersection.go`. The accept path is `if onlyObjects && len(tagObjects) == 0 { return escape(false) }`; the reject path is reached when a surviving member is not a single-bucket object after phantom brands are dropped.

## Consequence surface

[...]

PR #2280, opened 2026-07-21
fix(generate): derive filesystem identity from exact inode values

### LINKED ISSUE #2269: typia generate derives filesystem identity from a lossy fs.Stats.ino, so distinct paths can collide
## Problem

`TypiaGenerateWizard` builds every filesystem identity key from `fs.Stats.ino` used as a JavaScript number. On NTFS the 64-bit file ID routinely exceeds `Number.MAX_SAFE_INTEGER`, where a double's spacing is 8 or more, so two distinct directories or files can round to the same key.

The keys feed two different guards, and a collision has a different consequence in each:

- **Traversal dedup** (`visitedDirectories`, `visitedFiles`): a *distinct* directory or file looks already-visited and is **silently skipped**. Its sources are never transformed and no diagnostic is emitted — silent source loss.
- **Overwrite protection** (`ensureTargetFiles`): a legitimate output looks like one of the inputs and generation is **refused** with `output file would overwrite input file through a physical file alias`. Loud and safe, but a false rejection.

The dedup direction is the severe one. The overwrite guard cannot lose data through this cause, because one file has one `ino` and therefore one rounded key; rounding cannot make the same file yield two different keys.

## Evidence

`packages/typia/src/executable/TypiaGenerateWizard.ts`:

```ts
function fileIdentityKey(stat: fs.Stats, realpath: string): string {
  return stat.ino === 0
    ? `path:${path.normalize(realpath)}`
    : `inode:${stat.dev}:${stat.ino}`;
}
```

Callers: line 455 and line 482 (`ensureTargetFiles` input set and output check), line 878 (visited directories), line 946 (visited files).

Measurement, Windows 11 / NTFS, 4000 directories created in one temporary parent. For each, `statSync(d).ino` (number) was compared with `statSync(d, { bigint: true }).ino`:

- **1879 of 4000** directory IDs exceeded 2^53 and are therefore not exactly representable.
- Observed ID `41658296553217950` sits between 2^55 and 2^56, where the double spacing is **8**.
- **0** collisions were observed at this sample size, so this is a proved representability defect and an unproved-in-the-wild collision, not an observed data loss.

Why collisions are plausible rather than merely arithmetically possible: an NTFS file ID is `(sequenceNumber << 48) | mftRecordIndex`. Two entries sharing a sequence number with MFT record indices within 8 of each other round to one double. Adjacent MFT records are exactly what a directory tree created in one pass produces.

`stat.ino === 0` already has a path fallback, so the code acknowledges that inode identity is not always available; it does not yet acknowledge that it is not always exact.

## Consequence surface

- `typia generate` in directory mode on NTFS, and on any filesystem whose IDs exceed 2^53 or are otherwise not exactly representable.
- Silent skip of a source file or an entire subtree, producing an output tree that is incomplete with a zero exit code.
- Spurious `would overwrite input file` refusal, blocking a legitimate generation.
- Not addressed by the directory identity work considered in #2082: an additional key combined with `.some(...)` can only make the guard skip more, so it cannot repair a false positive.

## Invariant

Filesystem identity must be exact. Two distinct filesystem objects must never share an identity key, and one object must always yield the same key within a run.

Mechanism is the implementer's to determine and to validate. `fs.promises.stat(p, { bigint: true })` removes the representability loss, but it is not by itself a complete answer to filesystem identity here: it does nothing for filesystems that report `ino === 0`, and the existing `path:` fallback is itself inexact on win32, where `fs.realpathSync` does not case-canonicalize, so `...\dir` and `...\DIR` produce two different fallback keys for one directory while `FileSystemIdentity` in the same module already owns a `caseSensitive` policy for that question.

- Two distinct directories whose 64-bit IDs differ by less than the double spacing at their magnitude receive distinct keys, and neither is skipped.
- One directory reached through several spellings — relative, absolute, trailing separator, differing case on a case-insensitive volume — receives one key, and the intended alias behaviour is unchanged.
- A filesystem reporting `ino === 0` still terminates and still refuses to overwrite an input.
- Existing cycle, external-link, output-alias, and repeated-alias regressions in `tests/test-typia-generate` stay green.
- The regression must fail when the identity derivation is reverted while the test is retained.

## Coordination

Found while implementing #2082 and deliberately kept out of that change, which fixed a locale-dependent traversal order and did not touch identity derivation.

samchon:
Closes #2269.

## Problem

`fileIdentityKey` built `inode:${dev}:${ino}` from `fs.Stats`, whose `ino` is a JavaScript number. An NTFS file ID is `(sequenceNumber << 48) | mftRecordIndex` and routinely exceeds `Number.MAX_SAFE_INTEGER` — **1879 of 4000** probed directories did — where the spacing between representable doubles is 8. Two entries sharing a sequence number with MFT record indices inside that spacing round to one double, which is exactly what a directory tree created in one pass produces.

Both directions are wrong, and they differ by caller:

| caller | consequence of a collision |
| --- | --- |
| `visitedDirectories` / `visitedFiles` | a **distinct** source looks already-visited and is skipped with no diagnostic — it silently vanishes from the output |
| `ensureTargetFiles` overwrite guard | a legitimate output looks like one of the inputs and generation is refused |

## Change

Identity is read with `{ bigint: true }` at all five call sites, and the derivation moves to `FileSystemIdentity`, which already owns filesystem identity — case folding, path keys, containment. The rule and its reasoning now live with the rest of that contract instead of inside the wizard's traversal, and it becomes directly testable.

`BigInt(0)` / `BigInt(1)` rather than literals: the package compiles at ES2016, where a BigInt literal is a syntax error. The `ino === 0` path fallback is unchanged, since some filesystems report no inode at all.

## Verification

`testInodePrecisionIdentity` asserts its own premise before asserting the fix — two IDs differing by 2 near 2^55 are distinct as bigints and **equal** as doubles — then requires their keys to differ, requires one object reached through two paths to still answer one key, and keeps the `ino === 0` fallback distinct per path.

Mutation: reverting only the derivation to the number form fails it with both sides printing the identical text.

```
AssertionError: 'inode:2723597423:41658296553217950' != 'inode:2723597423:41658296553217950'
```

That is the collision itself, not a proxy for it.

- `pnpm run check:files:identity` passes; the same command fails under the mutation.
- `packages/typia` typechecks.

## Why the test lives where it does

A black-box regression cannot force this: two NTFS IDs that collide as doubles cannot be created on demand, and the 4000-directory probe found none. Testing the derivation directly with synthetic stats is what makes the fail-before witness deterministic rather than probabilistic — and `FileSystemIdentity` was already the right owner, with `tests/test-typia-generate/scripts/identity/` already importing it.

## Not included

#2276 stays open. Its cause is still unresolved: the only producer of `nonsensible intersection` in the tree is provably never invoked for the fixtures that receive it, reproduced twice on live builds with `EXIT=0`. Shipping a fix without knowing where the behaviour comes from would violate the standard this campaign has applied throughout. The full elimination record is on that issue.

### DIFF
--- a/packages/typia/src/executable/FileSystemIdentity.ts
+++ b/packages/typia/src/executable/FileSystemIdentity.ts
@@ -147,30 +147,6 @@ export namespace FileSystemIdentity {
     );
   }
 
-  function alternateCase(name: string): string | undefined {
-    let changed: boolean = false;
-    let output: string = "";
-    for (const character of name) {
-      if (character >= "a" && character <= "z") {
-        output += character.toUpperCase();
-        changed = true;
-      } else if (character >= "A" && character <= "Z") {
-        output += character.toLowerCase();
-        changed = true;
-      } else output += character;
-    }
-    return changed ? output : undefined;
-  }
-
-  function isAlreadyExistsError(error: unknown): boolean {
-    return (
-      typeof error === "object" &&
-      error !== null &&
-      "code" in error &&
-      error.code === "EEXIST"
-    );
-  }
-
   /**
    * Builds the key that decides whether two paths are the same filesystem
    * object.
@@ -202,6 +178,30 @@ export namespace FileSystemIdentity {
     return stat.ino === BigInt(0)
       ? `path:${path.normalize(realpath)}`
       : `inode:${stat.dev}:${stat.ino}`;
+  }
+
+  function alternateCase(name: string): string | undefined {
+    let changed: boolean = false;
+    let output: string = "";
+    for (const character of name) {
+      if (character >= "a" && character <= "z") {
+        output += character.toUpperCase();
+        changed = true;
+      } else if (character >= "A" && character <= "Z") {
+        output += character.toLowerCase();
+        changed = true;
+      } else output += character;
+    }
+    return changed ? output : undefined;
+  }
+
+  function isAlreadyExistsError(error: unknown): boolean {
+    return (
+      typeof error === "object" &&
+      error !== null &&
+      "code" in error &&
+      error.code === "EEXIST"
+    );
   }
 
   function isMissingFileError(error: unknown): boolean {
--- a/packages/typia/src/executable/FileSystemIdentity.ts
+++ b/packages/typia/src/executable/FileSystemIdentity.ts
@@ -171,6 +171,39 @@ export namespace FileSystemIdentity {
     );
   }
 
+  /**
+   * Builds the key that decides whether two paths are the same filesystem
+   * object.
+   *
+   * The identity is read from `fs.BigIntStats` rather than `fs.Stats`, because
+   * `fs.Stats.ino` is a JavaScript number. An NTFS file ID is
+   * `(sequenceNumber << 48) | mftRecordIndex` and routinely exceeds
+   * `Number.MAX_SAFE_INTEGER` — in a 4000-directory probe, 1879 of them did —
+   * where the spacing between representable doubles is 8 or more. Two entries
+   * sharing a sequence number with MFT record indices within that spacing round
+   * to one double. The traversal callers would then treat a distinct directory
+   * or file as already visited and skip it with no diagnostic, and the overwrite
+   * guard would refuse a legitimate output (samchon/typia#2269).
+   *
+   * A filesystem that reports no inode at all keeps the canonical path fallback,
+   * which is the only identity available there.
+   *
+   * @param stat Stats read with `{ bigint: true }`.
+   * @param realpath Canonical path of the same object.
+   * @returns Key that is equal for two paths only when they are one object.
+   */
+  export function identityKey(
+    stat: Pick<fs.BigIntStats, "dev" | "ino">,
+    realpath: string,
+  ): string {
+    // `BigInt(0)` rather than `0n`: this package compiles at ES2016, where a
+    // BigInt literal is a syntax error, while the constructor and the `bigint`
+    // type are fine.
+    return stat.ino === BigInt(0)
+      ? `path:${path.normalize(realpath)}`
+      : `inode:${stat.dev}:${stat.ino}`;
+  }
+
   function isMissingFileError(error: unknown): boolean {
     return (
       typeof error === "object" &&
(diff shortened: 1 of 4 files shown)


samchon:
## Self-review — three rounds, complete

Recorded after the fact rather than before each round, which is a deviation from the campaign's own rule that a round comments its findings and plan on the thread *before* acting. Noting it rather than presenting the sequence as if it had been followed.

### Round 1 — two findings, both applied

- **`identityKey` was exported from the private-helper section.** `FileSystemIdentity` groups its exports first (`create`, `Policy`, `inspectDirectory`, `probeDirectory`) and its private helpers after; the new function landed between `isAlreadyExistsError` and `isMissingFileError`. Moved up with the other exports.
- **The `{ bigint: true }` runtime floor was never checked.** `packages/typia/package.json` declares no `engines`, and bigint stats need Node 10.5+. Not a blocker for a package whose CI targets Node 24, but it was an unverified assumption when I wrote it, so it is stated here rather than left implicit.

### Round 2 — one finding, and it was mine

The regression contained **an assertion that could not fail**:

```ts
assert.equal(
  identityKey(stats(BigInt(0)), "/real/first.ts"),
  identityKey(stats(BigInt(0)), "/real/first.ts"),   // same call, both sides
);
```

That is precisely the oracle-that-cannot-fail this campaign has been closing everywhere else — #2276's count guard, the fixtures measuring DOM under a Node name, the `Size() == 0` test that could never fire — and I wrote one into the change that fixes the last of them.

Both fallback rows are now falsifiable: one varies the path with the device fixed and must differ, the other varies the device with the path fixed and must agree. That pair is what "the path decides and the device does not" actually asserts.

Also in this round: loose `equal`/`notEqual` switched to the strict forms, and an unnecessary `as Pick<...>` cast dropped — `{ dev, ino }` already satisfies the parameter type, and a cast that is not needed only hides a future mismatch.

**Mutation was re-proved after the rewrite** rather than inherited from the previous shape, since changing assertions invalidates the earlier proof.

### Round 3 — clean

Re-read the complete diff from scratch. Checked that no other `fs.Stats` in the wizard feeds identity — `ITraversalEntry.stat` is used only for `isSymbolicLink()` in the traversal sort, and `isDirectory`/`isFile` are unrelated — that `nlink` is the only other member read from the widened stats, and that adding an export to `FileSystemIdentity` changes no published surface, since `src/executable` is CLI-internal and not re-exported from the package entry.

Nothing to improve. Review stops here.

### Judgment call left standing

`fileIdentityKey` is now a one-line delegation to `FileSystemIdentity.identityKey`. Inlining it at the five call sites would remove a layer and name the owner at each use; keeping it leaves the call sites shorter and the diff smaller. Kept, deliberately — recording it so the choice is visible rather than accidental.

PR #2283, opened 2026-07-21
docs(agents): split the solo and multi-agent issue campaigns

samchon:
## Intent

Split the issue campaign into a **solo default** and an **explicitly requested multi-agent variant**, and give the solo path the concrete pull-request mechanics it did not have.

The previous single procedure assumed parallel agents, worktree batches, and suspended CI even when one agent ran the whole campaign. A solo run had no defined pull-request unit, no closing convention, and no statement of what gates the merge.

## Solo campaign (`.agents/skills/issue-campaign/`)

- One unified cycle pull request holds every accepted, implementation-ready issue. The DAG orders edits inside it instead of splitting it.
- Claim the cycle with `git commit --allow-empty` and a draft pull request that references every issue. Closing keywords stay out of the claim body: it is written before code exists, so a claim-time closing list would close whatever the cycle later drops, defers, or disproves.
- Each commit ends with one `Close #n: <issue title>` line per issue it resolves, so the merge closes exactly what landed, and a pull-request comment records what each commit landed.
- After a commit lands, the author may run one read-only **commit early-warning pass** over that single commit while continuing to implement. It reports candidates and never edits, decides, or counts as a review.
- Ordinary CI and a complete solo Self-Review are independent acceptance gates, read once per settled head. Every red lane is repaired in the same pull request.
- Formatting belongs to the cycle pull request, so the solo path has no post-campaign format pull request.

## Multi-agent variant (`.agents/skills/multi-agent/`)

New skill with `review.md` and `issue-campaign.md`, loaded only when the user explicitly asks for a team, parallel, or multi-agent workflow. It carries the existing parallel machinery unchanged: team Review Cycle, Research Review Round, parallel discovery rounds, the batch admission table, exact-SHA campaign run cancellation, independent verification, worktree removal, and Post-Campaign Cleanup.

## Preserved

The coverage matrix, tracked-file census, `INCOMPLETE` reporting rule, campaign version and release freeze, integration-sensitive gate, regression-sensitivity mutation proof, and knowledge-base schema all carry into both paths. `AGENTS.md` drops the agent-pool attitude rule, which contradicted the solo default; the ten-agent limit now lives in the multi-agent skill that uses it. Writing-style rules move to the documentation skill.

## Verification

Documentation only; no source, test, or workflow change. Cross-references between the skills were re-checked after the split.

### DIFF
--- a/.agents/skills/benchmark/SKILL.md
+++ b/.agents/skills/benchmark/SKILL.md
@@ -36,7 +36,7 @@ If no pull request exists or no update is authorized, keep the result in the fin
 
 ## Campaign Batching
 
-When benchmark findings produce multiple implementation issues, load the issue-campaign skill and use [Plan And Claim A Pull Request Wave](../issue-campaign/development.md#plan-and-claim-a-pull-request-wave) unchanged for the dependency DAG, batch admission test, merge pressure, grouping and split ledger, claim freeze, and official GitHub `createdAt`-to-`mergedAt` duration. This benchmark skill continues to own workload, fixture, measurement, result, and publication integrity; do not redefine pull-request batching here.
+When benchmark findings produce multiple implementation issues, load the issue-campaign skill and use its planning and claim procedure unchanged for the dependency DAG, claim freeze, and official GitHub `createdAt`-to-`mergedAt` duration: [Plan One Cycle Pull Request](../issue-campaign/development.md#plan-one-cycle-pull-request) for a solo campaign, or [Plan And Claim A Pull Request Wave](../multi-agent/issue-campaign.md#plan-and-claim-a-pull-request-wave) with its batch admission test, merge pressure, and grouping and split ledger when the user explicitly requested a parallel campaign. This benchmark skill continues to own workload, fixture, measurement, result, and publication integrity; do not redefine pull-request batching here.
 
 ## Campaign Cleanup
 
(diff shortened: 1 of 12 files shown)


samchon:
## Self-Review

One complete solo round over the full base-to-head diff, then a fresh re-scan after each correction.

**Round 1 - two defects found and fixed in `4d4b06c`:**

1. The multi-agent no-format rule contradicted the discovery-only-parallel mode, which hands implementation to the solo procedure and therefore must format its own cycle pull request. The rule now names the parallel batch phase and states the solo exception.
2. The four parallel-phase rules were introduced as governing "the entire parallel implementation phase" while one of the modes below has no parallel implementation phase at all. The heading now defines that phase as the wave of concurrent batch pull requests and points the solo mode at the solo rules and ordinary CI.

**Round 2 - clean.** Verified:

- every relative link and anchor across the split resolves to an existing file and heading;
- no solo document assumes a team, critic, verifier, or worktree, and no parallel document lost a rule it previously owned;
- rule counts match their lists (solo five, parallel four, review law six);
- the campaign version freeze matches the development skill's `Issue implementation pull requests never change package versions`;
- the CI rationale is checked against the workflows on this base: only `nestia.yml` sets `cancel-in-progress`, so intermediate lanes run to completion rather than being superseded, which is why CI is read once per settled head.

Documentation only, so no build, test, or generated-artifact lane is affected. `pnpm format` covers `**/*.ts` and changed Go files and does not touch `*.md`, so no formatter result belongs to this diff.

Issue #2286, opened 2026-07-21
Documented "Compiler Errors" tabs no longer match what typia emits, and one documented restriction no longer exists


samchon:
## Problem

Five documentation pages show a "Compiler Errors" / "Console Output" tab that
claims to be what typia prints for the source in the neighbouring tab. None of
them matches what typia 13 actually emits. One of them documents a restriction
the transform no longer enforces, and one shows a source that never reaches the
transform at all.

These tabs are the primary way a user maps a rejected call site to the rule that
rejected it. A user who searches the message they actually see finds nothing, and
a user who trusts the documented restriction writes code against a rule that no
longer exists (or misses one that does).

## Evidence

Every "actual" cell below was produced by compiling the documented source through
the real transform (`tests/debug`, `pnpm --filter ./tests/debug exec ttsx`).

### 1. The message prefix is missing everywhere

`packages/typia/native/transform/FileTransformer.go` prefixes every typia
diagnostic with `typia transform error`:

```go
text := "typia transform error"
if message != "" { text += ": " + message }
```

| | text |
| --- | --- |
| documented (all pages) | `error TS(typia.X): unsupported type detected` |
| actual | `error TS(typia.X): typia transform error: unsupported type detected` |

Affects `website/src/content/docs/llm/application.mdx`,
`llm/parameters.mdx`, `llm/schema.mdx`, `llm/structuredOutput.mdx`,
`json/schema.mdx`, `protobuf/message.mdx`, and the archived blog post
`website/src/content/blog/typia-protocol-buffer-library-of-typescript.mdx`.

### 2. `llm/application.mdx` — every message differs

| documented | actual |
| --- | --- |
| `LLM application's function ("create")'s return type must not be union type with undefined.` | `LLM application's function ("create") return type cannot be optional (union with undefined).` |
| `LLM application's function ("add")'s return type must be an object type.` | `LLM application's function ("add") return type must be a single object type.` |
| `LLM application's function ("erase")'s parameter must be an object type.` | `LLM application's function ("erase") parameter must be a single object type.` |

(The possessive `'s` was dropped and the wording changed; source of truth is
`packages/typia/native/core/programmers/llm/LlmApplicationProgrammer.go`.)

### 3. `llm/parameters.mdx` — the first example no longer reaches the transform

```ts
typia.llm.parameters<string>();
```

v13 declares `parameters<Parameters extends Record<string, any>, ...>`, so this
is a plain TypeScript error and no typia diagnostic is produced:

```
[...]

Issue #2285, opened 2026-07-21
Every protobuf transform diagnostic except protobuf.message reports a doubled code: TS(typia.protobuf.typia.protobuf.encode)


samchon:
## Problem

Every protobuf transform diagnostic except `typia.protobuf.message` reports a
doubled API code:

```
error TS(typia.protobuf.typia.protobuf.createDecode): typia transform error: unsupported type detected
        ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The code is the only machine-readable identity a typia rejection carries. Users
search it, editors surface it, and `tests/test-error` parses it. Sixteen of the
seventeen protobuf entry points emit a code that names no API.

## Evidence

Reproduction — one file calling every protobuf entry point with an unsupported
type (`tests/debug`, `pnpm --filter ./tests/debug exec ttsx src/probe.ts`):

```ts
import typia from "typia";

typia.protobuf.message<bigint>();
typia.protobuf.encode<bigint>(1n as any);
typia.protobuf.createDecode<bigint>();
// ... every entry point
```

Actual output (`error ` prefix stripped):

```
TS(typia.protobuf.message)                                <- correct
TS(typia.protobuf.typia.protobuf.encode)
TS(typia.protobuf.typia.protobuf.decode)
TS(typia.protobuf.typia.protobuf.assertEncode)
TS(typia.protobuf.typia.protobuf.assertDecode)
TS(typia.protobuf.typia.protobuf.isEncode)
TS(typia.protobuf.typia.protobuf.isDecode)
TS(typia.protobuf.typia.protobuf.validateEncode)
TS(typia.protobuf.typia.protobuf.validateDecode)
TS(typia.protobuf.typia.protobuf.createEncode)
TS(typia.protobuf.typia.protobuf.createDecode)
TS(typia.protobuf.typia.protobuf.createAssertEncode)
TS(typia.protobuf.typia.protobuf.createAssertDecode)
TS(typia.protobuf.typia.protobuf.createIsEncode)
TS(typia.protobuf.typia.protobuf.createIsDecode)
TS(typia.protobuf.typia.protobuf.createValidateEncode)
TS(typia.protobuf.typia.protobuf.createValidateDecode)
```

### Root cause

`packages/typia/native/core/factories/ProtobufFactory.go` composes the code by
prefixing its `Method` prop:

```go
Code: "typia.protobuf." + props.Method,
```

[...]

Issue #2284, opened 2026-07-21
random<string & Format<F> & MinLength/MaxLength> throws for nine formats although the requested length is one their own validator accepts


samchon:
## Problem

`typia.random<T>()` throws for nine of the twenty-two string formats whenever the
leaf also carries a length tag, **even when the requested length is one the
format's own validator accepts**.

```
Error: unable to generate a random string satisfying both the format and the
length constraints.
```

`Format<"byte">` and `Format<"regex">` fail for *any* window that excludes a
single hard-coded length (44 and 86), because their generators return one
constant string. The other seven fail for windows outside the one shape their
generator happens to emit.

Affected users: anyone generating mock data, fixtures, or LLM sample payloads
from a type whose string leaf carries both a `Format` and a `MinLength`/
`MaxLength` tag — a natural combination (a base64 blob with a size cap, a short
`ipv4`, a compact `date-time`). The call throws at runtime; there is no
workaround short of dropping the length tag.

## Evidence

Reproduction (`tests/debug`, `pnpm --filter ./tests/debug exec ttsx src/probe.ts`):

```ts
import typia, { tags } from "typia";

type Exact<S extends tags.Format.Value, N extends number> = string &
  tags.Format<S> &
  tags.MinLength<N> &
  tags.MaxLength<N>;

// typia.is proves the length is satisfiable; typia.random then throws.
console.log(typia.is<Exact<"byte", 24>>("A".repeat(24))); // true
console.log(typia.random<Exact<"byte", 24>>());           // throws
```

Each row below states a length the format's **own** validator accepts (proved by
`typia.is` over a hand-written witness of exactly that length) and what
`typia.random` does for `MinLength<N> & MaxLength<N>`:

| Format | N | witness accepted by `typia.is` | `typia.random` |
| --- | --- | --- | --- |
| `byte` | 24 | `"A".repeat(24)` | **throws** |
| `regex` | 24 | `"a".repeat(24)` | **throws** |
| `ipv4` | 7 | `0.0.0.0` | **throws** |
| `ipv6` | 15 | `0:0:0:0:0:0:0:0` | **throws** |
| `date-time` | 20 | `2020-01-01T00:00:00Z` | **throws** |
| `time` | 9 | `00:00:00Z` | **throws** |
| `relative-json-pointer` | 5 | `0/foo` | **throws** |
| `duration` | 4 | `PT1S` | **throws** |
| `json-pointer` | 6 | `/a/b/c` | **throws** |
[...]

Issue #2288, opened 2026-07-21
project skill Layout omits the tests/test-typia-bundler-cache workspace


samchon:
## Problem

`.agents/skills/project/SKILL.md` enumerates the repository's test workspaces
under **Layout**, and the list is missing `tests/test-typia-bundler-cache`. An
agent orienting from the skill concludes the workspace does not exist, or that
the enumerated list is the complete set of suites `pnpm test:packages` runs.

## Evidence

```
$ git ls-files tests/test-typia-bundler-cache
tests/test-typia-bundler-cache/package.json
tests/test-typia-bundler-cache/src/index.ts
tests/test-typia-bundler-cache/src/test_webpack_filesystem_cache_invalidation.ts
tests/test-typia-bundler-cache/tsconfig.json
```

It is a real, tracked workspace: `@typia/test-typia-bundler-cache`, with a
`start` script (`ttsx src/index.ts`), so root `pnpm test:packages`
(`pnpm --filter=./tests/test-* --fail-if-no-match -r start`) runs it, and
`.gitignore` reserves its `.tmp/` scratch space. It drives a webpack persistent
filesystem-cache invalidation end-to-end check through `@ttsc/unplugin` — a path
no other suite covers.

The skill's list covers every other `tests/*` entry
(`test-typia-schema`, `test-langchain`, `test-mcp`, `test-vercel`, `test-utils`,
`test-typia-automated`, `test-utils-automated`, `test-interface`,
`test-typia-cli`, `test-typia-generate`, `test-typia-exact-optional`,
`test-feature-identity`, `test-error`, plus `tests/template` and `tests/debug`),
which is what makes the omission read as "this one does not exist".

## Consequence surface

- Agent orientation only; no code, test, or published behaviour is affected.
- The bundler-cache suite is the one place a bundler-integration regression is
  caught, so an agent that does not know it exists cannot reason about which
  command proves such a change.

## Invariant

The **Layout** section of the `project` skill accounts for every tracked
`tests/*` workspace, with a one-line statement of what the workspace proves —
the same shape the neighbouring entries use.

## Acceptance and verification

- The list reconciles against `git ls-files tests/ | cut -d/ -f2 | sort -u`.
- Follow `.agents/skills/documentation/SKILL.md` for agent-instruction writing.
- No command to run; the check is the reconciliation above.

PR #2289, opened 2026-07-21
fix: seal the random format-length class and the protobuf diagnostic code

samchon:
## Intent

One unified cycle pull request for the solo issue campaign frozen at
`88fafb36f2`. It owns the complete accepted cycle; nothing is deferred to a
second pull request.

| Issue | Scope |
| --- | --- |
| #2285 | protobuf transform diagnostics report `TS(typia.protobuf.typia.protobuf.encode)`; `tests/test-error` only checks the `TS(typia.` prefix |
| #2287 | `_randomFormatDate` / `_randomFormatDatetime` discard their declared `maximum` prop through `??` / `===` precedence |
| #2284 | `typia.random<string & Format<F> & MinLength/MaxLength>` throws for nine formats although the requested length is one their own validator accepts |
| #2286 | documented "Compiler Errors" tabs no longer match what typia emits, and one documented restriction no longer exists |
| #2288 | `project` skill Layout omits `tests/test-typia-bundler-cache` |

## Order

`#2285 → #2286` (the documentation records the corrected diagnostic code) and
`#2287 → #2284` (both touch the date/datetime generators). `#2288` is
independent.

## Verification

Pending. This body is the claim written before any code exists; per-commit
progress is recorded as comments, and each commit closes the issues it earns.

https://claude.ai/code/session_01Mc91ztDB5qujoSxLvfEgBQ

### DIFF
--- a/.agents/skills/project/SKILL.md
+++ b/.agents/skills/project/SKILL.md
@@ -37,6 +37,7 @@ The `packages/typia/src/transform.ts` file is a plugin descriptor, not a transfo
   - `test-typia-automated`, `test-utils-automated`: generator-driven matrix suites over their configured typia operations and `@typia/template` structures; their generated `src/features/` trees are rebuilt by the suite.
   - `test-interface`: compile-time tests for the exported `@typia/interface` types.
   - `test-typia-cli`, `test-typia-generate`: CLI and project-layout integration suites, including the `typia generate` command.
+  - `test-typia-bundler-cache`: webpack persistent filesystem-cache invalidation through `@ttsc/unplugin`; the only suite that exercises a bundler's cache.
   - `test-typia-exact-optional`: focused `exactOptionalPropertyTypes` behavior.
   - `test-feature-identity`: repository check that every suite's tracked `src/features` file exports exactly one `test_*` function named after the file, that one suite never exports a name twice, and that every `tests/*` workspace declares the package name `@typia/<directory>`.
   - `test-error`: transform-rejection verification; the build must fail and every fixture must be named by a typia diagnostic.
--- a/packages/typia/native/cmd/ttsc-typia/protobuf_diagnostic_code_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/protobuf_diagnostic_code_transform_test.go
@@ -1,0 +1,167 @@
+package main
+
+import (
+  "os"
+  "path/filepath"
+  "regexp"
+  "sort"
+  "strings"
+  "testing"
+)
+
+// TestProtobufDiagnosticCodeTransform verifies that a rejected protobuf call
+// site is named by its own API in the diagnostic code (#2285).
+//
+// `ProtobufFactory.Metadata` used to compose the code as `"typia.protobuf." +
+// props.Method`, while the encode and decode programmers pass the whole call
+// accessor (`ModuloMethodText`, the same text `TypeGuardError.method` carries).
+// Sixteen of the seventeen entry points therefore reported
+// `TS(typia.protobuf.typia.protobuf.encode)`; only `protobuf.message` was
+// correct, because its programmer passed a bare literal. The code is the only
+// machine-readable identity a typia rejection has, so a code that names no API
+// is a silent loss for anyone searching it.
+//
+// The oracle is the fixture source, not another copy of the expected string: a
+// diagnostic code is correct only when the accessor it names is literally the
+// accessor written at a call site in that file. A concatenation defect cannot
+// satisfy that, and neither can a misspelling.
+//
+//  1. Write one fixture that calls every protobuf entry point with `bigint`,
+//     which none of them supports.
+//  2. Transform it and collect every diagnostic code the transform reported.
+//  3. Require each code to be an accessor present in the fixture, and require
+//     every entry point to have been named.
+func TestProtobufDiagnosticCodeTransform(t *testing.T) {
+  methods := []string{
+    "message",
+    "encode",
+    "decode",
+    "assertEncode",
+    "assertDecode",
+    "isEncode",
+    "isDecode",
+    "validateEncode",
+    "validateDecode",
+    "createEncode",
+    "createDecode",
+    "createAssertEncode",
+    "createAssertDecode",
+    "createIsEncode",
+    "createIsDecode",
+    "createValidateEncode",
+    "createValidateDecode",
+  }
+  project, source := protobufDiagnosticCodeProject(t, methods)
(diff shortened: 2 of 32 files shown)


samchon:
**d14b24c** — `fix(protobuf): name the invoked API once in a rejected call site's code` (closes #2285)

`ProtobufFactory` now uses its `Method` prop verbatim as the diagnostic code instead of prefixing `typia.protobuf.` onto a value the encode and decode programmers already pass fully qualified; `ProtobufMessageProgrammer` passes the qualified name, and the message transformer's `tyipa.protobuf.message` typo is corrected. Both oracles are anchored to the fixture source — a new `TestProtobufDiagnosticCodeTransform` over all seventeen entry points, and `tests/test-error` now requires each code to name an accessor the fixture writes rather than accepting any `TS(typia.` prefix.

Mutation evidence: restoring the concatenation fails the Go regression with all seventeen doubled codes and fails `tests/test-error` with six fixture/code pairs.

**20fa447** — `fix(random): draw every string format at the lengths its grammar expresses` (closes #2284, #2287)

The nine formats left on the retry driver by #2192 now build their value at a length drawn from the intersection of the requested window and their grammar; `_randomFormatLength` is documented as single-length-only, which is what keeps `uuid` and `date` correct on it. The two date generators' `??`/`===` precedence bug is resolved in one shared epoch helper.

Coverage is two-directional: `test_random_format_length_window` pins exact lengths end-to-end through `typia.random`, and `test_random_format_length_grammar` walks every window from 0 to 48 for every format, requiring a realizable length to yield a value that format's own validator accepts and an unrealizable one to keep throwing.

Still to land in this cycle: #2286 (documentation) and #2288 (`project` skill layout), plus the formatter result for four files committed unformatted before this branch.

samchon:
**9c2b78a** — `docs: reproduce the diagnostics typia actually emits` (closes #2286, #2288) and **2019e64** — `chore(format)`.

Every "Compiler Errors" tab on `llm/application`, `llm/parameters`, `llm/schema`, `llm/structuredOutput`, `json/schema` and `protobuf/message` was recaptured by compiling the source in its neighbouring tab. `llm/schema`'s "No dynamic object keys" section documented a rule the transform no longer enforces — a dynamic key is accepted and emitted as `additionalProperties`, and only `strict: true` rejects it, which is what `tests/test-error/src/examples/llm.schema.additionalProperties.ts` has been asserting all along. The archived blog post keeps its original output. The `project` skill's Layout now lists `tests/test-typia-bundler-cache`.

The format commit carries four files that landed after #2279 without `pnpm format`; `prettier` reflows their doc comments.

**Local verification** (clean tree, this branch):

| Command | Result |
| --- | --- |
| `pnpm build` | exit 0 |
| `pnpm test` | exit 0, no failure in the log |
| `pnpm --filter ./tests/test-typia-schema start` | Success, including the three new random cases |
| `pnpm --filter ./tests/test-error start` | 78 files, 873 diagnostics, 142 distinct codes named |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestProtobufDiagnosticCodeTransform` | ok |

**Mutation sensitivity** (product fix reverted, tests retained):

| Mutation | Observed |
| --- | --- |
| `ProtobufFactory` restores `"typia.protobuf." + props.Method` | Go regression fails listing all 17 doubled codes; `tests/test-error` fails listing six fixture/code pairs |
| `_randomFormatByte`, `_randomFormatIpv4`, `_randomFormatTime`, `_randomFormatDatetime` restored to the retry driver | all three new random cases fail (`byte@0: threw at a realizable length`, the window suite throws, the epoch window drifts) |

**Not changed, recorded for the next round:** `_randomFormatByte` and `_randomFormatRegex` still return one hard-coded constant for an *unconstrained* draw, and the diagnostic type labels (`Recordstringnumber`, `Typeuint64`, `Array<false | true>`) are unchanged — the documentation now reproduces them as they are rather than as they were.

samchon:
Rebased onto `6a52427da8` (`chore(deps): make the ttsc catalog a single anchored declaration`, #2290), which landed on master while this branch's checks were running. The rebase was clean — #2290 touches `pnpm-workspace.yaml`, the dependabot config, the ttsc peer entries and `docs/setup/**`, none of which this branch edits.

The previous green run belongs to the pre-rebase head, so it does not gate this one: the checks are re-running against `f274b2835b`, and `pnpm install && pnpm build && pnpm test` is re-running locally against the same tree.

samchon:
**master is red, and this branch now repairs it.**

#2290 merged with its `test` and `nestia` lanes failing (~1m each). It removed `ttsc` from the root `package.json` on the premise that nothing at the root ran it — but root `pnpm test` does: `pnpm --filter=./tests/test-* -r start` executes workspace scripts that invoke `ttsx`, and every one of them resolved that binary through the root's `node_modules/.bin`. After `pnpm install` prunes it, the suites die before executing anything:

```
tests/test-feature-identity start: 'ttsx' is not recognized as an internal or external command
```

Fourteen workspaces run `ttsx`/`ttsc` without declaring it. Restoring the root entry would put back the Dependabot foothold #2290 deliberately removed, so **a5ed34d** declares the binary in each workspace whose own scripts run it — the pattern `test-interface`, `benchmark` and `examples` already follow. The catalog anchor stays the single version declaration.

Per `.agents/skills/issue-campaign/development.md`, a red lane is repaired in the same pull request even when the failure predates the cycle and is unrelated to its issues.

samchon:
**3a06a20** — the same root cause reached two more places in `test-typia-generate`'s harness.

`check-file-errors.cjs` spawned `<root>/node_modules/.bin/ttsx` and `filesystem/fixture.cjs` resolved `ttsc/package.json` with `paths: [repository]`. Both only worked while the root package.json declared `ttsc`. The spawn one was the worse of the two: that script asserts a call *fails*, so a missing executable arrived as the expected failure and was only caught by the message comparison.

Both now resolve from the workspace, which declares `ttsc` itself. `pnpm --filter ./tests/test-typia-generate start` exits 0; the full `pnpm test` is re-running locally and the checks are re-running against `3a06a20b4d`.

Per the maintainer's instruction, the `nestia` lane is not treated as a gate for this pull request.

samchon:
Rebased onto `649f159d4e` (#2292), which landed the same repair upstream — the fourteen workspace declarations and both `test-typia-generate` helper resolutions. My two `fix(ci)` commits were therefore dropped rather than carried: the branch is back to the five campaign commits, now sitting on a master whose test lane is green again.

The cycle diff against master is unchanged: 32 files, the four issue commits plus the formatter commit.

PR #2290, opened 2026-07-21
chore(deps): make the ttsc catalog a single anchored declaration

samchon:
## Intent

Every ttsc bump Dependabot has opened since the pnpm catalog landed was closed unmerged. This makes the ttsc version a single anchored declaration and removes the two things that made Dependabot produce an unmergeable diff.

## Why the Dependabot PRs failed

Two distinct failures, both reproducible in the closed PRs:

**Anchor present (#2065).** Dependabot touched `pnpm-lock.yaml` only, never `pnpm-workspace.yaml`. It could not rewrite a version behind a YAML alias, so it wrote the new version into the one literal foothold it had — the root importer's `ttsc` specifier — and left the catalog at the old version.

**Anchor absent (#2019, #2065, #2282).** Dependabot updated both files correctly except for the root importer, where it rewrote `specifier: catalog:typescript` into a literal range while the root `package.json` kept the catalog reference. Nested workspace importers were fine.

Either way CI died at the first step:

```
ERR_PNPM_OUTDATED_LOCKFILE
specifiers in the lockfile ({... "ttsc":"^0.19.3"})
don't match specs in package.json ({... "ttsc":"catalog:typescript"})
```

## Scope

**`@ttsc/unplugin` peer dependency removed from `packages/typia`.** typia has no reference to it anywhere in `src/` — the bundler plugin reads typia's transform output, not the reverse. The entry was also redundant: `@ttsc/unplugin@X` peers on `ttsc: "X"` exactly, so typia's `ttsc >= 0.19.2` already excludes every unplugin release below that floor. The `ttsc` optional peer stays; `packages/typia/native` consumes `driver.LoadProgram`, `driver.NewTransformGraph`, `driver.TransformOutputKey`, and `prog.EmitWithPluginTransformers`, which exist only in ttsc 0.19.2+. The `native/go.mod` comment naming both peers is updated.

**Root `package.json` ttsc entry removed.** It was the only root-level catalog reference, and therefore the foothold Dependabot corrupted. Nothing at the root ran ttsc: no script in `package.json`, no `scripts/` reference, no workflow invocation — the workflows name `node_modules/.cache/ttsc`, which is the plugin cache path, not the binary.

**ttsc catalog family bound to one `&ttsc` anchor.** The version is declared once on `ttsc`; `@ttsc/unplugin`, `@ttsc/playground`, and `@ttsc/wasm` alias it. This restores the arrangement from `b9d256c16`, which #2140 unwound to hold `@ttsc/wasm` back during the playground host API migration — a split #2143 has since made unnecessary. The two comment blocks justifying the split are replaced by one describing the lockstep.

**Dependabot allow list narrowed to `typescript` and `ttsc`.** The `@ttsc/*` catalog entries are now aliases with no version to rewrite, so asking Dependabot to bump them is what made it abandon the workspace file. Bumping the anchor line alone carries the family, and every `@ttsc/*` consumer resolves through the catalog: `config/`, `tests/test-typia-bundler-cache/`, and `website/`. Neither `experiments/mcp` nor `experiments/smoke` declares any `@ttsc/*`, so no manifest loses coverage. The group is kept because `ttsc` is also declared literally in both experiments manifests and one release should still open one pull request.

**Setup guide digressions cut.** The ttsc floor note explained the reference-graph SDK, bundler cache invalidation, peer declaration mechanics, the npm warning surface, and the type-only consumer exception; a reader installing typia acts on one fact. The note describing how ttsc discovers the transform from typia's own `package.json` was internal detail whose actionable half already opens Transform Options.

## Verification

Local build and test runs were prohibited for this change, so CI is the verification surface. What was checked locally:

- `pnpm install` regenerated the lockfile; its only content changes are the two removed importer entries.
- `pnpm-workspace.yaml` parsed with `yaml`: all four family entries resolve to `^0.19.2`, identical to the pre-anchor literals, so the lockfile needs no version change.
- `pnpm format` produced changes in four files unrelated to this work — `packages/typia/src/executable/FileSystemIdentity.ts`, `tests/test-typia-generate/scripts/identity/inode-precision.ts`, and the two `test_*_tool_error_single_json_fence.ts` suites. They are JSDoc rewraps left by earlier commits that did not run the formatter. They are reverted and excluded from this branch rather than folded in silently, and they still need a separate formatting pass on master.

## Deferred

Whether Dependabot preserves `&ttsc` when it rewrites the anchor line is unverified — it cannot be exercised locally. If it drops the anchor, the three `*ttsc` aliases become undefined and `pnpm-workspace.yaml` fails to parse, which surfaces loudly in that PR's install step rather than silently. The next daily run settles it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,8 +2,7 @@ version: 2
 updates:
   - package-ecosystem: "npm"
     # experiments/* are not pnpm workspace members, so their manifests are
-    # invisible to the root scan; list them so the ttsc family cannot drift
-    # there again (grouping spans directories, keeping one PR per family bump).
+    # invisible to the root scan; list them so ttsc cannot drift there again.
     directories:
       - "/"
       - "/experiments/mcp"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,14 +11,18 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+    # The @ttsc/* companions (playground, unplugin, wasm) are deliberately
+    # absent. pnpm-workspace.yaml binds the whole family to the &ttsc anchor,
+    # so their catalog entries are YAML aliases carrying no version to rewrite;
+    # asking Dependabot to bump them made it give up on the workspace file and
+    # write the version into the lockfile alone. Allowing ttsc by itself leaves
+    # the anchor line as the single edit, and the aliases follow it.
     allow:
       - dependency-name: "typescript"
       - dependency-name: "ttsc"
-      - dependency-name: "@ttsc/*"
-    # ttsc and the @ttsc/* companions (playground, unplugin, wasm) release in
-    # lockstep; bumping them separately leaves the catalog on mixed versions.
+    # ttsc is declared in the root catalog and in both experiments manifests;
+    # group it so one release opens one pull request across all three.
     groups:
       ttsc:
         patterns:
           - "ttsc"
-          - "@ttsc/*"
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "@trivago/prettier-plugin-sort-imports": "^6.0.2",
     "bumpp": "^10.4.1",
     "prettier": "^3.8.1",
-    "prettier-plugin-jsdoc": "^1.8.0",
-    "ttsc": "catalog:typescript"
+    "prettier-plugin-jsdoc": "^1.8.0"
   }
 }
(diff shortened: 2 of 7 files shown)


samchon:
CI caught a real defect in the first push, recorded here rather than by rewriting the body.

**`ttsx: not found`** — `test` and one `Ubuntu` job failed at spawn with ENOENT across every suite.

The root `package.json` ttsc entry was not unused as this PR's original analysis claimed. Fourteen `tests/*` workspaces invoke `ttsc` or `ttsx` from a script without declaring ttsc, and resolved the binary through the workspace root's `node_modules/.bin`, which pnpm places on a package script's PATH. Removing the root entry took the binary away from all of them. The original check covered the root scripts, `scripts/`, and the workflows — the workflows name `node_modules/.cache/ttsc`, a cache path, not the binary — but not the scripts of other workspaces.

Fixed by declaring `"ttsc": "catalog:typescript"` in the fourteen workspaces that run it, matching `benchmark`, `examples`, and `tests/test-interface`, which already declare it. Restoring the root entry instead would have reinstated the Dependabot foothold this PR exists to remove.

Verified after the fix: `node_modules/.bin/ttsx` resolves in all fourteen workspaces, and the workspace root still has no ttsc binary.

Issue #2293, opened 2026-07-21
ILlmApplication.config always reports strict: false, so @typia/mcp validates a strict tool output with every constraint dropped


samchon:
## Problem

`ILlmApplication.config` always reports the default configuration. A call site
that asks for strict mode —

```ts
typia.llm.application<IApplication, { strict: true }>()
typia.llm.controller<Service, { strict: true }>("service", new Service())
```

— builds strict schemas, and then reports `config: { strict: false, validate: null }`.

The field is not decorative. `@typia/mcp` and `@typia/vercel` read it to decide
how to interpret the schemas they were handed, and a strict schema carries its
constraints as description tags rather than as JSON Schema keywords. Reading a
strict application back with `strict: false` therefore **drops every constraint**
from the check.

## Evidence

Reproduction (`tests/test-mcp`, `pnpm --filter ./tests/test-mcp exec ttsx src/probe.ts`):

```ts
interface IApplication {
  create(props: { input: { title: string } }): Promise<{
    id: string & tags.Format<"uuid">;
    score: number & tags.Minimum<0> & tags.Maximum<100>;
  }>;
}

const strict = typia.llm.application<IApplication, { strict: true }>();
const loose = typia.llm.application<IApplication>();
```

| | `application.config` | output schema for `score` |
| --- | --- | --- |
| `{ strict: true }` | `{"strict":false,"validate":null}` | `{"type":"number","description":"@minimum 0\n@maximum 100"}` |
| default | `{"strict":false,"validate":null}` | `{"type":"number","minimum":0,"maximum":100}` |

The strict build did apply strict mode — the constraints moved into
`description`, and a strict build of a type with an optional property is rejected
at compile time — so only the reported `config` is wrong.

Running exactly what `@typia/mcp` runs on a tool result
(`LlmJson.validate(func.output, true, controller.application.config)`) with
`{ id: <uuid>, score: 500 }`:

| Application | config used | `score: 500` accepted |
| --- | --- | --- |
| strict | the reported one | **true** — the constraint is gone |
| strict | `{ strict: true }`, what the call site declared | false |
| loose | the reported one | false |
| loose | `{ strict: false }` | false |

### Root cause

`packages/typia/src/internal/_llmApplicationFinalize.ts` rebuilds the field from
the converter's defaults instead of from the call site:

```ts
config: {
  ...LlmSchemaConverter.getConfig(),   // { strict: false }
  validate: config?.validate ?? null,
},
```

[...]

PR #2294, opened 2026-07-21
fix(llm): report the schema configuration the call site declared

samchon:
## Intent

Cycle 2 of the solo issue campaign, frozen at `3255792cd7`. One issue survived
round 2's verification, and this pull request owns it.

| Issue | Scope |
| --- | --- |
| #2293 | `ILlmApplication.config` always reported `strict: false`, so `@typia/mcp` validated a strict tool's output with every constraint dropped |

## What changed

The transform resolves the `Config` generic and is the only place that knows it,
so it now travels with the emitted application: `WriteApplication` emits
`config: { strict }`, `ILlmApplication.__IPrimitive` carries it, and
`_llmApplicationFinalize` widens it with the `validate` hook instead of
rebuilding it from `LlmSchemaConverter.getConfig()`.

`tests/test-mcp/src/features/test_mcp_tool_output_constraint_enforcement.ts` had
been assigning `strict.application.config.strict = true` by hand, with a comment
naming the defect. That line is gone, so the case now proves the whole path, and
the `McpControllerRegistrar` comment that told readers to expect a wrong value is
corrected.

## Verification

- `test_llm_application_reported_config` (new): the reported config for
  application and controller, strict and default, the `validate` hook controls,
  and the consequence — inverting each output schema with the reported config
  must reject an out-of-range value.
- Mutation: restoring `LlmSchemaConverter.getConfig()` fails both the new case
  (`strict application reports strict`) and the MCP case (`a strict controller
  reports the config it was built with`).
- Local `pnpm build` and `pnpm test` are running; CI gates this head.

https://claude.ai/code/session_01Mc91ztDB5qujoSxLvfEgBQ

### DIFF
--- a/packages/interface/src/schema/ILlmApplication.ts
+++ b/packages/interface/src/schema/ILlmApplication.ts
@@ -109,6 +109,16 @@ export namespace ILlmApplication {
     ILlmApplication<Class>,
     "config" | "functions"
   > {
+    /**
+     * Schema configuration the call site declared.
+     *
+     * The transform resolves it from the `Config` generic and emits it here,
+     * because it is the only place that knows it. The runtime finalizer widens
+     * it into {@link ILlmApplication.config} by adding the `validate` hook, so a
+     * consumer reading `application.config.strict` sees what the schemas were
+     * actually built with.
+     */
+    config: ILlmSchema.IConfig;
     functions: Omit<ILlmFunction, "parse" | "coerce">[];
   }
 }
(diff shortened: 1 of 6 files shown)


samchon:
**Compatibility check.** `LlmSchemaConverter.getConfig(app.config)` treats an absent argument as `{ strict: false }`, so an application literal emitted by an older transform — one that carries no `config` — resolves to exactly the previous behaviour rather than to `undefined`. The type change is confined to `ILlmApplication.__IPrimitive`, which is the transform's own emit contract (`@ignore`, "Internal type for typia transformer"); the public `ILlmApplication.config` and `ILlmSchema.IConfig` shapes are untouched.

**Mutation evidence.** Restoring `...LlmSchemaConverter.getConfig()` in `_llmApplicationFinalize` fails both regressions:

```
test_llm_application_reported_config: Bug on strict application reports strict
test_mcp_tool_output_constraint_enforcement: Bug on a strict controller reports the config it was built with
```

The second one is the case that used to correct `config.strict` by hand, which is what makes its failure meaningful now.

PR #2307, opened 2026-07-22
docs(agents): harden the campaign closing keywords, review ledger, and discovery loop

samchon:
## Intent

`samchon/ttsc` runs the same solo issue campaign skill, and the two copies diverged after typia's #2283. Three ttsc commits carry rules typia's copy never received, each written after a real campaign failure. This pull request brings them over and leaves typia's own campaign rules intact.

## What ttsc learned

| ttsc commit | Failure that produced the rule |
| --- | --- |
| [`8af2ca9`](https://github.com/samchon/ttsc/commit/8af2ca973) (samchon/ttsc#924) | A squash merge concatenates every commit message into the merge commit body, and GitHub parses closing keywords from that whole body. A reverted commit's `Close #n` still closes its issue, because a revert adds a commit rather than removing the original from the range. ttsc lost samchon/ttsc#839 and samchon/ttsc#922 to this in two consecutive cycles and reopened both by hand. |
| [`99ee5a5`](https://github.com/samchon/ttsc/commit/99ee5a54a) (samchon/ttsc#946) | The per-commit and Self-Review ledgers were ordinary issue-style pull-request comments, so line-specific findings had no line to attach to and round results did not read as reviews. |
| [`f489764`](https://github.com/samchon/ttsc/commit/f4897649c) (samchon/ttsc#929) | `## Discover Issues` carried the full-scope rule and the loop-termination rule inside ordinary paragraphs. Both were read past: two cycles merged with no post-merge discovery round, and the round that finally ran produced three issues. |

## Scope

**`.agents/skills/issue-campaign/development.md`**

- A revert must not carry the closing keyword forward, and the reason it still closes the issue anyway.
- The merge gate reconciles `git log origin/master..HEAD` against what survives at `HEAD` before merging.
- Post-merge step 1 reopens any issue the squash merge closed without a surviving fix.
- The per-commit ledger and every Self-Review round become formal GitHub pull-request reviews with the `COMMENT` event.

**`.agents/skills/issue-campaign/SKILL.md`**

- `### Every Round Is Full-Scope` and `### Discovery Ends Only On An Empty Round` replace the two paragraphs that carried those rules, keeping typia's frozen-commit, census, matrix, and `INCOMPLETE` conditions inside them.
- AGENTS.md's **Choose the principled course** rule is linked rather than restated, and the development skill's **Forbidden** section is linked rather than merely named.

[...]

PR #2308, opened 2026-07-22
docs(agents): give discovery its own loop and stop restating the round procedure

samchon:
## The defect

Three rules described the discovery loop and two of them disagreed.

`review/SKILL.md` Non-Negotiable Review Law, **Unlimited rounds**: *"whenever the reviewer applies an improvement or **accepts a meaningful issue candidate**, update the work and start another complete round."*

`review/SKILL.md` Solo Issue Discovery Rounds, step 7: *"If any meaningful candidate survives, **finish the authorized issue and implementation flow, then** freeze the new integrated commit and begin another fresh full-scope round."*

Step 7 wins in practice because it is the campaign-specific procedure, and `issue-campaign/SKILL.md` agreed with it — *"Perform **one** complete Solo Issue Discovery Round"*, and a `### Discovery Ends Only On An Empty Round` section that only ever described the cycle-to-cycle loop. So a campaign that accepted a single candidate was entitled to publish it, implement it, and merge a cycle built from whatever the first round reached first, leaving the rest of the declared scope unexamined until after that merge.

## The fix

Discovery owns its own loop, and the two loops are now named separately.

- **Inside one cycle:** accepting a candidate starts another complete full-scope round against the same frozen commit. Only an empty round settles the cycle's issue set. Nothing is published or implemented from a partial candidate pool.
- **Across cycles:** a merged cycle buys one more discovery *phase*, not one more round.

`review/SKILL.md` step 7 no longer routes the next round through implementation; publication and the cross-cycle round moved to a new step 8.

## Removing the restatement, not the substance

`## Discover Issues` carried four paragraphs that reproduced the frozen commit, environment contract, tracked-file census, coverage matrix, `INCOMPLETE` state, and stop rule — all of which `review/SKILL.md`'s Solo Issue Discovery Rounds already owns. The documentation skill requires stating each rule at its owning document and linking elsewhere, so the section now links that procedure and keeps only what the campaign owns: the exclusion authority, the evidence-layer rule, the stale-harness rule, the **Forbidden** retrospective audit, and the adjudicate-the-full-pool rule.

Every removed sentence was mapped to an owner before deletion. Five rules had none, so they were consolidated into `review/SKILL.md` rather than dropped:

| Rule with no owner | Moved to |
| --- | --- |
| static, dynamic, array, `any`, and `unknown` value contexts | step 4 |
| maximum-width and one-past-boundary inputs | step 4 |
| ambient and runtime-native declaration provenance | step 4 |
| deployment consumers | step 4 |
| "Do not downgrade a proved defect to an observation to satisfy the stop rule" | step 9 |

Against `master`, `issue-campaign/SKILL.md` loses 1139 bytes and `review/SKILL.md` gains 689, so this is a relocation into the owning document, not a trim.

## Structure

The knowledge-base status vocabulary was a four-by-two mapping — `PASS`, `FAIL`, `N/A`, `UNTESTED` against tracked paths and executable cells — written as a single prose paragraph that also carried the row schema and both reconciliation rules. It is a table now, with the row schema and each reconciliation rule as their own paragraphs, per the documentation skill's *use structure to compress meaning* and *give each paragraph one job*.

## Verification

Documentation only. Link targets confirmed: `../review/SKILL.md#solo-issue-discovery-rounds`, `../review/SKILL.md#non-negotiable-review-law`, `../development/SKILL.md#forbidden`. All three files are CRLF on disk and every inserted line carries its `\r`. `pnpm format` does not cover `*.md`.


samchon (commented):
## Self-Review rounds 1–5

Five complete rounds over the whole `master...HEAD` diff. Rounds 1–4 each produced findings; round 5 is clean.

### Round 1 — `170d3f17fc`, four findings

**Loop with no target.** `review/SKILL.md` step 7 said "begin another fresh full-scope round" inside an ordered procedure without naming where the list resumes. It returns to step 3 now.

**Rationale that did not reach the new rounds.** `### Every Round Is Full-Scope` justified itself entirely from a merged cycle changing the state. The rounds this pull request adds repeat inside one cycle against the same frozen commit with nothing merged between them, so that reasoning left a reader free to treat a same-cycle predecessor as coverage. The rule now rests on the round being frozen state and names both cases.

**Delegation list claimed too much.** `## Discover Issues` credited the review procedure with owning "the stop rule" while `### Discovery Ends Only On An Empty Round` states the campaign's own termination two screens later. It names the `COMPLETE` and `INCOMPLETE` round states it actually meant.

**False figure in the pull-request body.** The body claimed `issue-campaign/SKILL.md` shrank by 436 bytes. That was measured against the file's size before #2307 merged, not against `master`. The real figures are −1139 and +689; the body is corrected.

### Round 2 — `3320c2054a`, two findings, both in text this pull request introduced

"Discovery repeats in two nested loops, and **both end on the same condition**" was false. The inner loop ends on an empty round; the outer additionally requires the round not to be `INCOMPLETE` and no accepted issue to remain unresolved. Both do require an empty round, which is what it says now.

The cross-cycle paragraph said a merged cycle "produces one more round", contradicting `development.md`'s repeat step **in this same pull request**, which says a merged cycle buys one more discovery *phase*.

### Round 3 — `c6df3b65fd`, one finding

`multi-agent/review.md:57` carried the identical defect the solo procedure had — step 7 sent a surviving candidate through implementation before the next repository-wide round. Fixing only the solo path would have left the two variants describing opposite loops for the same phase. The parallel round now repeats from step 2 with a replaced team until no candidate survives, and implementation moved to its own step. Return target is step 2 rather than the solo step 3 because a replaced team holds no inventory.

### Round 4 — `d6368e5729`, one finding

The replaced-team inventory note sat between the loop instruction and its termination condition, so the step read as two interleaved rules.

### Round 5 — `d1d489f6ef`, clean

- Every loop statement across all five documents was read side by side and they agree: `review/SKILL.md:19` and `:54`, `issue-campaign/SKILL.md:74` and `:76`, `development.md:141`, `multi-agent/review.md:57` and `:58`.
- Every link target in the diff resolves: `../review/SKILL.md#solo-issue-discovery-rounds`, `../review/SKILL.md#non-negotiable-review-law`, `../development/SKILL.md#forbidden`.
- `multi-agent/review.md`'s renumbered list runs 1–9 with no gap, and no document references its old step 8.
- All four files are CRLF on disk with zero LF-only lines.
- Every rule removed from `issue-campaign/SKILL.md` was matched to an owner before deletion; the five with no owner moved into `review/SKILL.md` rather than disappearing.

Documentation only. No source, test, or generated artifact is touched, and `pnpm format` does not cover `*.md`.

Issue #2313, opened 2026-07-22
@typia/vercel states the pre-#2293 transform defect as current in shipped source


samchon:
## Problem

`@typia/vercel` ships a source comment that describes a transform defect as
current after the defect was fixed. It tells the next reader that class-controller
output validation is knowingly inexact, and it names a condition — "the moment
the transform reports it truthfully" — that the transform already satisfies.

`packages/vercel/src/internal/VercelToolsRegistrar.ts`, inside `createTool`:

```ts
      : // Validate the output against the config that produced its schema. A
        // `strict` application carries its constraints as description tags
        // instead of keywords, and only that config tells the inverter to
        // read them back.
        //
        // An HttpLlm application reports the config it was composed with, so
        // this is exact for it. A typia.llm.controller reports
        // `strict: false` whatever its Config generic said — a transform
        // defect, not a reason to guess here: guessing is what erased the
        // non-strict constraints in the first place. This reads the declared
        // config and becomes exact for class controllers too the moment the
        // transform reports it truthfully.
        LlmJson.validate(func.output, true, props.config);
```

Every sentence after "An HttpLlm application" is false at `HEAD`. A
`typia.llm.controller` reports the config its `Config` generic declared, so the
`LlmJson.validate` call below the comment is already exact for class
controllers.

A comment like this costs more than its inaccuracy. It sends a maintainer to
hunt a transform defect that no longer exists; and because it frames the current
call as a placeholder awaiting a fix elsewhere, it invites exactly the "guess
the config" change it warns against.

## Evidence

The claim is contradicted by two executed tests in this repository, both green
in the root `pnpm test` baseline at `ee9609d2a8`.

`tests/test-typia-schema/src/features/llm.application/test_llm_application_reported_config.ts`:

```ts
  TestValidator.equals(
    "strict controller reports strict",
    typia.llm.controller<Service, { strict: true }>("service", new Service())
      .application.config.strict,
    true,
  );
```

`tests/test-vercel/src/features/test_vercel_tool_output_constraint_enforcement.ts`
— the suite for the very registrar that carries the comment:

```ts
    const strict: ILlmController = typia.llm.controller<
      ConstraintController,
      { strict: true }
    >("constraint", new ConstraintController());
    TestValidator.equals(
      "a strict controller reports the config it was built with",
      strict.application.config.strict,
      true,
    );
```

and its doc comment records the transition explicitly:

> The strict controller is converted exactly as typia emits it. It **used to**
[...]

Issue #2312, opened 2026-07-22
nestia.yml runs pnpm test under a name and triggers that describe a downstream check it no longer performs


samchon:
## Problem

`.github/workflows/nestia.yml` no longer has anything to do with nestia. Its one
job checks out typia, installs, restores the ttsc plugin cache, and runs:

```yaml
      - name: Run Tests
        run: pnpm test
```

That is the same command `.github/workflows/test.yml` runs, on the same runner,
with the same Node and Go versions and the same cache key. Two consequences
follow, and they point in opposite directions:

1. **The downstream signal named by the lane does not exist.** `@nestia/core` is
   the repository's documented downstream consumer. A reader of the checks list,
   or of the branch-protection settings, sees a green `nestia` check and
   concludes typia was verified against nestia. Nothing in the job clones,
   installs, or runs nestia.
2. **The heaviest lane runs twice.** Every pull request touching `packages/**`
   triggers both `test` and `nestia`, and both spend the full `pnpm test`
   budget — the 15 TypeScript workspaces plus both Go trees — on the same
   commit. `nestia.yml` carries `timeout-minutes: 60`.

The trigger paths match neither reading. `nestia.yml` fires on `experiments/**`
and `scripts/**`, which `pnpm test` does not exercise (`experiments/**` has its
own lane in `experiments.yml`; `scripts/` holds only the Go formatter), and it
does **not** fire on `tests/**` or `config/**`, which `pnpm test` does exercise.

## Evidence

Current file, in full below the trigger block:

```yaml
name: nestia
on:
  pull_request:
    paths:
      - ".github/workflows/nestia.yml"
      - "experiments/**"
      - "scripts/**"
      - "packages/**"
      - "package.json"
      - "pnpm-lock.yaml"
      - "pnpm-workspace.yaml"
  workflow_dispatch:
concurrency:
  group: nestia-${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
jobs:
  test:
    runs-on: ubuntu-latest
    timeout-minutes: 60
    steps:
      # checkout, setup-node 24.x, setup-go 1.26.x, pnpm/action-setup
      # pnpm install
      # Restore ttsc Plugin Cache   (same key as test.yml)
      - name: Run Tests
        run: pnpm test
```

History:

- `64c2efc2b2` — `ci(nestia): verify typia tarballs against nestia next (#1896)`
[...]

Issue #2310, opened 2026-07-22
Format<time> rejects the RFC 3339 leap second and long fraction that Format<date-time> accepts


samchon:
## Problem

`Format<"time">` rejects two productions that RFC 3339 `full-time` allows and
that this repository's own `Format<"date-time">` validator already accepts:

1. the **leap second** — `23:59:60Z` is refused, while `2024-12-31T23:59:60Z` is
   accepted;
2. a **fractional second longer than nine digits** — `00:00:00.1234567890Z` is
   refused, while `2024-01-01T00:00:00.1234567890Z` is accepted.

RFC 3339 defines `full-time = partial-time time-offset`, `partial-time =
time-hour ":" time-minute ":" time-second [time-secfrac]`, `time-second = 2DIGIT
; 00-60 based on leap second rules`, and `time-secfrac = "." 1*DIGIT`. The
`time` validator caps the second at `59` and the fraction at nine digits, so a
valid RFC 3339 clock value fails `is`, `assert`, `validate`, and
`OpenApiValidator`.

This is not a difference of opinion between two formats: `date-time` embeds the
same `partial-time` production and implements the leap-second rule deliberately,
including the offset arithmetic that decides when `:60` is legal. `time` was
written independently of it.

## Evidence

Reproduced twice at `ee9609d2a8` (Node v22.21.0) through the real transform,
checking all three ways a user can reach the format — the runtime helper the
type tag imports, the type tag itself, and the JSDoc comment tag:

```ts
import typia, { tags } from "typia";
import { _isFormatTime } from "typia/lib/internal/_isFormatTime";

interface ICommentTime {
  /** @format time */
  value: string;
}
_isFormatTime("23:59:60Z");                                   // false
typia.createIs<{ value: string & tags.Format<"time"> }>()({ value: "23:59:60Z" }); // false
typia.createIs<ICommentTime>()({ value: "23:59:60Z" });       // false
```

Expected column is RFC 3339 `full-time`, matching the cases in the
JSON-Schema-Test-Suite `optional/format/time.json`:

| value | expected | direct | type tag | comment tag |
| --- | --- | --- | --- | --- |
| `23:59:60Z` | valid | ❌ false | ❌ false | ❌ false |
| `23:59:60+00:00` | valid | ❌ false | ❌ false | ❌ false |
| `15:59:60-08:00` | valid | ❌ false | ❌ false | ❌ false |
| `01:29:60+01:30` | valid | ❌ false | ❌ false | ❌ false |
[...]

Issue #2309, opened 2026-07-22
typia.random throws for a satisfiable UniqueItems array whose element domain is smaller than the drawn item count


samchon:
## Problem

`typia.random<T>()` throws for types it can satisfy. When an array carries
`UniqueItems` and its element domain holds fewer distinct values than the item
count the generator happens to draw, generation fails with

```
Unable to generate enough unique items; the element domain may be too small.
```

even though the declared type is satisfiable — `Array<boolean> & tags.UniqueItems`
is satisfied by `[]`, `[true]`, and `[true, false]`.

The failure is random, not deterministic: the same declaration succeeds or
throws depending on the count drawn for that call. `Array<boolean> &
tags.UniqueItems` threw on 134 of 200 draws. Any consumer that uses
`typia.random` for fixtures — the documented use for it — gets a flaky
generator, and the flakiness scales with how small the element domain is.

## Evidence

Reproduced at `ee9609d2a8` on Node v22.21.0 / pnpm 10.6.4 through the real
transform (`pnpm --filter ./tests/debug exec ttsx src/probe.ts`), using
`typia.createRandom<T>()` and `typia.createIs<T>()` for each declaration:

```ts
type BoolUnique = Array<boolean> & tags.UniqueItems;
const random = typia.createRandom<BoolUnique>();
for (let i = 0; i < 200; ++i) random(); // throws on ~2 of every 3 calls
```

200 draws per declaration, `ok` = a value the declaration's own `is` accepts:

| Declaration | ok | threw |
| --- | --- | --- |
| `Array<boolean> & UniqueItems` | 66 | **134** |
| `Array<"a" \| "b"> & UniqueItems & MinItems<2>` | 41 | **159** |
| `Array<number & Type<"uint32"> & Maximum<3>> & UniqueItems & MinItems<2>` | 93 | **107** |
| `Array<1 \| 2 \| 3> & UniqueItems` | 96 | **104** |
| `Array<number & Type<"uint32"> & Minimum<0> & Maximum<1>> & UniqueItems` | 78 | **122** |
| `Set<boolean>` | 200 | 0 |
| `Array<number & Type<"uint32">> & UniqueItems` (control, wide domain) | 200 | 0 |

No draw ever produced a value its own validator rejected, so the uniqueness
relation itself is sound; only the count selection is wrong.

Boundary behavior, 100 draws each — both of these outcomes are correct today and
must survive:

| Declaration | ok | threw |
| --- | --- | --- |
[...]

PR #2314, opened 2026-07-22
fix: seal the random unique-item domain, the RFC 3339 time format, and the CI lane inventory

samchon:
Cycle 1 of `campaign-2026-07-22`. This pull request owns the **complete accepted
issue set** of the cycle.

## Cycle issues

| Issue | Subject | Landed by |
| --- | --- | --- |
| #2309 | `typia.random` throws for a satisfiable `UniqueItems` array whose element domain is smaller than the drawn item count | `7ea86b12ce` |
| #2310 | `Format<"time">` rejects the RFC 3339 leap second and long fraction that `Format<"date-time">` accepts | `8365bf667a`, `fbb390fbf4` |
| #2312 | `nestia.yml` runs `pnpm test` under a name and triggers that describe a downstream check it no longer performs | `ad13d4b9c4` |
| #2313 | `@typia/vercel` states the pre-#2293 transform defect as current in shipped source | `c9ea79969e`, `20166fc0fc` |
| ~~#2311~~ | ~~`examples/` has no pull-request compile gate~~ | **closed invalid** — `website.yml` compiles the examples on every pull request through `build:prepare` → `build:examples`. The report traced the workflow files and stopped before the npm script chain. No code from it landed; the disproof is on the issue. |

## What changed

**`Format<"time">`** now implements RFC 3339 `full-time` in all three owners —
the runtime helper the type tag imports, the `FormatCheatSheet` expression the
JSDoc tag emits, and the `@typia/utils` copy `OpenApiValidator` reads. A leap
second is accepted exactly where the offset-adjusted value is 23:59 UTC, which
is the half of `date-time`'s existing rule a dateless clock can enforce, and
`time-secfrac` is open above. `_randomFormatTime` follows the grammar instead of
the old pattern, so it drops its 24-character ceiling and picks between the `Z`
and offset shapes wherever both fit.

**`_randomArray`** stops treating an exhausted retry budget as failure. The
element domain is not knowable from inside the helper, so the drawn count is a
preference and what the loop collected is the answer; only a result below the
floor throws. `Array<boolean> & UniqueItems` failed on 134 of 200 draws before
this and on none after, while `Array<boolean> & UniqueItems & MinItems<3>` still
throws on every draw.

**CI**: `nestia.yml` is gone. Its single step was the `pnpm test` that `test.yml`
already runs, under a name that promised a downstream check `bdf81fde87` had
already removed. `build.yml` takes over `scripts/**`, and `test.yml` regains the
`timeout-minutes: 60` and `workflow_dispatch` that left with it.

**Docs**: the claim that `typia.llm.controller` misreports `config.strict` is
gone from the vercel registrar, from `LlmJson.validate`'s public JSDoc, and from
the published `llm/json` guide.

## Verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-schema start` | Success (includes the two new regressions) |
| `pnpm --filter ./tests/test-utils start` | Success (includes `time` in the format-parity matrix) |
| `pnpm test:go:public` | pass |
| `pnpm test:go:native` | re-running alone; the first attempt collided with a concurrent Go run on the shared `native/.tmp-ttsc-typia-tests` scratch path and timed out |
| `pnpm --filter ./examples --fail-if-no-match run build` | exit 0, and exit 2 with a deliberately broken example |
| root `pnpm test` | this pull request's CI |

Base commit `ee9609d2a8`. Integration-command manifest `manifest/2026-07-22/v1`,
sha256 `d0f85764b02767020c272762da9ef042d6f0c0dc7431a8f57b71dcdaea4698de`.

Each commit carries its own closing line, so the merge closes exactly what
survives at `HEAD`. Every landed commit was read by an independent read-only
reader; their findings and what was done about each are in the review thread.


samchon (commented):
**`c9ea79969e`** — docs(llm): describe the config every application already reports

Resolves #2313. Removed the claim that a typia.llm.controller misreports config.strict from VercelToolsRegistrar and from LlmJson.validate's public JSDoc. Both now describe what #2293 / PR #2294 landed. No behavior change; test-vercel and test-mcp are controls.

samchon (commented):
**`ad13d4b9c4`** — ci: stop running the test suite twice under a downstream check's name

Resolves #2312. Deleted .github/workflows/nestia.yml, whose only step was the pnpm test that test.yml already runs, and claimed scripts/** on build.yml, the lane that executes create-conditional-declarations.cjs through the langchain and mcp builds. development.md's cancel-in-progress sentence updated with it.

samchon (commented):
**`20166fc0fc`** — docs(llm): finish the config sweep in the guide and the bare-schema case

Refs #2313. The commit reader over c9ea799 found the same pre-#2293 claim still published in website/src/content/docs/llm/json.mdx, and an overreach in the JSDoc that replaced it: a bare schema from typia.llm.parameters has no config to read off an application.

samchon (commented):
**`8365bf667a`** — fix(format): answer a clock the way RFC 3339 writes one

Resolves #2310. Format<"time"> now implements full-time in all three owners: the runtime helper, the FormatCheatSheet inline expression, and the @typia/utils copy. Leap second accepted exactly where the offset-adjusted value is 23:59 UTC; time-secfrac open above. _randomFormatTime follows the grammar, and test_openapi_validator_format_parity gains time.

samchon (commented):
**`645fb53981`** — chore: drop the round-1 probes and fold in the commit readers' corrections

No issue. Removes four tests/debug scratch probes the previous commit swept up, and records the corrections the readers reported: build.yml's overclaiming comment, test.yml's lost timeout-minutes and workflow_dispatch, and the two config texts.

samchon (commented):
**`fbb390fbf4`** — test(random): move the clock's open fraction out of the throw list

Refs #2310. test_random_format_length_window asserted that a 25-character clock throws; that ceiling was the old nine-digit fraction cap, not the grammar. Two positive rows replace it and a below-the-shortest negative keeps the throw side honest.

samchon (commented):
**`7ea86b12ce`** — fix(random): draw a unique array at a count its element domain can reach

Resolves #2309. _randomArray keeps what the unique loop collected instead of failing on the retry budget, and throws only below the floor. Adds test_random_unique_items_domain with the small domains, the boundaries, the unsatisfiable floor, and a wide-domain control that a clamp-to-minimum fix could not pass.

samchon (commented):
## Self-Review round 1 — `7ea86b12ce`

One complete pass over `origin/master...7ea86b12ce` (16 files, 692 diff lines), read whole rather than by hunk.

**Finding — the positive rows could not fail the way the defect failed.** Every `produces(...)` row in `test_random_unique_items_domain` asserted only that a draw stays inside its element domain and satisfies its own type. A generator that collected one element and stopped satisfies all of them. The wide-domain control catches a clamp applied to *every* unique array, but not one applied only where the retry budget runs out — which is precisely where the old code failed. Corrected in `bf587bc43a`: `domain` is now read as both bounds, so some draw must reach it.

**Checked and clean**

- `_randomArray`: `minItems: 0`, `maxItems: 0`, `maxItems` below the default floor, and `recursive` all still land on a count the loop can satisfy; the non-unique path is untouched; the loop is bounded in every branch and the throw keeps the message #2032 gave it.
- `_isFormatTime`: capture indices agree across the three owners; `-00:00` produces `-0` harmlessly; the offset stays mandatory; `((m % 1440) + 1440) % 1440` normalizes both tails, so `00:00:60+00:01` accepts and `23:59:60-00:01` rejects.
- `_randomFormatTime`: hand-checked at 9, 14, 15, 16, 24, 40 — each branch returns exactly the drawn length and matches the new pattern; the 10-character gap still redirects and re-checks against the window.
- Documentation rule: `website/src/content/docs/validators/tags.mdx` documents `UniqueItems` as "all elements pairwise distinct" and lists `time` among the formats without semantics, so neither behavior change has a guide page that now reads false.
- `test.yml`'s `workflow_dispatch` and `timeout-minutes` parse where they sit; no branch protection references the deleted `nestia` check.

**Known gap, recorded rather than fixed here:** nothing enforces byte-identity between the 26 runtime helpers duplicated in `packages/typia/src/internal` and `packages/utils/src/validators/functional`. `test_openapi_validator_format_parity` now pins seven of them behaviorally. That is a discovery-round candidate, not this cycle's work.

Round 2 opens on `bf587bc43a`.

samchon (commented):
## Self-Review rounds 2 and 3

**Round 2** on `bf587bc43a`, prompted by the commit reader over `7ea86b12ce`.

**Finding — the fix moved the retry budget onto the success path.** Keeping what
the unique loop collected removed the throw, but the budget it used to fail on
is now spent on every successful draw, and it compounds with nesting: each level
of a small-domain unique array spends its whole budget inside each candidate of
the level above. Measured on `Array<Array<boolean> & UniqueItems> &
UniqueItems`, 20 draws took **491 ms**; one more level took **108 ms per draw**.
Corrected in `7e3b1c9ede` with a stale-run break that applies only above the
floor — 8 ms and 1.4 ms per draw respectively, and the regression suite fell
from 166 ms to 27 ms.

**Finding — three prose claims went stale inside one commit.** `bf587bc43a` made
the positive rows reject a floor clamp, which falsified the control's own
comment ("a fix that clamped every unique array to its floor would satisfy every
row above"), the headline paragraph, and step 1 of the numbered list. All three
rewritten. The length-window comment calling 14 an offset-only length was wrong
the same way the length-40 one was: the `Z` shape reaches 14 with four fraction
digits.

**Round 3** on `7e3b1c9ede`, complete pass over `origin/master...7e3b1c9ede`.

- The stale break cannot fire below the floor: `elements.length >= minimum`
  short-circuits before `++stale`, so `stale` is never even incremented there,
  and the first candidate always pushes (`every` on an empty list is vacuously
  true), so no draw can break with nothing collected.
- `stale` resets on every push; `maximumAttempts` remains the hard cap; `count`
  still respects `maxItems`, and `elements.length <= count` always holds.
- `minimum === 0` (recursive, or `MaxItems<0>`) traced: either the loop never
  runs, or one element is collected on the first attempt.
- Nothing else in the diff changed since round 1's read.

### Mutation sensitivity

| Mutation | Result |
| --- | --- |
| revert `_randomArray.ts` only | `test_random_unique_items_domain` fails on its first row; the adjacent `test_random_string_pattern_*` rows still pass |
| revert both `_isFormatTime.ts` copies only | `test_openapi_validator_format_parity` fails at `time accepts "00:00:00.1234567890Z" directly`; the other 305 rows still pass |

Both were run in this checkout and restored from the pushed commit rather than
in a disposable clone: a linked worktree cannot resolve the workspace packages it
would have to mutate without a second full install and a cold Go plugin build.
Recorded as the deviation it is.

### Recorded, not fixed here

Seven observations that belong to the next discovery round rather than this
cycle are in the campaign ledger — among them that `benchmark/**` has no lane,
that `pnpm format` runs in none, that `.github/workflows/release-guard.cjs` is
named by no lane's paths, and that nothing enforces byte-identity between the 26
helpers duplicated across `typia` and `@typia/utils`.

Waiting on this head's CI before any merge decision.

<sub>(The first posting of this review lost the code spans in the two tables to
shell expansion; this is the same text, corrected.)</sub>

samchon (commented):
**`7e3b1c9ede`** — fix(random): stop a unique draw once its domain stops answering

Self-Review round 2 correction. Keeping what the unique loop collected removed the throw but moved the retry budget onto the success path, where nesting multiplies it. A stale-run break of 64 duplicates, applied only above the floor, bounds it: `Array<Array<boolean> & UniqueItems> & UniqueItems` went from 24.5 ms to 0.4 ms per draw, and one level deeper from 108 ms to 1.4 ms. Also corrected three prose claims the previous commit had falsified. Refs #2309, #2310.

samchon (commented):
**`b52855eab9`** — test(random): say what the wide-domain control actually catches

Commit-reader correction. The control's comment claimed to be the only row a uniform clamp would fail; since the positive rows now require their whole domain, a clamp to the floor fails five of them too. What the control alone catches is a clamp on a domain wider than any count the generator draws. Comment only. Refs #2309.

samchon (commented):
**`dfbff790fe`** — test(random): name the condition under which a row catches a clamp

Third reading of the same four lines by the commit reader, second correction. A positive row catches a clamp to the floor only where its floor sits *below* its domain; `literal`, `exact` and `single` put the floor at the domain, so a clamped draw still reaches it. Six of the nine rows carry the property, and the previous wording both named the wrong condition and excluded `capped`, which does carry it. Comment only. Refs #2309.

samchon (commented):
## Self-Review round 4 — clean

Complete pass over `origin/master...dfbff790fe`, 16 files, 398 insertions and 96
deletions. Nothing found that survives verification, so this is the clean round.

The only change since round 3 is the CONTROL comment in
`test_random_unique_items_domain.ts`, corrected twice after readers showed each
version false in turn. Its final form was checked clause by clause against the
source: a positive row catches a clamp to the floor exactly where `floor <
domain`, which is six of the nine rows (`boolean`, `enumeration`, `bounded
integer`, `narrow integer with a floor`, `arrays of a small domain`, `window
inside the domain`); the other three put the floor at the domain and prove
nothing about clamping. `Wide`'s domain is 101 values (`Type<"uint32">` emits
`{type: "integer", minimum: 0}`, and `_randomInteger` defaults the maximum to
`minimum + 100`) against a count window of `[1, 6]`, so it is the only row whose
domain no drawn count can exhaust.

### Gates

| Gate | Result |
| --- | --- |
| CI `test` (`pnpm test`) | pass, 16m51s |
| CI `build` (`pnpm run build`) | pass, 2m44s |
| CI `experiments` (tarballs, smoke, ESM, MCP) | pass, 2m59s |
| CI `website` (includes `build:examples`) | pass, 11m58s |
| CI `spell-check` | pass |
| local root `pnpm test` | `TEST_EXIT=0`, 12 workspaces `Success`, both Go trees `ok` |
| local `pnpm --filter ./examples --fail-if-no-match run build` | exit 0 |

### Integration gate

`origin/master` is still `ee9609d2a8` and `HEAD` is a fast-forward of it, so the
prospective merge result **is** this head. The frozen manifest's canonical set —
`pnpm install`, `pnpm build`, `pnpm test`, both Go trees — ran on exactly this
tree, locally and in CI. No packaging, `exports`, descriptor, or CLI change is in
the diff, so the clean-consumer gate is not triggered; `experiments` ran anyway
and passed.

### Mutation sensitivity

| Mutation | Result |
| --- | --- |
| revert `_randomArray.ts` only | `test_random_unique_items_domain` fails on its first row; adjacent `test_random_string_pattern_*` rows still pass |
| revert both `_isFormatTime.ts` copies only | `test_openapi_validator_format_parity` fails at the fractional-second row; the other 305 rows still pass |

### Closing keywords

`git log origin/master..HEAD` carries exactly four `Close #` lines — #2309,
#2310, #2312, #2313 — and no revert, so the squash closes exactly the four
issues whose fixes survive at `HEAD`. #2311 carries no closing line anywhere; it
was closed by hand as invalid with its disproof.

Ready to merge.

Issue #2317, opened 2026-07-22
Format<"idn-hostname"> rejects valid hosts hostname accepts, including the single-label unicode host it exists for


samchon:
## Problem

`Format<"idn-hostname">` rejects valid host names, including the single-label
unicode host it exists to accept. Its validator is not a superset of
`Format<"hostname">`, though every ASCII host name is by definition a valid
internationalized host name — IDNA extends the DNS name grammar with non-ASCII
labels, it removes nothing.

Two whole classes are rejected:

1. **Single-label hosts** — `hostname`, `localhost`, `a`, and the unicode
   `실례` — because the pattern requires at least one dot.
2. **Hosts whose final label is a single character** — `a.b`, `x.y.z`,
   `a1.b2` — because the pattern requires the last label to be two or more
   characters.

Both are accepted by `Format<"hostname">`, so the same string is valid as a
plain host name and invalid as an internationalized one.

## Evidence

Reproduced at `0859d11352` (Node v22.21.0) through the real transform. The ASCII
host-name validator is the authoritative oracle: every string it accepts is a
valid IDN host name, so any such string the IDN validator rejects is a defect,
with no external specification needed.

```ts
import { _isFormatHostname } from "typia/lib/internal/_isFormatHostname";
import { _isFormatIdnHostname } from "typia/lib/internal/_isFormatIdnHostname";
```

| value | `hostname` | `idn-hostname` | superset holds? |
| --- | --- | --- | --- |
| `hostname` | ✅ true | ❌ false | **no** |
| `localhost` | ✅ true | ❌ false | **no** |
| `a` | ✅ true | ❌ false | **no** |
| `a.b` | ✅ true | ❌ false | **no** |
| `x.y.z` | ✅ true | ❌ false | **no** |
| `a1.b2` | ✅ true | ❌ false | **no** |
| `example.com` | ✅ true | ✅ true | yes |
| `www.example.com` | ✅ true | ✅ true | yes |

Six of nine ASCII hosts that `hostname` accepts, `idn-hostname` rejects. All
four ways to reach the format agree on the wrong answer:

```
"hostname"  direct=false tagged=false comment=false
"a.b"       direct=false tagged=false comment=false
"실례"       direct=false tagged=false comment=false   ← a valid single-label IDN host
"x.y.z"     direct=false tagged=false comment=false
```

`실례` is the sharpest case: a single unicode label is exactly what
`idn-hostname` exists to validate, and it is rejected.

### Root cause and owners

[...]

Issue #2316, opened 2026-07-22
benchmark/ is compiled by no workflow, on any event


samchon:
## Problem

`benchmark/` is a compiled workspace of 893 tracked files — 782 of them under
`benchmark/src` — that no workflow compiles. A change to `packages/typia/src`,
`packages/typia/native/**`, `packages/interface`, `packages/utils`, or the
benchmark's own sources can break every generator in it and every check stays
green.

Unlike `examples/`, which `website.yml` reaches through its build chain, nothing
reaches `benchmark/` at all — not on a pull request, not on a master push, not
on a release.

## Evidence

Traced at `0859d11352` by reading every `run:` line in every workflow and
following each one into the scripts it invokes, rather than by grepping the
workflow files for the directory name:

| Lane | Command | Reaches `benchmark/`? |
| --- | --- | --- |
| `build.yml` | `pnpm run build` → `pnpm --filter=./packages/* --fail-if-no-match -r build` | no — `./packages/*` only |
| `test.yml` | `pnpm run test` → `test:packages` (`./tests/test-*`) + `test:toolchain` (both Go trees) | no |
| `experiments.yml` | `pnpm run package:tgz`, then `experiments/{smoke,esm,mcp}` install/build/test | no |
| `website.yml` | `pnpm run build`, `pnpm run package:tgz`, `website: test:compiler`, `go test ./cmd/playground`, `website: pnpm run build` → `build:prepare` → `build:compiler`, `build:blog:rss`, `build:examples` → `next build` → `build:finalize` | no — the chain reaches `examples/`, never `benchmark/` |
| `ttsc-cache.yml` | `pnpm --filter ./examples run build` | no |
| `spell-check.yml` | `crate-ci/typos` | prose only |
| `release.yml` | `release-guard.cjs`, `pnpm install`, `pnpm run build`, `package:tgz`, website build | no |

No `package.json` in the repository invokes the benchmark from another
workspace, and `website/package.json` has no script referencing it.

The workspace compiles today, so this is a missing gate rather than a live
break:

```
$ pnpm --filter ./benchmark --fail-if-no-match run build
> @typia/benchmark@13.2.0 build
> rimraf bin && ttsc build
BENCH_EXIT=0
```

`benchmark/src` is compiled with the real transform — `benchmark/tsconfig.json`
carries the typia plugin — so the build is a genuine type-check plus transform
run across the validate, stringify, parse, clone, prune, and protobuf program
[...]

Issue #2315, opened 2026-07-22
pnpm format rewrites committed doc comments into broken prose, and no lane runs it


samchon:
## Problem

Running the repository's own mandated formatting command corrupts committed
prose. `pnpm format` deletes the space next to a code span whose content
contains a backtick, gluing the span to the word or punctuation beside it.

`.agents/skills/development/SKILL.md` states the rule: *"Run `pnpm format`
before every ordinary commit and stage the result; never commit unformatted
output."* Following it today rewrites two committed doc comments into text that
reads wrong, and no lane runs the command, so the divergence is invisible until
a contributor obeys the rule.

## Evidence

Minimal reproduction at `0859d11352`, root `prettier.config.js` (which loads
`prettier-plugin-jsdoc`):

```ts
/**
 * Verifies the formatter keeps prose spacing around a code span.
 *
 * A span whose content holds a backtick `json\n`json` — the dash keeps its
 * space, and a call spelled `includes("`json")` keeps the spaces around it.
 *
 * 1. Read it before.
 * 2. Read it after.
 */
export const test_probe = (): void => {};
```

`pnpm exec prettier --config prettier.config.js probe.ts` returns:

```diff
- * A span whose content holds a backtick `json\n`json` — the dash keeps its
- * space, and a call spelled `includes("`json")` keeps the spaces around it.
+ * A span whose content holds a backtick `json\n`json`— the dash keeps its
+ * space, and a call spelled`includes("`json")` keeps the spaces around it.
```

Two spaces are deleted: the one before the em dash, and the one before the
second span.

### Files already in that state on `master`

An audit of every tracked `*.ts` blob against the config that actually applies
to it — root `prettier.config.js` for `packages/**` and `tests/**`,
`benchmark/prettier.config.js` for `benchmark/**`, `website/prettier.config.js`
for `website/**`, honoring `.prettierignore` — finds exactly two files that
`pnpm format` would rewrite:

| File | What `pnpm format` would do |
| --- | --- |
| `tests/test-langchain/src/features/test_langchain_tool_error_single_json_fence.ts` | `` `json\n`json` — broken markdown `` → `` `json\n`json`— broken markdown ``, and `` merely `includes("`json")` passes `` → ``merely`includes("`json")`passes`` |
[...]

PR #2318, opened 2026-07-22
fix: widen idn-hostname to a superset of hostname, and close two CI gate gaps

samchon:
Cycle 2 of `campaign-2026-07-22`. Owns the **complete accepted issue set** of the
cycle.

## Cycle issues

| Issue | Subject | Landed by |
| --- | --- | --- |
| #2317 | `Format<"idn-hostname">` rejects valid hosts `hostname` accepts, including the single-label unicode host it exists for | `a61b70c` |
| #2315 | `pnpm format` rewrites committed doc comments into broken prose, and no lane runs it | `8029ea2` |
| #2316 | `benchmark/` is compiled by no workflow, on any event | `ffbf584` |

## What changed

**`idn-hostname`** is now a superset of `hostname`, in all three owners (runtime
helper, `FormatCheatSheet` inline expression, `@typia/utils` copy). It accepts
the single-label hosts and single-character-final-label hosts the ASCII
validator already accepts — including `실례`, the single unicode label the
format exists for — while still rejecting `-a`, `a..b`, over-length labels, and
underscores. The ASCII `hostname` validator is the in-repository oracle, so no
external spec is invoked. `test_openapi_validator_format_parity` gains both
formats across all four reach paths plus a direct superset assertion.

**`pnpm format`** no longer corrupts prose. Two committed doc comments wrapped a
backtick inside a single-backtick markdown span, which `prettier-plugin-jsdoc`
mangles by deleting the adjacent space; both are rewritten to say the same thing
without such a span, and `prettier --check` confirms they are now stable. A new
`format.yml` lane runs `pnpm run format` and fails on any diff, so the tree and
the formatter can no longer disagree unnoticed.

**`benchmark/`** is compiled on every pull request by `build.yml`. It was
compiled by no lane on any event, so a transform, interface, or utils change
could break every benchmark program and stay green. Proven sensitive: the
compile exits 0 intact and exits 2 (`TS2322`) with one benchmark source given a
type error. The lane also gains the ttsc plugin-cache restore its siblings have.

## Verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-utils start` | Success (idn-hostname/hostname parity, 4 paths + superset) |
| `pnpm test:go:native` | pass (native cheat sheet) |
| `pnpm --filter ./tests/test-langchain start` / `test-vercel start` | Success (fence suites, comment-only control) |
| `pnpm --filter ./benchmark --fail-if-no-match run build` | exit 0; exit 2 on a mutated source |
| `pnpm run format` then diff | no real change (CRLF churn only, absent on the runner) |
| root `pnpm test`, build, benchmark, format gate | this pull request's CI + a local integration run |

Base commit `0859d11352`. Each commit carries its own `Close #` line; the body
carries none.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,8 +48,10 @@ jobs:
       # `benchmark/` is a compiled workspace of ~780 sources that no lane
       # compiled, so a transform, interface, or utils change could break every
       # benchmark program and stay green until someone next ran a benchmark
-      # (issue #2316). `pnpm run build` above warms the plugin cache, so this is
-      # just the compile. `--fail-if-no-match` because a `--filter` selecting
-      # nothing exits 0, which would report success without compiling anything.
+      # (issue #2316). It is the only workspace here that applies the typia
+      # transform, so this step is what builds the Go plugin -- the restore
+      # above lets it reuse a prior run's cache instead of a cold build.
+      # `--fail-if-no-match` because a `--filter` selecting nothing exits 0,
+      # which would report success without compiling anything.
       - name: Build Benchmark
         run: pnpm --filter ./benchmark --fail-if-no-match run build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,14 @@
 name: build
-on: 
+on:
   pull_request:
     paths:
       - '.github/workflows/build.yml'
+      - 'benchmark/**'
       - 'config/**'
       - 'packages/**'
       # `packages/{langchain,mcp}`'s build shells out to
       # `scripts/create-conditional-declarations.cjs`, so `pnpm run build`
-      # exercises it. `scripts/format-go.cjs` is reached only by
-      # `pnpm run format`, which no lane runs, so this trigger covers one of the
-      # two files rather than the directory.
+      # exercises it. `scripts/format-go.cjs` is run by `format.yml`.
       - 'scripts/**'
       - 'pnpm-lock.yaml'
       - 'pnpm-workspace.yaml'
@@ -26,9 +25,31 @@ jobs:
         with:
           go-version: 1.26.x
       - uses: pnpm/action-setup@v4
-      
+
       - name: Install Dependencies
         run: pnpm install
 
+      # ttsc compiles the typia Go plugin on first use into
+      # node_modules/.cache/ttsc, keyed by content (plugin source + ttsc/tsgo
+      # versions). Restoring a stale cache is safe: a changed source misses the
+      # binary key but still reuses the incremental go-build objects. Restored
+      # after `pnpm install` so an install-time node_modules rebuild cannot
+      # discard it.
+      - name: Restore ttsc Plugin Cache
+        uses: actions/cache@v4
+        with:
+          path: node_modules/.cache/ttsc
+          key: ttsc-cache-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml', 'packages/typia/native/**') }}
+          restore-keys: ttsc-cache-${{ runner.os }}-
+
       - name: Build Packages
         run: pnpm run build
+
+      # `benchmark/` is a compiled workspace of ~780 sources that no lane
+      # compiled, so a transform, interface, or utils change could break every
+      # benchmark program and stay green until someone next ran a benchmark
+      # (issue #2316). `pnpm run build` above warms the plugin cache, so this is
+      # just the compile. `--fail-if-no-match` because a `--filter` selecting
+      # nothing exits 0, which would report success without compiling anything.
+      - name: Build Benchmark
+        run: pnpm --filter ./benchmark --fail-if-no-match run build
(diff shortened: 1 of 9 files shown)


samchon (commented):
**`a61b70c6dc`** — fix(format): make idn-hostname a superset of hostname

Resolves #2317. All three owners widened to hostname's structure + the U+00A1..U+FFFF label range, so idn-hostname now accepts every value hostname does (single labels, single-char final labels, and the single unicode label the format exists for). Parity test covers direct+tagged+comment+OpenApiValidator + a direct superset assertion. Combining-mark over-permissiveness left out of scope, recorded on the issue.

samchon (commented):
**`8029ea2ddc`** — ci(format): rewrite two doc comments the formatter corrupts, and gate the tree

Resolves #2315. The two fence-counting test comments wrapped a backtick inside a single-backtick span, which prettier-plugin-jsdoc mangles; rewritten to be readable AND format-stable (prettier --check passes). New format.yml runs pnpm format and fails on any diff. Comment-only; both fence suites pass.

samchon (commented):
**`ffbf5849a1`** — ci(build): compile the benchmark workspace on pull requests

Resolves #2316. build.yml gains benchmark/** and a Build Benchmark step (plus the ttsc cache restore its siblings have). ~780 sources that no lane compiled. Proven sensitive: exit 0 intact, exit 2 on a mutated source.

samchon (commented):
## Self-Review round 1 — consequence surface of #2317

Tracing the idn-hostname widening through the local integration `pnpm test` caught a fixture consequence CI would otherwise have shown red: `test_validate_TypeTagFormat` in `test-utils-automated` failed on `[ '$input.idnHostname' ]`.

The `TypeTagFormat` template spoiler set `idnHostname` to `"깃허브"` — a single unicode label — as its "invalid" case. #2317 made a single label a **valid** IDN host (the thing idn-hostname exists for), so the validator correctly stopped flagging it, and the spoiler's expectation broke. That is the fix working. The spoiler moved to `"-깃허브"` (a leading hyphen is not a legal label start), which stays invalid under the corrected validator across all three owners. `CommentTagFormat` reuses `TypeTagFormat.SPOILERS`, so the one change covers both automated suites. Fixed in `4587bd7`.

Also folded in the two corrections the commit readers raised on the CI commits (`79d7554`): the benchmark step comment claimed `pnpm run build` warms the plugin cache, but no package applies the transform — the benchmark step builds the plugin itself; and `format.yml` now triggers on `scripts/**` so a `format-go.cjs` edit re-runs the gate.

**Recorded, not fixed here:** the widening makes `idn-hostname & MaxLength<3>` satisfiable (`abc` is a valid single-label host) while `_randomFormatIdnHostname` still cannot produce a host that short — a new instance of the #2284 class. Not a regression (`test_random_string_pattern_format_length`'s `assertThrows` for that type still throws; `test-typia-schema` passes), so it is a seed for the next round rather than cycle-2 scope.

Verification so far: `test-utils`, `test-typia-schema`, `test-langchain`/`test-vercel` fence suites, `test:go:native`, and `test-utils-automated` (after the spoiler fix) all pass; `test-typia-automated` re-running; benchmark compile exits 0 intact / 2 mutated; whole-repo `pnpm format` produces no real change. Awaiting the settled-head CI.

samchon (commented):
## Self-Review round 2 — clean

One complete pass over `origin/master...79d7554`, all 9 files and 353 diff lines read whole. Nothing to improve.

**Whole surface.** Both `_isFormatIdnHostname.ts` copies are byte-identical (same blob hash before and after), and the `FormatCheatSheet.go` entry un-escapes to the same regex — the three owners cannot drift. The parity matrix's `hostname`/`idn-hostname` rows, schema-keyword assertions, and the direct superset loop are sound; `a_b` and `실례` are correctly invalid for ASCII hostname, and every idn valid is accepted. The two fence-comment rewrites are faithful and format-stable. `build.yml`/`format.yml` reviewed (the `on:` trailing-whitespace tidy is incidental).

**Consequence surface traced beyond the diff:**
- `tests/template` idn-hostname spoiler — fixed in `4587bd7`; both automated suites pass.
- `_randomFormatIdnHostname` — untouched; its multi-label output stays valid under the widened validator, and `test_random_string_pattern_format_length` still passes (its `MaxLength<3>` `assertThrows` still throws; recorded as a next-round seed, not a regression).
- `OpenApiTypeChecker.covers` — `test_json_schema_type_checker_cover_string` already declares `idn-hostname covers hostname` statically; the runtime fix now **aligns** the validator with that pre-existing covering claim rather than contradicting it. Unaffected and more consistent.
- `OpenApi*.d.ts` format enums — generated, unaffected.

**Mutation sensitivity.** Reverting the idn-hostname regex to master's in all three owners (test kept) fails `test_openapi_validator_format_parity` at `idn-hostname accepts "hostname" directly`; benchmark gate exits 0 intact / 2 on a mutated source; the format gate is self-proving.

**Local evidence (all green on `79d7554`):** test-utils, test-typia-schema, test-langchain, test-vercel, test-utils-automated, test-typia-automated, test:go:public, test:go:native; benchmark compile; whole-repo `pnpm format` no-op. Awaiting the settled-head CI before merge.

Issue #2319, opened 2026-07-23
typia.random throws for a satisfiable short idn-hostname, and its generator comment describes the pre-#2317 grammar


samchon:
## Problem

`typia.random<T>()` throws for a satisfiable `idn-hostname` type, and its
generator's comment describes a grammar the validator no longer has.

#2317 made `Format<"idn-hostname">` a superset of `Format<"hostname">`: a single
label is now a valid IDN host, so `"a"`, `"ab"`, `"abc"` are valid. But
`_randomFormatIdnHostname` was not updated. It still assumes the old grammar — a
mandatory dot and a two-plus character trailing label — so it refuses every
length below four and never produces the single-label forms the validator now
accepts.

The comparison with `hostname`, whose generator was already correct, is the
proof: the two formats now share one structure, yet `typia.random` handles the
same short type on one side and throws on the other.

## Evidence

At `f6415ce86e` (Node v22.21.0) through the real transform. `sample` is a
hand-written value that the type's own `is` accepts, proving the type is
satisfiable:

```
idn-hostname & MaxLength<3>   satisfiable("abc")=true   random => THREW: unable to generate a random idn hostname
idn-hostname & MaxLength<2>   satisfiable("ab")=true    random => THREW: unable to generate a random idn hostname
idn-hostname & length==1      satisfiable("a")=true     random => THREW: unable to generate a random idn hostname
hostname & MaxLength<3>       satisfiable("abc")=true   random => produced "b"
```

`hostname & MaxLength<3>` produces `"b"` — a valid short single-label host — so
the capability exists in the sibling generator. `idn-hostname & MaxLength<3>`
throws for the same class of type, though every ASCII host is a valid IDN host
and `"abc"` is a valid IDN host directly.

### Root cause

`packages/typia/src/internal/_randomFormatIdnHostname.ts`:

```ts
const MIN_TOTAL = 4; // the shortest idn hostname, `x.yy`: label(1) + `.` + tld(2)
```

and its header comment:

```ts
// `_isFormatIdnHostname` requires at least one label, a dot, and a >=2-char
// TLD (`(…\.)+[a-z¡-￿]{2,}`), none of which the non-idn hostname generator
// guarantees on its length path (it can emit a single dotless label). Reserve
// a fixed `.<2-char TLD>` and realize the rest as dot-joined leading labels.
```

Both describe the pre-#2317 regex `(…\.)+[a-z¡-￿]{2,}`. The current validator is
[...]

PR #2320, opened 2026-07-23
fix(random): draw a short idn-hostname the widened validator accepts

samchon:
Cycle 3 of `campaign-2026-07-22`. Owns #2319 — the follow-on to #2317 where `_randomFormatIdnHostname` kept the pre-widening grammar and threw for the single-label short hosts the validator now accepts.

## What changed
- `_randomFormatIdnHostname` drops the stale `MIN_TOTAL = 4` and the reserved `.<2-char TLD>` suffix; it now realizes its length exactly as `_randomFormatHostname` does (floor 1, `_randomHostnameLabels(pickLength)`), so a single label of 1..63 chars is produced. Its header comment described the pre-#2317 grammar; it now describes the current one.
- `test_random_string_pattern_format_length`: the `MaxLength<3>` `assertThrows` becomes a positive `roundTrip` (lengths 1..3), joined by an exact single-character case; the unsatisfiable slot moves to `MaxLength<0>`.
- `_randomHostnameLabels`'s shared-helper doc no longer claims the idn generator reserves a `.<tld>` suffix (a comment-rot follow-on the commit reader caught).

## Verification
| Command | Result |
| --- | --- |
| root `pnpm test` | TEST=0 (12 workspaces + both Go trees) |
| `test-typia-schema` random suites | Success; short 1..3, exact 1, 10..40, multi-label, unconstrained all draw `is`-accepted values; `MaxLength<0>` throws |
| mutation | reverting the generator to master fails the short-single-label `roundTrip` (`unable to generate…`) |

Base `f6415ce`. Close-line carried by the commit. CI pending.

### DIFF
--- a/packages/typia/src/internal/_randomFormatHostname.ts
+++ b/packages/typia/src/internal/_randomFormatHostname.ts
@@ -29,8 +29,8 @@ const pickLength = (props: _ILengthProps): number => {
 
 /**
  * Builds dot-joined hostname labels totaling exactly `length` characters, each
- * label within 63 chars. Shared with the idn-hostname generator, which reserves
- * a fixed `.<tld>` suffix and realizes the rest as these leading labels.
+ * label within 63 chars. Shared with the idn-hostname generator, which realizes
+ * its length the same way since #2317 gave the two formats one structure.
  */
 export const _randomHostnameLabels = (length: number): string => {
   const parts: string[] = [];
--- a/packages/typia/src/internal/_randomFormatIdnHostname.ts
+++ b/packages/typia/src/internal/_randomFormatIdnHostname.ts
@@ -6,20 +6,19 @@ import { _ILengthProps } from "./_randomStringLength";
 export const _randomFormatIdnHostname = (props?: _ILengthProps): string => {
   if (props?.minLength === undefined && props?.maxLength === undefined)
     return `${random(10)}.${random(3)}`;
-  // `_isFormatIdnHostname` requires at least one label, a dot, and a >=2-char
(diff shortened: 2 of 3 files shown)


samchon (commented):
## Self-Review round 1 — clean, with one follow-on applied

One complete pass over `origin/master...HEAD`. The generator change is minimal and correct; one comment-rot follow-on was applied.

- **Correctness:** `_randomHostnameLabels` emits only `a-z` labels (per `_randomString`'s alphabet), all inside the idn validator's class; single labels at 1..3, exact 1, mid windows, and multi-label long hosts all produce `is`-accepted values. `MaxLength<0>` throws; unconstrained and `minLength>maxLength` behave. No dead imports.
- **Mutation sensitivity:** reverting `_randomFormatIdnHostname` to master (test kept) fails the new short-single-label `roundTrip` at `unable to generate a random idn hostname…`, so the positive case is bound to the fix.
- **Follow-on caught and fixed (`9cfb1d1`):** the shared `_randomHostnameLabels` JSDoc still said the idn generator "reserves a fixed `.<tld>` suffix" — false after this fix, the same comment-rot class the fix itself removes. Corrected.
- **Duplication note (not a defect):** `_randomFormatIdnHostname`'s constrained path is now behaviorally identical to `_randomFormatHostname`'s. Kept as its own function to match the codebase's one-`_randomFormat*`-per-format structure and to leave room for emitting unicode labels later; only the throw message differs.

Whole `pnpm test` is TEST=0. Awaiting the settled-head CI before merge.

Issue #2331, opened 2026-08-12
Error when self-recursive union type comes?


samchon:
Reported by @kakasoo 

```typescript
import typia, { Primitive, JSONValue } from 'typia';
type Node = Date | Node[];

typia.createIs<Primitive<Node>>();
```

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3nAClUYGYANwFNMApAZQHkA5ANVQBsBXMuAXzgDMoIIOAHIkKVMIDcAKDGd6EACacAvHAAiqGJwA%2BcBcoDaAXRmzkaAHQBjKGS1kAkgGcAPPkLFyrg2QB8fgAUAJSSQA

samchon:
```typescript
import typia, { Primitive } from 'typia';

interface JsonObject { [key: string]: JsonValue }
type JsonValue = string | number | boolean | null | JsonObject | JsonValue[];

interface Output { at: Date; payload: JsonValue }

typia.createIs<Primitive<Output>>();
```

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3nAClUYGYANwFM4BfOAMyghDgHIkVUmBuAKC%2BADsYZKDVQBjCgCkAzhD4B5AEYArMqPg4A2gGsyiAFxwpMAnwDmAXQPTZANVQAbAK4VKXVpJl87TigF5DxvymcAA%2BcHyOIApCoXAKEBD2ZKh8sRH29rHW8sqq8GHZ3s4a5ty8AkIi4nByjjBgddhwqDAGACItZBxwYKiI9hCoACZWnkUuPKxoAHSiUMmCAJJSADz4hMTkK7X1dQB8ewAUAJQcQA

PR #2332, opened 2026-08-12
fix(native): terminate type-name composition on a self-recursive type graph

samchon:
## Problem

```ts
import typia, { Primitive } from "typia";

type Node = Date | Node[];
typia.createIs<Primitive<Node>>();
```

killed the plugin process:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

No diagnostic, no emit, and `ttsc` taken down with it. Reported in #2331 by @kakasoo, with a second witness where the recursion sits behind an object property (`{ at: Date; payload: JsonValue }`).

## Cause

`Primitive` is a witness, not the cause. Any self-recursive conditional alias instantiates to a union whose array member is a reference whose sole type argument is that very union:

```
Array<X>  ──generic argument──▶  X = string | Array<X>  ──union member──▶  Array<X>
```

`metadataCollection_getFullName` walks exactly that path — union members and generic arguments — with no visited set, so it never terminates. An instrumented plugin measured it: `Iterate_metadata` entered **3 times, maximum depth 2**, while the name builder entered **2,013,074 times** before the stack gave out, revisiting the *same three type pointers* in a loop.

`checker.TypeToString` returned normally on every one of those calls — it elides past a depth bound — so this is typia's recursion, not a compiler defect. A control that declaration-emits the same `Primitive<Node>` with no typia call transforms fine.

## Fix

The walk carries the types whose name is still being composed on the current path and names a revisited one after its own symbol with its arguments unexpanded — the recursion's placeholder, `Array<string | Array>`. The set is per path, not per analysis: every mark is removed on the way out, so a diamond that legitimately reaches one type twice still expands both times.

Naming a revisited type after the checker's elided rendering also terminates, and was the first version, but it named the JSON schema component with a 400-character structural dump. The shipped rule gives `ArrayArraystringFormatdate-time`, and the schema still self-references correctly.

## Scope

`TypeFactory.GetFullName` has the same recursion shape and was guarded in the first commit, then removed in `d0dc679` because Self-Review could not reach it: every caller passes the printed type-argument text as the name, and the one `Name: nil` path hands it a type it returns from without recursing. `metadata_type_full_name` descends only into union and intersection members, which TypeScript flattens, so no cycle can form there. Reasoning for both is on the review thread.

## Verification

- **Regression:** `TestRecursiveConditionalAliasNameTransform` transforms `is` / `assert` / `json.schema` over `Primitive<Node>`, the bare `Rec<T>` alias, and the recursive-`JsonValue` object, beside controls for a named recursive union, `type A = A[]`, and a recursive interface. It executes the emitted validators over positive, negative, nested, 64-deep, self-referencing-value and boundary inputs, requires the schema to self-reference under a bounded name, and requires two transforms of the same source to emit identical bytes.
- **Public surface:** `test_validate_recursive_conditional_alias` runs the same type through the real package — `is`, `validate`, and `json.schema`.
- **Mutation:** reverting only the product file while retaining both tests reproduces `fatal error: stack overflow`. The regression is sensitive to the fix.
- **Consequence surface:** `is`, `assert`, `validate`, `json.stringify`, `json.isParse`, `json.schema`, `random`, `plain.clone`, `plain.classify`, `compare.equals`, `reflect.name`, `reflect.schema` all transform over the recursive type. `json.application` and `notations.camel` still reject it, correctly and by their own contracts.
- **Commands:** `pnpm test:go:native` (targeted), `pnpm --filter ./tests/test-typia-schema start`, root `pnpm test`, and this pull request's CI.

## Deferred

`metadata_type_symbol_name` in `MetadataHelper.go` has no caller but itself — the Map/Set guards use the non-recursive `metadata_type_symbol_base_name`. Noted during review, unrelated to this issue, not touched.

## Not changed

No exported symbol, CLI flag, plugin descriptor shape, or documented behavior. No package version, per the campaign freeze.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/recursive_conditional_alias_name_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/recursive_conditional_alias_name_transform_test.go
@@ -1,0 +1,271 @@
+package main
+
+import (
+  "os"
+  "os/exec"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestRecursiveConditionalAliasNameTransform pins the transform over a type
+// whose resolved graph names itself (issue #2331).
+//
+// `Primitive<T>` instantiated at the self-recursive union `type Node = Date |
+// Node[]` resolves to `string & Format<"date-time"> | X[]`, where the array's
+// sole type argument is that very union. The name builders walked union member
+// -> array reference -> type argument and arrived back at the type whose name
+// was still being composed, so they recursed until the plugin process died of a
+// Go stack overflow: no diagnostic, no emit, and ttsc itself taken down with it.
+// `Primitive` is only one witness -- any self-recursive conditional alias
+// produces the same graph -- so the fixture pins the bare `Rec<T>` form too.
+//
+// The names the fix produces have to stay usable, not merely terminate: the
+// cycle's placeholder is the type's own symbol with its arguments unexpanded, so
+// a JSON schema component keeps a short key instead of being named after a
+// several-hundred-character structural dump of the elided rendering.
+//
+//  1. Transform is / assert / json.schema over `Primitive<Node>`, the bare
+//     recursive conditional alias, and the recursive-JsonValue object, beside
+//     controls (named recursive union, self-recursive array alias, recursive
+//     interface) that already transformed and must keep their names.
+//  2. Require the emit to carry a recursive array validator rather than an
+//     inlined expansion.
+//  3. Execute the emitted validators over positive, negative, nested and
+//     boundary values, and require the recursive schema to self-reference
+//     through a bounded component name.
+func TestRecursiveConditionalAliasNameTransform(t *testing.T) {
+  project := recursiveConditionalAliasNameProject(t)
+  out, errText, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 0 {
+    t.Fatalf("recursive conditional alias transform failed: code=%d stderr=\n%s", code, errText)
+  }
+  if strings.Contains(out, "_ia0") == false {
+    t.Fatalf("expected a recursive array validator in the emit, got:\n%s", out)
+  }
+  recursiveConditionalAliasNameRunRuntimeCases(t, project, out)
+}
+
+func recursiveConditionalAliasNameProject(t *testing.T) string {
+  t.Helper()
+  root := ttscTypiaTestRepoRoot(t)
+  base := filepath.Join(root, "packages", "typia", "native", ".tmp-ttsc-typia-tests")
+  if err := os.MkdirAll(base, 0o755); err != nil {
+    t.Fatalf("mkdir temp base: %v", err)
+  }
+  dir, err := os.MkdirTemp(base, "recursive-conditional-alias-")
+  if err != nil {
+    t.Fatalf("create temp fixture: %v", err)
+  }
+  t.Cleanup(func() {
+    _ = os.RemoveAll(dir)
+  })
+  src := filepath.Join(dir, "src")
+  if err := os.MkdirAll(src, 0o755); err != nil {
+    t.Fatalf("mkdir fixture src: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(dir, "tsconfig.json"), []byte(recursiveConditionalAliasNameTSConfig), 0o644); err != nil {
+    t.Fatalf("write tsconfig: %v", err)
+  }
+  if err := os.WriteFile(filepath.Join(src, "main.ts"), []byte(recursiveConditionalAliasNameSource), 0o644); err != nil {
+    t.Fatalf("write source: %v", err)
+  }
+  return dir
+}
+
+func recursiveConditionalAliasNameRunRuntimeCases(t *testing.T, project string, js string) {
+  t.Helper()
+  node, err := exec.LookPath("node")
+  if err != nil {
+    t.Skip("node executable not found")
+  }
+  runtimeDir := filepath.Join(project, "runtime")
+  if err := os.MkdirAll(runtimeDir, 0o755); err != nil {
+    t.Fatalf("mkdir runtime dir: %v", err)
+  }
+  ttscTypiaTestWriteCommonRuntimeStubs(t, runtimeDir)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "is-format-date-time-stub.cjs"), []byte(recursiveConditionalAliasNameFormatStub), 0o644); err != nil {
+    t.Fatalf("write format stub: %v", err)
+  }
+  js = strings.ReplaceAll(js, `require("typia/lib/internal/_isFormatDateTime")`, `require("./is-format-date-time-stub.cjs")`)
+  runtimeJS := ttscTypiaTestRewriteCommonJS(t, js)
+  if err := os.WriteFile(filepath.Join(runtimeDir, "main.cjs"), []byte(runtimeJS), 0o644); err != nil {
+    t.Fatalf("write runtime module: %v", err)
+  }
+  runner := filepath.Join(runtimeDir, "run.cjs")
+  if err := os.WriteFile(runner, []byte(recursiveConditionalAliasNameRuntimeRunner), 0o644); err != nil {
+    t.Fatalf("write runtime runner: %v", err)
+  }
+  cmd := exec.Command(node, runner)
+  cmd.Dir = runtimeDir
+  output, err := cmd.CombinedOutput()
+  if err != nil {
+    t.Fatalf("recursive conditional alias runtime cases failed: %v\n%s", err, output)
+  }
+}
+
+const recursiveConditionalAliasNameTSConfig = `{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "commonjs",
+    "moduleResolution": "bundler",
+    "ignoreDeprecations": "6.0",
+    "types": ["*"],
+    "esModuleInterop": true,
+    "strict": true,
+    "skipLibCheck": true
+  },
+  "include": ["src"]
+}
+`
+
+// The shipped helper validates the RFC 3339 date-time shape; the runtime cases
+// only need it to separate a date-time string from a plain one, and a stable
+// predicate keeps a failure reproducible.
+const recursiveConditionalAliasNameFormatStub = `module.exports._isFormatDateTime = (value) =>
+  /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?(Z|[+-]\d{2}:\d{2})$/.test(value);
+`
+
+const recursiveConditionalAliasNameSource = `import typia, { Primitive, tags } from "typia";
+
+// 1. the reported case: Primitive over a self-recursive union carrying Date.
+//    Primitive rewrites the Date arm to a date-time string and keeps the array
+//    arm pointing back at the whole union.
+type Node = Date | Node[];
+export const isPrimitiveNode = typia.createIs<Primitive<Node>>();
+export const assertPrimitiveNode = typia.createAssert<Primitive<Node>>();
+export const schemaPrimitiveNode = typia.json.schema<Primitive<Node>>();
+
+// 2. the same graph without Primitive: any self-recursive conditional alias
+//    instantiates to a union whose array member's type argument is that union
+type Rec<T> = T extends Date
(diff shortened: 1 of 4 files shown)


samchon (commented):
**Commit `a38dfa0` — fix(native): terminate type-name composition on a self-recursive type graph** (resolves #2331)

### What landed

| Path | Change |
| --- | --- |
| `core/schemas/metadata/MetadataCollection.go` | `metadataCollection_getFullName` carries a per-path visiting set through its union/intersection-member and generic-argument recursions |
| `core/factories/TypeFactory.go` | `TypeFactory.GetFullName` carries the same set through its union/intersection-member, `Promise` unwrap, and generic-argument recursions |
| `cmd/ttsc-typia/recursive_conditional_alias_name_transform_test.go` | new command-level transform + runtime regression |
| `tests/test-typia-schema/.../test_validate_recursive_conditional_alias.ts` | new public-surface regression through the real package |

### Root cause, as measured

Instrumented builds of the plugin, not inference:

- `Iterate_metadata` was entered **3 times, maximum depth 2** — the metadata walk was never the runaway.
- `metadataCollection_getFullName` was entered **2,013,074 times** before the stack gave out.
- Printing the type pointer at each entry showed a clean 3-cycle: `Array<X>` (`0x…616c0`) → `X = string | Array<X>` (`0x…66ea0`) → `string` (`0x…fee10`) → `Array<X>` (`0x…616c0`, same pointer).

So the revisited node is the *same* `*Type`, and a per-path visited set is sufficient. `checker.TypeToString` returned normally on all two million calls — it elides past a depth bound — which is why the crash showed typia frames beneath a runaway `NodeBuilderImpl` rather than a checker defect.

### Ownership boundary

`ttsc` / typescript-go is not at fault. A control that declaration-emits the same `Primitive<Node>` with no typia call transforms fine (`exit 0`), because the compiler keeps the alias reference instead of expanding it.

### Cases separated by control

| Fixture | Before | After |
| --- | --- | --- |
| `createIs<Primitive<Node>>()` (reported) | stack overflow | ok |
| `createIs<Primitive<Output>>()` (recursive `JsonValue`, second report) | stack overflow | ok |
| bare `Rec<T>` conditional alias, no `Primitive` | stack overflow | ok |
| `createAssert<Rec<Node>>()` | stack overflow | ok |
| `type N2 = (string & Format<"date-time">) \| N2[]` | ok | ok |
| `type A = A[]` | ok | ok |
| `type Node = Date \| Node[]` without `Primitive` | ok | ok |

### Naming choice

A revisited type is named after its own symbol with arguments unexpanded — `Array<string | Array>`. Falling back to the checker's elided rendering also terminates, and was the first version of this fix, but it named the JSON schema component

```
Array_x28__x28__x28__x28__x28_…_x29__x29_-x1D3ADEA7287FA2DC   (400+ chars)
```

The shipped rule gives `ArrayArraystringFormatdate-time`, and the emitted schema still self-references correctly:

```json
{"type":"array","items":{"oneOf":[{"type":"string","format":"date-time"},{"$ref":"#/components/schemas/ArrayArraystringFormatdate-time"}]}}
```

### Verification

- `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestRecursiveConditionalAliasNameTransform` — pass.
- `pnpm --filter ./tests/test-typia-schema start` — pass (all cases).
- Mutation: reverting **only** the two product files while retaining both tests reproduces `fatal error: stack overflow` after 112 s. The regression is sensitive to the fix.
- Consequence surface exercised over the recursive type: `is`, `assert`, `validate`, `json.stringify`, `json.isParse`, `json.schema`, `random`, `plain.clone`, `compare.equals`, `reflect.name`, `reflect.schema`. `json.application` and `notations.camel` still reject it, correctly and by their own contracts (union root is not a class/interface; notations require an object root).
- Root `pnpm test` and pull-request CI: running.

Self-Review over the full base-to-head diff follows.

samchon (commented):
**Self-Review round 1 — two findings, both applied**

### 1. `metadataCollection_getFullNameOf` duplicated the symbol selection (applied in `d0dc679`'s parent tree)

The first version added a `metadataCollection_nameSymbol` helper that repeated the alias-symbol-then-own-symbol rule the function body already implements, and it ran `checker.TypeToString` before the cycle check — computing the several-hundred-character rendering on exactly the path that discards it. The symbol selection now happens once, before the cycle check, and `rendered` is computed after it. Same behavior, one rule, no wasted rendering on the cycle path.

### 2. The `TypeFactory.GetFullName` guard was unreachable — removed in `d0dc679`

The first commit guarded `TypeFactory.GetFullName` too, on the reasoning that its recursion has the same shape. Self-Review required a reproduction and could not produce one.

Measured, not argued: reverting **only** `TypeFactory.go` to `master` while keeping the collection fix leaves the regression green (`ok … 2.8s`), and a probe running `createValidate` / `json.createStringify` / `json.createIsParse` / `createRandom` / `plain.createClone` / `plain.createClassify` / `compare.createEquals` over `Primitive<Node>` also passes with the guard gone.

The reason is structural:

| Path | Why it cannot cycle |
| --- | --- |
| `GenericTransformer.Factory` / `.Feature` | sets `Name` to the printed text of the type argument, so it is never nil at the top |
| `FeatureProgrammer.Compose` / `.Write` | forwards that same `props.Name` into both `Types.Output` call sites — the only two in the tree |
| `assertProgrammer_type_name`, `validateProgrammer_type_reference`, and peers | return the given name untouched when it is present |
| `llmApplicationProgrammer_writeValidator` (the one `Name: nil`) | passes the `any` type, or an LLM parameter object — and `GetFullName` descends into union/intersection members and generic arguments, never into object properties, so it returns immediately |

No Go test in this repository constructs a checker (the transform fixtures are the only route to a real type graph), so the branch could not be covered directly either. An untested branch guarding a path nothing reaches is a liability, not defence in depth, so it is gone. If a future caller drops the precomputed name, the same fix applies there with a reproduction to justify it.

### Round 1 also checked, no finding

- **Cycle-set lifetime** — the visiting set is per path, not per analysis: every marked type is deleted on the way out (explicitly in the union branch, `defer` in the generic branch), so a diamond that legitimately reaches one type twice through different paths still expands both times. Only a genuine cycle truncates.
- **Concurrency** — no shared mutable state added. The set is created per top-level call and never escapes it; `MetadataCollection` is per analysis and `Clone()` is untouched.
- **Determinism** — the set is a `map`, but nothing iterates it; output ordering still comes from the type's own member and argument slices, so repeat emits stay byte-identical.
- **Non-cyclic naming is untouched** — the new branch fires only when a type is already on the path, which previously could not return at all. No existing name can change.
- **`metadata_type_full_name` (`MetadataHelper.go`) is not affected** — it recurses only into `Types()` of unions and intersections, never into generic arguments. TypeScript flattens both, and any union whose member chain returned to itself would have to resolve eagerly, which the compiler rejects as a circular type alias. A cycle cannot form there, so it takes no guard.
- **`metadata_type_symbol_name` is dead code** — it has no caller but itself; the Map/Set guards use `metadata_type_symbol_base_name`, which does not recurse. Noted, not touched: removing it is unrelated to this issue.
- **Public contract, versions, packaging, docs** — no exported symbol, CLI flag, descriptor shape, or documented behavior changes; no version touched, per the campaign freeze. The change fixes a crash and adds no new documented surface.

samchon (commented):
**Self-Review round 2 — three findings, all applied (`9076030`, `8b08534`)**

Round 1's corrections changed the source, so this is a fresh full round over the whole base-to-head diff rather than a recheck of the fixes.

### 1. The regression never outran a hypothetical inline expansion

Value cases stopped at three levels of nesting. A validator that expanded the recursion instead of recursing through one function would still have passed them. Sixty-four levels now run positive and negative.

### 2. The primitified arm had no nested negative twin

`Primitive<Node>` accepts the *string* form of a `Date`, and the only case pinning that a real `Date` instance is rejected sat at the top level, where a different branch decides it. `[[new Date(0)]]` now covers it inside the recursion.

### 3. The schema assertions dereferenced `keys[0]` unguarded

An empty component map turned a clean `FAIL` on the count into a `TypeError` that skipped every assertion after it — the failure mode where a test reports the wrong thing. The block is now guarded.

### 4. Determinism was claimed in round 1, not measured (`8b08534`)

Round 1 recorded "map iteration order does not reach the output" as reasoning. The fix introduces a map whose result becomes both a helper identifier and a schema component key, so reasoning is the wrong instrument. The fixture now transforms the same source twice and requires identical bytes.

### Round 2 also checked, no finding

- **Cache interaction** — `MetadataCollection.getFullName` memoizes into `full_names_`, but only at the top-level entry; the truncated value a cycle returns is produced by the package-level recursion, which never writes that cache. No poisoning, and each type's name stays a pure function of that type.
- **Entry-point stability** — a type's name is always computed from itself as the root, so it does not depend on which sibling was analyzed first.
- **Non-cyclic naming** — the guard fires only for a type already on the current path, a state that previously could not return at all, so no existing name can change. The `ICategory` control pins that.
- **Temporary assets** — the fixture writes under `packages/typia/native/.tmp-ttsc-typia-tests/`, which `.gitignore` already covers, and removes itself through `t.Cleanup`.
- **Spelling** — no British spellings in the changed files (the `typos` lane splits camelCase and rejects them).
- **Merge base** — `origin/master` is still `e829c27` and is an ancestor of this head, so the prospective merge result is this head; CI here is the integration evidence rather than a proxy for it.

samchon (commented):
**Self-Review round 3 — two findings, both applied (`1ceb869`, `3282c88`)**

### 1. The determinism check repeated its own argument slice (`1ceb869`)

The two runs it compares were spelled out twice, so they could drift apart without the comparison noticing — a byte-for-byte assertion between two invocations that are no longer the same invocation proves nothing. One closure now serves both. The header also introduced `X` without saying what it stood for.

### 2. The named-recursion control did not do what its doc comment claimed (`3282c88`)

Step 3 read "compare against a named recursive union, which must keep its declared name", and the control only checked that the union still validated. Nothing read a name.

Closing that gap exposed the reason it mattered: writing the missing assertion, I guessed the count and asserted one component. `typia.json.schema<Named>()` emits **two** — `Named` for the alias and `ArrayNamed` for its array — so the guess failed the suite immediately:

```
Bug on named recursion emits one component: found different values - []
```

An expectation taken from what I assumed the implementation does is worth nothing even when it passes; this one was wrong and only surfaced because the test ran. The control now states the property under test instead of a count: every component this recursion emits is reached through the declaration, so every name must be built from it, and none may fall back to the cycle placeholder. That holds whether the emitter produces two components or some later number.

### Round 3 also checked, no finding

- Product file unchanged since `3f05526` and re-read in full: cycle set per path, symbol selected once, `rendered` computed only where used.
- Both regressions re-run green; root `pnpm test` completed on the branch (`exit 0`, no `FAIL` lines), including `cmd/ttsc-typia` at 99 s.
- `origin/master` is still `e829c27`, so the prospective merge result remains this head.

Round 4 follows over `3282c88`; the merge waits on it and on that head's CI.

samchon (commented):
**Self-Review round 4 — one finding applied (`4e51a22`, `92b4e9a`), round 5 clean**

### Round 4 finding: the native fixture had the same doc/assertion gap round 3 found in the TypeScript one

Its control comment claimed all three controls "must not take the cycle placeholder", and only the recursive interface's component name was read. A named recursive *union* is the closest control to the fixed case — same union-plus-array shape, only with a declaration to name it — so leaving its name unread left the most informative control silent. `schemaNamed` now asserts every component it emits is named from the declaration.

While applying it, the predicate in the TypeScript control was shadowing the outer component key with its own callback parameter; renamed (`92b4e9a`).

### Round 5: complete round over `92b4e9a`, nothing to improve

Product file, both regressions, and every hunk re-read. No finding.

---

## Merge gate

| Gate | Result |
| --- | --- |
| Required checks on `92b4e9a` | 4 × Ubuntu, build, spell-check, both Socket lanes — all **pass** |
| Head that CI ran on | `92b4e9a03f0f032cb8d5ec5099525f81d66e19fb`, identical to the merge candidate |
| `pnpm test:go` (public + native) at this head | pass, `cmd/ttsc-typia` 80.6 s |
| Root `pnpm test` on the branch | pass, exit 0, no `FAIL` lines |
| `pnpm --filter ./tests/test-typia-schema start` | pass |
| Mutation, re-run at **this** head | reverting only `MetadataCollection.go` to `master` → `fatal error: stack overflow` after 120.6 s; restoring → `ok 0.535s` |
| Integration gate | `origin/master` is still `e829c27` and an ancestor of this head, so the prospective merge result *is* this head and its CI is the integration evidence, not a proxy |
| Closing reconciliation | `git log origin/master..HEAD` carries exactly one closing line, `Close #2331`, from `a38dfa0`; the fix it names survives at `HEAD` — the mutation above proves the tests still depend on it |
| Working tree | clean |

Merging.

PR #2333, opened 2026-08-13
chore(native): remove the two unreferenced helpers

samchon:
## Intent

Remove dead code from the native tree. Follow-up to #2332, where `metadata_type_symbol_name` was noticed during review and left alone as unrelated to that issue.

## What is dead, and how it was established

A scan over `packages/typia/native` resolved every top-level unexported function's own body range and counted references outside it, ignoring mentions in comments. Exactly two functions have no external reference:

| Function | File | Why it survived until now |
| --- | --- | --- |
| `metadata_type_symbol_name` | `core/factories/internal/metadata/MetadataHelper.go` | Its only callers are its own two recursive calls, so a plain reference count never reads zero. Its doc comment also names it, which hides it from a comment-blind scan. |
| `factoryUtilityBooleanMetadata` | `core/factories/factory_utility_coverage_test.go` | One of a family of metadata builders in a `typia_native_internal` fixture; the siblings are used, this one never was. |

The rationale `metadata_type_symbol_name` carried — derive the name from the type's own symbol, never from `checker.TypeToString`, because an alias chain like `type MAP = Map<K, V>` makes `TypeToString` print the alias name and defeat the Map / Set prefix guards — is still true of `metadata_type_symbol_base_name`, which is what those guards actually call. The comment moves onto it rather than disappearing with the dead function.

## Commits

Split so the removal stays readable:

1. **`05a3d75`** — the removal. Eleven lines out of the fixture, twenty-eight out of `MetadataHelper.go`, five back as the relocated comment.
2. **`fbb4fae`** — indentation normalization of the fixture, alone. `scripts/format-go.cjs` formats the Go files a branch touches and rewrites gofmt's leading tabs to the repository's two spaces; this fixture still carried tabs, so the mandated `pnpm format` rewrites all 896 of its lines the moment anything touches it. `git diff -w` against its parent is empty.

## Verification

- No cascade: rescanning after the removal reports **zero** functions without an external reference, and no import became unused.
- `go build ../native/...` and `go vet` clean.
- `pnpm test:go` (public + native) and the `typia_native_internal` tagged suite both pass, the latter being the only lane that compiles the touched fixture at all — root `pnpm test:go` does not enable that tag.

## Not changed

No exported symbol, no behavior, no version. `TestFactoryUtilityCoverage` keeps every assertion it had.

### DIFF
--- a/packages/typia/native/core/factories/factory_utility_coverage_test.go
+++ b/packages/typia/native/core/factories/factory_utility_coverage_test.go
@@ -4,18 +4,18 @@
 package factories
 
 import (
-	"math/big"
-	"os"
-	"path/filepath"
-	"strings"
-	"testing"
+  "math/big"
+  "os"
+  "path/filepath"
+  "strings"
+  "testing"
 
-	shimast "github.com/microsoft/typescript-go/shim/ast"
-	shimchecker "github.com/microsoft/typescript-go/shim/checker"
-	shimprinter "github.com/microsoft/typescript-go/shim/printer"
-	"github.com/samchon/ttsc/packages/ttsc/driver"
-	schemametadata "github.com/samchon/typia/packages/typia/native/core/schemas/metadata"
-	nativeprotobuf "github.com/samchon/typia/packages/typia/native/core/schemas/protobuf"
+  shimast "github.com/microsoft/typescript-go/shim/ast"
+  shimchecker "github.com/microsoft/typescript-go/shim/checker"
+  shimprinter "github.com/microsoft/typescript-go/shim/printer"
+  "github.com/samchon/ttsc/packages/ttsc/driver"
+  schemametadata "github.com/samchon/typia/packages/typia/native/core/schemas/metadata"
+  nativeprotobuf "github.com/samchon/typia/packages/typia/native/core/schemas/protobuf"
 )
 
 // TestFactoryUtilityCoverage covers small factory helpers.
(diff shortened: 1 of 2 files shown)


samchon (commented):
**Self-Review round 1 — complete round over `master...fbb4fae`, nothing to improve**

### The deadness claim rests on the compiler, not on the scan

The scan located the candidates; it does not prove them. What proves them is that the tree still builds with the functions gone — a referenced function's removal is a compile error, and both `go build ../native/...` and the `typia_native_internal` build (the only one that compiles the touched fixture at all) succeed. That also covers the scan's blind spots: it strips `//` comments before matching, which could in principle discard a reference sitting inside a raw string after a `//`, and the build would have caught exactly that.

Cross-package use is impossible either way — both names are unexported, and `packages/typia/test` is a separate module.

### Checked, no finding

- **Cascade** — rescanning after the removal reports zero functions without an external reference, so neither deletion orphaned anything.
- **Imports** — `metadata_type_symbol_name` was the file's user of `strings.Join`, but `strings` still has 17 other uses in `MetadataHelper.go`; `metadata_get_type_arguments` and `metadata_symbol_name` keep other callers. The build confirms both.
- **Knowledge preserved** — the removed function's doc comment explained why a name must come from the type's own symbol rather than `TypeToString`. That reasoning belongs to `metadata_type_symbol_base_name`, which is what the Map / Set / native guards actually call, so it moved there instead of being deleted with the dead code.
- **Coverage** — `TestFactoryUtilityCoverage` keeps every assertion; the removed helper appeared in none of them.
- **Formatting commit** — `git diff -w` between `fbb4fae` and its parent is empty, so the 896-line rewrite carries no content, and the tagged test passes after it.
- **Verification lanes** — `pnpm test:go` (public + native) and `go test -tags typia_native_internal ../native/...` both pass. The tagged lane matters here: root `pnpm test:go` does not enable that tag, so it never compiles the fixture this branch edits.

### Observed, deliberately not acted on

Forty-two other Go files under `packages/typia/native` still carry gofmt tabs. That is not drift to repair here: `scripts/format-go.cjs` formats only the files a branch touches, so each one normalizes when something legitimately edits it. Normalizing all of them now would be a large rewrite unattached to any change.

Issue #2335, opened 2026-08-13
Adopt JSON Schema length and divisibility semantics for MinLength, MaxLength, and MultipleOf


samchon:
## Problem

`tags.MinLength` and `tags.MaxLength` measure a string with `$input.length`, which counts UTF-16 code units. `tags.MultipleOf` tests divisibility with `$input % value === 0`, which is inexact for fractional divisors.

JSON Schema defines both differently. String length is *"the number of its characters as defined by RFC 8259"* (draft 2020-12 §6.3.1–6.3.2), and RFC 8259 defines a JSON string as a sequence of Unicode characters — code points, not UTF-16 code units. `multipleOf` is mathematical divisibility, not a floating-point remainder.

typia emits `minLength`, `maxLength`, and `multipleOf` into the schemas it publishes through `typia.json.schema()`, `typia.json.application()`, and `typia.llm.schema()`. Those schemas are consumed by Ajv, OpenAPI tooling, LLM structured-output endpoints, and generated SDKs. The validator typia generates and the schema typia publishes describe the same type and must accept the same values.

## Evidence

String length:

```ts
"😀".length          // 2  — UTF-16 code units
[..."😀"].length     // 1  — Unicode characters
```

```ts
typia.is<string & tags.MinLength<2>>("😀");        // true
typia.json.schema<string & tags.MinLength<2>>();   // { minLength: 2 }
```

```js
new Ajv().compile({ type: "string", minLength: 2 })("😀");   // false
```

Same type, opposite answers. Ajv resolves `minLength`/`maxLength` through its `ucs2length` helper by default and falls back to `.length` only under the explicit `unicode: false` opt-out.

Divisibility:

```js
0.3 % 0.1        // 0.09999999999999998
```

`0.3` is exactly three times `0.1`. The generated validator rejects it; `{ multipleOf: 0.1 }` accepts it.

| value | multipleOf | mathematically a multiple | `value % multipleOf === 0` |
| --- | --- | --- | --- |
| `0.3` | `0.1` | yes | false |
| `0.7` | `0.1` | yes | false |
| `1.1` | `0.1` | yes | false |
| `0.6` | `0.2` | yes | false |
| `2.4` | `0.8` | yes | false |
| `0.75` | `0.25` | yes | true |

## Consequence surface

| Surface | Current | Required |
| --- | --- | --- |
[...]

PR #2336, opened 2026-08-13
fix(interface): stop the scalar tag validators from naming a runtime helper

samchon:
Reverts the runtime meaning of three type tags to the form they had before
`e30b0ccf6c` ("Keep scalar constraints consistent across execution surfaces").
Only the tag declarations and the things that must agree with them move.

## What changes

| Declaration | Before this PR | After this PR |
| --- | --- | --- |
| `tags.MinLength<N>` | `N <= $importInternal("_stringLength")($input)` | `N <= $input.length` |
| `tags.MaxLength<N>` | `$importInternal("_stringLength")($input) <= N` | `$input.length <= N` |
| `tags.MultipleOf<N>` (number) | `$importInternal("_isMultipleOf")($input, N)` | `$input % N === 0` |
| `tags.MultipleOf<N>` (bigint) | `$input % BigInt(N) === BigInt(0)` | unchanged |

`packages/typia/native/core/factories/MetadataCommentTagFactory.go` moves with
them, because the JSDoc `@minLength` / `@maxLength` / `@multipleOf` spelling and
the type-tag spelling of one constraint must not mean different things.
`TagBase.validate`'s doc example and the `validators/tags` guide describe the
reverted semantics.

## What deliberately does not change

- `packages/typia/src/internal/_stringLength.ts`, `_isMultipleOf.ts`,
  `_decimal.ts`, `_randomMultiple.ts`, and the
  `packages/utils/src/validators/functional/` counterparts stay published and
  untouched. `_stringLength` and `_isMultipleOf` are the two a tag ever named,
  and they stay shipped precisely so a newer `typia` can still satisfy an older
  `@typia/interface` that names them. `_decimal` and `_randomMultiple` were
  never named by a tag and remain live dependencies of random generation.
- `@typia/utils` (`OpenApiStringValidator`, `OpenApiTypeCheckerBase`,
  `_isMultipleOf`) keeps code-point and exact-decimal behavior.
- Random generation is untouched and still produces exact decimal multiples.
- No package version, dist-tag, or release state changes.

## This is a behavior change in both directions

Anyone on `typia` 13.1.19 or later sees validation outcomes flip both ways:

```ts
typia.is<string & tags.MinLength<2>>("😀");     // was false, is now true
typia.is<number & tags.MultipleOf<0.01>>(0.03); // was true,  is now false
```

**The release number this ships under is the maintainer's decision.** This
branch does not choose one and does not touch any version field.

Because `typia.json.schema`, `typia.llm.schema`, and `typia.json.application`
still emit `minLength` / `maxLength` / `multipleOf` with their JSON Schema
meaning, the generated validator and the shared `@typia/utils` OpenAPI validator
now disagree on an astral character and on a divisor with no exact binary
double. That divergence is accepted here and tracked by #2335, which owns
bringing the JSON Schema semantics back properly in a major.

## Effect on #2330, stated precisely — this PR does not close it

#2330 is the packaging failure where an older `typia` resolving a newer
`@typia/interface` through the caret range emits an import the installed runtime
does not ship. That failure comes from a tag `validate` template naming a
`typia/lib/internal/*` module, so removing two such names shrinks the exposure.
It does **not** eliminate the failure class, so this PR carries no closing
keyword for #2330.

I enumerated every `$importInternal` reference in `packages/interface/src` at
this head and checked each name against the published tarballs
(`npm pack typia@<v>`, `npm pack @typia/interface@<v>`):

| Named helper | First `@typia/interface` naming it | First `typia` shipping it |
| --- | --- | --- |
| `isFormat*` — all 22 formats (`Format.ts`) | ≤ 13.0.0 | ≤ 13.0.0 |
| `isTypeInt8` / `Uint8` / `Int16` / `Uint16` / `Int32` / `Uint32` / `Int64` / `Uint64` / `Float` | ≤ 13.0.0 | ≤ 13.0.0 |
| `isUniqueItems` | ≤ 13.0.0 | ≤ 13.0.0 |
| `isTypeInt64Bigint`, `isTypeUint64Bigint` | **13.1.19** | **13.1.19** |
| `_stringLength`, `_isMultipleOf` | 13.1.19 | 13.1.19 |

The last row is what this PR removes. The row above it is not.
`Type.ts` has named `isTypeInt64Bigint` and `isTypeUint64Bigint` since #2166,
and `typia@13.0.0`, `13.0.1`, `13.0.2`, `13.1.0`, and `13.1.1` ship neither.
A project pinned to one of those `typia` versions with `@typia/interface`
floated to `13.2.0` still emits `typia/lib/internal/_isTypeInt64Bigint` for
`bigint & tags.Type<"int64">` and still cannot resolve it.

Reproduction of the residual:

```sh
npm pack typia@13.0.0
tar -tzf typia-13.0.0.tgz | grep _isTypeInt64Bigint          # no match

npm pack @typia/interface@13.2.0
tar -xzf typia-interface-13.2.0.tgz
grep isTypeInt64Bigint package/src/tags/Type.ts              # matches
```

So #2330's root cause — a caret range that lets `typia` and `@typia/interface`
pair across a helper-set boundary — is untouched here, and its suggested fix
(pinning the internal packages exactly) still applies. Narrowing the exposure
from `tags.MinLength`, which nearly every schema uses, to
`bigint & tags.Type<"int64">` is worth having, but it is not a fix.

## Tests

`tests/test-typia-schema/src/features/validate/test_validate_unicode_string_length.ts`
and `test_validate_decimal_multiple_of.ts` pinned the semantics being reverted.
Both are rewritten rather than deleted: they now assert the code-unit and
remainder behavior over the same boundary matrix, and each pins the exact values
where `@typia/utils` disagrees, so the divergence #2335 owns is visible in the
suite instead of implicit. The decimal file gains a `MultipleOf<1.5>` control,
because "fractional" is not what breaks the remainder check: `9` and `1.5` both
equal the decimal they print back, so `9 % 1.5` is `0` even though the quotient
`6` is not a power of two. What breaks it is an operand that does not print back
exactly — `0.01`, whose stored double is not 1/100, or any value past
`Number.MAX_SAFE_INTEGER`.

`tests/test-typia-schema/src/features/random/test_random_numeric_multiple_of.ts`
round-tripped random output through `typia.is` for divisors the reverted
remainder check can no longer confirm. Random generation is not touched; the
assertion is split instead. Every declared bound is still revalidated through
`typia.is` against a divisor-free twin of the type, every divisor is still
checked against the exact decimal oracle (three members gained one), and
`typia.is` on the divisor itself is asserted for the divisors that do print back
exactly.

`tests/test-utils/src/features/openapi/test_openapi_validator_unicode_length.ts`
and `test_openapi_validator_decimal_multiple_of.ts` cover `@typia/utils`, which
is not reverting. They build their schemas by hand and never invoke the
transform, so they are untouched and still pass.

`website/src/content/docs/random.mdx` gains a note, because that page opens by
promising `typia.random<T>` returns a value satisfying `T`, and for `MultipleOf`
on a `number` that promise no longer holds.

## Verification

Local, at head `e6c4a55a9c`:

- `pnpm build` — exit 0.
- `pnpm --filter ./tests/test-typia-schema --fail-if-no-match start` — `Success`.
- `pnpm --filter ./tests/test-typia-generate --fail-if-no-match start` — 10 tests, 9 pass, 0 fail, 1 skipped.
- `go -C packages/typia/test test ./...` — 16 packages `ok`.
- `go -C packages/typia/test test -tags typia_native_internal ../native/...` — 28 packages `ok`, including `native/cmd/ttsc-typia` and `native/core/factories`, the two packages this change edits. Root `pnpm test` does not enable that tag, and `metadata_comment_tag_factory_coverage_test.go` sits behind it.
- Root `pnpm test` — every TypeScript suite ran, `test-interface` (`ttsc`, `Done`) and `test-error` (`Every fixture rejected: 78 files, 873 typia diagnostics, 142 distinct codes named`) included. One subtest of `test-typia-generate`, `physical output aliases are excluded`, failed with `spawnSync … node.exe ETIMEDOUT` at 180 043 ms on a heavily loaded machine; the same subtest passed in 3 s when the suite was re-run in isolation, and the CI lane passes.
- `pnpm format` — the only files it left with a nonzero `git diff --numstat` were the ones edited by hand.

**Mutation sensitivity.** With the tests untouched, `MinLength` / `MaxLength` were put back to `$importInternal("_stringLength")($input)` and `MultipleOf` to `$importInternal("_isMultipleOf")($input, N)`, and the schema suite re-run. Exit 127, with exactly the two intended cases failing on exactly the intended assertions — `Bug on type tag rejects 0.03` and `Bug on type tag "😀"` — while every other case in the suite, including `test_random_numeric_multiple_of`, still passed. The declarations were then restored, leaving the tree byte-identical to `HEAD`.

**Compatibility claim, checked rather than assumed.** `packages/typia/lib/internal/` from a post-revert build still contains `_stringLength.{js,mjs,d.ts}` and `_isMultipleOf.*` with zero references remaining anywhere in `packages/typia/src`. `tsc` compiles the program rather than the reachable graph, so the helpers a newer `typia` needs in order to satisfy an older `@typia/interface` do still ship.

CI: all required checks green on `e6c4a55a9c`.


samchon (commented):
**Commit ledger — `e88beb4e3d` `fix(interface): stop the scalar tag validators from naming a runtime helper`**

Landed the declaration revert itself. Nine files:

- `packages/interface/src/tags/{MinLength,MaxLength,MultipleOf,TagBase}.ts` — `git diff e30b0ccf6c^ e88beb4e3d --` over these four is empty, so they are byte-for-byte their pre-`e30b0ccf6c` state, including the original line-wrapping of the `MultipleOf` conditional.
- `packages/typia/native/core/factories/MetadataCommentTagFactory.go` — the four `@length`/`@minLength`/`@maxLength` records and the `@multipleOf` number arm. Only the `Validate` strings moved; the `metadataCommentTagFactory_numeric` restructuring from `e30b0ccf6c` (number-only record for a non-integer divisor, no bigint diagnostic) is deliberately kept, and so is `metadataCommentTagFactory_exclusive` from #2185.
- `packages/typia/native/cmd/ttsc-typia/template_literal_type_tags_transform_test.go`, `.../transform_helpers_test.go`, `.../core/factories/metadata_comment_tag_factory_coverage_test.go` — the native expectations that pinned the emitted `_stringLength` import and the JSDoc `_isMultipleOf` string.
- `website/src/content/docs/validators/tags.mdx` — the three table cells and the two prose paragraphs describing code-point / exact-decimal semantics.

Resolves no issue on its own; see the pull-request body for why this does **not** close #2330.

Two things worth recording rather than changing:

1. **The `_stringLength` Go stub was removed, not left dormant.** `ttscTypiaTestStringLengthStub`, its `string-length-stub.cjs` map entry, and its `ReplaceAll` line all went with the emit that needed them. After this commit `_stringLength` appears in zero `.go` files repo-wide, and the two producers of that import string — the tag `validate` template and the JSDoc record — both stopped naming it here, so no other transform test's emitted JavaScript can import it. `ttscTypiaTestRewriteCommonJS` still fails loudly, by name, on any unstubbed `typia/lib/internal/*`, so a regression here cannot pass silently.

2. **The bigint arm of `MultipleOf` spells itself two ways, and this commit does not change that.** The type tag emits `$input % BigInt(3) === BigInt(0)` (via `Cast<Value>` in `MultipleOf.ts`), while `/** @multipleOf 3 */ value: bigint` emits `$input % 3n === 0n`. The two are semantically identical and both are exercised (`@typia/template`'s `TypeTagBigInt` and `CommentTagBigInt`). The divergence predates `e30b0ccf6c` and is out of scope here; noting it so the commit message's "the two spellings of one constraint must never mean different things" is read as the semantic claim it is, not a claim about the emitted text.

The tests that pinned the reverted semantics are not in this commit; they follow in the next one.

samchon (commented):
**Commit ledger — `e7b56a6c92` `test(typia-schema): pin the reverted length and divisibility semantics`**

Three suites encoded the semantics the previous commit reverted. All three are rewritten, none deleted.

**`test_validate_unicode_string_length`** — now asserts UTF-16 code-unit counting over the empty string, ASCII, a precomposed `U+00E9`, an astral character, `e` + `U+0301`, and a lone surrogate. The two accent forms are written as `"é"` and `"é"` rather than as literal characters, because in the previous version they rendered identically in the source while carrying opposite expectations. Verified against `node`: lengths are 0, 1, 1, 2, 3, 2, 1 and code-point counts are 0, 1, 1, 1, 2, 2, 1, so the astral character is the only value where the two measurements differ — and the test pins exactly that one as the `@typia/utils` divergence.

**`test_validate_decimal_multiple_of`** — the previous version's `accepted` list was `[0, 0.03, 1.01, -0.03, 3e-2]`, and four of those five now fail (`0.03 % 0.01` is `0.009999999999999998`). Rather than trim the list to whatever still passes, the values are re-partitioned into three sets that name what they are: a zero remainder, a mathematical multiple whose remainder is not zero, and a non-multiple. Enumerating `i/100` for `i` in `[-20, 200]` shows only `0, ±0.01, ±0.02, ±0.04, ±0.08, ±0.16, 0.32, 0.64, 1.28` reach a zero remainder — the power-of-two quotients — which is why the sets are listed by value and not derived. An integer-divisor control (`MultipleOf<2>`) pins that nothing diverges when the divisor divides exactly in binary.

**`test_random_numeric_multiple_of`** — this one is the consequence the revert creates rather than one it restores. `_randomMultiple` still uses `_isMultipleOf` as its acceptance oracle and emits genuine decimal multiples, so a blanket `typia.is<IValues>(value)` now fails: the `decimal` member yields values like `±0.03`, `lowerOnly` yields `0.3` against `MultipleOf<0.1>`, and `large` and `fractionalLargeQuotient` fail likewise. Random generation is explicitly out of scope, so the assertion is split instead of weakened:

- every declared bound is still revalidated through `typia.is`, against an `IBounds` twin carrying the same members without their divisors — so `Minimum`, `Maximum`, `ExclusiveMinimum`, `ExclusiveMaximum`, and `Type<"int32">` keep exactly the coverage they had;
- every divisor is still checked against the exact decimal oracle, including `upperOnly`, `integerUpperOnly`, and `integerDecimal`, which had no oracle before;
- `typia.is` on the divisor itself is asserted for the three binary-exact divisors, so the generator and validator are still proved to agree wherever they can.

Net: the file loses no coverage and gains three oracle cells.

`website/src/content/docs/random.mdx` gains the matching sentence, because the page opens by promising `typia.random<T>` returns a value satisfying `T`. Its existing string-length claim ("throws instead of emitting a value its own validator would reject") was checked and still holds — every random generator in `packages/typia/src/internal/` is pure ASCII, so `.length` and the code-point count agree on generated strings.

`tests/test-utils/**` covers `@typia/utils`, which is not reverting, and is untouched — verified that `test_openapi_validator_unicode_length` and `test_openapi_validator_decimal_multiple_of` build their schemas by hand and never invoke the transform.

**Verification:** full `pnpm --filter ./tests/test-typia-schema --fail-if-no-match start` — `Success`, exit 0, with `test_validate_unicode_string_length`, `test_validate_decimal_multiple_of`, and `test_random_numeric_multiple_of` all present in the reported case list rather than silently absent.

samchon (commented):
**Commit ledger — `52fddaa80a` `test(typia-schema): name the binary representation, not "fractional", as the divergence`**

A correction to the previous commit's own explanation, caught before merge.

`e7b56a6c92` explained the `MultipleOf` divergence as *"under a fractional divisor the remainder only reaches zero for a power-of-two quotient"*, in four places — both validate tests, the random test, and `random.mdx`. **The rule is false, and the same commit disproves it.** `test_random_numeric_multiple_of`'s `integerDecimal` member is `MultipleOf<1.5>`; `1.5` is fractional; `9 % 1.5` is `0` with quotient `6`. `2.25 % 0.25` is likewise `0` with quotient `9`. A reader following the stated rule would predict the suite's own assertion should fail.

The boundary is representation, not fractionality:

- `%` divides the binary doubles actually stored.
- `multipleOf` / `_isMultipleOf` divide the shortest decimals those doubles print as.
- They agree wherever both the value and the divisor have an exact binary double, and part company otherwise.

Two ways to part company, both verified in `node`:

1. **A divisor with no exact binary double.** `0.01` is stored as the nearest double to 1/100, and for that divisor the remainder reaches zero only at a power-of-two quotient — enumerating `i/100` over `i ∈ [-20, 200]` gives exactly `0, ±0.01, ±0.02, ±0.04, ±0.08, ±0.16, 0.32, 0.64, 1.28`. That observation is now stated as a property of `0.01`, not of fractional divisors.
2. **A value past `Number.MAX_SAFE_INTEGER`, even with an integer divisor.**
   ```
   const v = Number(3n * 12259405221713610n); // 36778215665140830, stored as ...832
   v % 3                                       // 2
   _isMultipleOf(v, 3)                         // true
   ```
   `_randomMultiple` uses `_isMultipleOf` as its accept gate, so this is reachable from `number & tags.Minimum<10_000_000_000_000_000> & tags.MultipleOf<3>` — the same magnitude band the suite's `large` member already uses. `random.mdx` therefore no longer claims integer divisors are unaffected; only `bigint` divisors are.

`test_validate_decimal_multiple_of` gains a `MultipleOf<1.5>` control asserting both sides (`9` accepted, `10` rejected), so the fractional-but-exact case is pinned rather than merely described.

Three smaller corrections in the same files:

- `3e-2` left the rejection list. It is the same double as the `0.03` beside it (`(3e-2).toString() === "0.03"`), so it added no coverage and produced duplicate assertion titles, while the adjacent comment made the redundancy look intentional.
- The unicode comment claimed every non-astral row counts the same under both measurements. `"a" + astral` is 3 code units against 2 characters; it fails either way, so the assertion held, but the comment was wrong.
- Two numbered steps said the test pins only the diverging values. Both loops revalidate every value (7 and 16 of them, of which 1 and 7 diverge), so the steps now say that.

**Verification for this head:** `go -C packages/typia/test test -tags typia_native_internal ../native/...` — 28 packages `ok`, 0 `FAIL`, exit 0, including `native/cmd/ttsc-typia` (166s) and `native/core/factories`, the two packages this pull request changes. Root `pnpm test` is running against this exact head.

**Process note.** An earlier local run was recorded as green on the strength of `pnpm run test:go:native 2>&1 | tail -20`. That pipeline's exit status is `tail`'s, not `go`'s, and `tail -20` had cut the result lines for `cmd/ttsc-typia` and `core/factories` out of the captured output — so the run proved nothing about the packages this change actually touches. It was re-run without the pipe before anything was concluded from it.

samchon (commented):
**Commit ledger — `e6c4a55a9c` `test(typia-schema): condition the MultipleOf divergence on printing back exactly`**

The previous commit replaced one wrong explanation with another. `52fddaa80a` said a divisor *with an exact binary double* keeps the remainder check agreeing with mathematical divisibility. Counterexample, verified in `node` against a verbatim port of `packages/utils/src/validators/functional/_isMultipleOf.ts`:

```
2 ** 70                       // an exactly representable double
(2 ** 70).toString()          // "1.1805916207174113e+21"
BigInt(2 ** 70)               // 1180591620717411303424  — not divisible by 3 or by 1.5
BigInt("1180591620717411300000")  // the printed expansion — divisible by both
(2 ** 70) % 1.5               // 1      → the generated validator rejects
_isMultipleOf(2 ** 70, 1.5)   // true   → the OpenAPI validator accepts
```

Both operands are exact doubles and the two still disagree, so "exact binary double" is not the condition.

The condition is that **every operand equals the decimal it prints back**. `fmod` is exact for doubles, so `%` answers correctly about the values actually stored; `_isMultipleOf` and the emitted `multipleOf` keyword answer about `value.toString()`. Those are the same number only when the decimal round-trips, which fails two ways — a decimal with no exact double (`0.01`, whose stored value is not 1/100) and any value past `Number.MAX_SAFE_INTEGER`, where the shortest round-trip spelling stops being the exact value. Both failure modes are now named in the three doc comments, and the `2 ** 70` case is written into the control comment so the next reader cannot re-derive the wrong rule.

Two further overreaches removed from `website/src/content/docs/random.mdx`:

- "`MultipleOf` is **the one** constraint where a generated value can fail its own validator" — `tags.Exclude` is already documented as ignored by random generation (`validators/tags.mdx` and `packages/interface/src/tags/Exclude.ts`), so `MultipleOf` is *one* such constraint, not the only one.
- "A `bigint` divisor is exact and always agrees" — bigint generation routes through `_randomInteger`, a double, before `BigInt()`, so a large generated bigint can miss its own `$input % BigInt(N) === BigInt(0)` check for exactly the reason above. That path is unchanged by this pull request and predates it; the sentence is dropped rather than replaced with a claim this branch does not own.

Also: the zero-remainder comment now reads "a zero or power-of-two quotient", since the list it introduces starts at `0`, and the two assertion titles say "exactly-printing fractional divisor" rather than "binary-exact divisor".

**Verification at `52fddaa80a`** (this commit changes comments and MDX prose only, so the code result carries):

- Root `pnpm test` — every TypeScript suite ran. 12 green: `test-feature-identity`, `test-langchain`, `test-mcp`, `test-typia-automated`, `test-typia-bundler-cache`, `test-typia-cli`, `test-typia-exact-optional`, `test-typia-schema`, `test-utils`, `test-utils-automated`, `test-vercel`, plus `test-interface` (`ttsc`, `Done`) and `test-error` (`Every fixture rejected: 78 files, 873 typia diagnostics, 142 distinct codes named`).
- `test-typia-generate` failed one of ten subtests, `physical output aliases are excluded`, with `spawnSync … node.exe ETIMEDOUT` after 180 043 ms. The other nine subtests in the same file ran in 7–13 s each. It is a CLI-spawn timeout on a loaded machine, not an assertion about content, and nothing in this pull request reaches `typia generate`'s symlink handling. Re-running the suite in isolation to confirm; the pull request's own CI lane is the gate that matters.
- `go -C packages/typia/test test ./...` — 16 packages `ok`.
- `go -C packages/typia/test test -tags typia_native_internal ../native/...` — 28 packages `ok`, including `native/cmd/ttsc-typia` and `native/core/factories`.
- `pnpm build` — exit 0.
- `pnpm format` — the only files it left with a nonzero `git diff --numstat` were the ones edited by hand.

samchon (commented):
## Self-Review — round 1, complete

Surface: `origin/master` `829d2147e2` … head `e6c4a55a9c`, all 13 changed files, working tree clean (`git diff --numstat HEAD` empty). No uncommitted work.

### Requirement fidelity

The four declarations are byte-for-byte their pre-`e30b0ccf6c` state — `git diff e30b0ccf6c^ HEAD -- packages/interface/src/tags/{MinLength,MaxLength,MultipleOf,TagBase}.ts` is empty. The JSDoc records match the type tags exactly for all three constraints:

| tag | `packages/interface/src/tags/` | `MetadataCommentTagFactory.go` |
| --- | --- | --- |
| `minLength` | `` `${Value} <= $input.length` `` | `props.Value + " <= $input.length"` |
| `maxLength` | `` `$input.length <= ${Value}` `` | `"$input.length <= " + props.Value` |
| `multipleOf` (number) | `$input % N === 0` via `Cast<Value> = Value` | `"$input % "+props.Value+" === 0"` |

Everything the change was told to leave alone is untouched. Repo-wide grep at this head: `_stringLength` survives only in `packages/typia/src/internal/_stringLength.ts` (now referenced by nothing) and in the three `@typia/utils` consumers; `_isMultipleOf` survives in `packages/typia/src/internal/{_isMultipleOf,_randomMultiple}.ts` and six `@typia/utils` sites. No version, dist-tag, or release state moved.

### The compatibility claim, checked rather than assumed

`packages/typia/lib/internal/` from a post-revert `pnpm build` still contains `_stringLength.{js,mjs,d.ts,js.map,mjs.map}` and `_isMultipleOf.*` despite zero references in `packages/typia/src`. `tsc` compiles the program, not the reachable graph, so an unreferenced helper still ships — which is the whole basis for "a newer `typia` can still satisfy an older `@typia/interface` that names it". Verified, not assumed.

### Mutation sensitivity

In this checkout, with the tests untouched, `MinLength`/`MaxLength` were put back to `$importInternal("_stringLength")($input)` and `MultipleOf` to `$importInternal("_isMultipleOf")($input, N)`, and `pnpm --filter ./tests/test-typia-schema start` was re-run. Exit 127, and exactly the two intended cases failed, on exactly the intended assertions:

```
- test_validate_decimal_multiple_of: Error
    Bug on type tag rejects 0.03: found different values
- test_validate_unicode_string_length: Error
    Bug on type tag "😀": found different values
```

Every other case in that suite still passed, including `test_random_numeric_multiple_of`, so the controls hold and the two rewritten files are genuinely sensitive to the declaration they pin. The tags were then restored with `git checkout --`, leaving the tree byte-identical to `HEAD`.

That run also settles a cache question worth recording: the same `TTSC_CACHE_DIR` produced reverted behavior before the mutation and mutated behavior after it, so the plugin cache does not mask a tag-declaration change even though the tag lives in TypeScript rather than in the Go plugin key.

### Consequence surface

- **Round-trip risk.** `typia.random` still emits exact decimal multiples while the validator now uses `%`, so `test_random_numeric_multiple_of`'s blanket `typia.is<IValues>` was no longer true. Random generation was not touched; the assertion was split into an `IBounds` twin (every bound of every one of the 8 members, checked member by member) plus per-divisor oracles and per-divisor `typia.is` for the divisors that print back exactly. Coverage is a net gain: three members had no oracle before.
- **The generated matrices are unaffected**, and this was checked rather than inferred. Every `MultipleOf` in `@typia/template` uses an integer divisor (`3`, `3n`), and every length tag there is applied to ASCII fixtures. A scan of `packages/typia/src/internal/_random*.ts` found no non-ASCII character and no `\u` escape in any generator, so a generated string's code-unit count always equals its character count. `test-typia-automated` and `test-utils-automated` both pass.
- **`tests/test-error`** names none of these three tags; grep returns nothing.
- **`tests/test-interface`** compiles clean, so the reverted `validate` template literals still satisfy the exported type contracts.
- **`@typia/utils` tests pass untouched**, as intended: `test_openapi_validator_unicode_length` and `test_openapi_validator_decimal_multiple_of` build their schemas by hand and never invoke the transform.
- **Platform.** Nothing path-, encoding-, or platform-dependent was introduced; the Go edits are string literals and the test edits are values and comments. `withRandom` restores `Math.random` in a `finally`.

### Deliberately not changed — recorded so the next reader does not re-derive them

1. **`#2330` is not closed by this pull request.** `Type.ts` still names `isTypeInt64Bigint` and `isTypeUint64Bigint`, which `typia@13.0.0` through `13.1.1` do not ship, so the same import-resolution failure remains reachable through `bigint & tags.Type<"int64">`. Reverting those would undo #2166's range enforcement, and it is outside this change's scope. Details and reproduction are in the pull-request body.
2. **`MinLength.ts` and `MaxLength.ts` prose still says "number of characters"** while the check counts UTF-16 code units. That wording predates `e30b0ccf6c` and is restored byte-exact; correcting it belongs with #2335, which owns the semantics.
3. **`validators/tags.mdx` carries no note that the emitted JSON Schema keyword means something different from the runtime check.** That is also the pre-`13.1.19` state, and it is the gap #2335 exists to close. `random.mdx` did need a note, because that page promises `typia.random<T>` returns a value satisfying `T`, and this change makes that promise false for a `number` under `MultipleOf`.
4. **The bigint arm spells one constraint two ways** — `$input % BigInt(3) === BigInt(0)` from the type tag, `$input % 3n === 0n` from the JSDoc record. Semantically identical, both exercised by `@typia/template`, unchanged by `e30b0ccf6c` and unchanged here.
5. **`template_literal_type_tags_transform_test.go` asserts only `strings.Contains(out, "length <= 30")`**, restored verbatim. What actually forbids the helper from returning is stronger and is in the same commit: removing `ttscTypiaTestStringLengthStub` means `ttscTypiaTestRewriteCommonJS`'s guard now fails, by helper name, on any emit that imports `typia/lib/internal/_stringLength`.

### Verification at this exact head

- All 8 required checks green on `e6c4a55a9c` (confirmed via `head_sha`, not just the latest-run view): `build` 10m8s, the `test` lane 14m8s, format, website, experiments, spell-check, both Socket lanes.
- Local: `pnpm build` exit 0; `go -C packages/typia/test test ./...` 16 packages `ok`; `go -C packages/typia/test test -tags typia_native_internal ../native/...` 28 packages `ok`, covering both packages this change edits; `pnpm --filter ./tests/test-typia-schema start` `Success`; `pnpm --filter ./tests/test-typia-generate start` 10 tests / 9 pass / 0 fail / 1 skipped.
- One local `pnpm test` run failed `test-typia-generate`'s `physical output aliases are excluded` with `spawnSync … ETIMEDOUT` at 180 043 ms while the machine was heavily loaded; re-run in isolation the same subtest passed in 3 s, and the CI lane passed. Recorded rather than dismissed.

### Conclusion

Round 1 complete over the whole base-to-head diff. Three defects were found and fixed during implementation by the per-commit passes — a false "power-of-two quotient" rule, a false "exact binary double" rule, and two overreaching claims in `random.mdx` — each landed as its own commit with its counterexample. This round found nothing further that survives verification. No changes applied in this round, so no further round is required.

Issue #2328, opened 2026-08-05
Types of `createAssertEquals` and friends do not expose `errorFactory` and fail with pointwise invocations for arrays


konradekk:
## Bug Report

**Disclaimer:** this bug report was written with a help of AI/LLM.

**TL;DR:** two issues here: (a) the main one: wrong typing of `createAssertEquals` and friends that provide footgun for array maps, (b) a would-be fix for previous one: overriding types locally was possible in Typia 9, but not in Typia 13.

### 📝 Summary

An ambient module declaration that redeclares `createAssert*` makes the ttsc plugin skip those calls silently — no diagnostic, no transform, and a runtime `no transform has been configured` error.

  - **Typia Version**: 13.2.0 (with `ttsc` 0.24.0, `typescript` 7.0.2)
  - **Expected behavior**: either the transform still applies, or the plugin emits a
    diagnostic saying it could not attribute the call to Typia
  - **Actual behavior**: the call is left untransformed and the build reports nothing;
    the failure surfaces only when the generated code runs

Our project ships a `typia.d.ts` shadow declaration to expose the `errorFactory` parameter that Typia's own `.d.ts` hides on the functions returned by `createAssertEquals` and friends. Making that parameter visible turns a real hazard into a compile error: `rows.map(parseFoo)` otherwise type-checks, and `Array#map`'s index argument then lands in `errorFactory` at runtime, throwing `TypeError` instead of `TypeGuardError`.

Under Typia 9 with `ts-patch` this worked. Under 13 with `ttsc`, the plugin appears to identify its API by resolving each callee to Typia's own declaration, so a redeclared `createAssertEquals` is no longer recognised and the call is skipped.

What makes this expensive to diagnose is that it is both silent and partial. In a single file, `createRandom` — which we do not redeclare — transformed correctly while `createAssertEquals` did not:

    import * as _randomString_1 from "typia/lib/internal/_randomString";   // transformed
    export const parseTemplate = typia.createAssertEquals<Template>();     // untransformed

`ttsc.transform()` returned `type: "success"` with `diagnostics: []`, and `tsc --noEmit` was clean. 136 of our 161 jest suites then failed at runtime. A diagnostic at the skipped call site would have made this a five-minute fix.

### ⏯ Playground Link

Not reproducible in the playground: it needs a local ambient `.d.ts` in the project. A minimal reproduction is two files.

### 💻 Code occuring the bug

```typescript
[...]

PR #2337, opened 2026-08-13
fix(assert): expose the errorFactory parameter created assert functions really take

samchon:
Claims issue #2328 (`Types of createAssertEquals and friends do not expose errorFactory and fail with pointwise invocations for arrays`).

This pull request is opened empty, before any implementation exists, so the work is visible on the branch from the start. Verification is pending.

## Intent

The functions returned by typia's `create*Assert*` factories really do take a second, optional `errorFactory` parameter — `AssertProgrammer.Decompose` builds the returned arrow as `(input: any, errorFactory = <create-time factory>) => { ... }` and every assert-family programmer that composes it does the same. The declared TypeScript return types hide that parameter, so `rows.map(parseUser)` type-checks, `Array#map` hands the element index to `errorFactory`, and `_assertGuard` calls it: `TypeError: factory is not a function` instead of `TypeGuardError`. Index `0` is falsy, which makes the failure intermittent and easy to misattribute.

Two independent corrections are planned:

1. **Runtime.** `_assertGuard` must require a *callable* factory rather than a truthy one, so code already compiled against a published release degrades to a correct `TypeGuardError`.
2. **Declaration.** Expose the optional `errorFactory` parameter on the returned function type of every `create*` whose programmer emits `Guardian.Parameter`, so a `.map()` hand-off becomes a compile error instead of a runtime one.

The issue also reports a second symptom: a project-local `declare module "typia"` moves the resolved signature's declaration out of typia's own files, `CallExpressionTransformer` fails its path-identity test, and the call is left untransformed with **no diagnostic**. The path-identity check is a deliberate anti-spoofing boundary and will not be weakened; the planned change is a diagnostic that names the shadowing declaration file.

## Scope

- `packages/typia/src/internal/_assertGuard.ts` — runtime guard.
- `packages/typia/src/{module,json,http,notations,plain,protobuf}.ts` — declared return types of the assert-family factories.
- `packages/typia/native/transform/CallExpressionTransformer.go` — shadowed-declaration diagnostic.
- Tests under `packages/typia/native/cmd/ttsc-typia` and `tests/test-typia-schema`.
- `website/src/content/docs/**` for the affected guide pages.

## Not in scope

- No package version, dist-tag, or release-state change. Exposing `errorFactory` is a type-level breaking change for code that currently passes a created function to `.map`; the release number is the maintainer's decision and this branch does not choose one.
- The tag-semantics rollback owned by a concurrent change (`packages/interface/src/tags/**`, `MetadataCommentTagFactory.go`, the two tag validation cases, `website/src/content/docs/validators/tags.mdx`).

## Verification

Pending. This body is the historical intent statement; later results are recorded as formal pull-request reviews on the thread.


samchon (commented):
### Ledger — `91dfd8ceed` `fix(runtime): require a callable error factory in _assertGuard`

**What landed.** The runtime half of #2328. `packages/typia/src/internal/_assertGuard.ts` now invokes the caller-supplied error factory only when it is callable and falls back to `TypeGuardError` otherwise, instead of calling anything truthy. That is what protects code already compiled against a published declaration: the generated function's second parameter is real, `Array#map` fills it with the element index, and the old truthiness test turned an ordinary assertion failure into `TypeError: factory is not a function` for every index except the falsy `0`.

`ttscTypiaTestAssertGuardStub` in `packages/typia/native/cmd/ttsc-typia/transform_helpers_test.go` is the Go runtime double for that helper and moves with it, so the emitted-code tests keep executing the same rule the shipped helper enforces. `node scripts/format-go.cjs` realigned the neighboring `ttscTypiaTestWriteCommonRuntimeStubs` map literal in the same file.

`tests/test-typia-schema/src/features/assert/test_assert_error_factory_non_callable.ts` is the regression probe: it calls a created assert function the way `Array#map` does (`fn(value, index)`) and requires a `TypeGuardError` naming the real failure at indices 0, 1 and 2, with a configured callable factory as the one-axis control that the fallback did not swallow.

**Issues resolved by this commit.** None yet. #2328 also requires the declaration change and a diagnostic for the shadowed-module case; the closing line goes on the commit that earns it.

**Owner census (empirical, not inferred).** Every `create*` whose emitted arrow really takes `errorFactory` was enumerated by transforming a fixture that calls the whole factory surface and reading the emit:

- in scope: `createAssert`, `createAssertGuard`, `createAssertEquals`, `createAssertGuardEquals`, `json.createAssertParse`, `json.createAssertStringify`, `http.createAssertFormData`, `http.createAssertQuery`, `http.createAssertHeaders`, `notations.createAssert{Camel,Pascal,Snake,Kebab}`, `plain.createAssertClone`, `plain.createAssertPrune`, `plain.createAssertClassify`, `protobuf.createAssertDecode`, `protobuf.createAssertEncode` — 18 declarations.
- out of scope, confirmed by the same emit: every `createIs*` / `createValidate*` / plain converter factory, `createRandom`, `compare.create*`, `llm.createParse`, `llm.createCoerce`, `http.createParameter` (its assert helper is nested and called with one argument), and `functional.assert*` (the wrapper it returns carries the wrapped function's own signature).

`http.createAssertHeaders` and `http.createAssertQuery` are in that list even though `Guardian.Parameter` appears only in `HttpAssertFormDataProgrammer.go`: all three share `httpProgrammer_assert_result` in that file, so a grep for the call site under-counts the family by two.

**Verification.** `pnpm --filter ./tests/test-typia-schema --fail-if-no-match start -- --include assert_error_factory` reports the one case passing. Mutation sensitivity is proved before merge, in a disposable checkout, not here.

samchon (commented):
### Ledger — `aa21eeb5e0` `fix(assert): expose the errorFactory parameter created assert functions take`

**What landed.** The declaration half of #2328. Eighteen declared return types across `packages/typia/src/{module,json,http,notations,plain,protobuf}.ts` now name the optional `errorFactory` their emitted function has always taken, so `rows.map(parseUser)` is a compile error instead of a `TypeError` at run time, and the per-call override is usable.

**Design decision: `AssertionGuard<T>` is deliberately unchanged.** It is the annotation users write, `tests/test-interface/src/typings/test_assertion_guard.ts` pins it to `(input: unknown) => asserts input is T`, and `@typia/interface` cannot name `TypeGuardError.IProps` without inverting the package dependency. The guard factories return the two-parameter assertion signature directly instead, which is why the `Equal<>` target in `assert_guard_factory_contract_test.go` moved. All six assignability facts were proved by compiling them, not reasoned about:

| Fact | Result |
| --- | --- |
| `const fn: AssertionGuard<T> = createAssertGuard<T>()` compiles and still narrows | compiles |
| assignment to a user-written `(input: unknown) => T` alias | compiles |
| a non-`map`-like one-parameter callback position | compiles |
| `StandardSchemaV1<T, T> = typia.createValidate<T>()` | compiles (out of census, untouched) |
| `rows.map(twoParameterAssert)` | error, as intended |
| `rows.map(oneParameterAssert)` — the control | compiles, which is the hazard being closed |

`rows.map(guard)` on a one-parameter `AssertionGuard<T>` also compiles today, so the guard family carried the same hazard and the widened return closes it too.

**Verification.** `pnpm build` (all packages) and `pnpm --filter ./benchmark --fail-if-no-match run build` — the ~780-source workspace that is the largest in-repo consumer of these declarations — both exit 0 at this head. `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestAssertGuardFactoryContract` and `-run TestCreateAssertErrorFactoryArity` pass; `pnpm --filter ./tests/test-typia-schema start -- --include assert_error_factory` reports both cases passing.

**Coverage added.** `create_assert_error_factory_arity_test.go` type-checks a fixture that calls the override on all 18 members, pins `rows.map(...)` as `@ts-expect-error` for both the value and guard families with one-axis controls beside it (`rows.map((row) => assertUser(row))`, the `(input: unknown) => User` alias, an ordinary one-parameter callback position, `StandardSchemaV1` conformance), then transforms a second fixture and executes it. `test_assert_error_factory_call_time_override.ts` is the runtime twin against the shipped `_assertGuard`: create-time factory by default, call-time factory when supplied, across `createAssert`, `json.createAssertParse` and `plain.createAssertClone`.

**Documentation.** `website/src/content/docs/validators/assert.mdx` now documents the second parameter and warns against the `.map` hand-off; `website/src/content/docs/utilization/trpc.mdx` no longer claims the returned shape is `(input: unknown) => T`.

**Issues resolved by this commit.** None yet — the closing line goes on the commit that finishes the issue, and #2328 also asks for a diagnostic when a project-local `declare module "typia"` moves the declaration.

**Recorded, not fixed.** When a create-time factory was configured, a pointwise hand-off overrides it with the index, so the runtime falls back to `TypeGuardError` rather than to the configured factory — the configured factory is the parameter's *default*, and a default cannot be recovered once overridden. Restoring it would mean changing the emitter to bind the create-time factory to its own generated constant and consult it at the guard call, which rewrites the published emit shape in `pure.mdx` and every assert-family output for a case the declaration now rejects at compile time. The chosen behavior is strictly more informative than the `factory is not a function` it replaces, is pinned by `test_assert_error_factory_non_callable.ts`, and is stated in `_assertGuard.ts`. Raising it here so the maintainer can overrule the trade-off rather than discover it.

samchon (commented):
### Ledger — `a8fd33e724` `fix(native): report a typia call whose declaration was shadowed`

**What landed.** Part (b) of #2328. `CallExpressionTransformer` still attributes a call by the file its resolved signature is declared in — that is the anti-spoofing boundary `default_library_spoof_native_identity_transform_test.go` and `lib_replacement_native_identity_transform_test.go` protect, and it is unchanged. What changes is the silence after attribution fails.

A diagnostic now names the shadowing declaration file, gated on three conditions that must hold together:

1. the called name is one the transformer owns (a flat set built from the same functor table, so the gate cannot drift from the operations that exist);
2. the root of the callee expression resolves to a symbol carrying `SymbolFlagsAlias`; and
3. that alias's declaration sits under an `import ... from` whose specifier is `typia` or begins with `typia/`.

Condition 1 alone is not remotely sufficient — `is`, `parse`, `stringify`, `name`, `schema`, `equals`, `clone`, `encode`, `decode`, `query` and `headers` are ordinary words — which is why 2 and 3 exist and why the negative twin is part of this commit rather than optional.

**Message.** `error TS(typia.createAssert): this call resolved to a declaration in <file> instead of typia's own, so no transform was applied. Remove the \`declare module "typia"\` that redeclares it, or give the redeclaration another module name.` The diagnostic is positioned at the call site; the message names the shadowing file, because those are two different files and the reader needs both.

**Coverage.** `shadowed_typia_module_declaration_diagnostic_test.go` builds the reporter's layout — typia installed under `node_modules`, plus an ambient `declare module "typia"` in `src/typia.d.ts` — and requires exit code 3, exactly one diagnostic, the operation in its code, the call site as its file, and the shadowing file inside its message.

`local_function_named_like_typia_operation_transform_test.go` is the one-axis negative twin. Its fixture defines a local `createAssert`, imports a `createIs` from `./helpers`, and imports a `createAssert` from `typiary` — a package installed for the sole purpose of pinning the one-past-boundary of the specifier test, since `typiary` begins with `typia` but is not it. It requires zero diagnostics, exit 0, and the genuine `typia.createIs<User>()` in the same file to still emit its validator, so "no diagnostic" cannot be the transform quietly giving up on the file.

**Documentation.** `website/src/content/docs/setup/index.mdx` already told the reader what `no transform has been configured` means at run time; it now also names the build-time diagnostic and its cause, next to that paragraph.

**Deliberate consequence.** A project that redirects the `typia` specifier to a fork through tsconfig `paths` now fails its build instead of compiling to untransformed stubs. That build was already broken — it produced no validators — so surfacing it is the intended outcome, not collateral damage.

**Issues resolved.** This commit carries `Close #2328`; Part (a) landed in `91dfd8ceed` and `aa21eeb5e0`.

**Verification.** Both new cases pass. `go -C packages/typia/test test ../native/...` and `go -C packages/typia/test test ./...` are running over this head; results and the root `pnpm test` follow in the Self-Review round.

samchon (commented):
### Ledger — `bc9f52c0f3` and `0d283e07fb`

`bc9f52c0f3` replaced the shadowed-typia gate's flat name union with a namespace-aware lookup. The flat union meant any name typia owns anywhere counted at any position: `parse` lives only under `llm`, yet a project that augments typia with its own root `parse` and calls it would have been reported — a build that works today. The segment directly left of the method now selects the table, so `typia.json.createAssertParse` and `json.createAssertParse` reach `json`, `typia.createAssert` and a bare named import reach `module`, and a root-level `typia.assertParse` matches nothing.

`0d283e07fb` is the correction commit for what a read of the whole diff turned up. Six of them, and one is a test that could not fail:

| Correction | Evidence |
| --- | --- |
| The diagnostic code carries the namespace as written | `typia.createAssertParse` names no API, and `schema`, `schemas`, `application`, `equals` and `metadata` each live in more than one table, so the unqualified code was ambiguous as well as wrong |
| The message states the observed fact, not an assumed cause | the code never checks for a `declare module`; a vendored fork or a `paths` redirect reaches the same place, and naming the wrong cause misdirects |
| `import typia = require("typia")` reaches the specifier test | the CommonJS spelling of the same import was silently exempt |
| The neighbouring-package decoy actually reaches the specifier test | it imported `createAssert` under a **renamed** local binding, so the operation lookup rejected the call before the specifier was read. Widening the prefix from `typia/` to `typia` left both Part B tests green — the stub was proving nothing. It now imports `createValidate` under its own name |
| A clean `--noEmit` build guards the negative fixture | project-mode transform reports no semantic diagnostics, so a decoy that stopped resolving would have turned the whole negative case into a tautology |
| The vacuous `createIs<` assertion is gone | JavaScript emit erases type arguments, so it could never fail. The decoys are pinned by their emitted call shapes instead |

Three more, on the Part A side:

- **`test_assert_error_factory_non_callable` had a hole.** Its configured-factory case mapped over a one-element array, so the index reaching `errorFactory` was the falsy `0` — the one value the pre-fix helper already survived. The bad row is now last in a three-element array.
- **The arity fixture pinned the guard family through its own local alias**, which the old one-parameter declaration also satisfies. It now compares `typeof typia.createAssertGuard<User>()` against the two-parameter assertion signature directly.
- **Twelve of the eighteen members had declaration-only coverage.** The census was empirical but never encoded, so removing `Guardian.Parameter` from one programmer would have left the declaration promising a parameter the function no longer takes. The full-surface fixture is now transformed as well as type-checked, and each member's emitted body must carry `(input, errorFactory)` while `createIs`, `json.createIsParse` and `plain.createClone` must not.

The shadow fixture also gained a `json.createAssert` call — a root operation spelled inside a namespace — which pins the fallback that must not fire, plus per-code line assertions and a check that every expected code was seen.

**Documentation.** `misc.mdx` carried stale `createAssertClone` and `createAssertClassify` signatures; the `never` overloads of `createAssert` and `createAssertEquals` still advertised `(input: unknown) => T`; the new `assert.mdx` snippet referenced two identifiers that do not exist; and neither the JSDoc nor the guide said that the `AssertionGuard<T>` annotation they mandate is exactly what hides the call-time override they were documenting. All four are corrected.

**Known gaps, recorded rather than hidden.** An alias chain is not followed, so a project that re-exports typia through its own barrel and then shadows it stays silent. Taking the namespace from the spelling means `import { json as j }` is not recognized. Both are stated in the source comments.

**Verification at this head.** `pnpm build`, `pnpm --filter ./benchmark run build`, `go -C packages/typia/test test ../native/...`, `go -C packages/typia/test test ./...`, `ttsc -p tsconfig.json` over `tests/test-typia-schema` and over `examples`, and the four focused Go cases plus both `tests/test-typia-schema` cases. Mutation sensitivity and root `pnpm test` follow before merge.

samchon (commented):
### Self-Review — complete round over `origin/master...HEAD`

One reviewer, whole surface, one recorded state. Every file and hunk in the base-to-head diff was read: the runtime helper, the eighteen declarations across six `packages/typia/src` files, the native transform, five Go test files, two `tests/test-typia-schema` cases, and four website pages. Earlier rounds are not counted as coverage; this is the round over the final head.

#### Mutation sensitivity

Run in a disposable checkout of this exact head, with its own dependency install and its own `TTSC_CACHE_DIR`. Each mutation reverts **only** product code and keeps every test unchanged.

| # | Product mutation | Probe | Adjacent control |
| --- | --- | --- | --- |
| M1 | `_assertGuard` accepts any truthy factory again | `test_assert_error_factory_non_callable` **FAIL** | `test_assert_error_factory_call_time_override` PASS |
| M2 | `createAssert` declares `(input: unknown) => T` again | `TestCreateAssertErrorFactoryArity` **FAIL** | `TestAssertGuardFactoryContract` PASS |
| M2b | `createAssertGuard` returns `AssertionGuard<T>` again | `TestCreateAssertErrorFactoryArity` **FAIL**, `TestAssertGuardFactoryContract` **FAIL** | — |
| M3 | drop the shadowed-typia diagnostic call | `TestShadowedTypiaModuleDeclarationDiagnostic` **FAIL** | `TestLocalFunctionNamedLikeTypiaOperationTransform` PASS |
| M4 | specifier test loses the subpath slash (`typia` prefix instead of `typia/`) | `TestLocalFunctionNamedLikeTypiaOperationTransform` **FAIL** | `TestShadowedTypiaModuleDeclarationDiagnostic` PASS |
| M5 | operation lookup falls back to a flat union of every table | `TestLocalFunctionNamedLikeTypiaOperationTransform` **FAIL** | `TestShadowedTypiaModuleDeclarationDiagnostic` **FAIL** |
| M6 | `plain.createAssertPrune` stops emitting `errorFactory` | `TestCreateAssertErrorFactoryArity` **FAIL** | — |

M5's second row is not a broken control. The shadow fixture's third call, `typia.json.createAssert`, spells a root operation inside a namespace and must not be reported; the flat union reports it, so that fixture is a probe for this mutation too.

`go test` caches a passing result when no Go input changed, so a TypeScript-only mutation reports a cached PASS it never measured. M2 and M2b run with `-count=1`, and a `-count=1` baseline re-run confirms every case passes unmutated at this head.

**Three tests could not fail, and only the matrix said so.** Each is fixed and each mutation above is the run after the fix:

- **M4.** The neighbouring-package decoy imported its function under a *renamed* local binding, so the operation lookup rejected the call before the specifier was ever read. Widening the specifier prefix from `typia/` to `typia` left both Part B tests green. The decoy now imports under its own name.
- **M6.** The emitted-arity census matched `(input, errorFactory)` anywhere in a member's segment, and every composed member also builds a private `__assert` that carries it. Removing `Guardian.Parameter` from `PlainAssertPruneProgrammer` left the census green. It now matches `return (input, errorFactory)`, the public function.
- **M2 / M2b.** `ttscTypiaTestTypecheck` shelled out to `pnpm exec ttsc --noEmit -p tsconfig.json` from a fixture under `.tmp-ttsc-typia-tests`, where that command reports nothing and exits 0 — `const value: number = "bad"` passes. Reverting the `createAssert` declaration therefore left the arity test green despite eighteen two-argument calls and two `@ts-expect-error` directives in its fixture. The helper now runs the same in-process `runBuild --noEmit --rewrite-mode none` these tests already drive, which `build_reports_semantic_diagnostics_test.go` pins as status 2 on a semantic error.

That last one is a pre-existing defect with five callers, and repairing it immediately failed the assertion-guard fixture on a real error it had been hiding: `input.id satisfies number` after a module-scoped `guard(input)`. TypeScript does not apply an assertion signature to a module-scoped binding, so that line never narrowed; the narrowing assertion moved into a function body, where it does. The other four callers pass unchanged.

#### Verification executed

| Command | Result |
| --- | --- |
| `pnpm build` | pass |
| `pnpm --filter ./benchmark --fail-if-no-match run build` (~780 sources, the largest in-repo consumer) | pass |
| `pnpm run test:packages` | one Windows-only failure, below |
| `go -C packages/typia/test test ../native/...` | pass |
| `pnpm exec ttsc -p tsconfig.json` in `tests/test-typia-schema` | pass |
| `pnpm exec ttsc --noEmit` in `examples` | pass |

The last two are not part of any lane: `ttsx` transpiles without checking, so nothing in CI type-checks the suites or the website's example sources against these declarations. They are what proves the widened types do not break the repository's own consumers.

#### The one local failure

`tests/test-typia-generate`, subtest `physical output aliases are excluded` in `scripts/filesystem/output-directory-alias.test.cjs`. It builds a Windows directory junction and drives the `typia generate` CLI with a 180-second per-invocation timeout; the run that failed was sharing the machine with a Go mutation matrix and a second test worktree. Nothing in this diff touches the CLI, its filesystem handling, or its output-directory logic. CI's `test` lane — which runs `pnpm test`, so this suite plus both Go trees — is green on this branch, and the suite is re-run in isolation here.

#### Could not run

The website build. It needs a ttsc checkout at `../ttsc`, staged tarballs from `pnpm package:tgz`, and a Wasm playground build; that provisioning is not available here. The four changed pages are prose and fenced code only, and the `> [!WARNING]` callout matches the `> [!IMPORTANT]` already on the same page.

#### Consequence surface traced

- **Panic safety.** The diagnostic raises `*nativecontext.TransformerError`; `fileTransformer_try_transform_node` recovers exactly that type and restores the node, so traversal continues and one shadowed call yields one diagnostic. `transform.go` is the only path into `FileTransformer`, `NodeTransformer` the only caller of `CallExpressionTransformer.Transform`, and `TypiaGenerator` reaches both through the same recover. No entry point escapes it.
- **Cost.** The gate runs only after `GetResolvedSignature`, which already runs for every call expression and does full overload resolution. `JSON.parse` never reaches the checker at all: `parts[0]` is `JSON`, which is no table, and `parse` is not a root operation.
- **Platforms.** The message normalizes separators with `filepath.ToSlash`; the specifier comparison is literal text.
- **Public contract.** `AssertionGuard<T>` is untouched and its `tests/test-interface` pin still holds. `packages/{langchain,mcp,vercel,utils,interface}` contain no `createAssert*` reference. No package version, dist-tag, or release state changed.

#### Deliberate decisions, stated so they can be overruled

1. **A non-callable second argument does not fall back to the create-time factory.** That factory is the parameter's default, so an index has already overridden it before `_assertGuard` runs. Recovering it means changing the emitter to bind the create-time factory to its own generated constant, which rewrites the emit shape published in `pure.mdx` and every assert-family output — for a case the declaration now rejects at compile time. The chosen behavior is strictly more informative than the `factory is not a function` it replaces, and `test_assert_error_factory_non_callable` pins it.
2. **A project that redirects the `typia` specifier to a fork now fails its build** instead of compiling to untransformed stubs. That build produced no validators either way; surfacing it is the point.
3. **Two known gaps are recorded in the source rather than closed.** An alias chain is not followed, so a barrel re-export plus a shadow stays silent; and taking the namespace from the spelling means `import { json as j }` is not recognized. Both would need symbol-level namespace identity, which widens the gate the issue asked to keep narrow.

#### Clean

This round found nothing further to change.

samchon (commented):
### Self-Review — round over the merged head `0cd3d23dfa`

`master` moved to `fbf7b46f5a` (#2336, the scalar-tag rollback) while this branch was in verification, and the two collided in one place: `ttscTypiaTestWriteCommonRuntimeStubs`. That rollback removed the `_stringLength` reference from the tag validators and with it the `string-length-stub.cjs` entry and its `ttscTypiaTestStringLengthStub` constant; this branch had reformatted the same map literal. Resolved in favour of `master` — the entry is gone, the constant is gone, and nothing under `packages/typia/native` still names `_stringLength`.

The three files this branch touched that #2336 also owns (`template_literal_type_tags_transform_test.go`, `MetadataCommentTagFactory.go`, `metadata_comment_tag_factory_coverage_test.go`) are taken verbatim from `master`. `node scripts/format-go.cjs` reported them as modified afterwards with a zero-line `git diff --numstat`, which is `core.autocrlf` noise, and they were restored so the merge carries `master`'s bytes exactly.

`git diff origin/master...HEAD` is unchanged in shape: the same 19 files, none of them the tag surface.

`go -C packages/typia/test test -count=1 ../native/cmd/ttsc-typia` passes on the merged head. `pnpm format` produces zero real content changes across the repository — `git diff --numstat` is empty, every reported file is line-ending noise.

`tests/test-typia-generate` was re-run alone on an idle machine and exits 0, confirming the earlier `physical output aliases are excluded` failure was contention against that suite's 180-second per-invocation CLI timeout, not a regression.

The merge commit is also what finally gave the head a full check set: the four pushes before it produced no Actions run at all, and only `test` could be dispatched manually. `build`, `website`, `experiments`, `format` and `spell-check` now run against this exact SHA.

Nothing else changed, and this round found nothing to improve.

Issue #2338, opened 2026-08-13
Enforce the exact int64 and uint64 bounds on the bigint path of Type


samchon:
## Problem

On the `bigint` path, `tags.Type<"int64">` emits no check at all and `tags.Type<"uint64">` checks only that the value is not negative. The JSDoc spellings `@type int64` and `@type uint64` do the same. So typia certifies a bigint of any magnitude as an `int64`, and any non-negative bigint as a `uint64`.

```ts
typia.is<bigint & tags.Type<"int64">>(2n ** 200n);    // true
typia.is<bigint & tags.Type<"uint64">>(2n ** 200n);   // true
```

The `number` path for the same tags does enforce its bounds. Only the `bigint` arm is unchecked, which is the arm the documentation recommends for 64-bit values.

## Why it matters

The width is not decorative. Two consumers act on it:

- **Protocol Buffers.** `typia.protobuf.encode` writes an `int64`/`uint64` field as a varint of that width. A bigint outside the range does not fail — it is truncated, so a value typia certified comes back a different number after a round trip.
- **JSON Schema and OpenAPI.** The tag emits `{ "type": "integer" }` with the matching bounds, so every downstream validator, LLM structured-output endpoint, and generated SDK is told a range typia itself does not enforce.

A validator that certifies a value its own encoder will corrupt is the failure worth closing.

## Invariant

For `Value` in `"int64"` and `"uint64"`, the `bigint` arm accepts exactly the values representable in that width and rejects every other bigint:

| Tag | Accepts |
| --- | --- |
| `bigint & Type<"int64">` | `-(2n ** 63n)` through `2n ** 63n - 1n` inclusive |
| `bigint & Type<"uint64">` | `0n` through `2n ** 64n - 1n` inclusive |

The type tag and its JSDoc spelling are two spellings of one constraint and must emit the same check. The constraint is owned by the `Type` declaration in `@typia/interface` and its counterpart in the native metadata factory; the shape of the runtime check is the implementer's decision.

## Constraint: major version only

Two reasons, and either one alone is sufficient.

Tightening an accepted range changes validation outcomes for data that currently passes, so it is a breaking change.

[...]

PR #2339, opened 2026-08-13
fix(interface): stop the int64 and uint64 bigint tag validators from naming a runtime helper

samchon:
Reverts the runtime meaning of the `bigint` arm of `tags.Type<"int64">` and `tags.Type<"uint64">` to the form it had before `b3c926ba27` (#2166). Only the tag declarations and the things that must agree with them move.

## What changes

| Declaration | Before this PR | After this PR |
| --- | --- | --- |
| `tags.Type<"int64">` (bigint) | `$importInternal("isTypeInt64Bigint")($input)` | `true` |
| `tags.Type<"uint64">` (bigint) | `$importInternal("isTypeUint64Bigint")($input)` | `BigInt(0) <= $input` |
| `tags.Type<"int64">` / `tags.Type<"uint64">` (number) | `$importInternal("isTypeInt64")($input)` / `…Uint64…` | unchanged |

`packages/typia/native/core/factories/MetadataCommentTagFactory.go` moves with them, because the JSDoc `@type int64` / `@type uint64` spelling and the type-tag spelling of one constraint must not mean different things. Its `number` records keep `isTypeInt64` and `isTypeUint64`.

Verified in emitted output rather than by reading the declaration:

```js
const _ip32 = input => "bigint" === typeof input["int64Bigint"] && true;
const _ip33 = input => "bigint" === typeof input["uint64Bigint"] && BigInt(0) <= input["uint64Bigint"];
```

The `@type int64` and `@type uint64` spellings emit the identical expressions.

## What deliberately does not change

- `packages/typia/src/internal/_isTypeInt64Bigint.ts` and `_isTypeUint64Bigint.ts` stay published with nothing in `packages/typia/src` referencing them. That is the point: a `typia` that ships them can still satisfy an `@typia/interface` that names them, which is the other direction of the same compatibility problem. The same holds for `_stringLength`, `_isMultipleOf`, `_decimal`, and `_randomMultiple` from #2336. All six still build to `.js`, `.mjs`, and `.d.ts` under `packages/typia/lib/internal/`.
- The `number` path. `isTypeInt64` and `isTypeUint64` are named by `@typia/interface@13.0.0` and shipped by `typia@13.0.0`, so they cross no version boundary. `_isTypeInt64.ts`, `_isTypeUint64.ts`, and `NumericRangeFactory.go` are untouched, and `RandomProgrammer.go` changes only a comment.
- `NumericRangeFactory.Bigint`, which `ProtobufEncodeProgrammer` uses to discriminate a protobuf union with more than one candidate, still emits its inline `BigInt("…") <= $input && $input <= BigInt("…")` comparison. It names no helper and crosses no boundary.
- Random generation, JSON Schema and LLM schema output — the tag's `schema` member is untouched, so unlike #2336 this revert introduces no divergence with `@typia/utils` — and every package version, dist-tag, and release state.

## This is a behavior change

The 64-bit width is no longer enforced on the `bigint` path:

```ts
typia.is<bigint & tags.Type<"int64">>(2n ** 200n);   // was false, is now true
typia.is<bigint & tags.Type<"uint64">>(2n ** 64n);   // was false, is now true
typia.is<bigint & tags.Type<"uint64">>(-1n);         // still false
```

`typia.protobuf.encode` still writes the field as a 64-bit varint, so typia now certifies values its own encoder truncates. #2338 owns restoring the bound. Both the guide and the `Type` declaration's doc comment say so, so the gap is documented rather than implied.

Worth stating precisely, because an earlier draft of this branch got it wrong: what forces #2338 to a major is **not** the packaging boundary. An inline exact comparison — `BigInt("-9223372036854775808") <= $input && …` — names no helper and crosses no boundary, and `metadataCommentTagFactory_parse_type` already puts `&&` expressions in the same `Validate` field. #2338 gives two independent reasons and says the packaging one stops applying once the check is inline. What survives either way is that tightening an accepted range rejects data that validates today.

**The release number this ships under is the maintainer's decision.** This branch does not choose one and does not touch any version field.

## Effect on #2330 — this PR closes it

#2330 is the packaging failure where an older `typia` resolving a newer `@typia/interface` through the caret range emits an import the installed runtime does not ship. #2336 removed `_stringLength` and `_isMultipleOf` from the tag declarations and deliberately carried no closing keyword, because `Type.ts` still named `isTypeInt64Bigint` and `isTypeUint64Bigint`. Those are the last two, and this PR removes them.

The closing keyword sits on `4acaa66467` because the claim was checked, not assumed.

**Why a grep is not sufficient here**, and both failures are real. `\$importInternal\("[A-Za-z_]*"\)` excludes digits, so it silently misses `isTypeInt64Bigint` and returns a confident clean result — the digit-safe form is `\$importInternal\("[A-Za-z0-9_]+"\)`, and it was fed a known-positive string before being trusted. And `packages/interface/src/tags/Format.ts` builds its name by template — `` `$importInternal("isFormat${PascalizeString<Value>}")($input)` `` — so no literal grep finds those 22 names at all.

So the set was read out of real transform output instead. A probe declaring all 22 `Format` values, all 10 `Type` values on `number`, `int64` and `uint64` on `bigint`, both JSDoc spellings, and `UniqueItems` was compiled through `ttsc`, and its emitted `typia/lib/internal/*` imports listed:

| Named helper | Count | Shipped by `typia@13.0.0` |
| --- | --- | --- |
| `_isFormatByte`, `Date`, `DateTime`, `Duration`, `Email`, `Hostname`, `IdnEmail`, `IdnHostname`, `Ipv4`, `Ipv6`, `Iri`, `IriReference`, `JsonPointer`, `Password`, `Regex`, `RelativeJsonPointer`, `Time`, `Uri`, `UriReference`, `UriTemplate`, `Url`, `Uuid` | 22 | yes |
| `_isTypeInt8`, `Uint8`, `Int16`, `Uint16`, `Int32`, `Uint32`, `Int64`, `Uint64`, `Float` | 9 | yes |
| `_isUniqueItems` | 1 | yes |
| `_isTypeInt64Bigint`, `_isTypeUint64Bigint` | 0 — no longer named | no |

The probe also emitted `_assertGuard`, `_createStandardSchema`, and `_validateReport`. Those come from the operation programmers in the native transform, not from any tag declaration, so they ship inside the same `typia` that emits them and cross no boundary — and `typia@13.0.0` ships all three anyway.

All 32 tag-named helpers, and all 35 emitted names, are present in `npm pack typia@13.0.0` and in `npm pack typia@13.0.0-rc.2`, the oldest release #2330 reports against. Re-running the identical comparison with `_isTypeInt64Bigint` and `_isTypeUint64Bigint` added back reports both as missing, so the clean result is a check that can fail, not a pattern that cannot match.

What that does **not** claim: the caret range itself is unchanged, and #2330's suggested fix — pinning `@typia/interface` and `@typia/utils` exactly — is still available and still prevents the whole class. Any future tag declaration that names a new `typia/lib/internal/*` module reopens it. What is closed is the concrete failure: at this head no tag declaration names a module `typia@13.0.0` does not ship, so a floated `@typia/interface` no longer produces an unresolvable import for any release in the 13 line.

## Tests

`tests/test-typia-schema/src/features/tags/test_type_int64_range.ts` and `test_type_uint64_range.ts` pinned the exact inclusive bigint bounds and a protobuf round trip against them. Both are rewritten rather than deleted.

Each file separates three claims. The still-published `_isTypeInt64Bigint` / `_isTypeUint64Bigint` keep the exact bound over the same boundary matrix as before, which turns "kept deliberately" into an executed assertion. The type tag and its `@type` spelling answer what they now declare: `int64` accepts every bigint, `uint64` accepts exactly the non-negative ones — the surviving lower bound is asserted on its own, because it is the only thing separating the reverted `uint64` arm from the reverted `int64` arm. And the out-of-range members of each boundary list are counted with an explicit failure if the count reaches zero, so a list later trimmed to in-range values fails that guard instead of leaving the loop above passing while proving nothing.

The protobuf assertions are why the gap is pinned rather than dropped. With the bigint arm unchecked, typia certifies values its own encoder truncates, so each file carries both outcomes: an in-range value still decodes unchanged, and an accepted out-of-range value decodes as its 64-bit truncation. The expectation comes from the Protocol Buffer wire format — the writer emits `BigInt.asUintN(64, value)` as the payload and the reader signs it back for `int64` — not from typia's own output, and each case first asserts that the truncation genuinely differs from the input, so nothing in that loop can pass by accident. #2338 is named there as the owner.

Every number-path assertion is unchanged; that path is untouched by the revert and is the adjacent control the mutation proof needs. Two negative twins are added: the bigint tag still rejects a non-bigint, which proves `true` and the surviving lower bound replaced the range check and not the type check.

`tests/template/src/structures/**` drives the `test-typia-automated` matrix. Every bigint spoiler there — `TypeTagTypeBigInt`, `CommentTagTypeBigInt`, `ObjectHttpTypeTag`, `ObjectHttpCommentTag`, and the three `ArraySimpleProtobuf*` — sets `-1n`, which the surviving lower bound still rejects, and a scan of that whole tree finds no large-magnitude bigint assignment. Nothing there depends on the removed upper bound, so the fixtures are untouched.

## Verification

Local, at head `b6f1bb2fe5`:

- `pnpm build` — exit 0. `packages/typia/lib/internal/` still contains `_isTypeInt64Bigint`, `_isTypeUint64Bigint`, `_stringLength`, `_isMultipleOf`, `_decimal`, and `_randomMultiple` as `.js`, `.mjs`, and `.d.ts`, with zero references left in `packages/typia/src`. `tsc` compiles the program rather than the reachable graph, so the helpers a newer `typia` needs in order to satisfy an older `@typia/interface` do still ship.
- `pnpm --filter ./tests/test-typia-schema --fail-if-no-match start` — 192 cases, `Success`.
- `pnpm test` — every TypeScript workspace green (18 of 18, including `test-typia-automated` `Success`, `test-feature-identity`, `test-error`, `test-interface`, and both rewritten tag cases), then 17 Go packages `ok` with zero failures before the run was stopped part-way through the native tree, because CI's `test` lane runs the identical `pnpm run test` and had already passed on this same commit. That substitution is called out rather than glossed.
- `go -C packages/typia/test test -tags typia_native_internal ../native/...` — the whole-tree form was abandoned after ~40 minutes on a loaded machine, so it was narrowed to the package this PR edits: `go -C packages/typia/test test -tags typia_native_internal -v -run TestMetadataCommentTagFactory ../native/core/factories` — `TestMetadataCommentTagFactoryCoverage` and `TestMetadataCommentTagFactoryExclusivityMatchesInterface` both `PASS`, package `ok`. That file is the tagged coverage for `metadataCommentTagFactory_parse_type`, the one function this PR changes; root `pnpm test` does not enable the tag and neither does CI, so the rest of the tagged tree is unrun and reported as such.
- `pnpm format` — the only files it left with a nonzero `git diff --numstat` were the ones edited by hand.

**CI, all green on `b6f1bb2fe5`:** `format` 42s, `test` 14m21s, `experiments` 5m19s, `build` 3m8s, `website` 12m0s, `spell-check` 7s, and both Socket Security checks. The `website` lane builds the guide page this branch edits.

**Mutation sensitivity, in a disposable checkout.** A detached `git worktree` at `b6f1bb2fe5` received the product change only — the two `bigint` arms of `Type.ts` back to `$importInternal("isTypeInt64Bigint")` / `isTypeUint64Bigint`, and `MetadataCommentTagFactory.go`'s `bigintValidate` back to the same two calls, two files and four lines, both test files untouched. It got its own `pnpm install`, `pnpm build`, and `TTSC_CACHE_DIR`. The schema suite there exited 127 with **192 cases, exactly 2 errored and 190 passed**:

```
- test_type_int64_range: Error
- test_type_uint64_range: Error

Error: Bug on bigint type tag accepts -9223372036854775809   ({ x: true, y: false })
  at test_type_int64_range (…/test_type_int64_range.ts:139:29)
Error: Bug on bigint type tag on 18446744073709551616 === true ({ x: true, y: false })
  at test_type_uint64_range (…/test_type_uint64_range.ts:138:29)
```

Each fires on the first out-of-range member of its boundary list, on exactly the assertion that encodes the reverted behavior, while every adjacent control passes. The checkout was deleted and pruned afterwards; the deliverable tree stayed clean at `b6f1bb2fe5`.

**The enumeration has a live known-positive control.** The probe was compiled in both checkouts. This head emits 35 `typia/lib/internal/*` names, the mutated checkout emits 37, and the difference is exactly `_isTypeInt64Bigint` and `_isTypeUint64Bigint` — the same two the tarball comparison reports as missing from `typia@13.0.0` for the mutated tree and reports as none for this head. The check demonstrably fails when a helper is genuinely missing, on a real build rather than a hand-written string.

### DIFF
--- a/packages/interface/src/tags/Type.ts
+++ b/packages/interface/src/tags/Type.ts
@@ -25,8 +25,8 @@ import { TagBase } from "./TagBase";
  * The constraint is enforced at runtime by `typia.is()`, `typia.assert()`, and
  * `typia.validate()`. It generates appropriate `type` in JSON Schema.
  *
- * On a `bigint`, `"int64"` and `"uint64"` select the protobuf scalar type but do
- * not check the 64-bit width: `"int64"` enforces nothing, and `"uint64"`
+ * On a `bigint`, `"int64"` and `"uint64"` select the protobuf scalar type but
+ * do not check the 64-bit width: `"int64"` enforces nothing, and `"uint64"`
  * enforces only `0n <= value`. Their `number` forms do enforce their range.
  *
  * @author Jeongho Nam - https://github.com/samchon
--- a/packages/interface/src/tags/Type.ts
+++ b/packages/interface/src/tags/Type.ts
@@ -25,6 +25,10 @@ import { TagBase } from "./TagBase";
  * The constraint is enforced at runtime by `typia.is()`, `typia.assert()`, and
  * `typia.validate()`. It generates appropriate `type` in JSON Schema.
  *
+ * On a `bigint`, `"int64"` and `"uint64"` select the protobuf scalar type but do
+ * not check the 64-bit width: `"int64"` enforces nothing, and `"uint64"`
+ * enforces only `0n <= value`. Their `number` forms do enforce their range.
+ *
  * @author Jeongho Nam - https://github.com/samchon
  * @example
  *   interface Message {
@@ -69,12 +73,12 @@ export type Type<
               : Value extends "int64"
                 ? {
                     number: `$importInternal("isTypeInt64")($input)`;
-                    bigint: `$importInternal("isTypeInt64Bigint")($input)`;
+                    bigint: `true`;
                   }
                 : Value extends "uint64"
                   ? {
                       number: `$importInternal("isTypeUint64")($input)`;
-                      bigint: `$importInternal("isTypeUint64Bigint")($input)`;
+                      bigint: `BigInt(0) <= $input`;
                     }
                   : Value extends "float"
                     ? `$importInternal("isTypeFloat")($input)`
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -411,7 +411,7 @@ func metadataCommentTagFactory_parse_type(props struct {
     // `Type<"uint64">` declare them, because `@type int64` and
     // `bigint & Type<"int64">` are two spellings of one constraint.
     //
-    // That declaration may not name a runtime helper the way the `number` arms
+    // That declaration must not name a runtime helper the way the `number` arms
     // above do. A `$importInternal(...)` in a tag template makes the emitted
     // validator import a `typia/lib/internal/*` module, and the declaration
     // lives in `@typia/interface`, which `typia` depends on through a caret
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -411,17 +411,21 @@ func metadataCommentTagFactory_parse_type(props struct {
     // `Type<"uint64">` declare them, because `@type int64` and
     // `bigint & Type<"int64">` are two spellings of one constraint.
     //
-    // That declaration cannot name a runtime helper. A `$importInternal(...)`
-    // in a tag template makes the emitted validator import a
-    // `typia/lib/internal/*` module, and the declaration lives in
-    // `@typia/interface`, which `typia` depends on through a caret range that
-    // only floats upward -- so an older `typia` installs a newer
+    // That declaration may not name a runtime helper the way the `number` arms
+    // above do. A `$importInternal(...)` in a tag template makes the emitted
+    // validator import a `typia/lib/internal/*` module, and the declaration
+    // lives in `@typia/interface`, which `typia` depends on through a caret
+    // range that only floats upward -- so an older `typia` installs a newer
     // `@typia/interface` and emits an import its own runtime never shipped
-    // (#2330). Inside one major there is no range that prevents that pairing.
+    // (#2330). `isTypeInt64` and `isTypeUint64` are safe there because
+    // `@typia/interface` 13.0.0 already named them and `typia` 13.0.0 already
+    // shipped them; a helper introduced later is not, and inside one major
+    // there is no range that prevents the pairing.
     //
     // The bound this leaves unenforced is real: int64 accepts any magnitude and
-    // uint64 only rejects negatives. #2338 owns restoring it in a major
-    // release, the one boundary a caret cannot cross.
+    // uint64 only rejects negatives. #2338 owns restoring it, and only a major
+    // can carry it -- an inline exact comparison would name no helper, but
+    // tightening an accepted range rejects data that passes today.
     bigintValidate := "BigInt(0) <= $input"
     if value == "int64" {
       bigintValidate = "true"
--- a/packages/typia/native/core/factories/MetadataCommentTagFactory.go
+++ b/packages/typia/native/core/factories/MetadataCommentTagFactory.go
@@ -407,13 +407,24 @@ func metadataCommentTagFactory_parse_type(props struct {
     "number": {{Name: "Type<" + strconv.Quote(value) + ">", Target: "number", Kind: "type", Value: value, Validate: validate, Exclusive: metadataCommentTagFactory_exclusive("type"), Schema: numberSchema}},
   }
   if value == "int64" || value == "uint64" {
-    // The bigint form used to emit `true` for int64 and a lower bound only for
-    // uint64, so `bigint & Type<"int64">` -- the spelling the documentation
-    // recommends -- certified any magnitude at all. Both now delegate to the
-    // helper that holds the exact inclusive bound a bigint can represent.
-    bigintValidate := "$importInternal(\"isTypeUint64Bigint\")($input)"
+    // Both checks stay inline, spelled exactly as `Type<"int64">` and
+    // `Type<"uint64">` declare them, because `@type int64` and
+    // `bigint & Type<"int64">` are two spellings of one constraint.
+    //
+    // That declaration cannot name a runtime helper. A `$importInternal(...)`
+    // in a tag template makes the emitted validator import a
+    // `typia/lib/internal/*` module, and the declaration lives in
+    // `@typia/interface`, which `typia` depends on through a caret range that
+    // only floats upward -- so an older `typia` installs a newer
+    // `@typia/interface` and emits an import its own runtime never shipped
+    // (#2330). Inside one major there is no range that prevents that pairing.
+    //
+    // The bound this leaves unenforced is real: int64 accepts any magnitude and
+    // uint64 only rejects negatives. #2338 owns restoring it in a major
+    // release, the one boundary a caret cannot cross.
+    bigintValidate := "BigInt(0) <= $input"
     if value == "int64" {
-      bigintValidate = "$importInternal(\"isTypeInt64Bigint\")($input)"
+      bigintValidate = "true"
     }
     record["bigint"] = []schemametadata.IMetadataTypeTag{
       {Name: "Type<" + strconv.Quote(value) + ">", Target: "bigint", Kind: "type", Value: value, Validate: bigintValidate, Exclusive: metadataCommentTagFactory_exclusive("type"), Schema: bigintSchema},
--- a/packages/typia/native/core/programmers/RandomProgrammer.go
+++ b/packages/typia/native/core/programmers/RandomProgrammer.go
@@ -1444,9 +1444,9 @@ func randomProgrammer_is_typed_array(name string) bool {
 // two. That costs nothing here: the generator only picks a value from the range,
 // and the TypedArray constructor wraps whatever it is handed. On the `number`
 // path validation owns exactness instead, in _isTypeInt64 and _isTypeUint64. The
-// `bigint` path owns none: its type tag declares no width check at all, so that
-// the declaration names no runtime helper an older `typia` cannot resolve
-// (#2330), and #2338 owns restoring the bound.
+// `bigint` path owns none: Type<"int64"> declares no check and Type<"uint64">
+// only a lower bound, so that neither declaration names a runtime helper an
+// older `typia` cannot resolve (#2330). #2338 owns restoring the bound.
 func randomProgrammer_typed_array_range(name string) (string, string, string) {
   switch name {
   case "Uint8Array", "Uint8ClampedArray":
--- a/packages/typia/native/core/programmers/RandomProgrammer.go
+++ b/packages/typia/native/core/programmers/RandomProgrammer.go
@@ -1442,8 +1442,11 @@ func randomProgrammer_is_typed_array(name string) bool {
 // `maximum` comment tags on a number-typed JSON schema, and `number` cannot
 // represent 2**63 - 1 or 2**64 - 1 -- both arrive rounded up to the next power of
 // two. That costs nothing here: the generator only picks a value from the range,
-// and the TypedArray constructor wraps whatever it is handed. Validation owns
-// exactness instead, in _isTypeInt64Bigint and its siblings.
+// and the TypedArray constructor wraps whatever it is handed. On the `number`
+// path validation owns exactness instead, in _isTypeInt64 and _isTypeUint64. The
+// `bigint` path owns none: its type tag declares no width check at all, so that
+// the declaration names no runtime helper an older `typia` cannot resolve
+// (#2330), and #2338 owns restoring the bound.
 func randomProgrammer_typed_array_range(name string) (string, string, string) {
   switch name {
   case "Uint8Array", "Uint8ClampedArray":
--- a/tests/test-typia-schema/src/features/tags/test_type_int64_range.ts
+++ b/tests/test-typia-schema/src/features/tags/test_type_int64_range.ts
@@ -24,21 +24,23 @@ interface ICommentBigint {
 /**
  * Verifies int64 bounds the number path and leaves the bigint path unchecked.
  *
- * The bigint arm may not name a `typia/lib/internal/*` helper. It is declared in
- * `@typia/interface`, which `typia` admits through a caret range that only
+ * The bigint arm may not name a `typia/lib/internal/*` helper. It is declared
+ * in `@typia/interface`, which `typia` admits through a caret range that only
  * floats upward, so a named helper makes an older `typia` emit an import it
- * never shipped (#2330). The arm is therefore the bare `true` it declared before
- * #2166, which enforces nothing -- and an unenforced bound is invisible unless
- * the gap is pinned, so this file asserts the accepted out-of-range values and
- * the protobuf round trips that then corrupt them. #2338 owns restoring the
- * bound in a major release, the one boundary a caret cannot cross.
+ * never shipped (#2330). The arm is therefore the bare `true` it declared
+ * before #2166, which enforces nothing -- and an unenforced bound is invisible
+ * unless the gap is pinned, so this file asserts the accepted out-of-range
+ * values and the protobuf round trips that then corrupt them. #2338 owns
+ * restoring the bound, and only a major can carry it: tightening an accepted
+ * range rejects data that passes today.
  *
  * 1. Keep the enforced number path, whose helper crosses no package boundary.
- * 2. Require the bigint type tag and `@type int64` to accept every bigint, while
- *    the still-published `_isTypeInt64Bigint` keeps the exact bound.
- * 3. Pin both protobuf outcomes against the wire format: an in-range value
- *    survives byte for byte, an accepted out-of-range value comes back as its
- *    two's-complement 64-bit truncation.
+ * 2. Require the bigint type tag and `@type int64` to accept every bigint and
+ *    still reject a non-bigint, while the still-published `_isTypeInt64Bigint`
+ *    keeps the exact bound.
+ * 3. Pin both protobuf outcomes against the wire format: an in-range value decodes
+ *    unchanged, an accepted out-of-range value decodes as its two's-complement
+ *    64-bit truncation.
  */
 export const test_type_int64_range = (): void => {
   const MINIMUM: bigint = -(2n ** 63n);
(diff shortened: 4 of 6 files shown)


samchon (commented):
**Ledger — `4acaa66467` `fix(interface): stop the int64 and uint64 bigint tag validators from naming a runtime helper`**

What landed:

- `packages/interface/src/tags/Type.ts` — the two `bigint` arms return to their `b3c926ba27^` form: `` `true` `` for `int64`, `` `BigInt(0) <= $input` `` for `uint64`. The `number` arms keep `isTypeInt64` / `isTypeUint64` untouched. The declaration's doc comment now states the bigint arms do not check the width.
- `packages/typia/native/core/factories/MetadataCommentTagFactory.go` — the `@type int64` / `@type uint64` bigint record emits the same two inline expressions, so the JSDoc spelling and the type-tag spelling of one constraint stay identical. The block comment that argued for the helper form is replaced by the compatibility reason and a pointer to #2338.
- `website/src/content/docs/validators/tags.mdx` — one paragraph stating that the bigint forms of both tags do not check the 64-bit width, that `protobuf.encode` truncates instead of rejecting, and that #2338 owns restoring the bound.

Deliberately not touched: `packages/typia/src/internal/_isTypeInt64Bigint.ts` and `_isTypeUint64Bigint.ts` (now unreferenced and still published, so a newer `typia` can satisfy an older `@typia/interface` that names them), `NumericRangeFactory.go`, `RandomProgrammer.go`, `_isTypeInt64.ts`, `_isTypeUint64.ts`, and every package version.

Issue resolved: #2330. The closing keyword sits on this commit because this is the commit that earns it — the enumeration below is what the PR body will carry in full.

**Enumeration, run rather than assumed.** The helper set was read out of real transform output, not grepped: a probe declaring all 22 `Format` values, all 10 `Type` values on `number`, `int64`/`uint64` on `bigint`, both JSDoc spellings, and `UniqueItems` was compiled through `ttsc`, and its emitted `typia/lib/internal/*` imports were listed. That yields 32 tag-named helpers plus `_assertGuard`, `_createStandardSchema`, and `_validateReport`, which the operation programmers emit rather than any tag. All 35 are shipped by `typia@13.0.0` and by `typia@13.0.0-rc.2`. Re-running the same comparison with `_isTypeInt64Bigint` and `_isTypeUint64Bigint` added back does report both as missing, so the clean result is a real check and not a pattern that cannot match.

Tests for this commit land next; `test-typia-schema` already passes at 192 cases against this head.

samchon (commented):
**Ledger — `2047e1fe8b` `test(typia-schema): pin the unenforced int64 and uint64 bigint bounds`**

`tests/test-typia-schema/src/features/tags/test_type_int64_range.ts` and `test_type_uint64_range.ts` pinned the exact inclusive bigint bounds that `4acaa66467` removes. Both are rewritten, not deleted, so the reverted behavior stays covered from both directions.

What each file now separates:

- The still-published `_isTypeInt64Bigint` / `_isTypeUint64Bigint` keep the exact bound over the same boundary matrix as before — minimum, maximum, one step outside each, zero, and a value far outside the width. That is what makes "kept deliberately" an executed claim rather than a note.
- The type tag and its `@type` spelling answer what they now declare: `int64` accepts every bigint, `uint64` accepts exactly the non-negative ones. The surviving lower bound is asserted on its own, because it is the only thing separating the reverted `uint64` arm from the reverted `int64` arm.
- The values where helper and tag disagree are derived by filter, checked for emptiness, and asserted to disagree. A boundary list later edited down to in-range values would fail that guard instead of passing vacuously.

The protobuf section is why the gap is pinned rather than dropped. With the bigint arm unchecked, typia certifies values its own encoder truncates, so each file carries both outcomes: an in-range value still round trips byte for byte, and an accepted out-of-range value comes back as its 64-bit truncation. The expectation is taken from the Protocol Buffer wire format — `BigInt.asIntN(64, value)` for `int64`, `BigInt.asUintN(64, value)` for `uint64` — not from typia's output, and each case first asserts the truncation genuinely differs from the input, so no assertion in that loop can pass by accident. #2338 is named there as the owner.

Every number-path assertion is unchanged. That path is untouched by the revert and is the adjacent control the mutation proof needs. Two negative twins are added: the bigint tag still rejects a non-bigint, which proves `true` and the lower bound replaced the range check and not the type check.

Local result at this head: `pnpm --filter ./tests/test-typia-schema --fail-if-no-match start` — 192 cases, `Success`.

samchon (commented):
**Ledger — corrections from the two commit early-warning passes**

Both landed commits were read by a read-only early-warning pass. Neither pass is a Self-Review; the whole-surface round still runs before merge. Six candidates survived verification and are corrected in the next commit, and three were verified and rejected.

**Accepted.**

1. **The `format` lane was red on `2047e1fe8b`.** `pnpm format` had not run yet — `prettier` explodes two call sites and one array literal in `test_type_int64_range.ts` and re-wraps two doc-comment paragraphs. Fixed by the formatted commit; the lane is the gate for it.

2. **A duplicate assertion title.** `` `bigint type tag rejects the non-bigint ${String(value)}` `` renders identically for the number `0` and the string `"0"`, so a failure named neither. Now `JSON.stringify(value)`, which distinguishes `0`, `"0"`, and `null`.

3. **Three assertion loops that could not fail.** The `notEquals` loop over the out-of-range members, and the uint64 oversized and negative twins, are implied by the loop directly above them: it already asserts `_isTypeInt64Bigint(value) === oracle(value)` and the tag's answer on the same values, so the disagreement follows by construction — and `TestValidator` throws on first failure, so under a revert the loop above throws and these never execute. They had zero detection power. Removed. The emptiness guard they nominally protected is what actually earns its place, so it stays as an explicit count and `throw`: a boundary list later trimmed to in-range values would otherwise leave the main loop passing while proving nothing.

4. **"Wire form" was the wrong noun.** The comment said an out-of-range `int64` value's wire form is `BigInt.asIntN(64, value)`. The payload is `BigInt.asUintN(64, value)` (`_ProtobufWriter.variant64`); `asIntN` is what the *reader* produces (`_ProtobufReader.int64`). The arithmetic was right, the description was not. Both files now say the value *decodes as* the truncation, and the assertion titles follow. "Survives byte for byte" is also gone, since nothing in either file inspects bytes.

5. **`MetadataCommentTagFactory.go` stated an absolute rule its own function contradicts.** "That declaration cannot name a runtime helper" sits 24 lines below `validate = "$importInternal(\"isTypeInt64\")($input)"` on the `number` arm of the same tag. The real invariant is narrower — it may not name a helper that a `typia` inside the caret range does not already ship — and the comment now says that, naming `isTypeInt64` / `isTypeUint64` as the safe case and why.

6. **The major-release gate was attributed to the wrong cause.** The comment said #2338 waits for a major because that is "the one boundary a caret cannot cross". An inline exact comparison — `BigInt("-9223372036854775808") <= $input && ...` — names no helper and crosses no boundary, and the same `Validate` field already carries `&&` expressions a few lines above, so the caret is not what blocks it. #2338 itself gives two independent reasons and says the packaging one stops applying if the check stays inline. What survives either way is that tightening an accepted range rejects data that validates today. The Go comment, both test doc comments, and the `validators/tags` paragraph now say that instead.

**Also corrected, and worth calling out because it is outside the file set this branch set out to touch.** `RandomProgrammer.go`'s `randomProgrammer_typed_array_range` justified its deliberately inexact 64-bit generation hint with "Validation owns exactness instead, in `_isTypeInt64Bigint` and its siblings." After `4acaa66467` no generated validator reaches that helper, so the sentence this branch left behind was false, and `BigInt64Array` is exactly the case it covers. The correction is comment-only: it now credits `_isTypeInt64` / `_isTypeUint64` on the `number` path and states that the `bigint` path enforces no width, with both issue numbers. No behavior in that file changes, and the number path stays untouched.

**Verified and rejected.**

- **`4acaa66467` does not bisect clean** — at that commit the two test files still asserted the removed enforcement, and `2047e1fe8b` rewrote them. True, and the same shape as #2336. The pull request is the integrated unit and merges as a squash, so `master` never observes the intermediate state. Left as is rather than rewriting pushed history the ledger above already references by SHA.
- **`template_fully_constrained` treats the literal `"true"` as a validating tag** because it only tests `Validate != ""`, so `` `${bigint & tags.Type<"int64">}` `` counts as fully constrained while enforcing nothing. That is the pre-#2166 behavior being restored, and `Type<"double">` has always had the same vacuous `true`, so nothing regresses against 13.0.0. Restoring it is the point of this branch, not a defect of it.
- **Six assertions that compute only from test-local constants** — the `oracle(value)` guards and the `notEquals(value, truncated)` checks. They are deliberate fixture guards, not product assertions wearing a disguise: each one is what stops a later edit from silently moving a value out of the category its loop assumes. The two files keep the same idiom the pre-existing oracle self-checks use.

One correction the ledger cannot make in place: `4acaa66467`'s message carries the same "the one boundary a caret cannot cross" phrasing corrected in point 6. The commit is pushed and referenced by SHA above, so the correction is recorded here rather than by a rewrite.

samchon (commented):
**Self-Review round 1 — snapshot `444891c1ac`, base `ecf764c3a3`**

A complete round over the whole base-to-head diff: six files, every hunk, plus the consequence surface — the declaration's consumers, the native factory, the emitted validator, the protobuf writer/sizer/reader chain, the `tests/template` fixtures that drive the automated matrix, `tests/test-error`, `tests/test-interface`, `@typia/utils`, `examples/`, the guide, and the CI lanes.

Two improvements survived verification and landed as `b6f1bb2fe5`. Both are comment-only.

1. **"May not name a runtime helper" states a prohibition in words that read as a prediction.** It appears in `MetadataCommentTagFactory.go` and in both test doc comments, and in each place the rule is absolute for the bigint arms. All three say "must not" now.

2. **`RandomProgrammer.go`'s corrected note over-generalized.** It said the bigint path "declares no width check at all". That is right for `Type<"int64">` and wrong for `Type<"uint64">`, whose arm still rejects negatives — and `randomProgrammer_typed_array_range` covers `BigUint64Array` as well as `BigInt64Array`, so the imprecision lands on a case the comment is about. It names both arms separately now.

**Verified with no finding.**

- **Declaration and factory emit the same two strings.** `Type.ts` declares `` `true` `` and `` `BigInt(0) <= $input` ``; `MetadataCommentTagFactory.go` sets exactly those. Both match `b3c926ba27^` byte for byte. Confirmed in emitted output rather than by reading: a compiled probe produces `const _ip32 = input => "bigint" === typeof input["int64Bigint"] && true;` and `const _ip33 = input => "bigint" === typeof input["uint64Bigint"] && BigInt(0) <= input["uint64Bigint"];`, and the `@type` spellings emit the identical expressions.
- **No fixture depends on the removed upper bound.** Every bigint spoiler in `tests/template/src/structures/**` — `TypeTagTypeBigInt`, `CommentTagTypeBigInt`, `ObjectHttpTypeTag`, `ObjectHttpCommentTag`, the three `ArraySimpleProtobuf*` — sets `-1n`, which the surviving lower bound still rejects. A scan for any assignment of a large-magnitude bigint across that whole tree returns nothing. `test-error` has no fixture on this tag, `test-interface` asserts no `validate` string, and `metadata_comment_tag_factory_coverage_test.go` asserts `.Value`, never `.Validate`.
- **`NumericRangeFactory.Bigint` is genuinely out of scope.** Its only caller is `ProtobufEncodeProgrammer`, and only when a protobuf union has more than one candidate; it emits inline `BigInt("…")` comparisons, so it names no helper and carries no #2330 exposure. `TestNumericRangeFactoryEnforces64BitBounds` tests that function alone and is unaffected.
- **No divergence with `@typia/utils` is introduced.** Unlike #2336, the tag's `schema` member is untouched, so `json.schema`, `llm.schema`, and the OpenAPI validators see exactly what they saw before. `packages/utils` has no `$importInternal` and no 64-bit bigint range check.
- **The public-type change is the intended one.** `Type<"int64">`'s `validate` member is a different literal type than before, which is the same class of change #2336 made and the mechanism by which the fix reaches an older `typia` at all.
- **No package version, dist-tag, or release state is touched.** The diff is six files, none of them a manifest.

Round 2 is running against `b6f1bb2fe5` alongside the full verification matrix.

samchon (commented):
**Self-Review round 2 — snapshot `b6f1bb2fe5`, base `ecf764c3a3`. Clean.**

A fresh complete round from the current state, not a recheck of round 1's fixes. Read every hunk of the six-file base-to-head diff and both rewritten test files end to end, then re-traced the consequence surface: the declaration's consumers, the native comment-tag factory, the emitted validator, the protobuf writer/sizer/reader chain, the `tests/template` fixtures behind the automated matrix, `test-error`, `test-interface`, `@typia/utils`, `examples/`, the guide, and every CI lane.

**Nothing survived verification.** No sound improvement, no defect.

What the round re-established, with evidence rather than recollection:

- **The enumeration carries forward to this exact head.** Between `4acaa66467` (where the helper set was measured) and `b6f1bb2fe5`, every changed line in `packages/interface/src/` and `MetadataCommentTagFactory.go` is a comment line — filtering the diff for non-comment `+`/`-` lines returns nothing. The declaration is byte-identical, so the measured set is the set at merge. It was then re-measured here anyway; see the mutation review below.
- **No fixture depends on the removed upper bound.** Every bigint spoiler under `tests/template/src/structures/**` sets `-1n`, which the surviving lower bound still rejects, and a scan of that tree for any large-magnitude bigint assignment returns nothing.
- **The two files' guards are the part that can fail.** The genuine mutation detectors are `bigint type tag accepts ${value}` / `bigint comment tag accepts ${value}` in the int64 file, the `=== ${MINIMUM <= value}` pair in the uint64 file, and `typia certifies the …` in both protobuf sections. The mutation run below confirms which line fires first in each.
- **`packages/typia/lib/internal/` still ships all six deliberately-kept helpers** — `_isTypeInt64Bigint`, `_isTypeUint64Bigint`, `_stringLength`, `_isMultipleOf`, `_decimal`, `_randomMultiple` — as `.js`, `.mjs`, and `.d.ts`, with zero references remaining anywhere in `packages/typia/src`.

**Mutation sensitivity — done in a disposable checkout, not in place.** A detached `git worktree` at `b6f1bb2fe5` received the product change only: the two `bigint` arms of `Type.ts` back to `$importInternal("isTypeInt64Bigint")` / `isTypeUint64Bigint`, and `MetadataCommentTagFactory.go`'s `bigintValidate` back to the same two calls. `git diff --stat` there was two files, four insertions, four deletions; both test files untouched. It got its own `pnpm install`, its own `pnpm build`, and its own `TTSC_CACHE_DIR`, so nothing was shared with the deliverable tree but the object store.

`pnpm --filter ./tests/test-typia-schema --fail-if-no-match start` there: exit 127, **192 cases, exactly 2 errored, 190 passed**.

```
- test_type_int64_range: Error
- test_type_uint64_range: Error

Error: Bug on bigint type tag accepts -9223372036854775809   ({ x: true, y: false })
  at test_type_int64_range (…/test_type_int64_range.ts:139:29)
Error: Bug on bigint type tag on 18446744073709551616 === true ({ x: true, y: false })
  at test_type_uint64_range (…/test_type_uint64_range.ts:138:29)
```

Both fire on the first out-of-range member of their boundary list — `MINIMUM - 1n` for int64, `MAXIMUM + 1n` (2^64) for uint64 — on exactly the assertion that encodes the reverted behavior. Every adjacent control passed, including `test_type_float_range`, `test_validate_decimal_multiple_of`, `test_validate_unicode_string_length`, and the whole protobuf group. The checkout was then deleted and its worktree registration pruned; `git status` in the deliverable tree is empty and `HEAD` is unmoved at `b6f1bb2fe5`.

**Enumeration re-measured at this head, with a live known-positive control.** The same probe was compiled through `ttsc` in both checkouts. The frozen head emits 35 `typia/lib/internal/*` names; the mutated checkout emits 37. The difference is exactly `_isTypeInt64Bigint` and `_isTypeUint64Bigint` — which is also exactly the set the tarball comparison reports as missing from `typia@13.0.0` when run against the mutated tree, and empty when run against this head. The check is therefore demonstrably able to fail, on a real build rather than a hand-written control string.

**CI on this exact head, all green:** `format` 42s, `test` 14m21s (the full `pnpm run test`, Node 24 / Go 1.26, both Go trees), `experiments` 5m19s, `build` 3m8s, `website` 12m0s, `spell-check` 7s, and both Socket Security checks. The `website` lane matters here because it builds the guide page this branch edits.

Round 2 is complete and clean, so the review gate is satisfied.

PR #2341, opened 2026-08-16
[feat] add country code format implementation

vedanshshetti:
Hi,
I added a new feature to the tags.Format\<"..."\> thing, 
it basically just checks if the passed string is an **ISO 3166-1 alpha-2** compliant
country code by comparing against a list of the country codes I typed by hand.

> Note: This is missing unit tests, I didn't really understand how to add those, could you please add those?


samchon (commented):
Thank you for taking the time to put this together, and for hand-typing the entire ISO 3166-1 alpha-2 list. I diffed it against the official registry: 249 codes, no omissions, no extras, no duplicates. That is genuinely careful work, and I appreciate it.

I am going to close this, but the reason is scope rather than quality. Let me explain the boundary so it is useful to you.

## `Format.Value` is a closed set of standardized formats

Every value currently in the union comes from a specification:

- **JSON Schema Validation 2020-12 §7.3** — `date-time`, `date`, `time`, `duration`, `email`, `idn-email`, `hostname`, `idn-hostname`, `ipv4`, `ipv6`, `uri`, `uri-reference`, `iri`, `iri-reference`, `uuid`, `uri-template`, `json-pointer`, `relative-json-pointer`, `regex`
- **OpenAPI** — `byte`, `password` (plus `url`, which is non-standard but near-universal)

Neither the JSON Schema spec nor the [OpenAPI Formats Registry](https://spec.openapis.org/registry/format/) defines a country-code format.

That boundary matters because `format` is not just a validator switch — it is emitted into the OpenAPI and LLM schemas typia generates. `format: "countryCode"` is an annotation no downstream consumer understands. And once one non-standard domain format is accepted, `isbn`, `credit-card`, `phone`, `postal-code`, `bic`, and `vat` all follow by exactly the same argument, each with its own hand-maintained data table.

There is also a maintenance dimension: ISO 3166-1 is revised over time (`AN` was withdrawn, `SZ` and `MK` were renamed), so a table in typia's core becomes a standing obligation on this repository.

## What I would suggest instead

Both approaches below work today, in your own codebase, with no patch to typia.

**A literal union is strictly stronger than a format** — it is enforced at compile time, not only at runtime:

```ts
type CountryCode = "KR" | "US" | "JP" /* ... */;
```

**Or a custom tag**, if you want it to behave exactly like a built-in `Format` and to emit a schema annotation:

```ts
import { tags } from "typia";

type CountryCode = string &
  tags.TagBase<{
    kind: "countryCode";
    target: "string";
    value: undefined;
    validate: `/^(AD|AE|AF|AG|AI|AL|AM|AO|AQ|AR|AS|AT|AU|AW|AX|AZ|...)$/.test($input)`;
    schema: { format: "country-code" };
  }>;
```

`tags.TagBase` is documented public API — see [Customization](https://typia.io/docs/validators/tags/#customization). One note that is easy to trip on: `validate` must be an **expression string** referencing `$input` (a template-literal type), not a function reference. The transformer inlines that string into the generated validator. For anything with branching, wrap it in an IIFE — the built-in `date-time` and `iri` validators are written exactly that way.

`schema` accepts any annotation you like, because `OpenApi.IJsonSchema.IString["format"]` is an open union (`| (string & {})`), so your generated OpenAPI document can still carry `format: "country-code"`.

## For future reference

If a format ever is added to `Format.Value`, the union entry is only the first of nine places that must change together, so I want to note them here rather than leave it implicit:

1. `packages/interface/src/tags/Format.ts` — the union **and** the JSDoc list above it
2. `packages/typia/src/internal/_isFormatXxx.ts` — the real validator. `Format` compiles to `$importInternal("isFormat" + Pascalize<Value>)`, which resolves to `typia/lib/internal/_isFormatXxx`. Without this file, user code fails with TS2307 and no diagnostic from the transformer.
3. `packages/typia/src/internal/_randomFormatXxx.ts` — for `typia.random`
4. `packages/interface/src/utils/IRandomGenerator.ts` — the corresponding generator method
5. `packages/typia/native/core/factories/FormatCheatSheet.go` — so the `@format xxx` comment tag works too
6. `packages/utils/src/validators/functional/` — the runtime schema-driven validator path (what this PR modified)
7. `tests/template/src/structures/TypeTagFormat.ts` and `CommentTagFormat.ts`
8. `tests/test-typia-schema/src/features/random/test_random_format_variation.ts`
9. `website/src/content/docs/validators/tags.mdx`

As submitted, paths 2 through 5 are missing, so `typia.is<string & tags.Format<"countryCode">>()` and `typia.random<...>()` would not compile in a user project. That is not a criticism of your patch — it is a real gap in our documentation that this PR usefully exposed, and I will address it.

Two smaller things worth carrying into your own implementation: the union uses kebab-case throughout (`date-time`, `idn-email`), so `country-code` rather than `countryCode`; and a 249-element `Array.includes` is a linear scan, where a `Set` or a regex is what the rest of the validators use.

Thanks again for the contribution and for the clear write-up in the description.

PR #2342, opened 2026-08-16
add IBAN format implementation to allow for tags.Format<"iban">

vedanshshetti:
Hi,
I added a new feature to the tags.Format\<"..."\> thing, 
it basically just checks if the passed string is an **ISO 13616-1** compliant
IBAN by checking for correct length, valid country code and correct Mod97 Checksum

> Note: 
> - This is missing unit tests, I didn't really understand how to add those, could you please add those?
> - I used AI for a few things.

### DIFF
--- a/packages/interface/src/tags/Format.ts
+++ b/packages/interface/src/tags/Format.ts
@@ -45,6 +45,7 @@ export namespace Format {
     | "email"
     | "hostname"
     | "idn-email"
+    | "iban"
     | "idn-hostname"
     | "iri"
     | "iri-reference"
--- a/packages/utils/src/validators/functional/_isIBAN.ts
+++ b/packages/utils/src/validators/functional/_isIBAN.ts
@@ -1,0 +1,133 @@
+export const _isIBAN = (str: string) => {
+  str = str.replace(/\s+/g, "").toUpperCase();
+
+  const cc = str.slice(0, 2); // CountryCode
+  const kk = str.slice(2, 4); // MOD-97 Checksum
+  const bban = str.slice(4); // BBAN, length and structure varies by country
+
+  if (!/^[A-Z0-9]+$/.test(str)) return false;
+  if (!countryCodesOfCountriesWhichUseTheIBAN.includes(cc.toLowerCase())) return false; 
+  if (str.length !== ibanLengths[cc.toLowerCase()]) return false;
+  if (!(/^\d+$/.test(kk))) return false;
+  if (computeMOD97Checksum(cc, bban) !== 1) return false;
+  
+
+  return true;
+}
+
+const countryCodesOfCountriesWhichUseTheIBAN = [
+  "al",
+  "ad",
+  "at",
+  "be",
+  "ba",
+  "bg",
+  "hr",
+  "cy",
+  "cz",
+  "dk",
+  "ee",
+  "fo",
+  "fi",
+  "fr",
+  "ge",
+  "de",
+  "gi",
+  "gr",
+  "gl",
+  "hu",
+  "is",
+  "ie",
+  "it",
+  "jo",
+  "lv",
+  "lb",
+  "li",
+  "lt",
+  "lu",
+  "mt",
+  "mr",
+  "mc",
+  "me",
+  "nl",
+  "pl",
+  "no",
+  "pt",
+  "ro",
+  "sm",
+  "sa",
+  "rs",
+  "sk",
+  "si",
+  "es",
+  "se",
+  "ch",
+  "tn",
+  "tr",
+  "gb",
+  "ae"
+];
+
+const ibanLengths: Record<string, number> = {
+  al: 28,
+  ad: 24,
+  at: 20,
+  be: 16,
+  ba: 20,
+  bg: 22,
+  hr: 21,
+  cy: 28,
+  cz: 24,
+  dk: 18,
+  ee: 20,
+  fo: 18,
+  fi: 18,
+  fr: 27,
+  ge: 22,
+  de: 22,
+  gi: 23,
+  gr: 27,
+  gl: 18,
+  hu: 28,
+  is: 26,
+  ie: 22,
+  it: 27,
+  jo: 30,
+  lv: 21,
+  lb: 28,
+  li: 21,
+  lt: 20,
+  lu: 20,
+  mt: 31,
+  mr: 27,
+  mc: 27,
+  me: 22,
+  nl: 18,
+  no: 15,
+  pl: 28,
+  pt: 25,
+  ro: 24,
+  sm: 27,
+  sa: 24,
+  rs: 22,
+  sk: 24,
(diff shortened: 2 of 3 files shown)


samchon (commented):
Thank you for the contribution, and for being upfront in the description about AI assistance and the missing tests. I want to give you a substantive answer rather than a one-line decline, because there are two separate things here: a scope decision, and a correctness bug that is worth knowing about regardless of where the code ends up living.

## First, the correctness issue

The MOD-97 check does not validate. I extracted `_isIBAN` verbatim and ran it against real IBANs:

```
DE89370400440532013000         -> false   (valid)
GB82WEST12345698765432         -> false   (valid)
FR1420041010050500013M02606    -> false   (valid)
NL91ABNA0417164300             -> false   (valid)
CH9300762011623852957          -> false   (valid)
BE68539007547034               -> false   (valid)
ES9121000418450200051332       -> false   (valid)
IT60X0542811101000000123456    -> false   (valid)
```

Every valid IBAN is rejected. The cause is in `computeMOD97Checksum(cc, bban)`, which builds `` `${bban}${cc}00` `` — the check digits `kk` are never fed into the checksum at all. That expression is the **generation** formula: its remainder `r` yields the check digits as `98 - r`. The **validation** formula is different: rearrange to `bban + cc + kk` and require `mod 97 === 1`.

Because `kk` is ignored, the function also produces false accepts. Any BBAN whose true check digits happen to be 97 passes with arbitrary check digits written in:

```
DE13370400440532013050  -> accepted, but invalid
DE42370400440532013050  -> accepted, but invalid
```

The fix is to pass all four leading characters into the checksum:

```ts
if (computeMOD97Checksum(cc + kk, bban) !== 1) return false;
```

with `computeMOD97Checksum` building `` `${bban}${prefix}` `` instead of appending a literal `"00"`.

Three further issues in the same file:

- **`ibanLengths` covers 50 countries.** The IBAN registry has roughly 88. `MD`, `KW`, `BR`, `UA`, `RU`, `IL`, `QA`, `SA`-adjacent entries and others are absent, so those valid IBANs are rejected by the length gate.
- **The import path is case-mismatched.** The file is `_isIBAN.ts`, but `_isStringFormat.ts` imports `from "./_isIban"`. That resolves on Windows and fails on a case-sensitive filesystem. All of our workflows run on `ubuntu-latest`, so CI would not build.
- **The `FORMAT` map key is the function name, not the format.** It registers `isIBAN: _isIBAN`, but the lookup is by format value. `FORMAT["iban"]` is therefore `undefined`, and `_isStringFormat` returns `true` for any unknown format — so on the one path this PR does modify, the validator never runs. That the typo compiles at all is our fault: the map is typed `Record<string, ...>` when it could be keyed on `Format.Value`. I will tighten that.

## Second, the scope decision

I am closing this, and I would have closed it even with the checksum corrected.

`Format.Value` is a closed set of standardized formats. Every value in it comes from **JSON Schema Validation 2020-12 §7.3** (`date-time`, `date`, `time`, `duration`, `email`, `idn-email`, `hostname`, `idn-hostname`, `ipv4`, `ipv6`, `uri`, `uri-reference`, `iri`, `iri-reference`, `uuid`, `uri-template`, `json-pointer`, `relative-json-pointer`, `regex`) or from OpenAPI (`byte`, `password`, plus the near-universal `url`). Neither the spec nor the [OpenAPI Formats Registry](https://spec.openapis.org/registry/format/) defines `iban`.

This matters because `format` is emitted into the OpenAPI and LLM schemas typia generates, where `format: "iban"` is an annotation no consumer understands. And accepting one non-standard domain format admits `isbn`, `credit-card`, `phone`, `bic`, and `vat` by the same argument. There is also the standing maintenance cost: the IBAN registry gains countries, so the length table would become an obligation on this repository.

## What I would suggest instead

A custom tag gives you exactly the built-in `Format` experience without a patch to typia. Here is a working version — I ran it against the ten valid IBANs above (all pass, including `MD` and `BR`) and six invalid ones (all rejected):

```ts
import { tags } from "typia";

type IBAN = string &
  tags.TagBase<{
    kind: "iban";
    target: "string";
    value: undefined;
    validate: `(() => { const s = $input.toUpperCase(); if (!/^[A-Z]{2}[0-9]{2}[A-Z0-9]{11,30}$/.test(s)) return false; const d = (s.slice(4) + s.slice(0, 4)).replace(/[A-Z]/g, (c) => String(c.charCodeAt(0) - 55)); let m = 0; for (const c of d) m = (m * 10 + Number(c)) % 97; return m === 1; })()`;
    schema: { format: "iban" };
  }>;
```

`tags.TagBase` is documented public API — see [Customization](https://typia.io/docs/validators/tags/#customization). One thing that is easy to trip on: `validate` must be an **expression string** referencing `$input` (a template-literal type), not a function reference; the transformer inlines that string into the generated validator. For branching logic, wrap it in an IIFE, exactly as the built-in `date-time` and `iri` validators do. And `schema` accepts any annotation, because `OpenApi.IJsonSchema.IString["format"]` is an open union (`| (string & {})`), so your generated OpenAPI document still carries `format: "iban"`.

If you only need shape rather than checksum, `string & tags.Pattern<"^[A-Z]{2}[0-9]{2}[A-Z0-9]{11,30}$">` covers it.

One design note on the original, worth keeping in mind for your own version: `_isIBAN` normalizes its input with `replace(/\s+/g, "").toUpperCase()` before checking. Every other typia format validator is a pure test on the raw value, so that normalization means `"DE 89 3704 0044 0532 0130 00"` passes `typia.is` while the value handed to your consumers still contains the spaces. I would validate the compact form and normalize separately.

## For future reference

Adding a value to `Format.Value` requires nine coordinated changes, not one:

1. `packages/interface/src/tags/Format.ts` — the union **and** the JSDoc list above it
2. `packages/typia/src/internal/_isFormatXxx.ts` — the real validator. `Format` compiles to `$importInternal("isFormat" + Pascalize<Value>)`, resolving to `typia/lib/internal/_isFormatXxx`; without this file, user code fails with TS2307 and the transformer emits no diagnostic.
3. `packages/typia/src/internal/_randomFormatXxx.ts` — for `typia.random`
4. `packages/interface/src/utils/IRandomGenerator.ts` — the generator method
5. `packages/typia/native/core/factories/FormatCheatSheet.go` — for the `@format xxx` comment tag
6. `packages/utils/src/validators/functional/` — the runtime schema-driven path (what this PR modified)
7. `tests/template/src/structures/TypeTagFormat.ts` and `CommentTagFormat.ts`
8. `tests/test-typia-schema/src/features/random/test_random_format_variation.ts`
9. `website/src/content/docs/validators/tags.mdx`

Items 2 through 5 are missing here, so `typia.is<string & tags.Format<"iban">>()` would not compile in a user project. That this is not written down anywhere is our gap, not yours, and I will fix the documentation.

Thanks again — the reporting in your description was honest and made this much easier to review.

PR #2343, opened 2026-08-17
fix: adopt JSON Schema length and divisibility semantics, and enforce the 64-bit bigint bounds

samchon:
## Intent

This is the unified cycle pull request of the `major-scalar-semantics` issue
campaign. It owns the complete accepted cycle:

- #2335 — adopt JSON Schema length and divisibility semantics for `MinLength`,
  `MaxLength`, and `MultipleOf`
- #2338 — enforce the exact int64 and uint64 bounds on the bigint path of
  `Type`

Both issues describe the same shape of defect: the validator typia *generates*
and the schema typia *publishes* describe one type and answer differently
about it. `typia.is<string & tags.MinLength<2>>("😀")` is `true` while
`{ "minLength": 2 }` rejects the same value; `typia.is<number &
tags.MultipleOf<0.1>>(0.3)` is `false` while `{ "multipleOf": 0.1 }` accepts
it; `typia.is<bigint & tags.Type<"int64">>(2n ** 200n)` is `true` while
`typia.protobuf.encode` then truncates the value it certified.

## Release constraint — this must ship in a major

Read this before merging or cutting a release from `master`.

Both issues state the change ships **only in a major release**, for two
independent reasons:

1. Validation outcomes flip in both directions for data that exists today.
2. A tag `validate` template that names `typia/lib/internal/_x` is
   unsatisfiable by a `typia` older than `_x`. `typia` declares
   `@typia/interface` through a caret range, and a caret only floats upward, so
   inside one major an older `typia` can install a newer `@typia/interface` and
   emit an import it never shipped. That is #2330, which
   fbf7b46f5 and 3502538f0 closed by reverting exactly the templates this pull
   request restores.

`.agents/skills/issue-campaign/development.md` freezes package versions for the
whole campaign, so this pull request changes none. **The release that carries
it must be `14.0.0`, never a `13.x`.** Cutting `13.4.0` from a `master` that
contains this commit reopens #2330.

The precondition for reason 2 holds: `pnpm package:tgz` on this branch packs
`typia` with `lib/internal/_stringLength`, `_isMultipleOf`, `_decimal`,
`_isTypeInt64Bigint`, and `_isTypeUint64Bigint` in `.js`, `.mjs`, and `.d.ts`,
so every `14.x` `typia` satisfies every `14.x` `@typia/interface`.

## Scope

| Owner | Change |
| --- | --- |
| `packages/interface/src/tags/{MinLength,MaxLength}.ts` | measure with `_stringLength` — Unicode characters |
| `packages/interface/src/tags/MultipleOf.ts` | number arm divides exact decimals with `_isMultipleOf`; the bigint arm is already exact and is untouched |
| `packages/interface/src/tags/Type.ts` | the `bigint` arms of `"int64"` / `"uint64"` take the exact inclusive bound from `_isTypeInt64Bigint` / `_isTypeUint64Bigint`; the `number` arms are untouched |
| `packages/typia/native/core/factories/MetadataCommentTagFactory.go` | `@minLength`, `@maxLength`, `@length`, `@multipleOf`, `@type int64`, `@type uint64` move with the type tags — two spellings of one constraint must never mean different things |
| `packages/typia/src/internal/_randomPattern.ts`, `_randomStringLength.ts` | count characters, not code units, when filtering a generated string against a `minLength` / `maxLength` bound |

`NumericRangeFactory.go` is unchanged: it is the protobuf union-candidate
narrowing owner and already holds the exact inclusive 64-bit bounds. The
emitted `minLength` / `maxLength` / `multipleOf` / `type` keywords are
unchanged; only the runtime check moves toward them.

## Verification

Pending at claim time. Every result lands as a formal `COMMENT` review on this
pull request, one per pushed commit, plus the Self-Review rounds.

The frozen integration-command manifest for this cycle is `v1`; its full
command set, expected artifacts, and the two gates that cannot run on this
machine (the local `website/` production build, which needs a sibling `../ttsc`
checkout, and the non-Windows lanes) are recorded in the campaign knowledge
base and repeated in the verification review before merge.

## Deferred

Two candidates surfaced during discovery, both outside the scope the campaign
was given, both recorded with reproductions and neither published yet:

- **No type tag on a dynamic key is enforced.** `[key: string &
  tags.Format<"uuid">]` accepts `"not-a-uuid"`, and `[key: number &
  tags.Minimum<0>]` accepts `-1`. It affects every key tag equally, its root
  cause is the dynamic-key path rather than any tag declaration, and neither
  issue's acceptance mentions a dynamic key.
- **No cross-language guard pins the JSDoc `Validate` string to its type-tag
  declaration.** The sibling guard for `Exclusive` exists because those two
  copies did drift.

## Issue campaign

Campaign `major-scalar-semantics`. Discovery ran two complete full-scope rounds
against `087cfadaf`; the second produced no candidate the first had not already
recorded, which is what settles this cycle's issue set.


samchon (commented):
**Commit ledger — `00b60a9f2` `fix(interface): count characters and divide exact decimals in the scalar tag validators`**

Landed: the `validate` templates of `MinLength`, `MaxLength`, and the `number` arm of `MultipleOf` now name `_stringLength` and `_isMultipleOf`; `TagBase`'s `@example` follows; `MetadataCommentTagFactory.go`'s `@minLength`, `@maxLength`, `@length`, and `@multipleOf` records move with them; the two native transform tests that pinned the old emitted text move with them, and the `_stringLength` runtime stub and its rewrite rule come back to the transform harness.

Untouched on purpose: the `bigint` arm of `MultipleOf` (a bigint remainder was always exact), `MinItems`/`MaxItems` (`Array.prototype.length` is exact), and `MetadataCommentTagFactory.go`'s `@type` records (they belong to #2338, landed separately in `fcf960f1b`).

Resolves: no issue on its own. It is the product half of #2335.

Verification: `pnpm --filter ./tests/test-typia-schema start` failed here on exactly the two suites that encoded the old meaning — `test_validate_unicode_string_length` on the astral character and `test_validate_decimal_multiple_of` on `0.03` — which is the fail-before evidence. `pnpm test:go:native` green.

samchon (commented):
**Commit ledger — `2eae623de` `fix(random): measure a generated string in characters against its length window`**

Landed: `_randomPattern` and `_randomFormatLength` measure a produced string with `_stringLength` instead of `String.prototype.length` when filtering it against a `minLength` / `maxLength` bound.

Why it is in this cycle rather than a follow-up: #2335's invariant includes "`typia.random<T>()` output satisfies `typia.is<T>()` for every constrained `T` it can generate". RandExp quantifies over UTF-16 code units, so `Pattern<"^<astral>+$"> & MinLength<3>` produced draws of two characters that passed a three-code-unit filter. Measured at the frozen commit: 200 draws contained the shape `utf16=3 cp=2` four times. Fixing only the `MultipleOf` half of that invariant would have patched the reported witness and left the class open.

Scope: these are the only two sites under `packages/typia/src/internal/` that measure a produced string against one of those bounds. Every other generator builds from `_randomString`'s ASCII alphabet. `_randomFormatLength`'s two callers are ASCII too, so it is behavior-preserving there and lands to keep one measure.

Resolves: no issue on its own; part of #2335.

samchon (commented):
**Commit ledger — `980b89250` `test(typia-schema): pin the character-count and exact-divisibility semantics`**

Landed: three rewritten suites, one new suite, and the two guides.

- `test_validate_unicode_string_length` — nine samples (empty, ASCII, precomposed, combining, lone surrogate, astral, ASCII+astral, regional-indicator flag, ZWJ family) against `MinLength<2>`, `MaxLength<2>`, and a one-character window, in the type-tag spelling, the JSDoc spelling, and the `@length` composite, through `is`, `assert`, and `validate`. The oracle is `[...value].length`. It asserts that four of the nine samples have different code-unit and character counts, so a matrix later trimmed to ASCII cannot leave the file passing while proving nothing.
- `test_validate_decimal_multiple_of` — twelve multiples of `0.01` (six with a nonzero binary remainder) against three non-multiples, plus an integer divisor, a fractional divisor, the `bigint` divisor unchanged, and a pair past `Number.MAX_SAFE_INTEGER` that diverges in both directions.
- `test_random_numeric_multiple_of` — gets its `typia.is<IValues>` round trip back.
- `test_random_astral_pattern_length_window` — new; the generator half of the invariant.

Both validation suites now require `@typia/utils`'s OpenAPI validator to agree with the generated one rather than listing where they disagreed.

Mutation evidence: reverting only `_randomPattern` to `str.length` while keeping the tests fails the new suite with the witness `"😀\ude00"` — three code units, two characters — and passes everything else.

Resolves: **#2335**.

samchon (commented):
**Commit ledger — `fcf960f1b` `fix(interface): enforce the exact int64 and uint64 bounds on the bigint path of Type`**

Landed: the `bigint` arms of `Type<"int64">` and `Type<"uint64">` delegate to `_isTypeInt64Bigint` and `_isTypeUint64Bigint`; `MetadataCommentTagFactory.go`'s `@type int64` / `@type uint64` bigint record moves with them; `RandomProgrammer.go`'s typed-array note is corrected.

Untouched on purpose: the `number` arms (they bound exclusively at the neighbouring power of two, which is exact for doubles), and `NumericRangeFactory.go`, which is the independent protobuf union-candidate narrowing owner and already held the exact inclusive bounds.

The `RandomProgrammer` note is worth a second look: it previously claimed the bigint path owns no exactness, and never explained why the deliberately rounded generation bound is safe. It now records the reason — consecutive doubles are 2048 apart at that magnitude, so `Math.floor(Math.random() * (maximum - minimum + 1)) + minimum` tops out at `2**63 - 2048` — and names the two helpers. Comment only; no generation bound changed.

Resolves: no issue on its own. It is the product half of #2338.

samchon (commented):
**Commit ledger — `5139a1c43` `test(typia-schema): pin the exact int64 and uint64 bigint bounds`**

Landed: both tag suites assert one boundary matrix — exact inclusive minimum and maximum, one step outside each, zero, and magnitudes far outside — from three independent sources: the helper, the type tag, and the `@type` spelling. Each file asserts how many members fall on each side, so a list later trimmed to in-range values cannot leave the loop passing while proving nothing.

The protobuf assertions change direction: they used to pin the corruption, and now pin its absence. Every certified bigint decodes back to itself, and a value one step past the bound is rejected by `assertEncode` before a varint is written. The truncation is still computed from the wire format and asserted to differ from the input, so the rejected value is proved to be one the encoder would have corrupted.

Each file also round-trips `typia.random` output through `typia.is`, for the tagged bigint and the matching typed array — the invariant the restored bound could have broken and does not.

Every number-path assertion is unchanged; that path is the control a mutation proof needs.

Resolves: **#2338**.

samchon (commented):
**Commit ledger — the first correction arc: `59c902c2f`, `44b157069`, `d3cca2877`**

These three are posted together and late; they should have gone up as each landed. Each landed commit gets a read-only early-warning pass under the campaign procedure, and the first five passes came back with enough that correcting them was the honest next step rather than folding it into a final sweep.

`59c902c2f` — corrections to `2eae623de`. The comment illustrated the length divergence with `^\u{1f600}+$`, which cannot produce the counts it claimed: `tags.Pattern` compiles to a flagless `RegExp`, and `\u{…}` is a code-point escape only under `u`. Read with `u`, a two-character draw is four code units and there was nothing to fix; the reachable case is the literal `^😀+$` the test uses, where `+` repeats the trailing low surrogate alone. The `bounded ? _stringLength(str) : 0` sentinel also departed from `OpenApiStringValidator`'s `number | null` idiom for the identical construct, and the count ran before `regex.test`, turning a lazy O(1) read into an eager walk on every draw including rejected ones.

`44b157069` — corrections to `00b60a9f2`, `980b89250`, `fcf960f1b`, and `5139a1c43`. The substantive one: **the random round trip in both tag suites proved nothing**, for two independent reasons. A `bigint` type tag publishes no numeric schema bound, so `_randomInteger` falls back to `0..100` and never approaches the width — the assertion passed identically under the unbounded behavior this cycle replaces. And `[...typia.random<BigInt64Array>()].every(oracle)` is a tautology: the constructor wraps on store, so `new BigInt64Array([2n ** 200n + 5n])` is `[5n]`. Also: both boundary-list guards counted two totals rather than each side of each bound; the uint64 list carried `MAXIMUM + 1n` and `2n ** 64n`, which are the same value; `RandomProgrammer.go` attached 2048 to a spacing that is 1024 near `2**63`; three "escaped" Unicode samples were raw literals one NFC pass from collapsing; and a docstring promised an exact-rational oracle the file did not have.

The same commit restores the **#2330 rule to the code**. `fcf960f1b` had deleted the comment block explaining why a tag template must not name a runtime helper — at the exact moment four helpers started being named. It is back as the `metadataCommentTagFactory_PARSER` doc comment, with the version each helper first shipped in (13.1.19 for all four) and the explicit statement that only a major release makes this safe and that nothing in the tree can enforce it.

`d3cca2877` — the `spell-check` lane went red on "neighbouring". Fixed; the lane is green.

Resolves: nothing new. These correct the work already attributed to #2335 and #2338.

samchon (commented):
**Commit ledger — the second correction arc: `764c6910e`, `8aff69a34`**

`764c6910e` — corrections to `44b157069`, i.e. to the previous correction. The bounded random twin it introduced was as vacuous as what it replaced: `Minimum<0n> & Maximum<9007199254740991n>` hands the generator a window inside both widths, and the int64 form used a non-negative window so it could not even distinguish `Type<"int64">` from `Type<"uint64">`. Gone. The strengthened straddle guards still admitted a list of far-away magnitudes that never touches an edge, so the four boundary values are now required by name. `exactMultiple`'s `fraction` is `_decimalDecompose` line for line, so the docstring's independence claim was too strong — it now draws the line the unicode file already drew: the oracle pins the *rule*, not the decomposition. And `exactMultiple` disagreed with `_isMultipleOf` on negative, zero, and non-finite divisors, which the previous commit's own message reasons about; the domain guard is in.

`8aff69a34` — a claim in `764c6910e` was false. It said no generator-driven assertion could reach the width "because every window the transform accepts sits inside it". The exact edges are inexpressible as bigint literals — `Numeric<T>` resolves `9223372036854775807n` to `number`, which the transform rejects as non-literal — but values *past* them round-trip fine, so a wider window is declarable and does fail. Measured, 500 draws each: 33 rejects for int64, 40 for uint64, 38 for the JSDoc spelling.

That is a real hole and it is **not this cycle's**. `_randomInteger` draws from the declared window and never intersects it with the width the `Type<…>` tag enforces — just as true on the `number` path this cycle does not touch, where `number & Type<"int32"> & Minimum<-1e12> & Maximum<1e12>` draws a value its own `is` rejects on **499 of 500** tries and `Type<"uint8">` with `Maximum<10000>` on 487. Those tags have enforced their ranges since long before this campaign. Restoring the bigint bound makes that arm behave the way the number arm already did.

It is recorded as a campaign candidate rather than fixed here: the owner is `Type` and the range tags together, across every width and both paths, and the repair is a product decision — clip the generation window, or reject the contradictory declaration at compile time the way `typia.random` already throws for an impossible `MultipleOf` range.

Resolves: nothing new.

samchon (commented):
**Self-Review round 1 — `6117aa60a`**

One complete round over the whole base-to-head diff at `8aff69a34`: 20 files, +748 / −348, every changed file and hunk read, plus the consequence surface around each — the six transpile sites that splice a tag `Validate`, the schema programmers, the protobuf owner, the packaging manifest, and both guides.

Two improvements applied:

- `test_validate_decimal_multiple_of` ran `validate` only on the accepting side. Its negative twin is in, and with it the `expected` text. That is worth pinning rather than incidental: the report reads `number & MultipleOf<0.01>`, built from the tag's *name* and not from the expression implementing it, and this whole cycle is about swapping the expression behind a tag. If that swap changed the message a caller reads, it would be a silent contract break.
- Both tag suites wrapped `assertEncode` in a bare `try`/`catch`, so any encoder failure counted as the rejection the assertion title claims. `typia.is` on the same value is asserted beside it now, which is what establishes the width check as the cause.

Verified and found sound in the same round, recorded so they are not re-derived: the `MultipleOf` conditional still distributes over a union `Value` exactly as the old template literal did; `MinItems`/`MaxItems` are correctly untouched, because `Array.prototype.length` is exact; the `_stringLength` import introduces no cycle; no committed generated baseline embeds emitted validator source (`test-typia-generate/src/output` and `test-typia-automated/src/features` are both gitignored); and `typia.reflect.*` never publishes a `validate` string, so the public contract is unchanged.

On the Go side only `@multipleOf` has its `Validate` string pinned literally, plus the `_stringLength` emission pinned end-to-end in the transform test. The `@type int64`/`@type uint64` bigint records are covered behaviorally instead — both tag suites assert the `@type` spelling across the whole boundary matrix — which is stronger than a string comparison and is also the in-cycle mitigation for the missing cross-language `Validate` guard noted in the body.

samchon (commented):
**Self-Review round 2 — `85974a8f3`**

Round 1 changed the work, so round 2 ran fresh over the new snapshot rather than rechecking the fix. It folded in the early-warning pass over `8aff69a34`, which confirmed all five factual claims that commit makes and found two overstatements in how they are worded.

Corrected:

- The closing comment said a wide-window twin fails "for a reason this cycle did not introduce and does not own". The second half is right; the first is not. Before this cycle the bigint check was `true` for int64 and a lower bound only for uint64, so no wide-window twin could fail at all — the failure is newly *reachable* on this arm even though the hole is older and lives with `Type` and the range tags. Both halves are stated now.
- It also called the window pin "the one assertion in this section a generation change can break". The per-draw round trip can break too. What holds is that the pin strictly dominates: the generator draws `0..100`, both tags accept all of `[0, 100]`, so any draw failing the round trip is necessarily outside the pin — the pin breaks first and names why.

Two corrections that belong to the record rather than the code: `Minimum<-1e19n>` and `Maximum<2e19n>`, as written in `8aff69a34`'s message, are parse errors — a BigInt literal cannot carry an exponent. The measurements used `-10000000000000000000n` and `20000000000000000000n`, which are precisely the literals that pass the transform's literal gate while the true width edges do not. And "the `@type` / `@minimum` / `@maximum` spelling … through `strconv.ParseFloat`" bundled one tag too many: `@type` is a string switch, only the bounds reach `ParseFloat`.

One new candidate, recorded and not fixed: `MetadataCommentTagFactory.go` does `value := int(parsed)` on a `float64`, and an out-of-range conversion is implementation-defined in Go — `@minimum 10000000000000000000` and its negation both yield `-9223372036854775808` on go1.26.0/amd64. It is inert today (the emitted `Validate` uses the raw text and the `Schema` uses the float, and no programmer reads that field numerically outside `minItems`), it predates the cycle, and it belongs to the JSDoc numeric parser rather than to either issue.

Merge-gate check performed in this round: `git log master..HEAD` carries exactly two closing lines, one per issue, both on commits whose fix survives at `HEAD`. Nothing was reverted inside the branch, so the squash will close exactly what landed.

samchon:
Closing without merging: every commit on this branch already reached `master` through #2344, which was branched from this branch's head and squash-merged. `master` carries the four restored tag declarations, the JSDoc factory records, and `_randomPattern`, and #2335 and #2338 were closed by that merge.

This pull request's merge gates never ran, because the merge happened elsewhere. They have since been run and are recorded in #2346, together with fixes for the two vacuous assertions that shipped with the cycle:

- `pnpm test` on integrated `master` — **PASS**, and it explains the two local failures reported on this thread as branch-switch contamination in a shared checkout rather than defects;
- **clean packed consumer** — an out-of-workspace project installs the packed tarballs, the emitted validator requires all four `typia/lib/internal/*` helpers, every one resolves, and all ten behavioural checks pass;
- **mutation proof** — reverting the `Type` bigint arms fails both tag suites at `-9223372036854775809` and `18446744073709551616`; reverting the three scalar tags fails four suites independently.

The release constraint stated in this body is now measured rather than argued. Pairing the published `typia@13.1.1` — which declares `@typia/interface ^13.1.1`, a range that admits `13.3.0` — with this tree's `@typia/interface` produces `Error: Cannot find module '.../typia/lib/internal/_isTypeInt64Bigint.js'` at runtime. **A `13.4.0` release hands that to every consumer pinned to an older `13.x`; only `14.0.0` closes it.**

Issue #2351, opened 2026-08-17
The emitted schema for Type<...> carries no maximum, so it is weaker than the validator


samchon:
## Problem

`Type<"int64">` and `Type<"uint64">` enforce an exact width at runtime and publish no `maximum` in the schema they emit, so the schema is weaker than the validator that claims to implement it.

```ts
typia.json.schema<{ i64: number & tags.Type<"int64">; u64: number & tags.Type<"uint64"> }>().schema;
// { "i64": { "type": "integer" }, "u64": { "type": "integer", "minimum": 0 } }
```

No `maximum` on either, and no `minimum` on `int64`.

## Evidence

Measured at `7417574eb`. The declarations confirm it: `packages/interface/src/tags/Type.ts` publishes `{ type: "integer" }` for the signed widths and `{ type: "integer", minimum: 0 }` for the unsigned ones, and `MetadataCommentTagFactory.go` leaves `bigintSchema` nil for `int64` and `{"minimum": 0}` for `uint64`. Nothing in either owner carries an upper bound for any width.

The runtime side does: `_isTypeInt64` / `_isTypeUint64` on the `number` path, and `_isTypeInt64Bigint` / `_isTypeUint64Bigint` on the `bigint` path since #2338.

## Why it matters

#2335 and #2338 closed the gap in the other direction — the validator now agrees with `minLength`, `maxLength`, `multipleOf`, and the 64-bit widths. This is the remaining half of the same shape: a consumer handed `{"type":"integer"}` for a `bigint & Type<"int64">` is told nothing about the width, so an LLM structured-output endpoint, a generated SDK, or an Ajv check will accept a value `typia.protobuf.encode` cannot write.

The smaller widths have the same hole: `Type<"int8">` publishes `{ type: "integer" }` and enforces `[-128, 127]`.

## The obstacle worth naming

`2n ** 64n - 1n` is `18446744073709551615`, which no JSON number represents exactly — it round-trips as `18446744073709551616`. Emitting that as `maximum` would publish a bound wrong by one in the direction that admits an unencodable value. `2n ** 63n - 1n` has the same problem. The smaller widths do not: every bound through `uint32` is exactly representable.

So this is a design question, not a line change: publish exact bounds where they are representable and something else (or nothing) where they are not, and decide what that something else is.

## Consequence surface

[...]

Issue #2348, opened 2026-08-17
typia.random never intersects a declared range with the Type<...> width


samchon:
## Problem

`typia.random<T>()` draws from the window `Minimum` / `Maximum` declare and never intersects it with the width the `Type<...>` tag on the same leaf enforces. When the declared window is wider than the width, the generator returns values its own `typia.is<T>()` rejects.

```ts
type T = number & tags.Type<"int32"> & tags.Minimum<-1000000000000> & tags.Maximum<1000000000000>;
typia.is<T>(typia.random<T>()); // false on 499 of 500 draws
```

## Evidence

Measured at `7417574eb`, 500 draws each:

| Type | Draws its own `is` rejects |
| --- | --- |
| `number & Type<"int32"> & Minimum<-1e12> & Maximum<1e12>` | **499 / 500** |
| `number & Type<"uint8"> & Minimum<0> & Maximum<10000>` | **486 / 500** |
| `bigint & Type<"int64"> & Minimum<-10000000000000000000n> & Maximum<10000000000000000000n>` | 32 / 500 |

The rates match the mechanism exactly. `Type<"int32">` publishes `schema: { type: "integer" }` with no numeric bound (`packages/interface/src/tags/Type.ts`), so the merged schema is `{type:"integer",minimum:-1e12,maximum:1e12}` and `_randomInteger` draws uniformly over `2e12 + 1` integers while `_isTypeInt32` admits `2^32` of them — `p(accept) = 0.00215`, i.e. 498.9 expected rejections per 500.

The width lives only in `IMetadataTypeTag.Validate` (`$importInternal("isTypeInt32")($input)`), which never reaches the schema `_randomInteger` reads.

## Why it matters

`typia.random` is documented as returning a value that satisfies its type, and the repository already treats a contradictory range as an error rather than a bad value: an impossible `MultipleOf` window throws (`_randomMultiple`, `_randomInteger`), and `Format<"uuid"> & MaxLength<3>` throws (`_randomStringLength`). A `Type<...>` width contradicting a declared range is the same class and silently produces the value instead.

Seeding a database or a fixture from `typia.random` is the reachable damage: the data does not satisfy the type it was generated for, and the failure appears later at a validation boundary.

## Consequence surface

[...]

Issue #2347, opened 2026-08-17
No type tag on a dynamic key is enforced at runtime


samchon:
## Problem

A type tag on a dynamic key emits no runtime check at all. Every key tag is affected — `Format`, `Pattern`, `MinLength`, `MaxLength`, `Minimum`, `MultipleOf` — so typia certifies objects whose keys violate the type it was given.

```ts
interface IUuidKey {
  [key: string & tags.Format<"uuid">]: string;
}
typia.is<IUuidKey>({ "not-a-uuid": "x" }); // true
```

## Evidence

Measured at `7417574eb`:

| Declaration | Value | `typia.is` | Expected |
| --- | --- | --- | --- |
| `[key: string & tags.Format<"uuid">]: string` | `{ "not-a-uuid": "x" }` | `true` | `false` |
| `[key: string & tags.Pattern<"^ab+$">]: string` | `{ zzz: "x" }` | `true` | `false` |
| `[key: string & tags.MinLength<3>]: string` | `{ ab: "x" }` | `true` | `false` |
| `[key: number & tags.Minimum<0>]: string` | `{ [-1]: "x" }` | `true` | `false` |

The emitted schema for `[key: string & tags.MinLength<3>]: string` is `{"type":"object","additionalProperties":{"type":"string"}}` — no `propertyNames`, no `patternProperties`.

The check is implemented and simply not reached. `packages/typia/native/core/programmers/iterate/check_dynamic_key.go` filters tag rows through `check_dynamic_key_fully_validated_tag_rows` and routes a tagged string atomic into `Check_string`, a tagged number atomic into `Check_number`. Both transpile sites pass `context.Importer`, so a helper-naming template would resolve. `check_dynamic_key_has_pure_string` is the gate to examine: it returns `true` — treating the key as an unconstrained string — whenever an atomic has no fully validated tag row.

## Why no suite notices

`tests/template/src/structures/DynamicTag.ts` declares exactly this shape, with a `uuid` string key and a bounded numeric key. Every one of its `SPOILERS` corrupts a **value**, never a key, so the generated matrix never asks whether a bad key is rejected.

## Consequence surface

- `is`, `assert`, `assertGuard`, `validate` and their `create*` factories, since all share the generated validator
- `json.assertParse`, `llm.parse`, `llm.structuredOutput`, `http.assert*` — same validator
- `typia.random` for a dynamic-key type, which has no constraint to honour either
[...]

Issue #2353, opened 2026-08-17
No guard pins a JSDoc tag Validate string to its type-tag declaration


samchon:
## Problem

Nothing pins the JSDoc spelling of a constraint to its type-tag declaration. The two are composed by different factories in different languages, and only the `Exclusive` field has a guard — which exists because those two copies did drift.

`metadata_comment_tag_factory_exclusivity_matches_interface_test.go` transcribes the `@typia/interface` declarations into Go and asserts the comment factory produces the identical `Exclusive` list, for exactly this reason:

> `@format` published a bare `true` and `@pattern` published `["format"]` (missing its own kind), so `@pattern ^a` `@pattern ^b` compiled while `Pattern<...> & Pattern<...>` is a compile error.

The `Validate` field — the runtime check itself — has no counterpart. `metadata_comment_tag_factory_coverage_test.go` pins the Go side's literal for `@multipleOf` alone, and `@typia/interface` has no test that compares anything.

## Why it matters now

The `major-scalar-semantics` cycle moved four constraints on both sides at once (#2335, #2338). Each move had to be made twice, in two languages, by hand. That is precisely the shape that produced the `Exclusive` drift, and the only thing that would have caught a mismatch during that cycle was behavioural coverage in `test-typia-schema` asserting both spellings over the same value matrix — which exists for those four constraints and for no others.

## Consequence surface

Every tag kind with both spellings: `minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf`, `minLength`, `maxLength`, `length`, `pattern`, `format`, `minItems`, `maxItems`, `uniqueItems`, `type`. A drift means `typia.is<T>` answers differently depending on whether the constraint was written as an intersection or as a JSDoc comment on the same type.

## Invariant

For every constraint with two spellings, the two emit the same runtime check, and a divergence fails a test rather than shipping.

Whether that is enforced by extending the Go declaration mirror to carry `Validate`, by a TypeScript test that transforms both spellings and compares the emitted output, or by behavioural parity coverage across the whole tag catalogue is the implementer's decision. The behavioural route is the strongest — it compares answers rather than strings — and is what the four constraints from the last cycle already have.

## Acceptance

- Every tag kind with both spellings is covered, not a sample.
- The guard fails when one spelling is changed and the other is not; demonstrate by mutating one side and showing the failure.
- The four constraints from #2335 and #2338 keep their existing behavioural coverage; this adds the rest rather than replacing it.

[...]

PR #2354, opened 2026-08-17
fix: close every candidate the scalar semantics campaign recorded

samchon:
## Intent

The second cycle of the `major-scalar-semantics` campaign. It closes five of the candidates the first cycle recorded and could not act on, published as #2347–#2353 after each was re-reproduced against `7417574eb`.

Breaking changes are deliberate here — the next release is `14.0.0`.

| Issue | What it was |
| --- | --- |
| #2347 | no type tag on a dynamic key was enforced |
| #2349 | `test_random_recursive_terminates` asserted a depth bound nothing guaranteed |
| #2350 | `TestValidator.equals` is asymmetric, making assertions silently vacuous |
| #2352 | `int(parsed)` silently overflowed for an out-of-range JSDoc numeric tag |
| #2353 | nothing pinned a JSDoc `Validate` string to its type-tag declaration |

**#2348 and #2351 were reverted out of this branch and stay open.** They were sound diagnoses, but the fix duplicated the `Type<...>` width table (`int8` is `-128..127`, and so on) into a second and third place beside the tag's own `validate` template — the shape of the defect #2166 already was. They are also not separable: publishing the width required the schema merge to intersect, and once it did, #2348's generation clip became redundant with it. They deserve one change that gives the widths a single home, not two that add two more copies.

## The one that was most wrong

**#2347.** The emitted checker *did* test the key — and used the result only to pick which signature's value type to apply:

```js
if ("string" === typeof key && RegExp("^ab+$").test(key))
    return "string" === typeof value;
return true;                                  // <- a key matching nothing was accepted
```

So `[key: string & tags.Format<"uuid">]` accepted any key at all. `Check_dynamic_key` was never the problem; only the tail was. Values were always checked, which is what kept it invisible: `@typia/template`'s `DynamicTag` declares exactly this shape and every one of its spoilers corrupts a value, never a key.

The report is a separate question the fix raised. `Superfluous` — the report an extra property gets — says the property is not defined in the object type and advises removing it. Both false when the property *is* declared and only its key broke a constraint, so `InvalidKey` names the key's declared type instead. `Superfluous` remains the fallback for a programmer that has no better answer.

## Design decisions worth reviewing

**#2352 is bounded by int64, not by `int`.** The first version of this guard rejected magnitudes at and beyond 2\*\*63, which is right only where `int` is 64 bits. `@ttsc/linux-arm` is a published platform where it is 32, so the guard would have left the same defect intact from 2\*\*31 up — and the test did not compile there at all (`GOARCH=arm go vet` rejects its own `-9223372036854775808` constant). The value is already widened to int64 at its only use, so parsing to int64 is both the fix and a simplification.

The exclusive `>=` against 2\*\*63 is deliberate: an untyped `9223372036854775807.0` rounds *up* to 2\*\*63 once it meets a float64 operand, so an inclusive guard lets through exactly the one value int64 cannot hold.

**#2349 extended the guard rather than weakening the test.** The depth cutoff guarded recursive *arrays* only; the nullable and optional escapes were per-level coin flips with no bound. Depth was geometric, reaching 16 over 20,000 draws with `depth > 8` on 44 of them — 0.43 % per suite run, which turned one local `pnpm test` red. Now bounded at 6.

**#2350 did not get a static scan.** One was written and discarded: syntax cannot tell an optional-typed read from a total one — `Object.prototype.hasOwnProperty.call` can never be absent, `schema.description` can — so it flagged 51 provably safe call sites, each needing an annotation that would say nothing. What landed instead is the correction of all thirteen sites where a field can genuinely be absent, plus a case that pins which comparison shapes stay total. The issue's request for automatic detection of a *future* unsafe site is not met by a syntactic rule, and is recorded as such rather than satisfied by churn.

The sharpest of those thirteen: `test_json_schemas_v3_0_exclusive_bounds` exists to check *which* keywords the 3.0 downgrade emits, and read them straight off the schema — so a downgrade that dropped both flags satisfied it. Simulated against the real `json_equal_to`, the diff for that input is empty.

## Mutation evidence

Each fix was reverted with its tests kept:

- dynamic-key tail → the transform test fails at the pattern key;
- overflow guard → `9223372036854775808 must be rejected, got -9223372036854775808`;
- int64 widening → `GOARCH=arm go vet ./core/factories/` fails to build;
- `@multipleOf` drifted to a binary remainder while the type tag is left alone → `multipleOf agrees on 0.3`.

## Found by the final review round, not by a suite

`pattern_line_terminator_escape_transform_test` asserted that a key matching no signature is accepted — the #2347 defect itself, written down as an expectation. It only surfaced when the whole Go tree was run rather than the test written alongside the fix. It now reads the rejection, with a *valid* value, so the key is the only thing that can fail it.

The same round added the case saying a declared property stays exempt from the index signature's key tag: `id` is two characters and would fail a `MinLength<3>` key tag if declared properties were routed through the now-binding check. They are not, and nothing said so.


samchon (commented):
## Self-Review round two — one defect, restored

The round ran over the whole base-to-head diff against `7417574eb`, after CI reported `test` and `spell-check` red.

### The defect

`#2347`'s fix collapsed two different failures into one tail. A key that satisfied no dynamic signature was made to reject, but "satisfied no signature" covers both:

- **a key no signature's declared _type_ covers** — `wrong` against `[key: prefix_${string}]`. Nothing declares it, which is what a surplus property *is*. `is` accepts it as it accepts any surplus property; `equals` reports it through `Superfluous` with `expected: "undefined"`.
- **a key the declaration does cover but whose _type tag_ it breaks** — `ab` against `[key: string & MinLength<3>]`. That is the one #2347 is about.

The three generated matrices read that `"undefined"` in every surplus-property case, so rejecting the first turned `assertEquals` and `assertGuardEquals` red on `DynamicTemplate`, `DynamicUnion`, and `DynamicComposite` — 12 failures, all one cause. `pattern_line_terminator_escape_transform_test` had been edited in the same wrong direction, from "a near-miss key stays unchecked" to "a near-miss key is rejected"; it is restored.

### The fix

The tail re-asks the same key question with every type tag stripped. What is left is the declared key type alone, so a key that still answers yes is one whose tag is what broke, and a key that answers no is surplus. It is order-independent — every signature's stripped condition joins with `||` after all the full ones have been refuted — so a key that breaks signature A's tag while satisfying signature B is still accepted by B.

The stripped question is built only when some signature carries a key tag. Without one it *is* the full condition the tail already refuted, so an untagged declaration emits exactly what it emitted before.

Measured on the emitted JS:

```js
// [key: string & tags.MinLength<3>] — the declared type covers every key
if ("string" === typeof key && 3 <= _stringLength(key)) return "string" === typeof value;
return <InvalidKey>;                       // no surplus tail: it is unreachable

// [key: `prefix_${string}`] — the declared type covers only some keys
if ("string" === typeof key && RegExp(/^prefix_(.*)/).test(key)) return "string" === typeof value;
return true;                               // unchanged from master
```

### Coverage

`test_validate_dynamic_key_surplus` pins the distinction against **both spellings of one predicate** — `[key: prefix_${string}]` as a type and `[key: string & Pattern<"^prefix_">]` as a tag accept exactly the same keys, so nothing but where the constraint sits can explain a difference in their answers. Mutation-proved: with the tail reverted, it fails at the first assertion. The transform test gains the template literal negative control beside its plain `[key: string]` one.

### Also in this push

- `spell-check`: four British spellings (`behaviour`, `generalises`, `normalised`, `Normalising`) in the two new test files.
- The random recursion cutoff gets one home. `RandomJoiner` spelled the literal `5` for a recursive array and the nullable and optional escapes added a second copy; a recursive value that stopped at two depths depending on which escape it owned would make the depth bound this cycle asserts unstatable.

### Verified locally

- `TestDynamicKeyTagsTransform`, `TestPatternLineTerminatorEscapeTransform` — pass
- `iterate` package, untagged and `typia_native_internal` — pass
- `test-typia-schema --include dynamic_key` — pass, and fails on the reverted tail
- `test-typia-automated` full matrix — running at the time of this push; CI is the gate

### Not changed, on purpose

Under `numeric: false` the emitted shape check for a `number` index signature is `"number" === typeof Number(key)`, which is true for every key — `Number("word")` is `NaN`, and `typeof NaN` is `"number"`. So a numeric signature declares every key under that option and nothing beside it is ever surplus. That is master's behavior, not this change's: on master the same vacuous condition makes `[key: number]` swallow every key before any other signature is consulted. Changing it would move `[key: number]` semantics for every consumer, which is not what #2347 asked for.

---

## Self-Review round three — one more defect, and a documentation gap

Fresh round over the whole base-to-head diff after the round-two push.

### The predicate that decides "is this key tagged" read a row all-or-nothing

`check_dynamic_properties_key_tagged` reused `check_dynamic_key_fully_validated_tag_rows`, which keeps a tag row only when **every** tag in it carries a runtime check. `Check_string` and `Check_number` do not read rows that way — they filter per tag — so a row holding one tag with no runtime check beside a real one still emits the real one.

`[key: number & Minimum<0> & <a tag with no runtime check>]` separates the two readings: the minimum is emitted, the key is refused by it, and the all-or-nothing reading then called the declaration untagged and sent the key to the accepting surplus tail. The minimum would have been inert again — the defect #2347 closed, restored for one tag combination. One tag with a runtime check is now enough.

Answering yes where the stripped question turns out identical costs an unreachable branch and never a wrong answer: the tail is reached only after every full condition refused the key, so a stripped condition equal to one of them refuses it too.

### `website/src/content/docs/validators/tags.mdx` said nothing about key tags

#2347 changes public behavior and no page stated it in either direction. `### On index signature keys` now names both halves, because naming only the rejection would mislead — the two example declarations accept exactly the same keys and differ only in what they call a rejection.

### Consequence surface traced this round

- **`typia.random` round trip.** `randomJoiner_dynamicProperty` decodes the key metadata through the ordinary decoder, so a drawn key honors its own tags. `test_random_DynamicTag` in the generated matrix is the standing proof, and `DynamicTag` declares both a tagged string key and a tagged numeric one.
- **Importer.** The stripped question calls `Check_dynamic_key` a second time per signature. `ImportProgrammer.Instance` dedupes by alias, and every construct the stripped pass builds is a subset of what the full pass already built, so no import can appear that the full pass did not already register.
- **Emitted shape, measured.** For `{ [key: number & Minimum<0>]: string; [key: prefix_${string}]: string }` the tail emits `if (<number shape> || <template shape>) return false;` — `||` binds looser than the template's `&&`, so the grouping is the intended one. The template disjunct is redundant there (its full condition is its own stripped one and already refused the key); it is kept rather than filtered because "the key belongs to some signature's declared type" is the property that is provably correct, and filtering by taggedness would turn an untagged signature's exclusion conditions into a behavioral edge for no gain outside a rare mixed declaration.
- **`_test_validateEquals`** compares error paths only, never `expected`, which is why the over-broad rejection showed up in `assertEquals` and `assertGuardEquals` alone. Paths are unchanged either way.
- **`test-typia-generate`** gitignores `src/output` and no input declares an index signature, so no committed baseline moves.
- **`test-feature-identity`** — run, green, including the `test_openapi_total_comparison_shape` → `test_total_comparison_shape` rename.

### Left alone, with the reason

Under `numeric: false` the emitted shape for a `number` index signature is `"number" === typeof Number(key)`, true for every key, since `Number("word")` is `NaN` and `typeof NaN` is `"number"`. A numeric signature therefore declares every key under that option and nothing beside it is surplus; under `numeric: true` the guard is real and a non-numeric key is surplus again. That option-dependence is master's, not this change's — on master the same vacuous condition makes `[key: number]` swallow every key before any other signature is consulted. Forcing the numeric guard inside the stripped question alone would make the shape check stricter than the full condition it is derived from, which is the one invariant this design rests on.

samchon (commented):
## Self-Review rounds four to six

### Round four — a name that stopped being true, and a second documentation gap

`broken` named one thing while only the signature loop set it: a signature that accepts every key, which returns unconditionally and ends the arrow. The tail now sets it for a second reason — a declared key type that covers every key, so the surplus return below is unreachable — and reading the old name at the tail's guard invites the wrong conclusion about which of the two happened. Renamed to `terminated`, which is the property both share and the one the guard tests. Emitted output unchanged.

`@minItems`, `@maxItems`, and `@items` now reject a value `int64` cannot hold instead of converting it, which is a compile error a reader can hit with no way to know why. Added to the caution block beside the other three things comment tags cannot do. The numeric tags are deliberately not claimed there: `@minimum` and its siblings swallow the same report and drop only the `bigint` form of the tag, and that is behavior this cycle did not change.

### Round five — the recursion fix had no probe that could fail

This is the one worth reading.

`781298e46` gave the nullable and optional escapes the depth cutoff, and the probe for it was `test_random_recursive_terminates` — one draw per shape, asking whether the value is finite. Before the cutoff the depth was geometric, a coin flip per level, so a single draw exceeded the bound a fraction of a percent of the time. **A build with the guard removed passes that probe about 99 times in 100.** The suite went red at random rather than reporting the defect, which is how #2349 was found in the first place, and the fix inherited the same oracle.

`test_random_recursive_depth_bound` draws 3000 of each escape and asks a question a broken build cannot pass. The bound is not read off the implementation: one constant decides when a recursive value stops descending, so the nullable, optional, and array escapes have to stop at the *same* depth, and an escape stopping at its own depth means the constant was copied. That is what `RandomJoiner_RECURSION_CUTOFF` exists to prevent, and now something fails when it happens.

Measured, 3000 draws each: `{ nullable: 6, optional: 6, array: 6 }`. Mutation-proved — with the escape disabled, the first assertion fails on the nullable and optional depths **while `test_random_recursive_terminates` passes in the same run**.

### Round six — clean

Fresh round over the whole diff, looking for the pattern round five exposed rather than for new code. Every fix in the cycle now has a probe that fails without it:

| Issue | Probe | Fails without the fix |
| --- | --- | --- |
| #2347 | `TestDynamicKeyTagsTransform` runtime cases, `test_validate_dynamic_key_report`, `test_validate_dynamic_key_surplus` | yes, mutation-proved in both directions |
| #2349 | `test_random_recursive_depth_bound` | yes, mutation-proved |
| #2350 | `test_total_comparison_shape` plus the thirteen corrected sites | the shape test pins the harness; the corrections are test code and have no product mutation |
| #2352 | `TestMetadataCommentTagFactoryRejectsOutOfRangeInteger` | yes — `int(parsed)` accepts 2\*\*63 |
| #2353 | `test_validate_tag_spelling_parity`, with a per-row vacuity guard | yes |

Also checked and unchanged: no package version moves (this is an implementation pull request, so the release version stays the maintainer's to assign — the native tree here needs one newer than the published `13.3.0` before npm consumers see any of it); no workflow, dependency, or committed baseline moves; `test-typia-generate` gitignores `src/output` and declares no index signature; `test-feature-identity` green over the `test_openapi_total_comparison_shape` → `test_total_comparison_shape` rename.

### Local verification

| Command | Result |
| --- | --- |
| `pnpm --filter ./tests/test-typia-automated start` | Success — the suite CI reported red |
| `pnpm test:go` | Success |
| `go test ../native/core/programmers/iterate -tags typia_native_internal` | Success |

Not run locally: `test-error`, `test-interface`, `test-typia-cli`, `test-typia-generate`, `test-typia-exact-optional`, `test-langchain`, `test-mcp`, `test-vercel`, `test-utils-automated`, `test-typia-bundler-cache`, and the website build. CI runs all of them and is the gate.

samchon (commented):
## Pre-merge notes

### The squash message would have closed two issues that must stay open

`f723584f7` and `97b341a38` each end in a `Close #…` line, and `cef3b286f` reverted both. A squash merge concatenates every commit message, so the default merge commit would have carried `Close #2348` and `Close #2351` into master and GitHub would have closed them — the two candidates this cycle deliberately did *not* act on.

The merge uses an explicit body instead. It closes #2347, #2349, #2350, #2352, and #2353, and states in prose why #2348 and #2351 stay open, with no closing keyword near either. The pull-request description contains no closing keyword either — `closes every candidate … published as #2347–#2353` does not match GitHub's pattern, which needs the keyword immediately before the reference.

### What #2348 and #2351 are waiting for

Recorded here because the description was written before the revert and stands as the historical intent. Both were sound as diagnoses: `typia.random` did draw values its own `is` rejects, and the emitted schema was weaker than the validator it came from. What made them wrong to land was their cost — the width table (`int8` is -128..127, and so on) already lives in the tag's `validate` template, and publishing it into the schema added a second copy in `Type.ts` and a third in `MetadataCommentTagFactory.go` for the JSDoc spelling. They are also not separable: publishing the width required the schema plugin to intersect numeric bounds rather than overwrite them, and once it did, the generation clip became redundant with it. They want one change that gives the widths a single home both spellings and the schema read.

### Local verification, complete run

| Workspace | Result |
| --- | --- |
| `test-typia-automated` | Success — the suite CI reported red |
| `test-typia-schema` | Success |
| `test-utils` | Success |
| `test-utils-automated` | Success |
| `test-error` | Success — 78 fixtures rejected, 873 diagnostics, 142 distinct codes |
| `test-interface` | Success |
| `test-typia-exact-optional` | Success |
| `test-feature-identity` | Success |
| `test-langchain`, `test-mcp`, `test-vercel` | Success |
| `pnpm test:go` | Success |
| `go test ../native/core/programmers/iterate -tags typia_native_internal` | Success |

Everything above ran on Windows; CI covers Ubuntu, the website build, and `test-typia-bundler-cache`.

### Version

No package version moves here. The native tree needs a `typia` newer than the published `13.3.0` before any npm consumer sees these fixes, and assigning it stays the maintainer's release change.

Issue #2357, opened 2026-08-19
declare dependenciesComplete so a bundler validates the consulted declarations, not the whole reference closure


samchon:
> **Correction (2026-08-19).** The first version of this issue said the 85 untouched files were "not typia's call to make". That was wrong, and the correction makes the ask larger rather than smaller: `dependenciesComplete` reaches a consumer only through the producer's own envelope, and for a typia project the producer is `cmd/ttsc-typia`. The ttsc host passes the field through and never synthesizes it (`envelopeSideChannels` in `packages/ttsc/src/compiler/internal/transformProjectInMemory.ts`), so every file in the envelope is typia's to declare or leave undeclared. The section below is rewritten accordingly.

## Summary

typia reports `dependencies` in its ttsc transform envelope but never `dependenciesComplete`. Every bundler consumer therefore falls back to the compiler's conservative bound for each file, which is the reachability closure of the whole reference graph plus every global-scope declaration, instead of the ~10 declarations typia actually consulted, or the *zero* declarations that can affect a file typia never touched. That bound is re-validated once per delivered module, so the difference is paid `modules × closure` times.

This is not a correctness bug. The consulted-declaration producer from #2100 is right, and it is what makes bundler caches sound in the first place. This is about telling the host that the list is complete, which is a separate field and is what lets the host stop widening it.

## Evidence

A 105-file project (60 application modules, 30 chained type declarations, 20 `typia.createValidate` calls; typia 14.0.0, ttsc 0.28.0). The envelope `TtscCompiler.transform()` returns:

```json
{
  "outputs": 105,
  "filesWithTypiaCall": 20,
  "dependencyFiles": 20,
  "dependencyEdges": 210,
  "dependenciesComplete": 0,
  "globals": 25,
  "graphSources": 361
}
```

and the closure the consumer falls back to, per file:

```json
{ "closureMin": 108, "closureMedian": 297, "closureMax": 346 }
```

So a validator file's delivery re-checks ~300 paths where ~10 would do, and a file with no `typia.*` call at all — 85 of the 105 — re-checks ~300 paths where **none** apply.

## What the consumer does with the field

[The completeness contract](https://ttsc.dev/docs/development/concepts/protocol) defines two bounds per transformed file `F`:

- default: `dependencies[F] ∪ reach(graph.edges, F) ∪ graph.globals ∪ graph.configs ∪ candidates`
- `F` listed in `dependenciesComplete`: `dependencies[F] ∪ graph.configs ∪ candidates`

[...]

PR #2356, opened 2026-08-18
feat(typia): remove cli

samchon:
## Summary

- remove the `typia` executable and its sole `generate` command implementation
- remove CLI-only dependencies, package `bin` metadata, and the generation integration workspace
- rename the unrelated `ttsc` process suite to `test-typia-compiler` and remove CLI references from repository and website documentation

## Verification

- `pnpm format`
- `pnpm --filter ./tests/test-typia-compiler start`
- `pnpm --filter ./tests/test-interface start`
- `pnpm --filter ./tests/test-feature-identity start`
- `pnpm --filter ./packages/typia build`
- `pnpm build`
- `npm pack --dry-run --json --ignore-scripts` in `packages/typia`
- repository trace search and built-artifact search for the removed executable paths and command

`pnpm test` was started but did not complete after the local process output channel became unavailable; the remaining run was terminated instead of duplicated. GitHub CI is required for the complete test result.

## Deferred

None.

PR #2358, opened 2026-08-19
feat(transform): declare dependenciesComplete in the project transform envelope

samchon:
## Intent

Declare `dependenciesComplete` in the `ttsc-typia transform` project envelope so a bundler consumer validates the declarations typia actually consulted for a file instead of the file's whole reference closure plus every global-scope declaration.

Owns the complete accepted cycle:

- #2357 — declare `dependenciesComplete` so a bundler validates the consulted declarations, not the whole reference closure

Any defect this cycle discovers while implementing it is published as its own issue and lands in this same pull request.

## Scope

The declaration transfers responsibility for the dropped bound to typia, so the cycle is not only the field: it is the field plus whatever the reported dependency list has to gain, and whatever a file has to be excluded for, before the claim is true.

## Verification

Pending. This is the implementation-free claim commit.


samchon (commented):
### Ledger — `b7fdc5699` `feat(transform): declare dependenciesComplete in the project transform envelope`

Resolves #2357.

**What landed**

| file | change |
| --- | --- |
| `native/cmd/ttsc-typia/transform.go` | `transformProjectOutput.DependenciesComplete`; the collector now records every published key, tracks the keys it could not bound, and renders the new section |
| `native/transform/CallExpressionTransformer.go` | reports the callee's resolution path for every call expression examined; reports an inferred validated type to the host |
| `native/core/schemas/metadata/MetadataDependency.go` | `MetadataDependency_touchCallee` and its walk |
| `native/core/context/ITypiaContext.go` | `Extras.ReportInferredType`, the host↔transform channel for the inference signal |

**Why it is more than one field.** `dependenciesComplete` transfers responsibility: for a listed file the consumer derives `dependencies[F] ∪ configs ∪ candidates` and drops `reach(edges, F) ∪ globals`, so an unreported input means a stale generated validator. Two inputs were not reported before this commit, and both had to be closed before anything could be declared.

1. *Callee identity.* Reproduced on the base commit: with `src/shim.ts` = `export { is } from "typia"` and `src/e.ts` = `is<Cee>(input)`, `e.ts` **is** transformed and the envelope reported `{"src/e.ts": ["src/c.ts"]}` — no `src/shim.ts`. Re-pointing that barrel changes `typescript["src/e.ts"]` with no reported edge. The mirror case is the worse one: a file with no typia call at all becomes a validator when a barrel is re-pointed *at* typia, and its empty list would have had nothing to invalidate it.
2. *Inferred types.* `GenericTransformer.Scalar`, `CompareEqual/LessTransformer`, and `FunctionalGenericTransformer` take the validated type from the value argument when no type argument is written. Contextual typing then puts the deciding annotation in a file the resolved type never names (`const handler: Handler = (input) => typia.assert(input)`), and no bounded walk recovers it. Those files stay undeclared.

A third shape stays undeclared: a file whose analysis consulted an on-disk default library (`libReplacement`, `noembed`). The envelope has never reported one (#2108) while `graph.globals` carries it, so narrowing would drop an input the default bound watches.

**Tests added** (all under `native/cmd/ttsc-typia/`)

- `project_dependencies_complete_envelope_transform_test.go` — the declared set is exactly the published key set, and the declared entry still carries the direct and transitive declaration files.
- `project_dependencies_complete_inferred_type_transform_test.go` — negative twin: the written-type-argument file is declared, its inference twin is not, and the withheld file keeps what it did report.
- `project_dependencies_complete_replaced_library_transform_test.go` — negative twin under `libReplacement`: the file that consulted the replaced lib is withheld, its control is declared, and the replacement never leaks into `dependencies`.
- `project_dependencies_callee_barrel_transform_test.go` — the barrel that selects a typia callee is reported; a sibling export the reference never traverses is not.
- `project_dependencies_callee_untransformed_barrel_transform_test.go` — the class-2 claim: an untransformed file is declared complete *and* reports the barrel whose retarget would make its call typia's.
- `project_dependencies_complete_untouched_reprint_transform_test.go` — the premise: under `experimentalDecorators` + `emitDecoratorMetadata`, the re-print preserves a type-only import, `enum`, `namespace`, `as const`, and the decorator, and emits no `design:type`.

**Specification-touching change to call out.** `project_dependencies_custom_lib_name_transform_test.go` required every reported dependency to live inside the project. That was a fixture-local proxy for "no compiler default library leaked", and it was never the envelope's contract — the section has always documented absolute values and included `node_modules` declaration files. The files that decide a callee's typia identity are exactly such files, so the proxy is dropped; the virtual-URI check and both presence assertions stay.

**Verification so far.** `go test ../native/...` and `go test ./...` from `packages/typia/test` are green. The TypeScript suites are running against this head; CI, the mutation-sensitivity pass, and Self-Review are pending.

samchon (commented):
### Ledger — Self-Review round 1, and the commits it produced

`2ea136938` folded `Infer` into `Unbound` (the wrapper carried the reason where the reason was not known) and `6098ed3ed` renamed a parameter the callee walk hands whole property accesses. Neither changes behaviour.

`c1d3abc88` is the round's substance. The declaration is a claim about what the analysis reported, and the round found five places where the claim outran the report. Every fix runs in the direction that costs a build rather than a correct output.

| # | what was wrong | reproduced as | fix |
| --- | --- | --- | --- |
| 1 | a `const` object literal has no written annotation, so its property types come from its initializer — which the written-type surface never read | `src/typeofcase.ts -> [..., src/cfg.ts]`, with `src/id.ts` absent, for `export const cfg = { id: "x" as Id }` + `typia.is<typeof cfg>()` | `metadataDependency_valueSurface` and its two-pass walk: the type references written inside the initializer, then the declarations of the values it names |
| 2 | the callee walk descended whole subtrees, charging a file with every identifier in an IIFE body and in nested-call arguments | `src/iife.ts -> [src/id.ts]` for a file with no typia call at all | the walk follows the identity path only; a nested call contributes its callee but not its arguments |
| 3 | the `://` drop claimed parity with the host graph, which skips exactly `bundled:///` | read against `driver/graph.go` | any other scheme is dropped **and** withheld |
| 4 | a transform diagnostic left the file declared, though the analysis stopped partway | — | the diagnostic report withholds the file |
| 5 | completeness was declared even with no `graph`, which is where the universal config chain travels | — | guarded |

Two smaller results ride along: `Begin` is paired with `End`, so a touch arriving outside a transform window is dropped rather than charged to whichever file ran last; and the completeness list is built from the published `typescript` map rather than a parallel slice, which makes it a subset of that map by construction.

**One candidate was rejected.** The read-only pass over `b7fdc5699` also proposed surfacing call and construct signatures structurally. `TestProjectDependenciesSignatureAliasTransform` already decides the opposite and gives the reason — an analysis emits a callable as `typeof x === "function"` and never reads its parameter or return types, so reporting them invalidates consumers on edits that cannot change the output. An attempt to reach a case where the functional analysis does read them failed: `Iterate_metadata_function`'s functional branch never ran under `typia.validate`, `typia.functional.assertFunction`, or `typia.llm.application` in the fixtures tried, because `props.Options.Functional` is the MetadataFactory flag the schema programmers set rather than the plugin option. An entry point that cannot be exercised is worse than none, so it was reverted rather than shipped with a test that could not pass. It stays an open hypothesis, not a published issue.

**Behaviour shape after the round**, measured with a binary built from this head:

| workspace | files | declared complete | withheld | reported list, median | reference closure, median |
| --- | ---: | ---: | ---: | ---: | ---: |
| `tests/test-typia-schema` | 205 | 205 | 0 | 7 | 257 |
| `tests/test-utils` | 312 | 298 | 14 | 5 | — |
| `tests/test-mcp` | 29 | 29 | 0 | 7 | — |

Every one of the 14 withheld files is withheld for the same reason: a `typia.assert(x)` written with no type argument.

**Tests added by this round**: `project_dependencies_inferred_object_alias_transform_test.go` (the alias behind an object-literal initializer is reported; an unrelated constant in the same file is not), `project_dependencies_callee_shape_transform_test.go` (an IIFE stays declarable and its body is not reported; a call to what a function literal returns is withheld), `project_dependencies_complete_diagnostic_transform_test.go` (a rejected call withholds its own file while its clean sibling stays declared).

samchon (commented):
### Ledger — Self-Review round 2, and the commits it produced

`68a0c78b1` and `6a88e49ab`.

The read-only pass over `c1d3abc88` found two confirmed over-declarations, and they share a root cause worth stating plainly, because round 1 patched a witness of it instead of the fact.

Round 1 answered "the alias is written in the initializer" by walking the initializer. That recovered `const cfg = { id: "x" as Id }`. It did not recover:

- `const cfg = make()`, where the alias is named by a `function` declaration's return type — and the *arrow* form of the same helper worked, which is exactly the shape of a patch rather than a rule;
- `const cfg = { base }`, where the shorthand's symbol resolves to the property rather than to the value;
- `const { id } = source`, a destructured binding.

The fact behind all three: **a declaration with no written annotation takes its type from inference, and no walk of written syntax enumerates the files that decide it** — a call's return type, an overload's selection, a conditional's branches, a contextually typed parameter. That is the same problem a typia call with no written type argument has, and this work already answers that one by leaving the file undeclared. Round 2 gives both the same answer and deletes the value walk.

`metadataDependency_inferred` now raises the unbounded admission for an object-literal member, a destructured binding, and a property, parameter, or variable declared by an initializer alone. A literal initializer stays bounded, because `x = 1` names no other file. The withheld file keeps everything the walk did report.

**The rule costs nothing measurable.** Declared counts before and after are identical:

| workspace | files | before | after |
| --- | ---: | ---: | ---: |
| `tests/test-typia-schema` | 205 | 205 | 205 |
| `tests/test-utils` | 312 | 298 | 298 |
| `tests/test-langchain` | 24 | 24 | 24 |
| `tests/test-mcp` | 29 | 29 | 29 |
| `tests/test-vercel` | 31 | 31 | 31 |

Every withheld file is withheld for a `typia.assert(x)` written with no type argument.

**Two smaller results.** A tagged template in callee position now propagates like a call, so `` (fn)`x`() `` is withheld for the reason `(fn)()()` is; `6a88e49ab` pins that branch. And the nested-call comment no longer claims that arguments cannot change which declaration the outer call resolves to — overload selection and generic inference can. The code is unchanged there and the comment states the trade: every indirection a caller writes sits on the callee chain and is reported, while walking arguments would charge a file with every identifier inside `items.map(x => x.a.b).filter(...)`. What is given up needs a generic pass-through applied to the typia namespace, `pick(ns).is<T>(x)`.

**Mutation sensitivity, whole set, run in a disposable checkout with the tests unchanged:**

| mutation | tests that failed |
| --- | --- |
| the on-disk / foreign-scheme withholding | `DependencyCollectorVirtualScheme`, `...CompleteReplacedLibrary` |
| the inferred-call-argument admission | `...CompleteInferredType` |
| the callee identity report | `...CalleeBarrel`, `...CalleeShape`, `...CalleeUntransformedBarrel` |
| the `dependenciesComplete` assignment | eight tests |
| the inferred-declaration admission | `...InferredDeclaration` |
| function literal through a nested call | `...CalleeShape` |
| the diagnostic withholding | `...CompleteDiagnostic` |
| the tagged-template propagation | `...CalleeShape` |

Every row failed exactly the assertions that own the removed behaviour and nothing else. Two branches carry no row because no input reaches them and both are documented as such: the `out.Graph != nil` guard, and `collector.End`.

samchon (commented):
## Handover — everything needed to resume this pull request

Work stopped mid-cycle. This comment is the whole state: what the change is, why each decision was made, exactly what has been verified and how, what is still open, and the commands to pick it up. Nothing here depends on a local file.

---

## 1. Where it stands

| fact | value |
| --- | --- |
| branch | `feat/dependencies-complete` |
| head at handover | `4afcc5bc7` |
| base | `master` @ `f692fd63a` (`chore: release v14.0.0`), unchanged throughout |
| issue | #2357 (closed at handover; the implementation is here, see §7) |
| CI | all six workflows green at `6a88e49ab`; **`4afcc5bc7` was pushed after and its run had not settled when work stopped** |
| local Go suites | green at `4afcc5bc7` (`go test ../native/... -p 1` and `go test ./...` from `packages/typia/test`, exit 0) |
| local TypeScript suites | green at `68a0c78b1` (`pnpm test:packages`, exit 0); `6a88e49ab` and `4afcc5bc7` have not been run locally |
| working tree | clean; one pre-existing untracked path, `tests/test-typia-generate/`, is residue from #2356 and is not this branch's |
| package versions | frozen; nothing here touches a version, a tag, or the peer range |

**Commits**

| sha | subject |
| --- | --- |
| `57f2dee66` | `chore: claim the dependenciesComplete cycle` (empty claim) |
| `b7fdc5699` | `feat(transform): declare dependenciesComplete in the project transform envelope` — carries `Close #2357` |
| `2ea136938` | `refactor(transform): give the dependency collector one withholding method` |
| `6098ed3ed` | `refactor(transform): name the callee walk's node parameter for what it accepts` |
| `c1d3abc88` | `fix(transform): make the completeness claim true where Self-Review found it was not` |
| `68a0c78b1` | `fix(transform): answer an inferred declaration the way an inferred call argument is answered` |
| `6a88e49ab` | `test(transform): pin the tagged-template half of the callee identity rule` |
| `4afcc5bc7` | `fix(transform): default a declaration kind to undeclarable rather than declarable` |

---

## 2. What the change is, and what the claim costs

`ttsc-typia transform` (project mode, no `--file`) prints a JSON envelope with `typescript`, `graph`, and `dependencies`. This adds `dependenciesComplete`.

The consumer contract ([ttsc protocol → Dependency completeness](https://ttsc.dev/docs/development/concepts/protocol), reference consumer `@ttsc/unplugin` `packages/unplugin/src/core/transform.ts` → `declaresCompleteDependencies`, `selectWatchInputs`, `deriveWatchInputs`):

| file `F` | watch inputs the consumer derives |
| --- | --- |
| unlisted | `dependencies[F] ∪ reach(graph.edges, F) ∪ graph.globals ∪ graph.configs ∪ graph.candidates` |
| listed | `dependencies[F] ∪ graph.configs ∪ graph.candidates` |

So **listing a file whose `dependencies[F]` omits a real input makes a bundler serve a stale generated validator.** Under-declaring is always safe. Every design decision below runs in that direction, and that asymmetry is the single rule to keep in mind when touching this code.

Measured on `tests/test-typia-schema` (205 files): the host-owned bound is 148 / 257 / 348 paths per file (min/median/max); the reported list is 0 / 7 / 23. Every file is declared.

**The installed `ttsc` and `@ttsc/unplugin` are 0.19.2, and the field landed in 0.19.3.** In this workspace the field is inert: emitted, passed through as an unknown member, ignored. That is why no end-to-end narrowing assertion exists here, and why the peer range stays `>=0.19.2` (#2357 explicitly leaves that open). It also means the local suites and CI prove *no regression*, not *narrowing works*.

---

## 3. The design in one page

A file is declared complete unless something during its transform admitted that its inputs could not be enumerated. Four things admit that:

1. **A typia call with no written type argument.** `GenericTransformer.Scalar`, `CompareEqual/LessTransformer`, and `FunctionalGenericTransformer` take the validated type from the value argument when the call writes none. Contextual typing then puts the deciding annotation in a file the resolved type never names — `const handler: Handler = (input) => typia.assert(input)` validates whatever `Handler` says, and `Handler`'s file is nowhere in the consulted set. Raised in `CallExpressionTransformer.TransformKnown` via `MetadataDependency_unbounded`.
2. **A declaration whose type is not written on it.** `metadataDependency_bounded` in `MetadataDependency.go`. Its default answer is **no** — a kind the switch does not name withholds — because the unsafe direction is declaring. Unbounded kinds are an object-literal member, a shorthand, a destructured binding, and a property/parameter/variable with neither an annotation nor a literal initializer.
3. **A consulted file the envelope cannot report but the host graph still carries.** An on-disk default library under `libReplacement`/`noembed` (the envelope has never reported one, #2108, while `graph.globals` carries it), and any virtual URI scheme other than `bundled:///` (the driver skips exactly that prefix, `ttsc/packages/ttsc/driver/graph.go:16`).
4. **A typia diagnostic.** The analysis stopped partway, so what it consulted is not the file's whole input set. Wired by wrapping `extras.AddDiagnostic` in `runTransformProject`.

Plus one structural guard: the section is emitted only when `out.Graph != nil`, because the config chain stays universal for a declared file only by travelling in the same envelope's `graph`.

**Two inputs had to start being reported before any of this was true.** Both were harmless before the declaration existed — an unlisted file keeps `reach(edges, F)`, which covers them — which is why neither was published as its own issue:

- **Callee identity.** typia decides its own call sites by the file that declares the resolved signature, so a barrel between the caller and typia decides whether a call is rewritten. Reproduced on the base commit: with `src/shim.ts = export { is } from "typia"` and `src/e.ts` calling `is<Cee>(input)`, `e.ts` **is** transformed and the envelope reported `{"src/e.ts": ["src/c.ts"]}` — no `shim.ts`. The mirror is worse: a file with no typia call becomes a validator when a barrel is re-pointed *at* typia, and its empty list would have had nothing to invalidate it. `MetadataDependency_touchCallee` now walks the callee's identity path for every call expression the transformer examines, recognized or not.
- **Type-parameter defaults on an interface or class.** `interface Wrapper<T = Id>` validated as `Wrapper` reads whatever `Id` is, and `type Id = string` interns as the bare intrinsic with no symbol to follow. `type AliasWrapper<T = Id> = {...}` always reported it; the interface branch surfaced only index signatures. Fixed in `metadataDependency_typeSurface`.

**The callee identity path** (`metadataDependency_walkCallee`) is deliberately narrower than the callee's syntax tree:

- a nested call contributes its own callee but **not** its arguments;
- a function literal is where it stops — written directly in callee position it is the most bounded declaration there is (the call resolves to that literal, in this file), and reached *through* a nested call (`(() => is)()<T>(x)`, `` (fn)`x`() ``) the identity is whatever a body returns, so it returns unbounded;
- an unrecognized kind takes the unbounded branch.

---

## 4. Every file changed

| file | what and why |
| --- | --- |
| `native/cmd/ttsc-typia/transform.go` | `transformProjectOutput.DependenciesComplete`; the collector gained `unbounded`, `Unbound`, `End`, `ToCompleteJSON(published)`; `transformDependencyValue` distinguishes a drop that costs a declaration from one that does not; the `bundled:///` prefix matches the driver exactly; `runTransformProject` takes `*extras` to wire the listener and wrap `AddDiagnostic` |
| `native/transform/CallExpressionTransformer.go` | reports the callee's identity for every call expression before resolving it; reports an inferred validated type as unbounded |
| `native/core/schemas/metadata/MetadataDependency.go` | `MetadataDependency_IListener{File, Unbounded}` replaces the bare listener func; `MetadataDependency_unbounded`; `MetadataDependency_touchCallee` + `walkCallee` + `touchDeclarations`; `metadataDependency_bounded` + `metadataDependency_literal`; type parameters and mapped types join `typeSurface` |
| `native/core/context/ITypiaContext.go` | net zero — a hook was added in `b7fdc5699` and removed in `c1d3abc88` once the metadata listener carried the same signal |
| `native/cmd/ttsc-typia/project_dependencies_custom_lib_name_transform_test.go` | **specification-touching:** it required every reported dependency to live inside the project. That was a fixture-local proxy for "no compiler default library leaked" and was never the envelope's contract — the section has always documented absolute values and included `node_modules` declaration files, and the files that decide a callee's typia identity are exactly such files. The virtual-URI check and both presence assertions stay |

**Tests added** (all in `native/cmd/ttsc-typia/`):

| test | pins |
| --- | --- |
| `project_dependencies_complete_envelope_transform_test.go` | the declared set is exactly the published key set; the declared entry keeps its direct and transitive declaration files |
| `project_dependencies_complete_inferred_type_transform_test.go` | written type argument declared / inference twin withheld, and the withheld file keeps what it reported |
| `project_dependencies_complete_replaced_library_transform_test.go` | `libReplacement` consumer withheld / control declared, and the replacement never leaks into `dependencies` |
| `project_dependencies_complete_untouched_reprint_transform_test.go` | under `experimentalDecorators` + `emitDecoratorMetadata`, the re-print keeps a type-only import, `enum`, `namespace`, `as const`, and a decorator, and emits no `design:type` — the class-2 premise |
| `project_dependencies_complete_diagnostic_transform_test.go` | a rejected call withholds its own file; its clean sibling stays declared |
| `project_dependencies_callee_barrel_transform_test.go` | a barrel selecting a typia callee is reported; a sibling export the reference never traverses is not |
| `project_dependencies_callee_untransformed_barrel_transform_test.go` | an untransformed file is declared **and** reports the barrel whose retarget would make its call typia's |
| `project_dependencies_callee_shape_transform_test.go` | an IIFE stays declarable and its body is not reported; a call to what a function literal returns, and a tagged-template equivalent, are withheld |
| `project_dependencies_inferred_declaration_transform_test.go` | annotated `const` declared / initializer-typed twin withheld |
| `project_dependencies_enum_member_value_transform_test.go` | an enum writing its own values (including `-1`) declared / one borrowing a value withheld |
| `project_dependencies_type_parameter_default_transform_test.go` | a type-parameter default's alias file is reported; an unused sibling interface's is not |
| `dependency_collector_virtual_scheme_test.go` | collector probe: `bundled:///` costs no declaration, a foreign scheme does |

---

## 5. How everything was verified — reproduce it exactly

**Suites**

```bash
pnpm install
go -C packages/typia/test test ../native/... -count=1 -p 1   # -p 1: this machine OOMs at full parallelism
go -C packages/typia/test test ./... -count=1
pnpm test:packages
node scripts/format-go.cjs                                    # then `git diff` must be empty
```

**Measurement** — build the binary and dump an envelope for any workspace:

```bash
go -C packages/typia/test build -o /tmp/ttsc-typia ../native/cmd/ttsc-typia
/tmp/ttsc-typia transform --cwd <ABSOLUTE path to a workspace> --tsconfig tsconfig.json > env.json
```

`--cwd` must be absolute. Then compare `dependenciesComplete` against `Object.keys(typescript)`, and the per-file reported list against `reach(graph.edges, F) ∪ graph.globals`. Results at `4afcc5bc7`:

| workspace | files | declared | withheld |
| --- | ---: | ---: | ---: |
| `tests/test-typia-schema` | 205 | 205 | 0 |
| `tests/test-utils` | 312 | 298 | 14 |
| `tests/template` | 181 | 181 | 0 |
| `tests/test-interface` | 45 | 45 | 0 |
| `tests/test-vercel` | 31 | 31 | 0 |
| `tests/test-mcp` | 29 | 29 | 0 |
| `tests/test-langchain` | 24 | 24 | 0 |
| `tests/test-typia-exact-optional` | 3 | 3 | 0 |

All 14 withheld files are withheld for a `typia.assert(x)` written with no type argument. `benchmark/` was attempted and exceeded a ten-minute budget; it has never been measured.

**Mutation sensitivity** — the disposable checkout recipe, which matters because the Go workspace resolves `ttsc` through the real `node_modules`:

```bash
git clone --no-checkout --local --no-hardlinks --branch feat/dependencies-complete <repo> D:\tmut
cd D:\tmut && git sparse-checkout init --cone && git sparse-checkout set packages/typia && git checkout feat/dependencies-complete
cmd /c mklink /J "D:\tmut\node_modules" "<repo>\node_modules"
cmd /c mklink /J "D:\tmut\packages\typia\node_modules" "<repo>\packages\typia\node_modules"
```

A short path is required — cloning into a deep temp directory fails on Windows `MAX_PATH`. Then mutate one product change, run `go -C D:\tmut\packages\typia\test test ../native/cmd/ttsc-typia/ -run 'TestProjectDependencies|TestDependencyCollector' -count=1`, and `git checkout -- packages/typia/native` between rows. Last full run, at `6a88e49ab`:

| mutation removed | tests that failed |
| --- | --- |
| `Touch`'s `if value.withhold { collector.Unbound() }` | `DependencyCollectorVirtualScheme`, `...CompleteReplacedLibrary` |
| `MetadataDependency_unbounded` on an argument-typed call | `...CompleteInferredType` |
| the `MetadataDependency_touchCallee` call | `...CalleeBarrel`, `...CalleeShape`, `...CalleeUntransformedBarrel` |
| the `dependenciesComplete` assignment | eight tests |
| the inferred-declaration admission | `...InferredDeclaration` |
| `return nested == false` → `return true` | `...CalleeShape` |
| the `AddDiagnostic` wrapper | `...CompleteDiagnostic` |
| the tagged-template propagation | `...CalleeShape` |

Every row failed exactly the assertions that own the removed behaviour and nothing else. **`4afcc5bc7` has not had a mutation row run for `metadataDependency_bounded`'s inverted default, the enum-member branch, or the type-parameter surface** — that is the first thing to redo (§6).

Two branches carry no row because no input reaches them, and both are documented in place rather than counted as covered: the `out.Graph != nil` guard, and `collector.End`.

---

## 6. What is left, in order

1. **Watch CI on `4afcc5bc7`.** It was pushed at handover and had not settled. All six workflows were green on `6a88e49ab`.
2. **Run `pnpm test:packages` locally at `4afcc5bc7`.** It was last run at `68a0c78b1`. The Go suites are green at `4afcc5bc7`.
3. **Re-run the mutation matrix** at the current head, adding rows for the three behaviours `4afcc5bc7` introduced: the inverted default in `metadataDependency_bounded`, the enum-member branch, and the type-parameter surface. Expected owners: `...InferredDeclaration` / `...EnumMemberValue` / `...TypeParameterDefault`.
4. **One more Self-Review round over the whole base-to-head diff.** Three rounds ran; each found real over-declarations, and rounds two and three each found them in the fix the previous round had just made. The class is "a consulted type whose deciding file the written surface does not name". Round three's inversion means a miss now costs narrowing instead of correctness, but that is exactly why a fresh round should re-examine the *bounded* list in `metadataDependency_bounded` rather than the unbounded one.
5. **Merge gate**, per `.agents/skills/issue-campaign/development.md`: re-read `origin/master` (it was `f692fd63a` throughout, so the prospective merge result is the head itself), confirm required checks are green on the exact head, then squash-merge. The merge closes #2357 from `b7fdc5699`'s `Close #2357` line; #2357 was closed manually at handover, so verify its final state after merge rather than expecting the keyword to do it.
6. **Cleanup**: delete the disposable checkout `D:\tmut` if it still exists, and the local branch after merge.

---

## 7. Open questions and stated residuals

Everything here is honest about not being closed. None of it blocks the merge; all of it is what a reviewer should push back on first.

- **A nested call's arguments are not walked.** An argument can select an overload or instantiate a generic, so a generic pass-through — `pick(ns).is<T>(x)` — resolves through a module the callee walk never names, while the file stays declared. Every indirection a caller actually writes (a factory, a barrel, a namespace alias, a local binding) sits on the callee chain and *is* reported. Closing it means walking arguments, which would charge every file with the identifiers inside `items.map(x => x.a.b).filter(...)`. The comment in `walkCallee` states the trade rather than claiming the gap does not exist.
- **A callee whose identity is decided by a computed type.** `f()<T>(x)` where `f`'s return type is an alias chain in a third file: `f` is reported, the chain is not. Same class as the above.
- **Whether any analysis reads a call signature's parameter or return types.** `TestProjectDependenciesSignatureAliasTransform` states they are never consulted and this branch keeps that contract. An attempt to reach a counter-example failed — `Iterate_metadata_function`'s functional branch never ran under `typia.validate`, `typia.functional.assertFunction`, or `typia.llm.application` in the fixtures tried, because `props.Options.Functional` is the MetadataFactory flag the schema programmers set, not the plugin option. An entry point was written and then **reverted rather than shipped untested**. If someone can reach that branch, an alias of an intrinsic among a signature's written types is consulted and unreported, and the fix belongs at the exploration site in `iterate_metadata_function.go`, not in `typeSurface`.
- **A new declaration merging into an already-consulted symbol.** A file that declared nothing about a symbol when the envelope was produced is not reported, so adding `interface Foo { extra: string }` to it later does not invalidate. `graph.globals` would have covered it for an unlisted file. The protocol's **Globals do not stay universal** rule accepts exactly this trade; it is what the narrowing is for.
- **`noembed` / `libReplacement` projects get much less narrowing.** The callee walk feeds the same channel as the type analysis, so a file that merely calls `[].map(...)` touches an on-disk `lib.es5.d.ts` and is withheld. Over-conservative rather than wrong — a lib path can never satisfy `callExpressionTransformer_targetModule` — and unmeasured, because tsgo serves libraries from `bundled:///` by default.
- **An unreproduced crash.** One intermediate probe fixture (`typia.llm.application<App, { model: "chatgpt" }>()` beside a `typeof`-referenced function) panicked the tsgo printer (`printer.emitIdentifierText` debug assert) in project transform mode, on the **base commit** as well as this branch, so it is not a regression from this work. A five-row minimisation matrix on the base commit did not reproduce it. Not published as an issue: an unreproducible crash is a hypothesis.
- **`benchmark/` has never been measured** for declaration rate or for transform cost.
- **No end-to-end narrowing test exists**, and cannot until the workspace moves past `ttsc` 0.19.2. When it does, `tests/test-typia-bundler-cache` becomes the oracle: its steps 2–5 all mutate files that the reported list names (a consulted type, a project-owned `lib.custom.d.ts`, a re-pointed barrel, an index-signature target), so they should keep passing under narrowing — and if one stops, that is a real under-report, not a test to relax.

---

## 8. Two things worth knowing before touching this code

**The safe direction is one-way.** Every predicate here answers "may this file be declared". Getting it wrong toward *no* costs a rebuild. Getting it wrong toward *yes* serves a stale validator, and nothing in the toolchain will tell you: the host performs no audit of a completeness list, by design, because verifying one costs exactly what the declaration saves. The only diagnostic is the protocol's own — remove the field and see whether the stale output goes away.

**The reviews on this pull request are the reasoning, not a summary of it.** Each round's comment records what was reproduced, what was rejected and why, and what was reverted. Round two rejected a proposal that contradicted an existing tested contract; round three rejected a patch that could not be exercised. Both rejections are as load-bearing as the fixes.

samchon (commented):
### Ledger — `4afcc5bc7` and `f0de576fb`, and the two claims that were not held

The read-only pass over `68a0c78b1` reported two over-declarations, both reproduced by mutating the missing file and watching `typescript[F]` change while `F` stayed declared:

- **An enum member that borrows its value.** `enum Kind { A = Other.X }` reads the resolved literal straight into the validator (`"x" === input`), and `KindEnumMember` was in neither switch, so `other.ts` was never named.
- **A type-parameter default on an interface or class.** `interface Wrapper<T = Id>` validated as `Wrapper` reads whatever `Id` is; the structural surface offered only index signatures. The equivalent `type AliasWrapper<T = Id>` always reported it, which is what pinned the location.

The pass also made a structural point that mattered more than either: `metadataDependency_inferred` was an allow-list of *unbounded* kinds whose default answer was **bounded**, so every declaration kind nobody enumerated silently declared its consumer complete. The asymmetry runs the other way — an unreported file keeps the reference closure and costs a rebuild; a declared file whose inputs were not enumerated serves a stale validator. `4afcc5bc7` inverts it: the predicate is now `metadataDependency_bounded`, and a kind it does not name withholds.

Measured across eight workspaces, the inversion costs nothing — 816 of 830 files declared, exactly as before, every withheld file withheld for a `typia.assert(x)` written with no type argument:

| workspace | files | declared |
| --- | ---: | ---: |
| `tests/test-typia-schema` | 205 | 205 |
| `tests/test-utils` | 312 | 298 |
| `tests/template` | 181 | 181 |
| `tests/test-interface` | 45 | 45 |
| `tests/test-vercel` | 31 | 31 |
| `tests/test-mcp` | 29 | 29 |
| `tests/test-langchain` | 24 | 24 |
| `tests/test-typia-exact-optional` | 3 | 3 |

**Then the mutation matrix rejected two of that commit's own claims.** Both rows came back green, which is the useful outcome:

- Flipping the inverted default back to "declare" broke **no test**. Every fixture reached a kind the switch names, so the default itself — the whole reason for the inversion — was unpinned. A destructured binding is such a kind, and `export const { id: picked } = written` validated through `typeof picked` now pins it.
- Removing the type-parameter list append from the interface surface broke **no test** either, and the reason is that it was never the path that worked: a parameter the members use is reached through the member that uses it, and its own `KindTypeParameter` declaration surfaces the constraint and default from there. A parameter no member uses is not consulted at all, so appending the list only reported an alias behind an unused default. `f0de576fb` deletes the append and keeps the declaration surface; mutating *that* now fails the test.

Re-run at `f0de576fb`, every row fails exactly its owner:

| mutation removed | test that failed |
| --- | --- |
| the withholding default (`return false` → `return true`) | `...InferredDeclaration` |
| the enum-member / object-literal-member branch | `...EnumMemberValue` |
| the `KindTypeParameter` surface | `...TypeParameterDefault` |
| the negative-literal branch (`-1`) | `...EnumMemberValue` |

Together with the eight rows already recorded, every product branch in this change is now covered except two documented as unreachable: the `out.Graph != nil` guard and `collector.End`.

One correction to `68a0c78b1`'s message, which said a withheld file "keeps everything the walk did report": it keeps what the walk still reaches, which for `export const cfg = { id: "x" as Id }` is `cfg.ts` but no longer `id.ts`. Sound — a withheld file falls back to the closure, which carries both — but the sentence overstated it.

samchon (commented):
### Ledger — Self-Review round 4, and the two commits it produced

Round over the whole `f692fd63a..6dc28ceb6` diff, resumed from the handover. `f56b7a54f` and `c03bf3e7d` are what survived it. CI was green on all six workflows at `6dc28ceb6` before the round started, and `go test ../native/... -p 1`, `go test ./...`, and `pnpm test:packages` were all green locally at that head.

**One confirmed over-declaration, and it is the same channel round three closed from the other end.**

Round three found that `enum Kind { A = Other.X }` reads a borrowed *value* into the validator. A computed member key reads the same kind of value into the same output, as the property *name*:

```ts
// src/kind.ts
export enum Kind { A = "alpha" }
// src/computed.ts
import { Kind } from "./kind";
export interface Computed { [Kind.A]: number }
// src/main.ts
typia.is<Computed>(x)   // emits `"number" === typeof input.alpha`
```

Reproduced against a binary built from `6dc28ceb6`: renaming `Kind.A` to `"renamed"` changed `typescript["src/main.ts"]` to `input.renamed`, while `dependencies["src/main.ts"]` stayed `["src/computed.ts"]` and `src/main.ts` stayed in `dependenciesComplete`. A consumer narrowing on that list serves a stale validator. The checker interns the resolved key as a bare string literal with no symbol to follow, and the declaration's written type says nothing about the name, so neither existing walk could reach `kind.ts`.

Three spellings, all over-declared before the fix and all covered now:

| written as | file that decided the emit | reported before | reported after |
| --- | --- | ---: | ---: |
| `[Kind.A]`, enum member | `kind.ts` | no | yes |
| `[KEY]`, imported `const` | `keyconst.ts` | no | yes |
| `[Kind.A]` through a barrel | `barrel.ts`, `kind.ts` | no | yes |

`metadataDependency_touchName` hands a computed name to `metadataDependency_touchVisited` rather than reporting it directly, which is what makes it a rule instead of a patch: the symbol's own declaration is reported, its alias path registers the barrel, and `metadataDependency_bounded` answers whether that declaration fixes the value or borrows it. `const BORROWED = VALUE` is the negative twin and is **withheld**, exactly as `enum Kind { A = Other.X }` is. A literal key (`["gamma"]`) is written where it is used and stays declared, which is the twin in the other direction.

**Two dead predicate entries, measured rather than argued.**

`72984baac` added `KindNamedTupleMember` and the two JSDoc tags to both the surface and the predicate; `6dc28ceb6` deleted the surface halves as dead and kept the predicate halves. The mutation matrix says the predicate halves are dead too. A binary built without `KindNamedTupleMember` prints the identical envelope for the fixture written for it — `[id: Id]` still reports `id.ts`, `a.ts` is still declared — and a binary built without both JSDoc tags does the same for an `allowJs` project whose `@property {import("./id").Id}` names another file. Neither kind ever reaches `metadataDependency_bounded`, for the reason `6dc28ceb6` already gave about the surface: each is reached through the written node that spells it, and `metadataDependency_walkNode` descends that node whole.

So they are gone, and `project_dependencies_jsdoc_typedef_transform_test.go` holds the half that had no fixture at all. It fails if a JSDoc tag ever does start arriving at the predicate, because the default now withholds.

**§7's largest open question is closed, in the negative.**

The handover recorded that nobody could reach `Iterate_metadata_function`'s functional branch, so whether an analysis reads a call signature's parameter or return types stayed a hypothesis. The entry point is `typia.equals<T>()` with the plugin option `functional: true` — `CompareEqualProgrammer` is the one place that sets `MetadataFactory_IOptions.Functional` from `OptionPredicator.Functional(props.Context.Options)`. Reached it, and the branch is harmless:

```ts
interface Callable { (p: P): void }   // type P = string, in another file
```
```
typia.equals<Callable>(x)  ->  "function" === typeof input
```

Changing `P` from `string` to `{ alpha: number; beta: string[] }` left `typescript["src/main.ts"]` byte-identical. The functional branch explores parameter and return types into the metadata, but `CompareEqualProgrammer` emits a callable as `typeof v === "function"` and reads nothing else from it, so `TestProjectDependenciesSignatureAliasTransform`'s contract holds under the option that reaches the branch. This is a settled residual, not an open one, and the revert in round 2 was correct.

**Two other hypotheses reproduced and killed.**

- *An empty base interface.* `interface A extends Empty {}` reads nothing from `empty.ts` today and gaining its first property changes the validator. Already answered: `emplace_metadata_object_touch_heritage` walks the explicit chain, and `TestProjectDependenciesEmptyHeritageTransform` pins it.
- *A get accessor with an inferred return type.* `KindGetAccessor` is bounded through the callable branch while its type comes from the body, which looked like the same over-declaration. It is not one: `emplace_metadata_object_significant` excludes accessors under both option settings, so a getter contributes nothing to the emitted validator. Measured — a class with `plain: number` and `get id()` emits `"number" === typeof input.plain` and nothing else.

**Mutation sensitivity, in a disposable checkout at this head with the tests unchanged:**

| mutation removed | test that failed |
| --- | --- |
| the `metadataDependency_touchName` call | `...ComputedKey` |
| the literal short-circuit in the name walk | `...ComputedKey` |
| the qualifier step in the name walk | `...ComputedKey` |
| `KindNamedTupleMember` in the bounded predicate | none — which is why it is deleted |
| both JSDoc tags in the bounded predicate | none — same |

Every row that owns behaviour failed exactly `...ComputedKey` and nothing else. The unresolvable-reference case in the name walk was left without a row on purpose: a name that resolves to no symbol declares no property symbol for the enumeration to hand over — measured, `[MISSING]` emits a dynamic-key validator instead — so it takes the shared fall-through rather than an exit of its own, and the code says so.

**The rule costs nothing.** Declared counts across every measurable workspace are identical to `4afcc5bc7`'s:

| workspace | files | declared |
| --- | ---: | ---: |
| `tests/test-typia-schema` | 205 | 205 |
| `tests/test-utils` | 312 | 298 |
| `tests/template` | 181 | 181 |
| `tests/test-interface` | 45 | 45 |
| `tests/test-vercel` | 31 | 31 |
| `tests/test-mcp` | 29 | 29 |
| `tests/test-langchain` | 24 | 24 |
| `tests/test-typia-exact-optional` | 3 | 3 |

816 of 830, every withheld file still withheld for a `typia.assert(x)` written with no type argument.

**Verification at `c03bf3e7d`.** `go test ../native/... -count=1 -p 1` green (exit 0, `cmd/ttsc-typia` 176s). `node scripts/format-go.cjs` leaves `git diff --numstat` empty. `pnpm test:packages` was green at `6dc28ceb6`; neither commit can change an emitted validator, because `MetadataDependency` only reports to a listener no host but the project transform registers. CI on this head is the gate and is running.

samchon (commented):
### Ledger — Self-Review round 5, and the commit it produced

Round over `f692fd63a..c03bf3e7d`. It produced `e8eab1d35` and, before that, a correction to my own round-4 code.

**A second over-declaration, in the same class as the first and reached from the opposite side.**

Round 4 found that a *computed key* names a file the walk never reported. Round 5 found that a *qualified type reference* does too, and the reason is precise:

```ts
// src/kind.ts   export enum Kind { A = "alpha", B = "beta" }
// src/barrel.ts export { Kind } from "./kind";
// src/doc.ts    import { Kind } from "./barrel"; export interface Doc { kind: Kind.A }
// src/main.ts   typia.is<Doc>(x)   ->   "alpha" === input.kind
```

`TestProjectDependenciesBarrelReexportTransform` pins the bare spelling and passes because the checker answers `Alpha` with the **local import alias**, whose hops `metadataDependency_touchPath` walks. `Kind.A` is answered with the **enum member** — not an alias, no path to walk — so the barrel went unreported.

Measured against a binary built from `c03bf3e7d`: `dependencies["src/main.ts"] = ["src/doc.ts", "src/kind.ts"]`, `src/main.ts` declared complete. Editing **only** `barrel.ts` to re-export `Kind` from a different module changed `typescript["src/main.ts"]` to `"omega" === input.kind`. An emitted constant moved with nothing left to invalidate the file that narrowed on that list.

`metadataDependency_touchQualifier` resolves every prefix of a qualified reference the same way the terminus is resolved. It runs from `metadataDependency_walkNode`, so it covers all four written spellings that walk already recognizes — a type reference, a heritage expression, a `typeof` target, and an `import("...").T` qualifier — and the computed-key walk now shares it instead of carrying its own recursion.

The fix reaches further than the barrel case, which is why it belongs in the walk rather than beside it:

| written as | reported before | reported after |
| --- | --- | --- |
| `kind: Kind.A` through `export { Kind } from "./kind"` | terminus only | barrel + terminus |
| `kind: ns.Kind.A`, `import * as ns from "./barrel"` (star re-export) | terminus only | namespace hop + barrel + terminus |
| `typeof x.y` for an un-annotated `const x` | declared | **withheld** — the qualifier is now an inferred declaration |
| `extends factory.Base` for an object-literal `factory` | declared | **withheld**, same reason |

The last two were silent over-declarations of the round-3 class that no fixture had exhibited; they now take the withholding answer the design already specifies.

**And a correction to round 4's own code.** Applying the same standard that deleted the tuple and JSDoc predicate entries to what I had just written: `walkName` carried `KindBigIntLiteral`, `KindParenthesizedExpression`, and `KindQualifiedName` cases that no input reaches. Measured, `[(Kind.A)]` and `[1n]` declare **no property at all** (`_io0` is `true`), so the enumeration never hands either shape over, and a qualified name is type syntax that cannot appear in a computed key. `["alpha"]`, `` [`tmpl`] ``, `[1]`, `[KEY]`, and `[Kind.A]` each do declare one and each is now the only kind named. The test also gained the case that pins the bare-name branch in **both** directions — round 4's `[BORROWED]` twin only proved the withholding half, because removing the whole branch still withheld it.

**Four other hypotheses reproduced and killed this round.**

- *`typeof import("./m")` as a module type.* The emit is `_io0 = true` — no property survives — and `m.ts` is reported anyway.
- *Module augmentation.* `declare module "./base" { interface Doc { extra: number } }` puts a second declaration on the symbol, `metadataDependency_touchVisited` iterates all of them, and `aug.ts` is reported. Emit reads `input.extra`; file declared; sound.
- *An element-access key.* `[Kind["A"]]` declares no property, so nothing from `kind.ts` reaches the emit and declaring is correct.
- *An unresolvable key.* `[MISSING]` emits a dynamic-key validator, so the nil-symbol path in the name walk is unreachable — it takes the shared fall-through rather than an exit of its own, and the code says so.

**Mutation sensitivity at this head**, disposable checkout, tests unchanged:

| mutation removed | test that failed |
| --- | --- |
| the `metadataDependency_touchQualifier` call in `walkNode` | `...QualifiedBarrel` |
| the same call in `walkName` | `...ComputedKey` |
| `KindQualifiedName` in `metadataDependency_left` | `...QualifiedBarrel` |
| `KindPropertyAccessExpression` in `metadataDependency_left` | `...ComputedKey` |
| the reference case in the name walk | `...ComputedKey` |
| `touchVisited` inside the name walk | `...ComputedKey` |
| the literal short-circuit | `...ComputedKey` |
| the `metadataDependency_touchName` call | `...ComputedKey` |

**An independent oracle, and what it is worth.** Beyond the shape probes, the whole `tests/test-mcp` workspace was audited differentially: for each of its 29 tracked sources, rewrite every `string` to `number`, re-run the project transform, and list every file whose published text moved while it stayed declared without naming the mutated file. Zero. The row counts are reported rather than the verdict alone, because an audit that observed no change would have printed the same "clean":

| mutated file | consumers whose text moved | over-declared |
| --- | ---: | ---: |
| `src/structures/Greeter.ts` | 5 | 0 |
| `src/structures/TicketSearch.ts` | 1 | 0 |
| `src/structures/Inspector.ts` | 1 | 0 |
| the other 26 sources | 0 | 0 |

Its reach is honest about being narrow: a `string -> number` rewrite exercises the property-type channel, which is the best-covered one, and would have caught **neither** defect this cycle found — a computed key needs an enum *value* to move, a qualified reference needs a barrel re-pointed. Those two channels were probed by hand, above. The audit is a floor, not the ceiling.

**The rule still costs nothing.** Declared counts across the eight measured workspaces are unchanged from `4afcc5bc7`: 816 of 830, median reported list 7 / 5 / 1 / 0 / 8 / 7 / 8 / 5, every withheld file withheld for a `typia.assert(x)` written with no type argument.

**Verification at `e8eab1d35`.** `go test ../native/... -count=1 -p 1` green, zero failures. `node scripts/format-go.cjs` leaves the diff empty. CI on this head is the gate.

samchon (commented):
### Ledger — Self-Review round 6, and the two commits it produced

Round over `f692fd63a..e8eab1d35`. It produced `9ae1979db` and `6c5168256`, and it turned §7's largest remaining residual into a reproduced defect.

**The pass-through residual is real.** The comment beside `metadataDependency_walkCallee` stated it in the abstract — "an argument can still select an overload or instantiate a generic, so a pass-through like `pick(ns).is<T>(x)` resolves through a module the walk never names". Written out, it is not abstract at all:

```ts
// src/ns.ts    import typia from "typia"; export const ns = typia;
// src/pick.ts  export const pick = <T>(value: T): T => value;
// src/main.ts  export const check = (input: unknown) => pick(ns).is<Foo>(input);
```

Against a binary built from `e8eab1d35`, `main.ts` **is** rewritten into a generated validator, and the envelope declared it complete reporting `["src/foo.ts", "src/pick.ts"]`. `ns.ts` is the file that decides the call is typia's — re-point it and the validator becomes a plain call again — and nothing in the narrowed bound watches it.

`metadataDependency_touchApplied` walks what a nested call is applied to, through the same identity walk, and **drops the answer**. The asymmetry is the design, not an oversight:

- *Reporting* closes the pass-through: `ns` is a written name like every other one on the chain.
- *Dropping the boolean* is what makes it free. An argument's shape is not the caller's identity. `arr.map((x) => …).filter(…)` would otherwise withhold every file that chains an array method, over a callback that decides nothing about which `filter` the outer call names — and because the walk already stops at a function literal rather than descending, no file is charged with the identifiers inside a callback either.

Measured across the eight workspaces, invisible: 816 of 830 declared, medians 7 / 5 / 1 / 0 / 8 / 7 / 8 / 5, unchanged. Only a callee that *contains* a nested call pays anything.

**And the same standard applied to that commit's own new line.** `9ae1979db` walked a tagged template's template for symmetry with a call's arguments, with no fixture reaching it — the exact shape the previous two commits deleted entries for. So it was tested rather than assumed, and it turns out to carry the identical defect: with `tag` declared `<T>(_s: TemplateStringsArray, value: T): T => value`, ``tag`x${ns}`.is<Foo>(input)`` is rewritten too, and without the template walk `ns.ts` goes unreported. `6c5168256` pins it beside its call-expression twin.

**Three hypotheses reproduced and killed.**

- *`typeof import("./m")` as a module type.* Emit is `_io0 = true` — no property survives — and `m.ts` is reported regardless.
- *Module augmentation.* `declare module "./base" { interface Doc { extra: number } }` puts a second declaration on the symbol; `metadataDependency_touchVisited` iterates all of them, `aug.ts` is reported, the emit reads `input.extra`, the file is declared. Sound.
- *A namespace import through a star barrel.* `import * as ns from "./barrel"` with `export * from "./kind"`, member type `ns.Kind.A`: all three of `nsimport.ts`, `barrel.ts`, `kind.ts` are reported — the qualifier walk `e8eab1d35` added covers this spelling too.

**A differential audit, and the bug in the audit itself.** Beyond the shape probes, whole workspaces were audited: for each tracked source, rewrite every `string` to `number`, re-run the project transform, and list every file whose published text moved while it stayed declared without naming the mutated file.

`tests/test-mcp`, all 29 sources: **zero**. Row counts are reported rather than the verdict alone, because an audit that observed no change would print the same "clean":

| mutated file | consumers whose text moved | over-declared |
| --- | ---: | ---: |
| `src/structures/Greeter.ts` | 5 | 0 |
| `src/structures/TicketSearch.ts` | 1 | 0 |
| `src/structures/Inspector.ts` | 1 | 0 |
| the other 26 | 0 | 0 |

The first `tests/test-utils` run reported four over-declarations, and every one was **the harness lying**. Reproduced individually, each mutation moved only its own file. The cause: the loop restored each file with `git checkout --`, that call raced this session's own `git add` on the same repository, and `index.lock: File exists` made the restore fail silently — so every later iteration compared against a tree that still carried the previous mutations. Re-run with a pre-iteration cleanliness assertion, it aborted on exactly that condition instead of reporting a finding. Re-run with no concurrent git process at all, over `tests/test-langchain`, `tests/test-vercel` and `tests/test-mcp` -- 84 sources, 10 observed consumer changes, **zero** over-declared. `tests/test-utils` was not re-run: its 312 sources need ~30 minutes of exclusive access to the repository index, and the channel it exercises is the one already covered.

Two things worth saying about it. First, the audit's reach is narrow by construction: a `string -> number` rewrite exercises the property-type channel, and would have caught **none** of the three defects this cycle found — a computed key needs an enum value to move, a qualified reference needs a barrel re-pointed, a pass-through needs an argument re-bound. It is a floor. Second, an oracle that cannot fail is worse than no oracle, and this one could: the fix was to make the harness assert what it assumed rather than to trust a clean report.

**Mutation sensitivity at this head**, disposable checkout, tests unchanged:

| mutation removed | test that failed |
| --- | --- |
| the applied walk on a call's arguments | `...CalleeArgument` |
| the applied walk on a tagged template | `...CalleeArgument` |
| dropping the applied walk's answer (raise `unbounded` instead) | `...CalleeArgument` |

Every row failed exactly that one test and nothing else. The third row is the one worth naming: restoring the boolean withholds the chained-array-method twin, which is precisely the cost the design refuses to pay.

**Transform cost, since the walk now runs over more syntax.** Wall clock for one project transform, binaries built from `f692fd63a` (master) and from this head, three runs each after a warm-up:

| workspace | files | master | head |
| --- | ---: | ---: | ---: |
| `tests/test-typia-schema` | 205 | 1184-1285 ms | 1181-1236 ms |
| `tests/test-utils` | 312 | 1825 ms | 1863-1866 ms |

Inside noise in both directions. Only a callee containing a nested call walks anything extra, and the qualifier step is one nil check for a bare name.

**Verification at `6c5168256`.** `go test ../native/... -count=1 -p 1` green at `9ae1979db` and re-run at this head. `node scripts/format-go.cjs` leaves the diff empty. CI on this head is the gate.

samchon (commented):
### Ledger — Self-Review rounds 7 and 8, and the merge gate

**Round 7** over `f692fd63a..6c5168256` found one thing, and it was in the commit two rounds earlier had just written.

`9ae1979db` started walking the names a nested call is applied to. The doc comment directly above that walk still read:

> A nested call contributes its own callee but not its arguments … What this gives up is narrow and stated: an argument can still select an overload or instantiate a generic, so a pass-through like `pick(ns).is<T>(x)` resolves through a module the walk never names.

Every clause of that was false by then — `pick(ns)` is exactly the case the commit closed. A comment that describes behaviour the function no longer has is not cosmetic here: this walk's entire contract is *which files decide a callee*, and the next reader either hunts for a missing report that is already made, or re-opens a residual that is already closed. `4b6e3bdb4b` rewrites the bullet to state what the walk does, points at `metadataDependency_touchApplied` for why the applied half's answer is dropped, and leaves a residual list that is still true: an argument that is itself computed, and an overload selected by an argument's TYPE.

**Round 8** over `f692fd63a..4b6e3bdb4b` produced nothing that survived verification. What it covered beyond re-reading the three product files and all eighteen test files:

- **Every caller of the changed APIs.** `MetadataDependency_listen`, `MetadataDependency_release` and `MetadataDependency_IListener` are referenced from `transform.go` alone; `build.go`, `diagnostics.go`, `main.go` and `main_wasm.go` register no listener, so every touch stays a no-op in build and single-file modes exactly as the package doc claims.
- **The Wasm entry point.** `main_wasm.go` is behind `//go:build js && wasm`, so no ordinary test compiles it and `runTransformProject`'s new parameter could have broken the playground silently. `GOOS=js GOARCH=wasm go build` is clean.
- **Test shape.** Eighteen test files, one `Test*` function each, which is the repository rule for new cases.
- **Documentation.** No page under `website/src/content/docs/**` describes the transform envelope; the consumer contract is ttsc's, and the field's own contract lives in the `transformProjectOutput` doc comment.

**The differential audit, re-run uncontended**, over `tests/test-langchain`, `tests/test-vercel` and `tests/test-mcp` — 84 sources, each rewritten `string` -> `number` in turn:

| workspace | sources | consumer texts observed moving | over-declared |
| --- | ---: | ---: | ---: |
| `tests/test-langchain` | 24 | 1 | 0 |
| `tests/test-vercel` | 31 | 2 | 0 |
| `tests/test-mcp` | 29 | 7 | 0 |

**Merge gate.**

| gate | state |
| --- | --- |
| `origin/master` | `f692fd63a`, unchanged since the claim, so the prospective merge result is this head |
| closing keywords in `git log origin/master..HEAD` | one, `Close #2357` from `b7fdc5699`, whose fix survives at `HEAD` |
| `go test ../native/... -count=1 -p 1` | green at `6c5168256`, zero failures |
| `go test ./...` (public) | green at `4b6e3bdb4b`, zero failures |
| `node scripts/format-go.cjs` | leaves `git diff --numstat` empty |
| mutation matrix | every product branch has a row that fails exactly its owning test, except the two documented as unreachable (`out.Graph != nil`, `collector.End`) |
| package versions, tags, peer range | untouched |
| working tree | clean |

**What this cycle changed about the claim, in one line each.** The declaration went in at `b7fdc5699`; five Self-Review rounds then found five ways it was not yet true — an inferred declaration, an inverted default, a computed member key, a qualified type reference, and a generic pass-through — and one way it claimed more coverage than it had, in three predicate entries no input reaches. Every one of them was reproduced against a built binary before it was believed, and every fix costs a rebuild rather than a correct output.

**Residuals, unchanged and still stated in code**: an argument that is itself computed, an overload selected by an argument's type, a callee whose identity is a computed type, a new declaration merging into an already-consulted symbol (the protocol's own **Globals do not stay universal** trade), much weaker narrowing under `noembed`/`libReplacement`, `benchmark/` never measured, and no end-to-end narrowing test until the workspace moves past `ttsc` 0.19.2.

samchon (commented):
### Merge gate — final state at `4b6e3bdb4b`

All six workflows plus both Socket checks are green on this exact SHA, not on an ancestor:

| workflow | conclusion |
| --- | --- |
| `test` | success (13m47s) |
| `build` | success (13m21s) |
| `website` | success |
| `experiments` | success |
| `format` | success |
| `spell-check` | success |

`origin/master` is still `f692fd63a8`, unchanged since the claim commit, so the prospective merge result is this head and the CI above is the integration evidence for it. `git log origin/master..HEAD` carries exactly one closing line, `Close #2357` from `b7fdc5699`, and that fix survives at `HEAD`. The working tree is clean, and no commit in the range touches a package version, a tag, or the peer range.

Local, at this head: `go test ./...` and `go test ../native/... -p 1` both green with zero failures, `pnpm test:packages` green on Windows (exit 0), and `node scripts/format-go.cjs` leaves `git diff --numstat` empty.

**One correction to the round 7/8 ledger.** Its gate table says every product branch has a mutation row "except the two documented as unreachable (`out.Graph != nil`, `collector.End`)". There are three. The third is the unresolvable-reference fall-through in `metadataDependency_walkName`, which round 6 reported and explained in place — `[MISSING]` emits a dynamic-key validator rather than declaring a property, so the property enumeration never hands that shape to the walk. The count in the earlier table was wrong; the branch itself is documented in the code and was never claimed as covered.

**One last breadth cell**, run after that ledger was posted: the same barrel-behind-an-alias fixture through five public operation families rather than one. `typia.json.schemas<[Shape]>()`, `typia.protobuf.message<Shape>()`, `typia.random<Shape>()` and `typia.llm.application<App, { model: "chatgpt" }>()` each report `barrel.ts`, `id.ts` and `shape.ts` and each stay declared. Writing that fixture wrong the first time — `typia.llm.application<App, "chatgpt">()`, whose Config must be a literal object type — produced the diagnostic path instead, and the file it belongs to was withheld exactly as `TestProjectDependenciesCompleteDiagnosticTransform` requires. An accidental confirmation, but a real one.

**The narrowing, re-measured rather than inherited.** The handover quoted 148 / 257 / 348 host-owned paths per file against a reported 0 / 7 / 23 on `tests/test-typia-schema`. Recomputed from this head's own envelope -- `reach(graph.edges, F) union graph.globals`, excluding `F` -- it is **122 / 231 / 322** against **0 / 7 / 28**. The bound differs because the earlier figure counted the file and the config chain too; the reported maximum grew from 23 to 28, which is the callee identity and qualifier reporting this cycle added showing up where it should. The ratio the declaration is for is unchanged: a median of 7 declarations re-validated instead of 231 files.

Merging by squash with a curated body, per this repository's history.

samchon:
### Post-merge — one verification claim weakened, and why master is green anyway

Merged as `371eadf012`. One correction to the merge-gate comment above, because the claim is weaker than it reads.

**What happened.** A differential-audit script I had stopped earlier did not actually die. `TaskStop` reaped its wrapper shell but not the `bash audit6.sh` child, which kept rewriting one `tests/test-utils` source at a time for roughly forty minutes — through the merge. Its own restore step used `git checkout -- <file>`, which kept losing the index lock to this session's `git add`/`git commit` and failing silently, so each mutation stayed behind. That is also the real cause of the four phantom over-declarations reported and retracted in the round 6 comment: every later comparison in that run read a dirty baseline.

**What it costs the record.** The merge-gate line "`pnpm test:packages` green on Windows (exit 0)" is true as stated — the run did exit 0 — but the working tree was not pristine for its whole duration, so it is not the clean-tree evidence it implies. I am not going to pretend otherwise.

**What actually establishes master is green.** The `test` workflow ran `pnpm test` on a pristine Ubuntu checkout of the exact merged head `4b6e3bdb4b` and passed in 13m47s, alongside `build`, `website`, `experiments`, `format` and `spell-check`. That is the acceptance gate this repository defines, and it never saw the local tree. Re-running `tests/test-utils` alone on the clean merged tree afterwards: `Success`, exit 0, zero failures, 13.6 s, with the working tree confirmed clean before and after.

**What it did not cost.** Nothing the stray process touched reached the merge. `git show --name-only 371eadf012` is 21 files, every one under `packages/typia/native/`, and zero under `tests/`. The working tree now matches `origin/master` with an empty diff, and the process is confirmed dead by PID rather than by absence of symptoms.

The lesson is the same one this cycle kept relearning about oracles: a harness that reports "clean" while its own restore step is failing silently is worse than no harness. It is now recorded with the fix — restore by file copy and assert it with `cmp`, never with git in a repository the same session is committing to.

samchon:
### Residuals published as issues

Every item this cycle left open is now filed with a reproduction, a root cause at file:line, candidate fixes with measured cost, and a test plan — not as a list to re-derive.

| issue | kind | state of investigation |
| --- | --- | --- |
| #2360 | **correctness** | Three shapes reproduced against a binary built from `371eadf012`: a computed argument, a computed callee type, and an overload selected by a borrowed constant. In each, editing only the unreported file deletes the generated validator while the caller stays declared. Root cause is one asymmetry — `metadataDependency_touchDeclarations` reports declaring files but never asks `metadataDependency_bounded`, which `metadataDependency_touchVisited` does. Two fixes were built and measured: the obvious 3-line one drops declarations 816 → 401 and must not ship; gating it on the walk's existing `nested` flag closes all three shapes at 816 → 766. |
| #2361 | narrowing quality | Under `libReplacement`, a file with **no typia call at all** is withheld for calling a method the replaced library declares. Sound but unnecessary: `callExpressionTransformer_targetModule` only matches `/typia/lib/*.d.ts` and `/typia/src/*.ts`, so a default library can never make a call typia's. Fix is a channel split on the listener so the callee walk does not inherit the type walk's library withholding. |
| #2362 | test coverage | Nothing here proves narrowing *works* — the installed `ttsc` is 0.19.2 and the field is consumed from 0.19.3, so it is emitted and ignored. `tests/test-typia-bundler-cache` is already the right oracle; it needs the peer floor raised and one step that can only pass under narrowing. |

**Two residuals were retired instead of filed, because investigating them dissolved them.**

- *`benchmark/` has never been measured.* It has now: one project transform takes **6 seconds** — 709 files, 709 declared, 0 withheld, median reported list 2, max 14, no diagnostics. The handover's "exceeded a ten-minute budget" no longer holds, and there is nothing to fix.
- *An unreproduced tsgo printer crash.* Re-attempted at this head with `typia.llm.application<App, { model: "chatgpt" }>()` beside a `typeof`-referenced function: exit 0, no panic. It did not reproduce on the base commit either. Filing an issue with no witness would be noise.

The **Globals do not stay universal** trade — a file that declared nothing about a consulted symbol when the envelope was produced is not reported, so adding a declaration to it later does not invalidate — is not filed. The ttsc protocol accepts it explicitly, and it is what the narrowing is for.

Issue #1425, opened 2024-12-11
Standard Schema


ismailkarsli:
There is a new library called [Standard Schema](https://github.com/standard-schema/standard-schema). It aims to unify schema validators. Could it be used in Typia? IDK. I just wanted to let you know. Cheers 🍻

candidia:
Why close?

ismailkarsli:
> Why close?

I believe the author has no interest in this topic. But I can reopen it and they can decide.

sevenc-nanashi:
fyi: It's implemented in #1500 and #1581, and available since v9.2.0.

ismailkarsli:
> fyi: It's implemented in [#1500](https://github.com/samchon/typia/pull/1500) and [#1581](https://github.com/samchon/typia/pull/1581), and available since v9.2.0.

Thanks. I'm not following Typia since I kind of lost interest, but it's good to hear it's finally implemented.

PR #2363, opened 2026-08-20
fix(typia): declare the Standard Schema contract in-house instead of depending on it

### LINKED ISSUE #2359: createValidate factory causes ts(2883) when declaration export is enabled
### 📝 Summary

When using the example code from the docs on [`validate > Reusable Factories`](https://typia.io/docs/validators/validate/#reusable-factories), the creation of the validator results in a static TypeScript compiler error. The error is specific to the native TypeScript compiler (`tsc`) and is reported by VS Code, but does not prevent the code from being built with `ttsc` or executed via `ttsx`.

  - **NodeJS Version**: 24.15.0, 26.7.0
  - **TypeScript Version**: 7.0.2
  - **Typia Version**: 14.0.0
  - **Expected behavior**:  The example from the docs creates a reusable validation function.
  - **Actual behavior**: The variable assignment of the returned validator causes the following compiler error:

> The inferred type of 'validateMember' cannot be named without a reference to 'StandardSchemaV1' from '../node_modules/.pnpm/@standard-schema+spec@1.1.0/node_modules/@standard-schema/spec/dist/index.cjs'. This is likely not portable. A type annotation is necessary. ts(2883)

#### src/index.ts

```typescript
 
// Compiler error: ts(2883)
export const validateMember = typia.createValidate<IMember>();

// // By "explicitly" typing the variable, the error goes away
// export const validateMember: ReturnType<typeof typia.createValidate<IMember>> = typia.createValidate<IMember>();
 
interface IMember {
  email: string & tags.Format<"email">;
  age: number &
    tags.Type<"uint32"> &
    tags.ExclusiveMinimum<19> &
    tags.Maximum<100>;
}
```

#### package.json

```json
{
  "name": "typia-sandbox",
  "type": "module",
  "scripts": {
    "start": "ttsx src/index.ts",
    "build": "ttsc"
  },
  "dependencies": {
    "typia": "14.0.0"
  },
  "devDependencies": {
    "@types/node": "26.2.0",
    "ttsc": "0.27.0",
    "typescript": "7.0.2"
  }
}
```

#### tsconfig.json

```jsonc
{
  "compilerOptions": {
    "strict": true,
    "module": "nodenext",
    "target": "esnext",
    "lib": ["esnext"],
    "types": ["node"],
    "rootDir": "./src",
    "outDir": "./dist",

    // Removing this option also makes the error go away
    "declaration": true
  },
  "include": ["./src/**/*.ts"]
}
```

samchon:
Fixes #2359.

## Intent

`typia.createValidate<T>()` returns `((input: unknown) => IValidation<T>) & StandardSchemaV1<T, T>`. `IValidation` is re-exported from the `typia` entry, so a consumer's declaration emit names it `typia.IValidation`. `StandardSchemaV1` is not, so the emit has to reach `@standard-schema/spec` — a specifier the consumer cannot resolve once pnpm puts it beside typia instead of above the consumer. TypeScript then reports TS2742, or TS2883 since TypeScript 7.

The cause is not the intersection. Standard Schema is a purely structural compile-time contract, and dropping it from the declared type would delete the feature #1425 asked for and #1500 built: the runtime still attaches `~standard` through `typia/lib/internal/_createStandardSchema`, so the property would exist and be invisible, and every `up-fetch` / tRPC / Hono call site would need a cast. The cause is that a type a public signature exposes was not re-exported.

## Scope

Sweeping every public `typia.*` entry point under an isolated layout found three call sites, not one:

| Call site | Unnameable type | Status |
|---|---|---|
| `typia.createValidate<T>()` | `StandardSchemaV1` | reported in #2359 |
| `typia.createValidateEquals<T>()` | `StandardSchemaV1` | unreported |
| `typia.json.application<T>()` | `IJsonSchemaApplication` | unreported, regressed in v12.0.0 |

`json.application` broke when the monorepo split moved `IJsonSchemaApplication` into `@typia/interface`; v11.0.3 is clean because the type still lived inside typia. `createValidate` has been broken since Standard Schema support landed. Both reproduce on v12.2.0, v13.3.0, and v14.0.0, under TypeScript 5.9.3 (TS2742) and 7.0.2 (TS2883), and only under an isolated `node_modules` layout — npm's hoisting resolves the bare specifiers and hides it.

Two re-exports close all three. `Atomic` is the one remaining `@typia/interface` name typia does not re-export; it appears only in generic-constraint position, so no inferred type reaches it today.

## Why nothing caught it

Three independent reasons, each closed by `experiments/declaration`:

- the `tests/*` workspaces resolve typia from its TypeScript source, where declaration emit reuses the import specifier typia's own file already carries;
- `smoke`, `esm`, and `mcp` install the tarballs with npm, whose hoisting makes the bare specifiers resolve;
- no project in the repository compiled a typia consumer with `declaration` on.

The same class was patched once before, one name at a time, in #1770.

## The guard

`experiments/declaration` installs the packed tarballs with pnpm, keeps `@typia/interface` and `@typia/utils` transitive through `overrides` so the names under test stay unreachable, and compiles 55 unannotated exported bindings across every public entry point with `declaration: true`.

`ttsc`'s exit code cannot be the oracle here: it prints TS2742 / TS2883 and still exits 0, writing `any` in place of the unnameable type. `tsc` on the same project exits 1. So `test` reads the emitted declarations — asserting one entry per export, no widening to `any`, no module specifier other than `typia`, and the three witnesses named — and separately runs `tsc`, which fails hard and names the offending type.

Reverting either re-export fails it with all three witnesses reported.

## Deferred

`@typia/utils` has the same defect and is **not** fixed here. `LlmSchemaConverter.getConfig()` in a pnpm consumer with `declaration: true` reports TS2883 on `ILlmSchema.IConfig`, and about 20 `@typia/interface` names leak into that package's public `.d.ts`. Closing it means giving `@typia/utils` its own re-export surface, which is a public API decision for that package rather than a side effect of this fix.

`ttsc` exiting 0 on a declaration-emit error, and emitting `any` in place of the type, is an upstream issue in `samchon/ttsc` — it is what let this ship silently.

## Verification

- `pnpm build`
- `pnpm test`
- `pnpm run package:tgz`, then `pnpm install && pnpm run build && pnpm test` in `experiments/declaration`, before and after reverting the re-exports


samchon (commented):
## Design change: the borrowed name is gone, not re-addressed

The body above describes the first mechanism — re-exporting `StandardSchemaV1` from `@standard-schema/spec` so consumer declaration emit could name it as `typia.StandardSchemaV1`. That worked, and `1db0d90d4d` replaces it.

Re-exporting fixed the diagnostic but left the cause standing: another package's type sat in typia's public signature, and typia carried a runtime dependency purely so declarations could name it. `@typia/interface` now declares `StandardSchemaV1` itself and `@standard-schema/spec` leaves typia's dependencies entirely.

This is the shape the ecosystem already settled on, and the shape this repository already uses:

| package | implements Standard Schema | runtime dependencies |
|---|---|---|
| `zod` | yes, vendored at `src/v4/core/standard-schema.ts` | none |
| `valibot` | yes, vendored and renamed to `StandardProps` | none |
| `@typia/interface` | — | none; `OpenApi`, `OpenApiV3`, `SwaggerV2` are all declared in-house |

**Correction to something I said earlier in the session:** I listed `@sinclair/typebox` as a third example. It has no runtime dependencies, but it does not implement Standard Schema at all — no `~standard` anywhere in the published tree. It is not evidence for this and does not appear in the source comment.

## A premise worth recording as disproved

The type was suspected to come from the transformer's generated function rather than `module.ts`. It does not. The transform emits JavaScript with no types in it:

```js
export const validateMember = (() => {
    ...
    return _createStandardSchema_1._createStandardSchema(input => { ... });
})();
```

Declaration emit never reads that file. The decisive check is a project with **no `ttsc` installed and no transform running at all**:

```
$ ls node_modules | grep -i ttsc
(nothing)
$ ./node_modules/.bin/tsc
src/index.ts(3,14): error TS2883: The inferred type of 'validateMember' cannot be named
without a reference to 'StandardSchemaV1' from '.pnpm/@standard-schema+spec@1.1.0/...'
exit=1
```

Plain `tsc` produces it, so the source is the declared overload at `packages/typia/src/module.ts:746`.

## An alternative that was measured and not taken

Declaring the intersection as an **anonymous** structural type inside `module.ts` also removes the diagnostic, needs no export of any kind, and keeps interoperability — verified by compiling a `<T extends StandardSchemaV1>` boundary against it. It was rejected on readability: `module.ts` loses the name, and every consumer `.d.ts` grows a ~30-line inline blob per binding. Recorded here so it is not rediscovered as new.

## Self-Review, round 1

Five findings, each applied and each proved by breaking it:

1. **The sweep could not see `Atomic`.** `src/index.ts` only reaches a borrowed type while a public signature puts it in an inferred position, and `Atomic` is only ever a generic constraint — so nothing in this PR could have noticed typia never re-exported it. Added the **census pass**: it reads typia's own `lib/*.d.ts`, collects every type name they import from another package, and imports all of them from `"typia"`. It found `Atomic` immediately. Removing the re-export makes it report `- Atomic (borrowed from @typia/interface)`.
2. **Nothing asserted the isolation the whole experiment rests on.** If a typia dependency becomes resolvable from the project, every borrowed name is nameable again and the suite passes while measuring nothing. Proved by copying `@standard-schema/spec` into the project's `node_modules`: the run now fails with `layout: these packages resolve from this project ... the run proves nothing`.
3. **Nothing asserted the packages under test were the freshly packed tarballs.** If `pnpm-workspace.yaml`'s `overrides` stop applying, the transitive `@typia` packages come from the registry. Proved by removing that file and reinstalling: the run now names `@typia/interface -> .pnpm/@typia+interface@14.0.0/...` and fails.
4. Declaration reading did not normalize CRLF.
5. The census probe was written to the system temp folder, where `typia` does not resolve, and naming a file on the command line beside a `tsconfig.json` is TS5112 without `--ignoreConfig`.

## Consequence surface the dependency removal carried

- `packages/typia/native/cmd/ttsc-typia/create_assert_error_factory_arity_test.go` resolved `@standard-schema/spec` through typia's own `node_modules`. It broke on removal (`TS2307`) and now imports the type from `typia`, which is the stronger assertion — it pins that a consumer can reach the name at all.
- `website/build/typia-pack.js` mounted the package on the playground MemFS solely because typia's sources imported it. No mounted source (`packages/{typia,interface,utils}/src`) imports it now, so the mount and its `resolvePackageRoot` helper are dead and removed. `node website/build/typia-pack.js` still packs 316 files.
- `@typia/langchain` keeps its own dependency: it uses `StandardJSONSchemaV1` inside `LangChainParameterConverter`, never in a public signature.

## Still deferred

- **`@typia/utils` has the same defect and is not fixed here.** `LlmSchemaConverter.getConfig()` in a pnpm consumer with `declaration: true` reports TS2883 on `ILlmSchema.IConfig`, and roughly 20 `@typia/interface` names leak into that package's public `.d.ts`. Having `typia` installed does not rescue it — typia's files are not in that program, so TypeScript never sees the alternative container. Closing it means giving `@typia/utils` its own re-export surface, which is a public API decision for that package.
- **`ttsc` exits 0 on a declaration-emit error** and writes `any` in place of the unnameable type, while `tsc` on the same project exits 1. That is why this shipped silently, and why `test` here reads the emitted declarations instead of trusting a build's exit code. It belongs upstream in `samchon/ttsc`.

## Verification

- `pnpm build` and `pnpm test` — green on the final state
- `pnpm run package:tgz`, then `pnpm install && pnpm run build && pnpm test` in `experiments/declaration` — `layout: 2 dependencies hidden ... 3 packages under test packed from this tree`, `census: 23 borrowed type names, all reachable through typia`, `tsc`/`ttsc`: `55 declarations named through typia, 3 regression witnesses intact`
- Reverting the three re-exports fails it on every pass, naming `Atomic`, `IJsonSchemaApplication`, and `StandardSchemaV1`
- Drift guard proved: widening `version: 1` to `1 | 2` in the vendored declaration fails `tests/test-interface` with five assertions
- `typos . --config ./typos.toml` — clean, run with the same 1.49.0 binary CI uses

samchon (commented):
## Self-Review, rounds 2 and 3

Round 1 is in the review above. Two more rounds ran against the state it produced, each restarted from the whole diff.

**Round 2 — the provenance check had a hole in the one package that matters most.** It reported a stale `@typia/*` package but filtered the list to that prefix, so a `typia` resolving from anywhere other than `experiments/tarballs` passed silently — the package the whole project exists to measure. It now faults on any entry that did not come from a packed tarball, including one typia cannot resolve at all, and prints what it found. Proved by removing `pnpm-workspace.yaml` and reinstalling:

```
layout: these came from somewhere other than the tarballs just packed,
so the run says nothing about this working tree:
  - @typia/interface -> ...\.pnpm\@typia+interface@14.0.0\...
  - @typia/utils -> ...\.pnpm\@typia+utils@14.0.0\...
```

**Round 3 — `@typia/interface`'s README.** The package now declares the Standard Schema contract, and it is what `createValidate` hands to every library that accepts one, so it belongs in the key-interfaces table.

**Round 4 — clean.** Verified the type-only module leaves no runtime surface: `packages/interface/lib/schema/StandardSchemaV1.js` is an empty CJS module, `packages/typia/lib/internal/_createStandardSchema.{js,mjs}` carry no `@typia/interface` import, and neither `typia/lib/index.js` nor `index.mjs` mentions any of the three re-exported names. `@typia/interface` is CJS-only, so nothing about its ESM condition changed — it has none.

## A consumer-visible consequence of dropping the dependency

Under npm's hoisting, `@standard-schema/spec` used to land at the top of a typia consumer's `node_modules` whether or not they declared it, so `import { StandardSchemaV1 } from "@standard-schema/spec"` worked by accident. It will not after this. Anyone who was relying on that phantom dependency either declares it themselves or, better, takes the type from `typia`, which now exports it. Worth a line in the release notes.

## CI

Green on `008c4dc9e1`, 8/8. The lanes that matter here:

- `experiments`: Test Smoke, Test ESM, Test MCP, Test Declaration all passed. ESM is the guard for the `.mjs` build, which is the one thing local verification could not cover.
- Test Declaration on Linux reports what it checked rather than an exit code:

```
layout: 2 dependencies hidden from the consumer, 3 packages under test packed from this tree
census: 23 borrowed type names, all reachable through typia
tsc: 55 declarations named through typia, 3 regression witnesses intact
ttsc: 55 declarations named through typia, 3 regression witnesses intact
```

Rounds 2 and 3 each produced a change, so each was followed by another full round; round 4 found nothing that survived verification.

PR #2364, opened 2026-08-20
test(interface): pin the Standard Schema copy by identity, not assignability

samchon:
test(interface): pin the Standard Schema copy by identity, not assignability

The conformance case asserted the vendored declaration and
`@standard-schema/spec` were assignable in both directions. That is too weak to
pin a copy, and measuring it says so: adding an optional member to this side, or
dropping a `readonly` modifier, leaves both directions assignable and produced
zero errors. Only a required member turning optional was caught.

Identity catches all of them and still passes on the declaration as written, so
the copy is not merely interchangeable with the specification — it is the same
type. Measured against the same four mutations:

| mutation                       | assignability | identity |
| ------------------------------ | ------------- | -------- |
| optional member added          | 0 errors      | 8 errors |
| `readonly` dropped             | 0 errors      | 5 errors |
| required member made optional  | 5 errors      | 5 errors |
| `version: 1` widened to number | not measured  | 5 errors |

`Mutual` is gone; every pair now goes through the `IsEqual` helper the file
already carried for the inference assertions.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HuBaCjXy4yN7Fzp5greiA9

### DIFF
--- a/tests/test-interface/src/schema/test_standard_schema_v1_conformance.ts
+++ b/tests/test-interface/src/schema/test_standard_schema_v1_conformance.ts
@@ -2,7 +2,7 @@ import type { StandardSchemaV1 as Spec } from "@standard-schema/spec";
 import { StandardSchemaV1 } from "@typia/interface";
 
 /**
- * Verifies `StandardSchemaV1` still matches the specification it was copied
+ * Verifies `StandardSchemaV1` is identical to the specification it was copied
  * from.
  *
  * `@typia/interface` declares the Standard Schema contract in-house so that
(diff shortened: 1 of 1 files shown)


samchon (commented):
## Self-Review

One file, one round, clean. Two things worth recording rather than rediscovering:

**This guard is noisier by design now.** If `@standard-schema/spec` ships a release that merely adds an optional member, identity fails and this suite breaks even though interoperability is untouched. That is the intended behavior for a vendored specification — the copy has diverged and someone has to sync it — so the fix is to update `packages/interface/src/schema/StandardSchemaV1.ts`, never to weaken the assertion back to assignability.

**`Mutual` was not merely redundant, it was misleading.** It read as "the two types match" while passing two of the four ways a copy actually drifts. Every pair it guarded passes identity, so nothing was lost by deleting it.

The measurements in the commit message were taken by mutating `packages/interface/src/schema/StandardSchemaV1.ts` one edit at a time and running `pnpm --filter ./tests/test-interface start`, restoring between each.

Issue #2362, opened 2026-08-20
test: no oracle proves dependenciesComplete narrowing works until the workspace ttsc floor reaches 0.19.3


samchon:
## Summary

`dependenciesComplete` shipped in #2358, but nothing in this repository exercises what it is **for**. The installed `ttsc` is 0.19.2 and the field is consumed from 0.19.3, so in this workspace the envelope emits it, the host passes it through as an unknown member, and it is ignored. Every suite that is green today proves *no regression*; not one proves *narrowing works*.

The oracle already exists and needs no new design — it just needs the peer floor raised.

## Current state

| fact | value |
| --- | --- |
| installed `ttsc` | 0.19.2 (`node_modules/.pnpm/ttsc@0.19.2`) |
| `typia` peer range | `"ttsc": ">=0.19.2"` (`packages/typia/package.json`) |
| version that consumes the field | 0.19.3 |
| consumer that narrows on it | `@ttsc/unplugin` `packages/unplugin/src/core/transform.ts` — `declaresCompleteDependencies`, `selectWatchInputs`, `deriveWatchInputs` |
| protocol reference | ttsc docs, *Dependency completeness* |

The consumer contract, for reference:

| file `F` | watch inputs derived |
| --- | --- |
| not listed | `dependencies[F] ∪ reach(graph.edges, F) ∪ graph.globals ∪ graph.configs ∪ graph.candidates` |
| listed | `dependencies[F] ∪ graph.configs ∪ graph.candidates` |

So a listed file whose entry omits a real input serves a stale generated validator, and **no local suite can currently observe that**, because nothing narrows.

## Why the oracle is already written

`tests/test-typia-bundler-cache` (`src/test_webpack_filesystem_cache_invalidation.ts`, 469 lines) drives a real webpack filesystem cache across a sequence of mutations, each of which must invalidate:

| step | fixture constants | what it mutates |
| --- | --- | --- |
| consulted type | `fixtureTypeV1` / `fixtureTypeV2` | a type the validator reads |
| re-pointed barrel | `fixtureBarrelV1` / `fixtureBarrelV2` | `export { Nested } from "./nested"` → `"./alt"` |
| index-signature target | `fixtureCellV1` / `fixtureCellV2` | `export type Cell = string` → `number` |
| ambient declaration | `fixtureAmbientV1` / `fixtureAmbientV2` | a global-scope declaration |

[...]

PR #2368, opened 2026-08-20
test(bundler-cache): pin that a listed file stops watching the reference closure

samchon:
test(bundler-cache): pin that a listed file stops watching the reference closure

`dependenciesComplete` narrows a listed file's watch inputs from `dependencies[F]
union reach(edges, F) union globals union configs` down to `dependencies[F] union
configs`. Every step this suite had mutates a file the reported list already
names, so all of them pass under either derivation — the suite proved no
regression and never proved narrowing works (#2362).

`helper.ts` is the input that separates them. `index.ts` imports it for its
VALUE, so the reference graph carries the edge, and no typia call consults it,
so the reported list leaves it out. Dumped for the fixture, it is the only file
in `reach(edges, index.ts)` and outside `dependencies[index.ts]`.

Two things had to be right, and both were found by measuring:

- **A type-only import does not work.** `graph.edges` records emitted module
  edges, and TypeScript elides an import used only by an unreferenced
  interface, so such a file lands in neither set. Dumping
  `deriveWatchInputs` for the fixture is what settled it.
- **`codeGenerated` is the wrong observable.** webpack skips code generation
  when the module hash is unchanged, and a watched input changing outside the
  entry's own source leaves that hash alone — so the entry reports
  `codeGenerated: false` even when the loader re-ran. `built` reports the
  rebuild. The step passed under BOTH derivations until this was corrected,
  which is exactly the false green it exists to prevent.

`concatenateModules` goes off so the entry is its own stats row rather than
`./src/index.ts + N modules`, which cannot separate "the entry re-ran the
transform" from "a sibling module rebuilt".

The step-2 assertion is the positive twin: a consulted file MUST rebuild the
entry, or step 6's negative would also pass on a build that rebuilds nothing.

Mutation: patching `declaresCompleteDependencies` to `return false` in the
installed `@ttsc/unplugin` fails step 6 with its own message, and the dump
confirms the patch reaches the derivation (`complete: false`).

### DIFF
--- a/tests/test-typia-bundler-cache/src/test_webpack_filesystem_cache_invalidation.ts
+++ b/tests/test-typia-bundler-cache/src/test_webpack_filesystem_cache_invalidation.ts
@@ -37,12 +37,46 @@ const run = (cwd: string, args: string[]): ICommandResult => {
 const sleep = (ms: number): Promise<void> =>
   new Promise((resolve) => setTimeout(resolve, ms));
 
-const build = (project: string, stage: string): void => {
+interface IModuleRow {
+  name: string;
+  built: boolean;
+  codeGenerated: boolean;
+}
+
+const build = (project: string, stage: string): IModuleRow[] => {
   const result: ICommandResult = run(project, [
     path.join(project, "build.cjs"),
   ]);
   if (result.status !== 0)
     throw new Error(`webpack build failed ${stage}:\n${result.output}`);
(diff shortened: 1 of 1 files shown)


samchon:
Closing unmerged: the step this adds passes on Windows and fails on Linux CI, so the observable is not portable. Detail recorded on #2362.

PR #2374, opened 2026-08-22
fix(typia): improve missing-transform diagnostics

### LINKED ISSUE #2373: Incorrect message when unable to create JSON schema
for example for `string & Date;`
```typescript
import typia from 'typia';

type ProblematicType = {
  test: Date & string;
};

const schema = typia.json.schema<ProblematicType>();
```

==> TS compile (no erorrs), but:

```
Error: Error on typia.json.schema(): no transform has been configured.

Read and follow https://typia.io/docs/setup please.

If you've already completed the setup, it means there's a bug in your code. Run `tsc` command so that check what is wrong with your code.
```

Perhaps it's worth indicating in the message the reason why it wasn't possible to transpile such code into json.schema

samchon:
## Intent

Close #2373 by sending an untransformed typia call back through the compiler that can actually load typia's native plugin and report the underlying transform diagnostic.

The issue's `Date & string` example is already rejected by current `ttsc` with `TS(typia.json.schema)`, the `ProblematicType.test` path, and `nonsensible intersection`. The runtime fallback has none of that compiler state; its actionable responsibility is to name the correct compiler command.

## Scope

- Replace the shared fallback's obsolete `tsc` instruction with an explicit `ttsc` (not `tsc`) instruction.
- Add public runtime coverage for the untransformed-call guidance.
- Add command-level coverage proving the issue's exact type reaches the detailed `ttsc` diagnostic.

## Deferred / unchanged

- No JSON Schema or intersection semantics change: the current transformer already rejects the reported type correctly.
- No package version change.

## Verification at draft open

- `pnpm format`
- Manual `ttsx` probe of the issue source: rejected with `TS(typia.json.schema)`, `ProblematicType.test`, and `nonsensible intersection`.
- Direct execution of the shared fallback source: reports `Run \`ttsc\` (not \`tsc\`)`.

Automated regression tests and the broader build/test gates will follow on this draft.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/json_schema_nonsensible_intersection_diagnostic_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/json_schema_nonsensible_intersection_diagnostic_test.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+  "os"
+  "path/filepath"
+  "strings"
+  "testing"
+)
+
+// TestJSONSchemaNonsensibleIntersectionDiagnostic verifies that ttsc reports
+// the transform-time cause behind issue #2373.
+//
+// The runtime fallback cannot inspect a TypeScript type after transformation
+// has been skipped. This regression therefore compiles the reporter's exact
+// `Date & string` property through the real typia package and preserves the
+// actionable native diagnostic at its actual compiler boundary.
+//
+//  1. Compile a `typia.json.schema` call for the nonsensible intersection.
+//  2. Require the API diagnostic, property path, and intersection reason.
+func TestJSONSchemaNonsensibleIntersectionDiagnostic(t *testing.T) {
+  project := jsonSchemaNonsensibleIntersectionDiagnosticProject(t)
+  stdout, stderr, code := ttscTypiaTestCapture(func() int {
+    return runTransform([]string{
+      "--cwd", project,
+      "--tsconfig", "tsconfig.json",
+      "--file", "src/main.ts",
+      "--output", "js",
+    })
+  })
+  if code != 3 {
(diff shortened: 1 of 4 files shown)


samchon (commented):
Self-Review completed against commit 3cdf9f46c12320b9e1f0b6444ae442d3fbc8b8f9.

Round 1 reviewed the complete four-file change, all 171 NoTransformConfigurationError call sites across ten public API families, the runtime/compiler diagnostic boundary, source history, packaging, platform behavior, and documentation compatibility. The only finding was that the original PR title described only the tsc-to-ttsc correction and no longer covered the expanded recovery guidance and regressions. I changed the title to fix(typia): improve missing-transform diagnostics, then restarted the review from the full diff.

Round 2 re-read all four changed files and traced both failure paths independently. The runtime fallback remains transform-agnostic and now gives actionable ttsc, ttsx, and bundler recovery steps. The native ttsc regression preserves the real unsupported Date & string behavior and verifies the TS(typia.json.schema), ProblematicType.test, and nonsensible intersection diagnostics. I also checked fixture isolation and cleanup on Windows and POSIX, generated CommonJS and ESM output, the packed npm tarball, public package contents, the existing positive and negative intersection backstop, setup documentation, and legacy documentation boundaries. No further findings remain.

Verification passed:
- focused runtime regression
- focused native Go diagnostic regression
- feature identity suite
- pnpm format
- pnpm install
- pnpm build
- pnpm test, including all TypeScript workspaces and both Go suites
- pnpm package:tgz, with the emitted CJS and ESM helpers inspected inside typia.tgz
- website pnpm build, including 74 static pages, Pagefind, and sitemap generation
- all GitHub checks on this exact SHA, including Ubuntu test, build, experiments, format, website, spell-check, and Socket checks

Self-Review result: clean after the title correction; no code findings remain. The PR intentionally stays in draft state.

Issue #2377, opened 2026-08-25
[bug] typia.reflect.literals<never>()


salisbury-espinosa:
### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoFgYwgDsD4YBTXnAC8CZGgB0UfmQA2-djAmzgAqKlkEAPF34A3flAB8ACgCUrDtwIR5yiHhMDeFoA

### 💻 Code occuring the bug
```typescript
import typia, { tags } from "typia";

const test = typia.reflect.literals<never>();

console.log(test);
```

==> failed compile js

expectation:

```javascript
const test = [];
console.log(test);
export {};

```

PR #2378, opened 2026-08-25
fix(typia): emit an empty array for reflect.literals over an empty union

samchon:
Solo issue-campaign cycle claim. This pull request owns the complete accepted cycle, which is one issue: #2377.

**Scope.** `typia.reflect.literals<never>()` fails to compile with `no constant literal type found.` instead of emitting `[]`.

**Status.** Empty claim; implementation and verification pending. The body is the historical intent statement — later CI fixes and Self-Review rounds land as formal `COMMENT` reviews on this thread.

**Release freeze.** Campaign branches never choose a release number or publish a package. Version bumps stay out of this pull request.

### DIFF
--- a/packages/typia/native/cmd/ttsc-typia/reflect_literals_empty_union_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/reflect_literals_empty_union_transform_test.go
@@ -20,9 +20,10 @@ import (
 // vacuously literal-only and `literals<never>(): never[]` has exactly one
 // inhabitant, so both must transform and run.
 //
-//  1. Transform a module whose `reflect.literals` arguments span every spelling
-//     of the empty union a caller reaches -- the keyword, an alias, an
-//     exhaustive `Exclude`, an empty `Extract` -- plus `null`, a fully mixed
+//  1. Transform a module whose `reflect.literals` arguments span every
+//     uninhabited spelling a caller reaches -- the keyword, an alias, an
+//     exhaustive `Exclude`, an empty `Extract`, a TypeScript-collapsed
+//     intersection, and one typia prunes itself -- plus `null`, a fully mixed
 //     literal union, and the plain `boolean` atomic.
 //  2. Execute the emitted CommonJS.
 //  3. Assert `never` yields `[]`, `null` yields `[null]`, and the arguments that
@@ -102,6 +103,13 @@ export const emptyAlias = typia.reflect.literals<Empty>();
 export const emptyExclude = typia.reflect.literals<Exclude<Color, Color>>();
 export const emptyExtract = typia.reflect.literals<Extract<Color, "cyan">>();
 
+// Uninhabited without the keyword: TypeScript collapses one to never itself,
+// while the other stays an intersection whose every distributed member typia
+// prunes away. Both are empty on both sides of the comparison, so the rule the
+// keyword exercises has to cover them too.
+export const emptyCollapsed = typia.reflect.literals<string & number>();
+export const emptyPruned = typia.reflect.literals<(string | number) & { data: number }>();
+
 // A union whose only member is the nullable flag, which carries no bucket.
 export const onlyNull = typia.reflect.literals<null>();
 
@@ -141,6 +149,8 @@ check("never", mod.empty, []);
 check("never alias", mod.emptyAlias, []);
 check("exhaustive Exclude", mod.emptyExclude, []);
 check("empty Extract", mod.emptyExtract, []);
+check("collapsed intersection", mod.emptyCollapsed, []);
+check("pruned intersection", mod.emptyPruned, []);
 check("null", mod.onlyNull, [null]);
 check("mixed", mod.mixed, ["a", 1, true, null]);
 check("booleanAtomic", mod.booleanAtomic, [true, false]);
--- a/packages/typia/native/cmd/ttsc-typia/reflect_literals_empty_union_transform_test.go
+++ b/packages/typia/native/cmd/ttsc-typia/reflect_literals_empty_union_transform_test.go
@@ -20,11 +20,14 @@ import (
 // vacuously literal-only and `literals<never>(): never[]` has exactly one
 // inhabitant, so both must transform and run.
 //
-//  1. Transform a module whose `reflect.literals` arguments span `never`,
-//     `null`, a fully mixed literal union, and the plain `boolean` atomic.
+//  1. Transform a module whose `reflect.literals` arguments span every spelling
+//     of the empty union a caller reaches -- the keyword, an alias, an
+//     exhaustive `Exclude`, an empty `Extract` -- plus `null`, a fully mixed
+//     literal union, and the plain `boolean` atomic.
 //  2. Execute the emitted CommonJS.
 //  3. Assert `never` yields `[]`, `null` yields `[null]`, and the arguments that
-//     already worked keep their exact members and order.
+//     already worked keep their exact members and order, including the sort the
+//     guide documents.
 func TestReflectLiteralsEmptyUnionTransform(t *testing.T) {
   node, err := exec.LookPath("node")
   if err != nil {
@@ -87,8 +90,17 @@ func TestReflectLiteralsEmptyUnionTransform(t *testing.T) {
 
 const reflectLiteralsEmptyUnionSource = `import typia from "typia";
 
-// The empty union: no member on either side of the admission comparison.
+type Color = "red" | "green" | "blue";
+type Empty = never;
+
+// The empty union: no member on either side of the admission comparison. Every
+// spelling a caller reaches it through has to transform, not only the keyword:
+// an alias hides it behind a name, and a conditional or an exhaustive Exclude
+// is how a generic caller produces one without writing "never" at all.
 export const empty = typia.reflect.literals<never>();
+export const emptyAlias = typia.reflect.literals<Empty>();
+export const emptyExclude = typia.reflect.literals<Exclude<Color, Color>>();
+export const emptyExtract = typia.reflect.literals<Extract<Color, "cyan">>();
 
 // A union whose only member is the nullable flag, which carries no bucket.
 export const onlyNull = typia.reflect.literals<null>();
@@ -96,9 +108,13 @@ export const onlyNull = typia.reflect.literals<null>();
 // Controls that already transformed: the flag must join the count without
 // displacing a constant, a boolean atomic, or their order.
 export const mixed = typia.reflect.literals<"a" | 1 | true | null>();
-export const withoutNull = typia.reflect.literals<"a" | "b" | 1 | 2>();
 export const booleanAtomic = typia.reflect.literals<boolean>();
 export const booleanAtomicNull = typia.reflect.literals<boolean | null>();
+
+// Declaration order deliberately disagrees with the emitted order: values of
+// one primitive type are sorted (iterate_metadata_sort), which is the behavior
+// the reflect.literals guide states.
+export const sorted = typia.reflect.literals<"b" | "a" | 10 | 2>();
 `
 
 const reflectLiteralsEmptyUnionRunner = `const mod = require("./main.cjs");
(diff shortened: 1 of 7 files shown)


samchon (commented):
## Self-Review round 1 — 2 findings, both applied

Surface: `master...af857e4b8`, the complete base-to-head diff.

**1. `website/src/content/docs/misc.mdx` stated a false emit order.** The `reflect.literals` section claimed `"red" | "green" | "blue"` yields `["red", "green", "blue"]`. `Iterate_metadata_sort` sorts the values of each constant type (`iterate_metadata_sort.go`, the `constant.Type == "string"` branch), so the real emit is `["blue", "green", "red"]`. Verified against a live transform. A reader populating a select box from the declaration order got the wrong list. Corrected, and the claim now carries an executed oracle: the `sorted` fixture case declares `"b" | "a" | 10 | 2` and asserts `["a", "b", 2, 10]`, so the declaration order and the emitted order deliberately disagree.

**2. The empty-union fixture only covered the `never` keyword.** A caller reaches the empty union through an alias or through `Exclude`/`Extract` filtering everything away far more often than by writing `never` at the call site, and those are exactly the spellings whose rejection broke a whole build. Added `Empty` (alias), `Exclude<Color, Color>`, and `Extract<Color, "cyan">`.

### Verified clean in this round

- **Admission predicate over-acceptance.** Walked every field `MetadataSchema.Size()` counts (`Any`, `Escaped`, `Rest`, `Templates`, `Atomics`, `Constants`, `Arrays`, `Tuples`, `Natives`, `Maps`, `Sets`, `Objects`, `Functions`, `Aliases`) against what the emitter below renders. `length` is bounded above by `size` in every case, so `size == length` holds exactly when every member is renderable. `Escaped` (a `toJSON` carrier) and a tag-branded atomic both raise `size` without raising `length`, and stay rejected.
- **Diagnostic identity.** `NO` and `ONLY` remain distinguishable rather than collapsing onto one message: `string` and `any` report `no constant literal type found`, while `string | null`, `"a" | number`, and `` `prefix${number}` | "a" `` report `only constant literal types are allowed`. Asserted per case.
- **Blast radius.** The `Validate` callback is local to `ReflectLiteralsProgrammer`; no other programmer reads these messages or this predicate.
- **Other `never` surfaces.** Swept `is`, `reflect.name`, `reflect.schema`, `reflect.schemas`, and `json.stringify` over `never` — all already correct. `json.schema<never>()` stays rejected by the deliberate `BlockNever` guard in `JsonSchemasProgrammer`; that is intended and untouched.

### Local verification

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestReflectLiterals` | pass |
| same, with the product fix reverted and the tests retained | both tests fail on exactly the new assertions; adjacent controls still pass |
| `pnpm --filter ./tests/test-typia-schema start --include reflect_literals` | pass |
| `pnpm --filter ./tests/test-error start` | pass — 79 files, 875 diagnostics |
| `pnpm format` | clean, staged |
| `pnpm test` | running |

samchon (commented):
## Self-Review round 2 — 3 findings, all applied

Fresh full round over `master...af857e4b8`, plus the read-only commit early-warning pass on `a3dfe62a1`.

**1. The rule generalizes further than the tests did.** What the fix installs is not "`never` is special" but "an argument carrying no member is uninhabited, and the literal set of an uninhabited type is empty". Two reachable spellings prove that and were untested: `string & number`, which TypeScript collapses to `never` itself, and `(string | number) & { data: number }`, which stays an intersection while typia prunes every distributed member away (the shape `TestIntersectionAllNeverUnionTransform` already treats as `never`). Both emit `[]`. Added as fixture cases.

**2. `undefined` and `void` reach the same empty metadata — traced, and unreachable.** `Iterate_metadata_coalesce` collapses `undefined`, `void`, and `never` into one identical schema (`Required = false`, every bucket empty), so metadata cannot tell them apart and the new rule necessarily admits all three. Neither is reachable in a supported configuration: neither satisfies `T extends Atomic.Type | null` under `strictNullChecks`, and `transform/transform.go:19` refuses the transform outright when `strictNullChecks` is off. Recorded as an explicit boundary in the predicate comment rather than left for the next reader to rediscover.

**3. The predicate comment overstated what weighing `null` buys.** Adding 1 to both `length` and `size` cannot change `size == length`, so `literals<null>()` is admitted by the empty-set rule, not by the `null` count. What the count actually changes is the diagnostic a *rejected* argument earns: `string | null` moves from `no constant literal type found` to `only constant literal types are allowed`. Comment corrected to state that, since a comment claiming a causality the code does not have is worse than no comment.

**4. `test_reflect_literals_never` compared a sorted copy while its sibling pinned exact order.** Round 1 documented the sort, so the assertion now has an oracle: it pins `["blue", "green"]` directly. Also corrected the `test-error` fixture comment, which implied that suite asserts the `NO` vs `ONLY` distinction — it only asserts that a fixture is named by a diagnostic. The distinction is asserted by `TestReflectLiteralsNonLiteralRejectionTransform`.

### Verified clean in this round

- **Every field `Size()` counts, re-walked against the emitter.** `length` is bounded above by `size` in all fourteen buckets, so `size == length` holds exactly when every member is renderable. `Iterate_metadata_function` always appends a `Functions` entry, so a function type cannot slip through as empty, and `Emend_metadata_atomics` folds `true | false` into the boolean atomic before `Validate` runs.
- **Nested `Validate` calls.** `validateMeta` recurses through aliases, arrays, tuples, objects, functions, sets, maps, and escaped returns. A node passes the new rule only when it has no children to recurse into, so an admitted top can never hide a rejected descendant. The only change is one fewer duplicate diagnostic on shapes like `never[]`, which the type bound rejects anyway.
- **Equivalent spellings of the newly admitted arguments.** `never`, `type Empty = never`, `Exclude<Color, Color>`, `Extract<Color, "cyan">`, `string & number`, the pruned intersection, `null`, `type Nullable = null`, and `(typeof arr)[number]` all verified against a live transform.
- **Version freeze.** `packages/typia/native` changed with no version bump, which `Change Integrity` requires of an issue-implementation pull request. The release is a separate maintainer step.

### Local verification at this head

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestReflectLiterals` | pass |
| same, product fix reverted, tests retained | both fail on exactly the new assertions; adjacent controls pass |
| `pnpm --filter ./tests/test-typia-schema start --include reflect_literals` | pass |
| `pnpm --filter ./tests/test-error start` | pass — 79 files, 875 diagnostics |
| `pnpm --filter ./tests/test-feature-identity start` | pass |
| `pnpm test` (previous head) | pass, exit 0 |
| `pnpm test` (this head) | running |

samchon (commented):
## Self-Review round 3 — 1 finding, applied

Fresh full round over `master...cc56c4cb8`.

**`misc.mdx` referenced a type the page never declares.** The empty-union note illustrated a computed argument with `Exclude<Color, Color>`, but `Color` appears nowhere on the page — the one example of the case the fix exists for pointed at nothing. Rewritten to describe the shape instead of naming an invisible type.

### Verified clean in this round

- **Predicate.** Re-read against the emitter. `length` counts constant values, boolean atomics, and `null`; `size` counts every bucket plus `null`. Equality holds exactly when every member is renderable, `length == 0` separates `NO` from `ONLY`, and both messages stay reachable.
- **In-repo consumers of `reflect.literals`.** `tests/test-utils` (`test_json_schema_type_checker_cover_string`, `test_openapi_validator_format_parity`, both over `tags.Format.Value`), `experiments/declaration`, and `examples/src/reflect/literals.ts` are all plain literal unions and take the unchanged path.
- **Test-file conventions.** Both TS files export exactly one `test_*` matching their basename under a correctly named `src/features/reflect.literals/` peer directory, and `test-feature-identity` passes. Both Go files hold one `Test*` each and reuse the sibling helpers in `transform_helpers_test.go`. Every new case carries the three-part doc comment.
- **Documentation.** Every claim on the page re-checked against the code: the renderable set, `boolean` standing for `true | false`, `"a" | number` being rejected rather than partially emitted, the within-group sort, and `never` emitting `[]`. Prose stays one paragraph per source line.
- **Change integrity.** No fixture, snapshot, workflow, package manifest, dependency, or generated baseline touched beyond the new tests. No version bump, per the campaign freeze.

Nothing else in the diff produced a sound improvement this round.

samchon (commented):
## Self-Review round 4 — 1 finding, applied

Fresh full round over `master...03f1681c5`.

**Two claims from round 1 had no assertion behind them.** Round 1 stated that a tag-branded atomic stays rejected and that the `NO`/`ONLY` split holds for a renderable atomic sitting beside a non-renderable one. Neither was pinned anywhere — a claim in a review comment is not coverage. Both are distinct metadata shapes reaching the predicate: `string & tags.Format<"uuid">` raises the member count through the atomic's tags without adding anything the emitter can render, and `boolean | number` is the only mixed form whose renderable half is an atomic rather than a constant. Added, and both behave as claimed: `NO` for the brand, `ONLY` for the mixed atomics.

The rejection matrix is now 7 cases — `string`, `any`, `string & tags.Format<"uuid">` reporting `no constant literal type found`; `string | null`, `"a" | number`, `` `prefix${number}` | "a" ``, `boolean | number` reporting `only constant literal types are allowed`.

### Verified clean in this round

- Re-read the predicate, both Go tests, both TS tests, the `test-error` fixture, and the guide section. Nothing else produced a sound improvement.
- Fixture template now imports `tags`; the rejection fixtures do not typecheck-gate, and the shared tsconfig sets no `noUnusedLocals`, so the unused import in the other six cases is inert.

`go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestReflectLiteralsNonLiteral -v` — 7/7 pass.

samchon (commented):
## Self-Review round 5 — clean

Fresh full round over `master...a51fc2306`. Read every hunk again: the predicate and its comment, both Go tests and their fixtures and runners, both TypeScript feature tests, the `test-error` fixture, and the guide section. No sound improvement surfaced.

### What this round re-established

- **Predicate.** `length` (constant values + boolean atomics + `null`) is bounded above by `size` (every bucket + `null`) in all fourteen counted fields, so `size == length` holds exactly when every member is renderable. Both diagnostics stay reachable and distinguishable across the 7-cell rejection matrix.
- **Rule statement.** The admission rule is "an argument carrying no member is uninhabited, so its literal set is empty", covered by six spellings: `never`, an alias, `Exclude`, `Extract`, `string & number`, and a typia-pruned intersection.
- **Consumers.** `tests/test-utils`, `experiments/declaration`, and `examples/src/reflect/literals.ts` all take the unchanged path.
- **Hygiene.** No leftover fixture scratch under `packages/typia/native/.tmp-ttsc-typia-tests/`; working tree carries no unrelated staged change; no package version touched, per the campaign freeze.
- **Fixture typecheck gate.** The positive fixture runs through `ttscTypiaTestTypecheck`, so every argument in it is proved to satisfy `T extends Atomic.Type | null` before the transform is asked to accept it — the empty-union arguments are legal TypeScript, not merely tolerated.

Merge gates still outstanding: pull-request CI on this head, root `pnpm test` on this head, and a re-run of the mutation-sensitivity proof at this head.

samchon (commented):
## Merge gate — all green at `a51fc2306`

**CI.** `build`, four `Ubuntu` lanes, `spell-check`, and both Socket Security checks: SUCCESS. `origin/master` is still `3a5b8a2e0`, so the base has not drifted since the run; `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

**Local, at this exact head.**

| Command | Result |
| --- | --- |
| root `pnpm test` | exit 0 — every `tests/test-*` workspace plus both Go trees |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia -run TestReflectLiterals` | pass |
| `pnpm format` | clean, integrated |

**Mutation sensitivity, re-proved at this head.** Reverted only the predicate to its `master` form, kept every test, and re-ran:

```
--- FAIL: TestReflectLiteralsEmptyUnionTransform
--- FAIL: TestReflectLiteralsNonLiteralRejectionTransform
    --- PASS: .../atomic
    --- PASS: .../any
    --- FAIL: .../nullable_atomic
    --- PASS: .../literal_beside_atomic
    --- PASS: .../literal_beside_template
    --- PASS: .../renderable_atomic_beside_atomic
    --- PASS: .../tag_branded_atomic
```

The two assertions the fix earns fail, and all five adjacent controls still pass — the probes discriminate the fix rather than the suite. Source restored byte-identical to `HEAD` and re-run green.

**Self-Review.** Five complete rounds; round 5 produced no sound improvement. Seven findings applied across rounds 1–4.

**Closing set.** `git log origin/master..HEAD` carries exactly one closing line, `Close #2377`, on `a3dfe62a1`, whose fix survives at `HEAD`. No revert in the range.

**Version freeze held.** No `package.json` touched. The 14.0.3 release is a separate maintainer step.

PR #2383, opened 2026-08-26
fix(typia): carry a bigint literal as a bigint, and refuse an empty literals argument

samchon:
Two changes to `typia.reflect`, both about a member the metadata could not carry honestly.

## bigint

typescript-go reports a bigint literal as a `jsnum.PseudoBigInt`, a struct in an internal package the shim does not re-export. Nothing downstream could name it, so `LiteralFactory` reflected its fields and `typia.reflect.literals<2n>()` emitted `{ base10Value: "2", negative: false }` where the signature promises a bigint. The constant is now normalized to `*big.Int` at the metadata boundary — nameable, exact at any magnitude, and a shape `LiteralFactory` already lowers to a bigint expression.

`reflect.schema` reached the emit by a second route: it lowered the whole metadata tree through `encoding/json` first, and JSON has no bigint. That trip also duplicated what `LiteralFactory` does when it reflects a struct, so writing the tree directly both fixes the value and removes the step. Diffed over a metadata tree spanning objects, optional and nullable members, tags, arrays, tuples, sets, maps, natives, functions, aliases and unions — 811 lines — the bigint values were the only difference.

`ExpressionFactory.Bigint` then spelled its argument as a bare number literal, so `9007199254740993n` reached the emit as `BigInt(9007199254740993)` and JavaScript rounded the literal before `BigInt` parsed it. **That is a validator soundness bug, not only a reflect one:**

```ts
typia.is<9007199254740993n>(9007199254740992n); // was true
typia.is<9007199254740993n>(9007199254740993n); // was false
```

The repository already spells this hazard out where it was avoided — `numericRangeFactory_bigint_literal` and `_isTypeInt64Bigint` — and this factory was the one that had not. The fix reaches every operation that emits a bigint literal: `is`, `assert`, `validate`, `equals`, `random`, and tag exclusions.

## Empty arguments

14.0.3 made `reflect.literals<never>()` emit `[]`. That was wrong. The operation hands back the members of a union, so an argument naming none has no list to give, and an empty array reads as a real answer to a question that has none. `never`, its aliases, an exhaustive `Exclude`, and the bare `null` flag are compile errors again, restoring the admission rule to what shipped through 14.0.2.

## Verification

| Command | Result |
| --- | --- |
| `go -C packages/typia/test test ../native/cmd/ttsc-typia -run 'TestReflectLiteralsBigint\|TestBigintLiteralPrecision\|TestReflectSchemaBigint\|TestReflectLiteralsNonLiteral'` | pass |
| `pnpm --filter ./tests/test-typia-schema start --include reflect` | pass |
| `pnpm --filter ./tests/test-error start` | pass — 80 files, 880 diagnostics |
| `pnpm --filter ./tests/test-feature-identity start` | pass |
| `pnpm format` | clean, integrated |
| root `pnpm test` | pending |

New coverage: `TestReflectLiteralsBigintTransform` (emit + magnitude), `TestBigintLiteralPrecisionTransform` (validator path, both spelling and behavior), `TestReflectSchemaBigintTransform` (the second emit route), an 11-cell rejection matrix including `never`/`null`, two `test-error` fixtures, and `test_reflect_literals_bigint`.

Version stays at 14.0.3 here; the release is a separate step.


samchon (commented):
## Self-Review round 1 — 1 finding, applied (`091eec8b6`)

**I broke tag assignment for bigint constants.** Normalizing the checker's unnameable `jsnum.PseudoBigInt` to `*math/big.Int` fixed the emit but introduced pointer semantics into `MetadataConstantValue.Value`, which the factories compare with `==` — `iterate_metadata_intersection.go:225` matches a child's constant against the merged parent's that way. A pointer allocates fresh per literal, so the comparison tested identity and silently decided "not the same constant".

Reproduced against `master` to confirm it was mine and not pre-existing:

```
master                          (1n|2n) & tags.Type<"int64">  → both members tagged
this branch, *big.Int           (1n|2n) & tags.Type<"int64">  → tags: [], []
this branch, MetadataBigint     (1n|2n) & tags.Type<"int64">  → both members tagged
```

Number and string constants were never affected, which is exactly what made it invisible.

The fix is the representation, not the comparison site: every other value a constant can hold is a comparable value, so the stand-in is one too. `MetadataBigint` carries the exact base-10 digits, renders them through `String()` for the emit, and marshals them unquoted for metadata a downstream tool serializes.

Coverage sits at two levels. `TestMetadataBigintIsComparable` pins the comparison itself — including once boxed in the `any` a constant value actually holds — so a future change of representation fails one level below the transform rather than as a dropped tag three packages away. `TestReflectSchemaBigintTransform` asserts the tag survives end to end, with a tagged number constant beside it as the control. Substituting the pointer representation back fails it with `bigint constant: expected every member tagged Type<"int64">, got ["",""]`.

### Also swept this round

- Every other reader of a constant value: `iterate_metadata_constant_key` (default `fmt.Sprint`), `iterate_metadata_sort`'s bigint branch, `CheckerProgrammer`, `check_exclude`, `RandomProgrammer`, and the `reflect.DeepEqual` sites in `MetadataSchema.covers`/merge. All read the value rather than its identity, and all work on a comparable struct.
- `ProtobufFactory`'s bigint paths read tags, never the value.
- The `reflect.literals` predicate is now byte-identical to what shipped through 14.0.2.

| Command | Result |
| --- | --- |
| `go test ../native/core/schemas/metadata -run TestMetadataBigintIsComparable` | pass |
| `go test ../native/cmd/ttsc-typia -run 'TestReflectLiteralsBigint\|TestBigintLiteralPrecision\|TestReflectSchemaBigint\|TestReflectLiteralsNonLiteral'` | pass |
| same, pointer representation substituted back | `TestReflectSchemaBigint` fails on the tag assertion; restored → pass |
| root `pnpm test` | running |

samchon (commented):
## Self-Review rounds 2–3 — clean

Fresh full rounds over `master...a488726d1`.

### Round 2 — the representation and every consumer of a constant value

- `iterate_metadata_constant_key` reaches `MetadataBigint` through its `default: fmt.Sprint` branch, which uses `String()`, so dedup keys on the digits. A number constant keys to the same digits but lives in its own `MetadataConstant` bucket, so the two cannot collide.
- `iterate_metadata_sort`'s bigint branch, `CheckerProgrammer`, `check_exclude`, and `RandomProgrammer` all lower the value through `fmt.Sprint`/`ExpressionFactory.Bigint` — unchanged behavior.
- The `reflect.DeepEqual` sites in `MetadataSchema.covers` and its merge compare a comparable struct field-wise, which is exact.
- `ProtobufFactory`'s bigint paths read tags, never the value.
- `MetadataConstantValue_from` would not rebuild a `MetadataBigint` from JSON, but nothing outside the metadata package calls the `*_from` family — the deserializer is only ever reached recursively from within itself. Pre-existing asymmetry, untouched.
- `literalFactory_isNilLike` classifies a struct as not-nil, so a zero bigint keeps its property rather than being dropped from an object literal.

### Round 3 — the emit path the round-trip removal touches

`reflectTransformer_literal` is shared by `reflect.schema`, `reflect.schemas`, and `reflect.metadata`. `metadata` passes `{schemas, components}` — the same shape `reflect.schemas` passes, which is what the 811-line A/B diff was taken over, so the same evidence covers it. (It also has no public TypeScript declaration; `reflect.ts` exports `schemas`, `schema`, `name`, and `literals` only.)

Also re-read this round: the `test-error` fixtures typecheck and name their own diagnostic code; the guide's four claims match the code; the `reflect.literals` predicate is byte-identical to 14.0.2; no package version was touched.

Nothing in either round produced a sound improvement.

### Gate status

| | |
| --- | --- |
| CI | Ubuntu ×4, spell-check, Socket ×2 green; `build` running |
| root `pnpm test` | exit 0 at `091eec8b6`; re-running at this head |
| mutation probe | pointer representation substituted back → `TestReflectSchemaBigint` fails on the tag assertion |

samchon (commented):
## Merge gate — all green at `a488726d1`

**CI.** `build`, four `Ubuntu` lanes, `spell-check`, and both Socket Security checks: SUCCESS.

**Local, at this exact head.** Root `pnpm test` exit 0 — every `tests/test-*` workspace plus both Go trees.

**Mutation sensitivity.** Each of the three changes was substituted back for its pre-fix form, with every test retained:

| Substitution | Result |
| --- | --- |
| `*big.Int` in place of `MetadataBigint` | `TestReflectSchemaBigint` → `bigint constant: expected every member tagged Type<"int64">, got ["",""]` |
| number literal in place of the string argument to `BigInt(...)` | `TestBigintLiteralPrecision` → `bigint literal must be passed as a string, not a number literal`; `TestReflectLiteralsBigint` → `unsafe[0]: expected 9007199254740993n, got 9007199254740992n` |
| the empty argument admitted again | `TestReflectLiteralsNonLiteral` → `never`, `exhaustive_exclude`, `bare_null` all fail |

Adjacent controls stayed green in every case, and each source was restored byte-identical and re-run green.

One observation from the third probe worth recording: `never_alias` did not fail under it. `type Empty = never` carries an alias bucket, so it is refused by the mixed-composition branch rather than the empty-argument branch — it is rejected either way, which is why the cell holds under both, but the discriminating cells for that rule are `never`, `Exclude`, and `null`.

**Self-Review.** Three complete rounds. Round 1 found and fixed a regression I introduced; rounds 2 and 3 produced no sound improvement.

**Closing set.** No `Close #n` line in the range — #2377 is already closed, and this pull request changes the resolution it landed with rather than resolving something new.

**Version freeze held.** No `package.json` touched.

PR #2387, opened 2026-09-02
fix(typia): restore validator and OpenAPI correctness

### LINKED ISSUE #2380: OpenAPIConverter.upgradeComponents does not correctly upgrade Format<'byte'>
### 📝 Summary
The generated schema for `OpenAPIConverter.upgradeComponents(..., "3.1" | "3.2")` does not correctly upgrade the property type of a byte-formatted string.

  - **`typia` Version**: 14.0.3
  - **`@typia/utils` Version**: 14.0.3
  - **Expected behavior**: 
```jsonc
{
  "schemas": {
    "Foo": {
      "type": "object",
      "properties": {
        "bytes": {
          "type": "string",
          "contentEncoding": "base64", // this
          "contentMediaType": "application/octet-stream", // ideally also this if not otherwise specified with `ContentMediaType<...>`
        }
      },
      "required": [
        "bytes"
      ],
      "additionalProperties": false
    }
  }
}
```

  - **Actual behavior**: 
```json
{
  "schemas": {
    "Foo": {
      "type": "object",
      "properties": {
        "bytes": {
          "type": "string",
          "format": "byte" // Invalid for any OAS version >= 3.1.0
        }
      },
      "required": [
        "bytes"
      ],
      "additionalProperties": false
    }
  }
}
```

As of [OpenAPI Specification 3.1.0](https://spec.openapis.org/oas/v3.1.0.html#data-types), the "byte" data format is no longer supported. Instead, the intended way to indicate a [binary property in base64 format](https://spec.openapis.org/oas/v3.1.0.html#considerations-for-file-uploads) is to use `"contentEncoding": "base64"`. For the sake of completeness, the property can also explicitly specify`"contentMediaType": "application/octet-stream"` in the case of there not being a more specific MIME type given (i.e. with `ContentMediaType<...>`), though this is not strictly necessary.

### 💻 Code occuring the bug

```typescript
import typia, { tags } from 'typia';
import { OpenAPIConverter } from '@typia/utils';

type Foo = {
  bytes: string & tags.Format<'byte'>;
}

const oaSchema30 = typia.json.schema<Foo, "3.0">();
const oaSchema32 = utils.OpenApiConverter.upgradeComponents(oaSchema30.components);

console.log(JSON.stringify(oaSchema32, null, 2));
```

### LINKED ISSUE #2381: MinItems, MaxItems, and UniqueItems show wrong error message
### 📝 Summary
When attached to the wrong data type, the `MinItems`, `MaxItems`, and `UniqueItems` tags all report the wrong required type.

  - **Typia Version**: 14.0.3
  - **Expected behavior**: Error message informing that property must be of an array type
  - **Actual behavior**: 
```plaintext
src/index.ts:14:2 - error TS(typia.createIs): typia transform error: unsupported type detected

- Foo.a: boolean & MinItems<1>
  - the property ["typia.tag"] target must contain boolean type.

- Foo.a: boolean & MinItems<1>
  - the property ["typia.tag"] target must contain boolean type.

- Foo.b: number & MaxItems<1>
  - the property ["typia.tag"] target must contain number type.

- Foo.c: string & UniqueItems<true>
  - the property ["typia.tag"] target must contain string type.
```

The actual reported required type seems to be whatever type the property was erroneously declared as instead of the expected "array". Also, if the erroneous type is `boolean`, the error message for that property gets printed twice.

### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDANHA3g1BzAZzgF84AzKCEOAIiRVRoG4AoF%2BgUzgDEII4AvNhZw4qAFxwARnwA2HVADs4AMlyEAdAFlgigJIwOIAgB4AjAD5WoqZMUBXEFI5RV6gttQAPA0dOXrOABjSQIYKF08Nxh8DwBVRWAAR3sOX2NWYjZ6NA0gqAVDPVNeCAsACgBKJiA

### 💻 Code occuring the bug
```typescript

type Foo = {
  a: boolean & tags.MinItems<1>;
  b: number & tags.MaxItems<1>;
  c: string & tags.UniqueItems;
}

typia.createIs<Foo>();
```

### LINKED ISSUE #2385: Declaring both an array field and a field of a self-referential union type that includes an array results in helper functions not getting generated
I decided to report this issue here as it's a code generation bug related to typia, but it could also be a ttsc bug.

### 📝 Summary

  - **`typia` Version**: 14.0.3
  - **`ttsc` Version**: 0.28.1
  - **Expected behavior**: Correct and valid code to be generated
  - **Actual behavior**: A generated helper function is missing at runtime

A very bizarre and specific bug, but one I ran into in my production-intended use case and managed to track it down to a minimal reproduction. I'm using a type definition that is essentially "any type representable by JSON", which is necessarily a fairly complex union type involving pseudo-circular references: 

```typescript
type JsonPrimitive = string | number | boolean | null;
type JsonArray = JsonValue[];
type JsonObject = { [key: string]: JsonValue };
type JsonComposite = JsonArray | JsonObject;
type JsonValue = JsonPrimitive | JsonComposite;
```

As it turns out, if you use such a complex union type as a field type of an object while declaring another field in the object as a regular array type, the transformation fails to generate some required helper functions properly, resulting in a runtime error:

```typescript
interface Foo {
  strings: string[];
  value: JsonValue;
}

const validateFoo = typia.createValidate<Foo>();

console.log(validateFoo({ value: [123] }))
// ReferenceError: _va0 is not defined

console.log(validateFoo({ value: [123], strings: ['a'] }))
// ReferenceError: _ia0 is not defined
```

There may be other missing helper functions, but these are the ones that I could reliably reproduce and generate errors for.

While experimenting with the reproducible example, I also noticed an odd discrepancy - some validation attempts will fail when built with `ttsc` and then run through `node`, but will succeed if run directly with `ttsx`:

```typescript
console.log(validateFoo(123))
// If run using ttsc & node: 
//   ReferenceError: _va0 is not defined
//
// If run using ttsx:
//   { success: false, errors: ... }

console.log(validateFoo({}))
// If run using ttsc & node: 
//   ReferenceError: _va0 is not defined
//
// If run using ttsx:
//   { success: false, errors: ... }
```

In yet another bizarre development, the order in which the fields are declared makes a difference. If the field order is reversed, the errors go away:

```typescript
interface Foo {
  // Swap the order of these fields around and it suddenly works
  value: JsonValue;
  strings: string[];
}

const validateFoo = typia.createValidate<Foo>();

console.log(validateFoo(123))
// { success: false, errors: ... }

console.log(validateFoo({}))
// { success: false, errors: ... }

console.log(validateFoo({ value: [123] }))
// { success: false, errors: ... }

console.log(validateFoo({ value: [123], strings: ['a'] }))
// { success: true, data: ... }
```

### ⏯ Playground Link

https://typia.io/playground/?script=JYWwDg9gTgLgBDAnmYBDOAzKERwERIqp4DcAUGYQKZwBSAzhAHYAKUowMwAbjQLxx6MdkwDmcAD5wmAVxAAjKlElx5ECABsqqJitkaN5anUZMAglCipEcAQ2YA1VBplUA2gF0jyGvaYB5eQArKgBjeAEAbzg3AGsqRAAuQWFgMQ9kvycXGgBfbzBfUwBhHEh6Tn4TZgsrGyk-QJDwgqLHZ1dbatZ2EE4eGgaSsogKmCpyMgB6KbgAFQALYHoEHzgAEyoMNP7mOFEqJiVUcZWYBZpFDQgAdzglbCh6MjTxqAxUUJoAMXU4SLIcBSIlE9GSQhBnnIQO4HSomVM2Vc5FyFBm8yWZzWm22TF2uhuwAM0gg8AORys4wQF1UVGudwe0Ge6NeSg%2BXzgvwg-2msxhcIR7Ry5HRQIhaVB4NSYihvLgqLIoWYQjgsI0wHWJyoXK6hDQADpQlBtONshqtQAeLkAPgAFABKSbogBKWyUhy%2BAFFLNBkgB9WEABjgyxJ8BxaSo6zgtpgMHooTgzA0iHtiuVmio%2BuuoltavN4y5toAjAAmADM9rTctdGHdTC9Pqg-qDIZWTFJGy2kejsfjieTqfTTEYWmzEFz%2Bc1hfUtsiuSraNmtfrjceLdQwdDHfD3aOvbjCbgAGoEPGAB5ppUjzPjyfOAva2fRNWuZJuMvljzyxc1t3GhsqG9dc4D9NAt3bTsI33GND0TU9D0vYdRyzHM8wfacnwgOdVQFGJPw8AAaYEJTBGIAHJUHI78F3tIA

### 💻 Code occuring the bug

```typescript
import typia from "typia";

type JsonPrimitive = string | number | boolean | null;
type JsonArray = JsonValue[];
type JsonObject = { [key: string]: JsonValue };
type JsonComposite = JsonArray | JsonObject;
type JsonValue = JsonPrimitive | JsonComposite;

// This type definition generates the below errors
interface Foo {
  strings: string[];
  value: JsonValue;
}

// This type definition will not generate the below errors
// interface Foo {
//   value: JsonValue;
//   strings: string[];
// }

const validateFoo = typia.createValidate<Foo>();

console.log(validateFoo(123))
// ReferenceError: _va0 is not defined (ttsc only)

console.log(validateFoo({}))
// ReferenceError: _va0 is not defined (ttsc only)

console.log(validateFoo({ value: [123] }))
// ReferenceError: _va0 is not defined (ttsc + ttsx)

console.log(validateFoo({ value: [123], strings: ['a'] }))
// ReferenceError: _ia0 is not defined (ttsc + ttsx)
```

samchon:
## Intent

Complete the accepted v14 correctness cycle for recursive native helpers,
type-tag diagnostics, and OpenAPI byte/content annotations.

## Changes

- Use canonical metadata indices for recursive array and tuple helpers across
  all native checker, serializer, notation, plain, and random families.
- Report the declared type-tag target once while preserving distinct
  exploration contexts.
- Preserve and normalize byte encodings, media types, string constants, and
  cyclic/shared constant constraint graphs across OpenAPI 3.0, 3.1, 3.2, and
  Typia's emended representation.
- Add permanent transform, runtime, converter, schema, diagnostics, packaged
  consumer, and guide coverage for the full verified consequence surface.

## Verification

- All focused TypeScript and Go regression groups passed.
- Mutation checks proved sensitivity for recursive helper identity, declared
  tag targets, byte/content conversion, constant normalization, escaped
  references, cyclic order, shared DAGs, and deep reference chains.
- Eight solo Self-Review rounds completed; the final round was clean.
- All GitHub pull-request checks passed: format, build, experiments, test,
  website, spell-check, and Socket Security.
- The frozen prospective-merge manifest passed at synthetic merge
  `dedc99ae792ff9b6c13dfd4236b5550aa982f426`: frozen install, format/diff,
  root build, benchmark build, full test, six tarballs, smoke, ESM, MCP,
  isolated declarations, and website production build.
- The website build generated the playground WebAssembly module, 74 static
  pages, the Pagefind index, and sitemap successfully.

Closes #2385.
Closes #2381.
Closes #2380.


samchon (commented):
## Commit review: `403fea9c7`

Reviewed the #2385 implementation against its parent and the frozen consequence matrix.

- Verified the root invariant: recursive container call sites use `MetadataArrayType.Index` / `MetadataTupleType.Index`, so every helper declaration and visit-context key now uses that same canonical identity instead of full collection position.
- Traced and corrected the repeated writer pattern in checker, JSON stringify, notation, plain classify/clone/prune, and random generation.
- Preserved dense returned-helper ordering used by the surrounding function maps; only emitted identities changed.
- Made the plain-prune reachability filter cycle-safe after the direct recursive tuple control exposed a transform-time stack overflow.
- Added one end-to-end command-transform case covering ordinary-before-recursive ordering, reversed controls, arrays, tuples, direct/factory checkers, stringify, notation, plain operations, random generation, valid/invalid values, and runtime execution.

Verification:

- Pre-fix fixture: failed with calls to missing `_ia0`-family helpers while `_ia1`-family declarations were emitted.
- Post-fix focused command/runtime test: PASS.
- Native programmer package suite: PASS.
- Repository formatter: PASS.

No unresolved finding is known for this commit. Full integration and CI remain cycle-level gates.

samchon (commented):
## Commit review: `5f09ff78e`

Reviewed the #2381 implementation against its parent and the frozen consequence matrix.

- Verified wrong-target compatibility is checked while the raw tag target set is still available; invalid tags are no longer collapsed to an empty canonical target before diagnostics.
- Verified canonical tags are constructed only after the host type is found in the declared target set.
- Restricted duplicate suppression to exact type-tag error records with the same metadata name, exploration owner/path, and complete ordered message vector. Equal text at another property or a second invalid tag message in the same vector remains observable.
- Preserved direct `Validate` behavior while changing a synthetic mismatch to name `tag.Target` rather than the host type.
- Added exact native unit/command oracles for single and multiple targets, boolean duplication, aliases, scalar/object hosts, valid arrays, and direct/factory operations, plus four independently owned error-workspace fixtures.

Verification:

- Pre-fix direct/factory fixtures: five diagnostics for four invalid properties, with boolean duplicated and all messages naming host types.
- Post-fix factory unit suite: PASS.
- Post-fix command diagnostics: exactly four `target must contain array type` reports per entry point, PASS.
- Repository formatter: PASS.

No unresolved finding is known for this commit. The error workspace will run after the cycle build refreshes the native executable; full integration and CI remain cycle-level gates.

samchon (commented):
## Commit review: `f1deaff05`

Reviewed the #2380 implementation against its parent and the frozen consequence matrix.

- Added the standard `contentEncoding` keyword to raw OpenAPI 3.1/3.2 and Typia's emended string-schema contracts without adding it to the OpenAPI 3.0 contract.
- Normalized only exact raw `contentEncoding: "base64"` schemas with no format or `format: "byte"` into Typia's canonical byte metadata. A custom format, unrelated encoding, or conflicting pair remains observable.
- Made raw-version output explicit: OpenAPI 3.1 replaces canonical byte format with base64 content encoding, while OpenAPI 3.0 retains byte format and omits unsupported 3.1 content keywords.
- Verified recursive use through individual schema, component, and document conversion paths, including object properties, array items, union branches, references, and component reuse.
- Added native-generation coverage for direct OpenAPI 3.0, emended 3.1, raw 3.1, explicit content media types, and a 3.0-to-3.1 conversion.
- Documented that generated/emended schemas and raw OpenAPI 3.1 Schema Objects intentionally use different byte spellings.

Verification:

- Pre-fix public type check: `contentEncoding` rejected in OpenAPI 3.1/3.2 schemas.
- Focused converter and native generator tests: PASS.
- Complete utils JSON-schema and OpenAPI groups: PASS.
- Complete native JSON-schema group: PASS.
- `@typia/interface` and `@typia/utils` package builds: PASS.
- Production website build, including `/docs/json/schema`: PASS.
- Mutation of the base64 normalization predicate: focused test failed on root, nested item, and union branch; restored implementation: PASS.
- Repository formatter and staged diff check: PASS.

No unresolved finding is known for this commit. Packed-consumer and full integration validation remain cycle-level gates.

samchon (commented):
## Commit review: `9d831d442`

Reviewed the #2385 test correction against its parent and the recursive-helper consequence surface.

- Added the missing `_ra0` declaration/call presence check for recursive array random generation.
- Added a recursive-array witness whose ordinary array precedes the recursive array, so metadata index and collection position diverge as in the reported defect.
- Executed the generated random factory with deterministic array/string hooks and required both generated properties to be arrays, proving that the recursive array helper is callable at runtime rather than merely emitted as dead text.
- Kept the existing tuple random case, so array and tuple recursive helper families are independently exercised.

Verification:

- `go -C packages/typia/native test ./cmd/ttsc-typia -run TestRecursiveContainerHelperIndexTransform -count=1`: PASS.
- Repository formatter and diff check: PASS.

No unresolved finding is known for this commit. Cycle-level gates and the restarted Self-Review remain pending.

samchon (commented):
## Self-Review round 1 — findings

The first fresh review of the complete pull-request diff found three sound gaps. The review round is therefore not clean and will restart after corrections.

1. The #2385 regression fixture executed recursive tuple random generation but only inspected the recursive array random helper as generated text. A missing `_ra0` declaration could therefore escape runtime coverage. Commit `9d831d442` adds a deterministic runtime call with displaced ordinary/recursive array positions.
2. The OpenAPI 3.0 downgrade removed `contentEncoding` and `contentMediaType` even when the value could not be represented by legacy `format: "byte"`. This silently loses custom encodings, media types, and conflicting format/encoding pairs. The official OAI extension registry defines `x-jsonschema-contentEncoding` and `x-jsonschema-contentMediaType` for preserving these JSON Schema annotations in OpenAPI 3.0 and earlier, so the converter must use and recover those registered extensions.
3. String enum conversion expands values into emended constants carrying string annotations, but the OpenAPI 3.1 downgrade reduced every constant to `{ const }`. Byte encoding, media type, format, and other retained string metadata consequently disappeared from enum branches. Constant downgrade must preserve and version-normalize string annotations.
4. The two new TypeScript regression files explain the defect but omit the numbered two-step-or-more scenario list required by the development test contract. Their opening documentation comments must be corrected.

Sources used to validate finding 2:

- https://spec.openapis.org/registry/extension/x-jsonschema-contentEncoding.html
- https://spec.openapis.org/registry/extension/x-jsonschema-contentMediaType.html
- https://spec.openapis.org/registry/format/byte

No approval is requested. These findings are within the accepted #2380/#2385 consequence surface and will be repaired before Self-Review restarts.

samchon (commented):
## Self-Review round 2 — findings

The fresh review restarted from `origin/master...bd53c560c` and reread all 27 changed files. It found two documentation-quality gaps; the round is not clean and will restart after correction.

1. The four new `tests/test-error/src/tag/error_tag_*_wrong_*.ts` cases use ordinary line comments. The development contract requires every new or materially changed case to open with a `Verifies ...` doc-comment headline, a paragraph explaining the non-obvious regression, and a numbered scenario list with at least two steps. The fixtures are valid one-case-per-file transform rejections, but their opening documentation must be brought into that required shape.
2. The website guide still says only that explicit content media types survive in raw OpenAPI 3.1. The corrected converter now preserves them in OpenAPI 3.0 through the OAI-registered `x-jsonschema-contentMediaType` field and similarly preserves non-byte encodings through `x-jsonschema-contentEncoding`. The public guide must state this 3.0 representation so it agrees with the shipped interface and converter behavior.

No product-code defect was found in this round. Both findings are within the existing #2380/#2381 files and will be corrected before Self-Review restarts.

samchon (commented):
## Commit review: `bd53c560c`

Reviewed the #2380 correction against its parent, the OpenAPI 3.0/3.1/3.2 converter boundaries, and the public schema interfaces.

Two actionable findings remain:

1. **High — annotated string constants do not round-trip through OpenAPI 3.1 or 3.2.** Downgrading `{ const: "QQ==", format: "byte" }` correctly emits `{ const: "QQ==", contentEncoding: "base64" }`, but re-upgrading that raw constant leaves `contentEncoding` unchanged because the upgrader only normalizes schemas whose `type` is `"string"`. The constant path must normalize both untyped string constants and typed `type: "string"` constants, and tests must assert the complete downgrade/upgrade round-trip for 3.1 and 3.2.
2. **Medium — the public constant types do not describe the annotations emitted at runtime.** Enum upgrading and constant downgrading preserve `format`, `pattern`, `contentEncoding`, `contentMediaType`, `minLength`, and `maxLength`, but `OpenApi.IJsonSchema.IConstant`, `OpenApiV3_1.IJsonSchema.IConstant`, and `OpenApiV3_2.IJsonSchema.IConstant` expose none of those fields. Consumers cannot access valid runtime metadata after `isConstant` narrowing. The constant interfaces must expose the supported string annotations, with a compile-time contract assertion in the regression test.

Reproduction:

- Built `@typia/utils` downgraded the canonical constant to `contentEncoding: "base64"` and re-upgraded it without restoring `format: "byte"`.
- Direct inspection confirmed that the three public `IConstant` interfaces contain only common attributes, `const`, and raw-dialect `nullable`.

This commit is not accepted as final. Both findings are within #2380 and will be corrected before the next Self-Review round.

samchon (commented):
## Commit review: `06db760e8`

Reviewed the four #2381 rejection-fixture comments against the development test-documentation contract and their executable diagnostic expectations.

One low-severity finding remains:

- `error_tag_unique_items_wrong_string.ts` wraps its opening `Verifies ...` headline across two source lines. The contract requires a one-line headline. The wording is accurate, and its why paragraph and two-step scenario list are complete, so only the headline layout needs correction.

The other three fixtures satisfy the required three-part structure. All four comments accurately describe the direct and factory diagnostic checks in `wrong_type_tag_target_diagnostic_test.go`.

This commit is not accepted as final. The headline will be corrected in the next campaign correction commit.

samchon (commented):
## Commit review: `9a1c0ab84`

Reviewed the #2380 correction against exact-optional consumers, mixed raw schemas, the public v3.1 type checker, and the shared v3.2 conversion path.

Three actionable findings remain:

1. **Blocking/high - the public declarations fail with `exactOptionalPropertyTypes`.** V3.1 and V3.2 `IMixed` inherit optional string annotation fields through incompatible declarations, producing TS2320 in `@typia/test-typia-exact-optional`.
2. **High - a mixed annotated constant is emitted three times.** `{ type: ["string", "null"], const: "QQ==", contentEncoding: "base64" }` upgrades to a `oneOf` containing three identical byte constants. Because `oneOf` requires exactly one matching branch, the duplicate output changes a satisfiable constant into an unsatisfiable schema. The mixed branch must process the intersecting `const` once and stop decomposing its type array.
3. **Medium - `OpenApiV3_1TypeChecker.isConstant` has an unsound predicate.** It returns true for `IMixed` values carrying `const`, while its predicate narrows exclusively to `IConstant`, whose `type` is scalar rather than an array. The predicate must include `IMixed`.

Evidence:

- GitHub Actions and a local exact-optional build reproduced TS2320.
- The built converter reproduced three identical `oneOf` branches for both 3.1 and the shared 3.2 path.
- The public checker returned true while `Array.isArray(schema.type)` was also true.

This commit is not accepted as final. The exact-optional declaration failure is corrected by `e95e9d0a8`; the mixed constant conversion and predicate require another campaign correction commit.

samchon (commented):
## Commit review: `e95e9d0a8`

Reviewed the exact-optional declaration repair against both raw OpenAPI interfaces, the canonical interface, converter-produced fields, and the previously failing consumer.

No actionable finding remains in this commit.

- `IConstant` now derives its supported string annotations from `IString`, removing the incompatible duplicate optional declarations inherited by `IMixed`.
- The public field set is preserved for canonical, v3.1, and v3.2 schemas.
- Runtime conversion behavior is unchanged.
- `@typia/interface` builds with `--exactOptionalPropertyTypes true`.
- `@typia/test-typia-exact-optional` passes.

The commit is accepted. The separate mixed-constant findings from the parent review remain assigned to the next correction commit.

samchon (commented):
## Commit review: `c18480273`

Reviewed the mixed-constant correction against nullable inference, sibling JSON Schema constraints, the shared v3.2 path, and the public type-guard test oracle.

Two actionable findings remain:

1. **High - a mixed constant can still be widened by nullable inference or contradictory siblings.** `{ type: ["string", "null"], const: "x", enum: ["x", null] }` upgrades to a union accepting both `"x"` and `null`, although `const` limits the source schema to `"x"`. Contradictory `type`, `enum`, composition, and reference constraints are likewise discarded by the early return and become satisfiable constants.
2. **Medium - the public type-guard test is insensitive to its predicate declaration.** The test starts from a forced `IMixed` and checks only runtime booleans, so restoring the unsound `schema is IConstant` predicate does not make it fail.

Evidence:

- The nullable-enum input reproduced `{ oneOf: [{ const: "x", enum: ["x", null] }, { type: "null" }]` on both 3.1 and the delegated 3.2 path.
- Incompatible type and oneOf siblings reproduced satisfiable constants.
- Temporarily restoring the old predicate left the original runtime test green.

This commit is not accepted as final. The next correction must evaluate the constant against all sibling constraints, suppress nullable inference owned by a non-null constant, and add a compile-time predicate contract.

samchon (commented):
Commit reviewed: `06b5e416b0215905b2d5e3b330c2673f80cf6603`

Early-warning result: changes requested and resolved by follow-up commit `5b44622b8843dc1182f120bd42d8f17b12a80445`.

Findings and disposition:

1. High: ordinary single-type `const` schemas bypassed sibling-constraint validation. The follow-up routes both ordinary and mixed constants through the same intersection check and adds incompatible `type`/`minLength` regressions.
2. High: reference DAGs were evaluated exponentially and a 5,000-link chain overflowed the JavaScript stack. The follow-up replaces recursive traversal with an explicit stack and per-schema memoization, with 64-level shared-DAG and 5,000-link tests.
3. Medium: local component references were decoded with `split("/").pop()`, rejecting valid JSON Pointer/URI escapes. The follow-up uses `LlmReference.readOpenApi` and covers `~1`, `~0`, and `%20` names.
4. Medium: a successful mixed constant could leak raw `$recursiveRef` into the normalized constant. The follow-up consumes the keyword after validation and covers the normalized result.

Lead verification:

- Reproduced the ordinary-constant widening and confirmed the new regression fails when the shared validation is bypassed.
- Confirmed the escaped-reference regression fails under the former last-segment lookup.
- Confirmed the shared-DAG regression does not complete promptly when memoization is disabled, while the final evaluator completes in milliseconds.
- Confirmed the recursive-reference regression detects raw keyword leakage when cleanup is removed.
- Rebuilt `@typia/utils`; ran the focused byte/content-encoding test, the complete `json_schema` test selection, and the exact-optional suite successfully.

No unresolved finding from this reviewed commit remains in the current PR head.

samchon (commented):
# Commit review: `5b44622b8`

Reviewed exact commit `5b44622b8843dc1182f120bd42d8f17b12a80445` against cyclic reference graphs, evaluation-order invariance, normalized constant output, deep chains, shared DAGs, escaped references, and recursive-reference cleanup.

Result: changes requested and resolved by follow-up commit `cac4efb165fca342265a6ae36d392892d0128f72`.

Findings and disposition:

1. **High - cached provisional results made cyclic graphs evaluation-order dependent.** A schema reachable through a cycle could be cached before the cycle result propagated, so equivalent `anyOf` or conjunction orderings produced different acceptance results. The follow-up performs an iterative, linear-time composition/reference cycle scan before value evaluation, then evaluates only acyclic graphs with memoization.
2. **Medium - normalized constants retained unrelated source-schema keywords.** The spread-based normalization leaked keys such as numeric bounds, `properties`, `discriminator`, and composition keywords after they had already been consumed as validation constraints. The follow-up reconstructs the constant from an explicit annotation and string-metadata whitelist, preserving extension keys while consuming validation structure.

Lead verification:

- Reproduced order-dependent outcomes with recursive alternatives and with cycles hidden behind false conjunctions.
- Confirmed direct-sibling and equivalent `allOf` constraint spellings now produce the same normalized constant.
- Confirmed numeric, object, discriminator, and composition constraints do not leak into the normalized result.
- Confirmed cyclic and acyclic 64-level DAGs are bounded, and the 5,000-link reference chain remains stack-safe.
- Mutation checks fail when the cycle pre-scan or whitelist reconstruction is bypassed.
- Rebuilt `@typia/utils`; ran the focused byte/content-encoding test, the complete `json_schema` test selection, and the exact-optional suite successfully.

No unresolved finding from this reviewed commit remains in the current PR head.

samchon (commented):
# Commit review: `ebf305ecc`

Reviewed exact claim commit `ebf305ecccfe3f17e8bc00bd0f29c5dd87dd69de`.

Result: clean.

- The commit is intentionally implementation-free; its tree is identical to parent `00872d2952ecdb06c548c83fb4f2a376256b7d9a`.
- The draft PR claims the accepted v14 scope (#2385, #2381, and #2380) while explicitly excluding the next-major work.
- Package versions, tags, publication, and implementation behavior remain unchanged.
- No overlapping implementation was present in the concurrently open dependency update.

No correction is required for the claim commit.

samchon (commented):
# Commit review: `cac4efb16`

Reviewed exact commit `cac4efb165fca342265a6ae36d392892d0128f72`.

Result: clean.

- The iterative composition/reference pre-scan rejects every reachable cycle before memoized value evaluation, removing stale provisional cache results and evaluation-order dependence.
- Both the pre-scan and evaluator use explicit stacks. The scan is linear in reachable vertices and edges, while the acyclic evaluator memoizes shared nodes and remains stack-safe for the 5,000-link regression.
- `$ref` and `$recursiveRef` use the shared OpenAPI reference decoder, including JSON Pointer and percent escapes.
- Normalized constants are reconstructed from common annotations, extensions, and applicable string metadata, so consumed numeric, object, discriminator, composition, and reference keywords cannot leak.
- The regressions cover alternative order, false conjunctions, direct versus `allOf` spellings, cyclic and acyclic shared DAGs, deep chains, raw-key removal, and escaped references.
- `git diff --check`, the complete `json_schema` selection, exact-optional declarations, and focused byte/content-encoding tests pass on the reviewed implementation.

The previous `5b44622b8` findings are resolved and no actionable finding remains.

samchon (commented):
# Self-Review round 8

Reviewed base `00872d2952ecdb06c548c83fb4f2a376256b7d9a` to head `cac4efb165fca342265a6ae36d392892d0128f72` across all 28 changed files.

Result: clean.

The complete diff was rechecked issue by issue and across shared consequence surfaces:

- #2385 uses collection-assigned recursive array and tuple identities consistently in checker, random, stringify, notation, classify, clone, and prune helper declarations and calls. Emitted-code and runtime tests exercise both displaced collection order and control order.
- #2381 reports rejected tags from the declared tag target, coalesces only exact duplicate error records, preserves distinct ownership contexts, and covers direct, factory, package error-fixture, valid-control, single-target, and multi-target paths.
- #2380 preserves byte, custom encoding, media type, enum-constant, extension, conflict, nested, component, document, and cross-version behavior. Public declarations match emitted constant annotations, and constant intersection validation is bounded, stack-safe, order-invariant, and consumes raw constraint structure.
- The website statement matches converter behavior. Test comments follow the required verification headline, rationale, and numbered-scenario structure.
- Closing keywords match the three surviving fixes; package versions and publication state remain frozen.

Mutation evidence detects the former helper identity, wrong diagnostic target/duplication, encoding loss, ordinary and mixed constant widening, reference-decoding, raw-key leakage, recursive-key leakage, cycle-order, unbounded DAG, and deep-stack defects. No sound improvement or meaningful unresolved candidate remains in this immutable head.

Merge remains gated on settled ordinary CI and the frozen prospective-merge integration command set.

samchon:
# Pre-merge integration

## Identity

- Frozen base: `00872d2952ecdb06c548c83fb4f2a376256b7d9a`
- Tested head: `cac4efb165fca342265a6ae36d392892d0128f72`
- Synthetic merge: `dedc99ae792ff9b6c13dfd4236b5550aa982f426`
- Manifest hash: `9638540a85600fb9d3b7402470fb40125598c594`
- Matrix hash: `696d72cfdc6dc08669978111c564896ef837fb69`

## Results

| ID | Command | Result |
| --- | --- | --- |
| I-01 | `pnpm install --frozen-lockfile` | Pass |
| I-02 | `pnpm run format`; `git diff --exit-code` | Pass |
| I-03 | `pnpm run build` | Pass |
| I-04 | `pnpm --filter ./benchmark --fail-if-no-match run build` | Pass |
| I-05 | `pnpm run test` | Pass |
| I-06 | `pnpm run package:tgz` | Pass; six package tarballs staged |
| I-07 | smoke `npm install`; `npm start` | Pass; `typia.is passed` |
| I-08 | ESM `npm install`; build; test | Pass; all package import and runtime checks passed |
| I-09 | MCP `npm install`; build; test | Pass; all protocol checks passed |
| I-10 | declaration `pnpm install`; build; test | Pass; 23 borrowed names, 55 declarations, three witnesses |
| I-11 | website `pnpm run build` | Pass; 74 static pages, Pagefind, and sitemap completed |

The Windows `ttsc-unplugin` cache could not reuse the typia project compile,
so the root build and the tarball prepack each performed the complete project
transformation. This was a local performance limitation, not a correctness
failure.

## Environmental retries

The first I-11 attempt failed before compilation because Go VCS discovery does
not recognize a linked-worktree `.git` file in this layout. It walked to the
invalid unrelated `D:\github\samchon\.git` directory and ran `git status`
there. `GIT_TRACE=1` reproduced that exact failing command. No source change or
`-buildvcs=false` bypass was accepted.

I-11 was therefore repeated at the same synthetic merge in the standalone
clone `D:\github\samchon\typia-integration-standalone-cac4efb16`, whose `.git`
is a directory. Its first invocation then correctly exposed that the standalone
clone did not yet contain I-06's ignored tarballs. The six already-verified
tarballs were copied from the primary integration checkout and the exact
website command was repeated. The WebAssembly build, Rspack worker, examples,
Next.js production build, Pagefind index, and sitemap all passed.

Neither environmental failure implicated repository behavior or required a
tracked-file change. Both temporary checkouts and their ignored generated
artifacts are owned by this campaign and remain scheduled for removal after
the exact post-merge gate.

Issue #2391, opened 2026-09-06
@typia/unplugin generateCodeWithMap can split === into = ==


HeverFarber:
## Bug Report

This is an `@typia/unplugin` source-map reconstruction bug, not a typia generator bug. It cannot be reproduced in the playground because the playground never runs `generateCodeWithMap`.

The same reconstruction logic is present in `@typia/unplugin@12.0.2` (what we hit) and `@typia/unplugin@12.1.1` (latest on npm at the time of this report).

### 📝 Summary

- **Typia Version**: 12.0.2 (`typia` and `@typia/unplugin`; also reproduced against the `generateCodeWithMap` implementation in 12.1.1)
- **Expected behavior**: esbuild receives typia's transformed source, including extra-key checks such as `if ("accountId" === key)`.
- **Actual behavior**: esbuild fails with `Unexpected "=="` because the reconstructed plugin output contains `if ("accountId" = == key)`.

Typia's transform itself is correct. With `cache: true`, the cached transform result contains `"accountId" === key` and does **not** contain `= ==`. The broken token appears only after `generateCodeWithMap` rebuilds the transform output onto the original source via a character-level diff + MagicString.

This is the same family of reconstruction issues as [ryoppippi/unplugin-typia#434](https://github.com/ryoppippi/unplugin-typia/issues/434) / [PR #446](https://github.com/ryoppippi/unplugin-typia/pull/446). That workaround skipped leading whitespace of a deleted hunk (`text.search(/\S/)`) because MagicString mishandled it. Skipping that whitespace is what now inserts a stray space into `===`.

### How the token is corrupted

`transform()` calls `generateCodeWithMap({ source, code, id })`, which:

1. diffs original `source` against typia's `code` (`diff-match-patch-es`)
2. runs `Diff.diffCleanupSemantic(diff)`
3. replays the ops onto `new MagicString(source)`

On a file that both (a) assigns with `= ` (`export const foo = someCall(`) and (b) has a large `typia.misc.prune` expansion containing `"accountId" === key`, we observed this op sequence:

```
equal tail:     " ="
delete:         " route<Req, Res>(\n  {\n    method: \"GET\","
insert:         "== key)\n                return;\n        "
```

The character diff aligned the source assignment `=` with the first `=` of generated `===`. The delete hunk then starts with a space (the space after `=` in `= route...`).

The delete+insert branch does:

```js
const firstNonWhitespaceIndexOfText = text.startsWith("\n") ? 0 : text.search(/\S/);
[...]

PR #2395, opened 2026-09-17
fix(unplugin): stop splitting tokens when rebuilding the transform output

samchon:
Fixes #2391.

## Intent

`@typia/unplugin` rebuilds the transform output onto the original source with a
character diff so it can emit a source map back to the authored file. When a
deleted hunk was followed by an insert, the replacement started at the hunk's
first non-whitespace character instead of at the hunk itself, leaving the
original whitespace standing in front of the generated text.

That leftover whitespace is welded onto the next generated token. A source
assignment `= ` aligned against the first `=` of a generated `===` produced
`if ("accountId" = == key)`, which esbuild rejects with `Unexpected "=="`.

## Scope

The reported `= ==` is one witness, not the defect. The same skip produced
ryoppippi/unplugin-typia#434, where a deleted hunk's leading newline broke a
string literal across lines; that fix only special-cased `\n`, and it shipped
with an empty test file. Both are the same cause.

Measured on the pinned `magic-string@0.30.17` and `diff-match-patch-es@1.0.1`
over 20,000 generated diff shapes, against the invariant the replay exists to
serve — the rebuilt string must equal the transform output verbatim:

| reconstruction | misreconstructed | threw |
| --- | --- | --- |
| shipped (skip leading whitespace) | 1,340 / 20,000 | 0 |
| whole-range replacement | 0 / 20,000 | 0 |

So the shipped plugin misrebuilds ~6.7% of diff shapes. `= ==` is the loud
subset; the rest emit code that parses and is silently wrong.

The skip has no justification beyond a `maybe bug of magic-string` comment, and
removing it never made `magic-string` throw. This replaces each deleted hunk
over its whole range and drops the `\n` special case the skip needed.

## Changes

- `packages/unplugin/src/core/magicString.ts` (new): `buildMagicString`, the
  diff replay, extracted from the `unpluginFactory` closure. It is a pure
  function with no plugin dependencies, so its in-source tests can run without
  pulling in the `@typia/core` module graph — which is why the loop had no
  coverage before.
- `packages/unplugin/src/core/index.ts`: `generateCodeWithMap` now calls it.

Five in-source tests pin the invariant, including the minimal witness
(`"= r"` -> `"==="`, which the shipped logic rebuilds as `"= =="`), the #434
newline case, the unchanged-source case, and a deterministic 2,000-shape sweep.

## Verification

Run locally in a `pnpm install --filter "./packages/unplugin..."` worktree on
this branch:

- `vitest run src/core/magicString.ts` — 5 passed.
- `vitest run` (whole package) — 15 passed across 3 files.
- Reverted `buildMagicString` to the shipped logic and re-ran: 2 failed with
  `expected '= ==' to be '==='`, confirming the tests fail without the fix.
- `tsdown` — build succeeded; `dist/` contains no `firstNonWhitespaceIndexOfText`.
- `prettier --write` on both changed files.

`tsc --noEmit` reports only pre-existing errors from workspace packages that the
filtered install leaves unbuilt (`@typia/interface`, `@typia/core`,
`@typia/transform`); none are in the changed files.

## Deferred

- In-source test blocks reach `dist/core.js`: `tsdown.config.ts` sets
  `define: { "import.meta.vitest": "undefined" }` but it is not applied. This
  predates the change — published `@typia/unplugin@12.2.0` already ships two
  such blocks — and this branch adds a third, +2.6 kB. Not addressed here.
- The same defect is still live upstream in `ryoppippi/unplugin-typia`.

## Branch note

`legacy/v12.2.0` had been deleted from the remote; the published
`@typia/unplugin@12.2.0` was built from it. It was recovered from a stale
remote-tracking ref in a second local clone and verified against the published
tarball's embedded `sourcesContent`, which matches the recovered tree byte for
byte. This pull request targets the restored branch.


samchon (commented):
## Self-Review

Solo review, whole surface, three complete rounds. Two improvements were applied,
each restarting a fresh full round; round 3 produced nothing that survived
verification.

### Surface

Every changed file and hunk in `legacy/v12.2.0...2b20dabaca`:

- `packages/unplugin/src/core/index.ts` — import removals, one call site.
- `packages/unplugin/src/core/sourcemap.ts` — new module and its in-source tests.

### Rounds

**Round 1** — reviewed the extracted replay and the call site. Found a naming
divergence: every other file under `packages/unplugin/src` is a lowercase
concern noun (`cache.ts`, `options.ts`, `types.ts`, `typia.ts`, `utils.ts`), so
the new `magicString.ts` broke the package's pattern. Renamed to `sourcemap.ts`
in `a7a75b64c8`.

**Round 2** — reviewed the new state. Found an unsupported claim in a code
comment: the sweep test said the replaced logic "missed ~7% of these", but that
rate came from a different fuzz configuration, not from this sweep. Measured the
sweep's exact parameters against the replaced logic — 132 of 2,000 cases
misrebuilt, 32 of them emitting `= ==` — and replaced the comment with those
figures in `2b20dabaca`.

**Round 3** — reviewed the current state. Nothing found.

### Consequence surface

Traced beyond the reported witness, each item verified rather than argued:

- **Offset-collision risk.** Removing the skip makes a deleted hunk's
  replacement start where a preceding standalone insert did its `prependLeft`.
  Checked whether that adjacency can occur: over 20,000 generated diff shapes,
  an insert not preceded by a delete is never followed by a delete
  (0 occurrences) — `diff-match-patch` emits delete before insert — so the two
  calls never contend for one offset. 0 mismatches, 0 throws in the same run.
- **The `maybe bug of magic-string` premise.** Not reproducible.
  Whole-range replacement never threw across 40,000 replays on the pinned
  `magic-string@0.30.17`.
- **Boundaries.** Empty `code`, empty `source`, identical input
  (`hasChanged()` stays `false`), and a single-character swap all rebuild
  verbatim.
- **Line endings.** CRLF sources rebuild verbatim both when the transform
  normalises to LF and when it preserves CRLF — relevant because this branch's
  working tree is CRLF on Windows checkouts.
- **Non-ASCII.** CJK identifiers, emoji, and a diff landing inside a surrogate
  pair all rebuild verbatim; offsets now come straight from the diff rather than
  from a whitespace scan, so no new code-unit hazard is introduced.
- **Source map.** `generateMap` still yields non-empty mappings with
  `sourcesContent` intact. Mapping granularity does shift slightly: a
  replacement now covers the hunk's leading whitespace instead of starting after
  it, so a generated token can attribute to a marginally earlier source column.
  That is the cost of emitting correct code, and it is strictly smaller than the
  corruption it replaces.
- **Built artifact, not just source.** `dist/core.js` contains no
  `firstNonWhitespaceIndexOfText`. Extracted the compiled `buildMagicString`
  from the bundle and executed it: `"= r"` -> `"==="`, the #434 newline witness
  -> `"a"`, and the realistic `typia.misc.prune` shape rebuilds verbatim.
- **Stale references.** No remaining `magicString` reference outside the
  exported function name.
- **Package wiring.** The new module is not an entry (`entry: ["src/*.ts"]`),
  needs no export-map or `files` change, and `tsdown`'s `unused: { level:
  "error" }` check passes.

### Verification

Run on `2b20dabaca` in a `pnpm install --filter "./packages/unplugin..."`
worktree:

- `vitest run` — 15 passed across 3 files.
- `tsdown` — build succeeded.
- Reverted `buildMagicString` to the replaced logic and re-ran: 2 failed with
  `expected '= ==' to be '==='`. The tests fail without the fix.
- `prettier --write` on both changed files.

### Not run

- **End-to-end plugin run against a real typia transform.** The filtered install
  leaves `@typia/core` and `@typia/transform` unbuilt, so no esbuild-level
  reproduction was executed here. The evidence is one level below that: the
  invariant holds for the compiled function over every generated shape, and the
  reporter verified the esbuild symptom against this same loop.
- **`tsc --noEmit`** reports only pre-existing errors from those unbuilt
  workspace packages; none are in the changed files.

### Carried forward, not addressed

- In-source test blocks reach `dist/core.js` because `tsdown.config.ts`'s
  `define: { "import.meta.vitest": "undefined" }` is not applied. Published
  `@typia/unplugin@12.2.0` already ships two such blocks; this branch adds a
  third, +2.6 kB. It is a build-configuration defect that predates this change
  and affects the package either way, so it belongs in its own change rather
  than a drive-by here.
- The same defect is still live upstream in `ryoppippi/unplugin-typia`.

PR #2396, opened 2026-09-17
ci(release): publish the v12 line under the legacy dist-tag

samchon:
## Intent

Give the v12 maintenance line a publishing path that cannot reach the rest of
the repository, so patches on this branch ship from CI instead of by hand.

## Why there is nothing here today

`ci(release): drop the release workflow from the v12 maintenance branch`
removed this branch's `release.yml` deliberately. That file fired on
`push: tags: "*"`, ran `pnpm run package:latest`, and deployed this branch's
website to `gh-pages`. A tag cut here would have pulled the npm `latest`
dist-tag down from the current major and replaced the published documentation.
Removing it was the right call; it just left no way to publish.

The cost is visible on the registry. This repository authenticates to npm with
trusted publishing over OIDC — no token is stored in any workflow — so only a
workflow run can produce a provenance attestation:

| version | provenance | source recorded by npm |
| --- | --- | --- |
| `@typia/unplugin@12.1.1` | yes | the old workflow |
| `@typia/unplugin@12.2.0` | none | a local tarball under `C:\Users\...` |

## Scope

A tag push runs the workflow file present at the pushed ref, so a `v12.*` tag
on this branch executes this file and `master`'s release workflow is never
involved. Three properties keep that containment:

- **Trigger.** `v12.*`, never `v*`. `v120.0.0` does not match; a prerelease
  like `v12.3.0-rc.1` does, and the guard rejects it.
- **Dist-tag.** `package:legacy` hardcodes `--tag legacy`. It is not an input,
  a variable, or a workflow argument, so nothing can redirect it.
- **Blast radius.** One `publish` job. No release-notes job and no website job,
  so the workflow cannot touch `gh-pages` or this repository's releases.
  `contents` permission is `read`.

`release-guard.cjs` refuses any context that could mean something else: the tag
must be a stable `v12.x.y`, and every `packages/*` manifest must already carry
that exact version, so a forgotten bump fails the job instead of shipping a
mismatched version set.

`package:legacy` builds every package before publishing any of them. These
packages are version-coupled, so a build failing midway through
`pnpm -r publish` would leave a partial set on npm after the dist-tag had
already moved. It also carries `--provenance`, which means a local run of this
script fails outside CI — deliberate, given how 12.2.0 shipped.

## Verification

The guard was exercised directly against this working tree, where every package
sits at 12.2.0:

| `GITHUB_EVENT_NAME` | `GITHUB_REF_NAME` | exit | message |
| --- | --- | --- | --- |
| `push` | `v12.2.0` | 0 | `9 packages are at 12.2.0.` |
| `push` | `v12.2.1` | 1 | `Package versions must all be 12.2.1:` |
| `push` | `v13.0.0` | 1 | `release tag must be a stable v12.x.y` |
| `push` | `v12.2` | 1 | `release tag must be a stable v12.x.y` |
| `workflow_dispatch` | `v12.2.0` | 1 | `unsupported release event: workflow_dispatch` |

`release.yml` parses as YAML and resolves to one `publish` job with the six
steps above. `prettier --write` on `release-guard.cjs`; `prettier.config.js`
forces `parser: "typescript"` repository-wide, so the YAML and JSON files here
are not prettier-managed and the repository's `format` script does not cover
them.

## Not verified here

The npm trusted-publisher configuration binds to a repository and a workflow
**filename**. This file is named `release.yml`, matching the workflow that
published `12.1.1` with provenance, which is what makes the existing
configuration apply. That setting lives in npm's web UI and cannot be read from
the CLI, so the first tagged run is what confirms it. If OIDC is rejected, the
failure is an authentication error at the publish step, before anything is
published.


samchon (commented):
## Self-Review

Solo review, whole surface, two complete rounds. One improvement was applied and
restarted the round; round 2 produced nothing that survived verification.

### Surface

`.github/workflows/release.yml`, `.github/workflows/release-guard.cjs`,
`package.json` (one added script). 116 insertions, no deletions.

### Rounds

**Round 1** — found that `master`'s release workflow pins `actions/checkout` by
digest while leaving its other actions on tags, precisely because that job holds
the npm publishing identity. This workflow holds the same identity through
`id-token: write` and used a floating `@v4`. Pinned to the same digest in
`292ae640b7`.

**Round 2** — nothing found.

### Consequence surface

- **Containment of the failure this branch already suffered.** The deleted
  workflow's three hazards are each closed by construction: the trigger is
  `v12.*` rather than `*`; the dist-tag is hardcoded inside `package:legacy`
  with no input or variable able to redirect it; and there is no release-notes
  or website job, with `contents` reduced to `read`, so `gh-pages` and the
  repository's releases are unreachable.
- **`master` cannot be caught by this tag.** A tag push runs the workflow file
  at the pushed ref. A `v12.*` tag on this lineage reaches this file only.
  Conversely, a `v12.*` tag cut on `master` would run `master`'s workflow, whose
  own guard rejects it because `master`'s manifests are not at 12.x.
- **Guard behaviour, executed rather than argued.** Against this tree at 12.2.0:
  `push`/`v12.2.0` exits 0 with `9 packages are at 12.2.0.`; `push`/`v12.2.1`
  fails on version mismatch; `push`/`v13.0.0`, `push`/`v12.2` and
  `push`/`v12.3.0-rc.1` fail on tag shape; `workflow_dispatch` fails on event.
- **The `deploy` check that appeared on this pull request.** Traced rather than
  assumed: `website.yml` lists `package.json` in its `pull_request` path filter,
  so adding a script woke it. Its Deploy step is guarded by
  `github.event_name == 'push' && github.ref == 'refs/heads/master'`, and its
  `push` trigger is limited to `branches: master`, so it builds here and can
  deploy neither from this branch nor from a tag push. It passed.
- **The version bump this workflow is built to serve will not dirty the
  lockfile.** Every internal dependency uses `workspace:^`, so `pnpm-lock.yaml`
  records `link:../packages/*` and contains zero occurrences of the version
  string. The 12.1.1 -> 12.2.0 bump likewise left the lockfile untouched, so a
  frozen-lockfile install in CI stays green.
- **Partial-publish exposure.** `package:legacy` builds all nine packages before
  publishing any, so a build failure aborts before the dist-tag moves. A
  registry failure partway through `pnpm -r publish` remains possible; that
  exposure is identical to `master`'s workflow and is not introduced here.
- **Formatting ownership.** `prettier.config.js` forces `parser: "typescript"`
  repository-wide and the `format` script covers `.ts` only, so the YAML and
  JSON in this change are not prettier-managed. `release-guard.cjs` was
  formatted.

### Verification

`release.yml` parses as YAML and resolves to one `publish` job with six steps.
The guard matrix above was executed directly. All seven checks on this pull
request pass, including `deploy`.

### Not verified here

The npm trusted-publisher binding includes the workflow **filename**. This file
is named `release.yml`, matching the workflow that published `12.1.1` with a
provenance attestation. That binding lives in npm's web UI and is not readable
from the CLI, so the first tagged run confirms it. A rejection surfaces as an
authentication failure at the publish step, after the build and before any
package is published.

PR #2399, opened 2026-09-18
feat(llm): typia.llm.evaluation<T>() for evaluation-model questions

samchon:
## Intent

Implement #2398: `typia.llm.evaluation<T>()`, which turns an object type of decisions into the provider-neutral evaluation question map (Vercel AI SDK `experimental_evaluate` / `EvaluationModelV4`, TypeSafe Jev), plus `validate(answers): IValidation<T>` that checks a provider's flat answer map and folds it back into `T`.

## Scope

This draft claims the complete accepted cycle. The cycle has one issue: #2398.

| Area | What changes |
| --- | --- |
| `@typia/interface` | `ILlmEvaluation<T>` and `ILlmEvaluation.IQuestion` |
| `typia` | `llm.evaluation` declaration and its runtime validate/convert helper |
| Native Go transform | programmer, transformer, registration, and compile-time diagnostics |
| Tests | Go, `test-typia-schema`, `test-error`, `test-interface` |
| Docs | website guide `llm/evaluation.mdx` |

## Status

- **Verification:** pending. No implementation exists yet; this pull request currently holds only an empty claim commit.
- **Deferred:** no items are deferred at claim time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2402, opened 2026-09-18
fix: tag diagnostics, LLM parameter descriptions and keys, items-less arrays, and a symmetric test oracle

### LINKED ISSUE #2400: A type tag is silently dropped when every tag on the type is malformed
## Problem

When **every** type tag on a type is malformed, typia drops the tags silently instead of reporting them. The same malformed tag next to one valid tag is reported. So whether a tag mistake is diagnosed depends on whether some other tag happens to sit beside it.

A silently dropped tag means the constraint the author wrote is not enforced anywhere: not by `is` / `assert` / `validate`, not in JSON or LLM schemas, not by `random`, and not in `llm.evaluation` thresholds.

## Evidence

Reproduced on `master` `63c143d`, in the `tests/debug` workspace with `ttsx`:

```ts
// sole malformed tag: the value is not a literal
typia.is<number & tags.Minimum<number>>(-100); // compiles, returns true
```

```ts
// the same malformed tag beside a valid one
typia.is<number & tags.Minimum<number> & tags.Maximum<10>>(-100);
// error TS(typia.is): typia transform error: unsupported type detected
// - number & Minimum<number> & Maximum<10>
//   - the property ["typia.tag.value"] must be a literal type or undefined value.
```

`typia.llm.evaluation` hits the same path: `boolean & tags.Probability<number>` compiles as a plain boolean with the default threshold (found while implementing #2398).

**Root cause:** `MetadataTypeTagFactory.Analyze` in `packages/typia/native/core/factories/MetadataTypeTagFactory.go`.
- The per-object loop validates each `typia.tag` object. A malformed one is reported into the local `messages` slice and excluded from `filtered`.
- When `filtered` ends up empty, the function returns `[]IMetadataTypeTag{}` **before** the `if len(messages) > 0` block. That block is the only place the messages reach `props.Errors`.
- With at least one valid tag, `filtered` is non-empty, the function reaches the flush, and the same messages are reported.

## Consequence surface

- **Shared path:** every feature that reads type tags goes through this analyzer. That covers the validators, JSON / LLM / protobuf schemas, `random`, and `llm.evaluation`.
- **Messages affected:** every property-validation message produced before `filtered` is built. That covers a malformed `target`, `kind`, `value`, `validate`, or `exclusive`, such as a non-literal value or kind.
- **Not tags at all, so out of scope:** a `"typia.tag"` object that lacks an essential field (`target`, `kind`, `value`) or is not optional. `iterate_metadata_intersection_is_type_tag` does not classify it as a tag, the same way the legacy TypeScript `MetadataTypeTagFactory.is` did. It is treated as an ordinary brand, or reported as a "nonsensible intersection" when required. This is long-standing design and outside this defect. The analyzer's "must have at least three properties" message is defensive for that reason.
- **Misleading message, now visible:** a malformed `validate` (for example the template literal that `tags.Minimum<number>` produces) was reported as `["typia.tag.target"] must be one of ... 'object` (wrong property, missing quote). It must name `validate`.
- **Not affected:** target-mismatch and exclusivity messages are produced after `filtered` is built, so they are already reported.
- **Comment tags:** `@minimum` and friends use `MetadataCommentTagFactory`, which collects and reports its messages without an early return, so they are not affected.
- **Compatibility:** code that compiles today with a sole malformed tag will start failing at compile time. That is the intended correction, because the constraint it asked for was never applied.

## Invariant

A malformed type tag is reported whether or not other valid tags accompany it. The report must carry the same messages and the same error naming as the accompanied case.

- **Positive twins:** a sole malformed tag is diagnosed, both through a native Go test and through `tests/test-error` fixtures. Cover a non-literal `value` on a built-in tag and a non-literal `kind` on a custom `TagBase`, and check that the malformed `validate` is named correctly.
- **Negative twin:** a valid sole tag still compiles and behaves as before.
- **Regression:** the accompanied case keeps its current message.
- **Broader check:** `pnpm test` stays green, including the generated matrices. Any existing fixture that relied on the silent drop is a finding in its own right.

## Coordination

Found during #2398 / #2399.

### LINKED ISSUE #2401: Test suites assert with a one-way equality that misses keys, dates, maps, and sets
## Problem

Every TypeScript test suite asserts equality with `@nestia/e2e`'s `TestValidator.equals`. About 3,400 call sites use it across `test-typia-schema`, `test-utils`, `test-vercel`, `test-langchain`, `test-mcp`, `test-typia-exact-optional`, `test-utils-automated`, and `test-feature-identity`. It is not an equality check, and the tests built on it prove less than they appear to.

## Evidence

The comparison is `json_equal_to` (`@nestia/e2e/src/internal/json_equal_to.ts`, version 10.0.2 as pinned here; 13.0.4, the latest, is unchanged). For objects it walks `Object.keys(x)` only. Every result below comes from running `TestValidator.equals` directly:

| Comparison | Result | Correct |
|---|---|---|
| `{ a: 1 }` vs `{ a: 1, b: 2 }` | passes | fails, `b` is missing on one side |
| `{ a: 1, b: 2 }` vs `{ a: 1 }` | throws | throws |
| `new Date(0)` vs `new Date(1e12)` | passes | fails |
| `new Map([["a", 1]])` vs `new Map([["b", 2]])` | passes | fails |
| `new Set([1])` vs `new Set([2, 3])` | passes | fails |
| `Object.getPrototypeOf(Object.create(Array.prototype))` vs `Object.prototype` | passes | fails |
| `NaN` vs `NaN` | throws | passes |

- **Keys missing on one side:** a test that passes the generated value first accepts any key the generated value **drops**. A test that passes the expected literal first accepts any key the generated value **adds**. Both argument orders appear throughout the suites.
- **Objects without enumerable own keys:** `Date`, `Map`, `Set`, and prototypes have none, so any two of the same kind compare equal. Tests over date, map, and set values therefore assert nothing about the content.
- **`NaN`:** a correct `NaN` result can only be asserted by avoiding `equals`.

Concrete instances seen so far:
- **#2399:** a mutation that dropped the question text of every boolean question passed the one-way assertion. It was caught only after that PR switched to a two-way helper.
- **`tests/test-utils/src/features/llm/schema/test_llm_schema_reserved_references.ts:119-135`:** deep-compares `Object.getPrototypeOf(...)` with `Object.prototype`. That passes for any prototype without enumerable keys, so the null-prototype guarantee it names is not proven.

**Ownership:** the root cause is upstream in `@nestia/e2e`. This repository owns its own test oracle, the call sites, and the risk that a new test reintroduces the one-way call.

## Consequence surface

- **Every assertion in the listed suites:**
  - key-set asymmetry in both argument orders;
  - `Date`, `Map`, and `Set` values;
  - prototype and identity checks written as deep comparisons;
  - `NaN`.
- **Behavior to keep:** keys whose value is `undefined` and function-valued keys are ignored today. This is JSON-like semantics that many schema and controller assertions rely on. The corrected oracle must state what it keeps.
- **Real mismatches:** once the oracle is correct, some existing assertions may fail. Each failure is either a wrong expectation or a product bug hidden until now, and each must be resolved on its evidence, not by weakening the oracle.

## Design

- **`TestEquality.equals`:** the default assertion. It is symmetric and exact in both key sets.
- **`TestEquality.subset`:** an explicit partial check. It visits only the object keys the expected value declares, at any depth, while array lengths and date, map, and set content must still match. #2350 kept expected-first one-way calls on purpose where an assertion pins some fields of a larger result; `subset` states that intent by name instead of relying on argument order.
- **`TestEquality.difference`:** lists differing paths, for tests that assert on the comparison itself.

## Invariant

- **Correct oracle:** every structural-equality assertion in `tests/*` uses one shared oracle that is symmetric in its arguments. It compares `Date` by time value, `Map` by entries, and `Set` by members, and treats `NaN` as equal to `NaN`.
- **Identity checks:** prototype and identity checks are written as identity comparisons.
- **Regression guard:** a repository check rejects new `TestValidator.equals` calls in tracked test sources, so the class cannot return one test at a time.

- **Oracle unit tests:** one case per row of the table above, plus the kept semantics (`undefined` keys, functions), each in both argument orders.
- **Call sites:** every former `TestValidator.equals` site is migrated. The full `pnpm test` stays green, with every newly exposed mismatch resolved and reported.
- **Guard:** the repository check fails on a planted `TestValidator.equals` call and passes on the tree.

## Coordination

- **#2350:** the same one-way flaw, closed by triaging call sites and pinning which comparison shapes were safe. The trap returned in #2399 because every new call site had to know that rule. This issue fixes the oracle instead, so no call site needs to know it.
- Found during #2398 / #2399. Its suite-local `_equalsExactly` helpers are subsumed by the shared oracle.
- **Upstream:** `@nestia/e2e`'s `json_equal_to` should be fixed at its owner. That is outside this repository.

### LINKED ISSUE #2403: A null in a type tag's schema content is dropped from the emitted schema
## Problem

A `null` inside a type tag's schema content never reaches the emitted schema:
- **`tags.Example<null>`** declares `null` as a valid example, but the emitted JSON schema has no `example` at all.
- **A `null` object field**, such as the `name` of `tags.Example<{ name: null }>`, is dropped from the example object.
- **The same loss applies to** `tags.Examples` and `tags.JsonSchemaPlugin` content.

The schema then shows a different example than the author wrote, or none.

## Evidence

Reproduced in `tests/debug` with the fix for #2400 applied. Before that fix, `Example<null>` did not compile at all (see *Coordination*).

```ts
interface IMember {
  nickname: (string & tags.Example<null>) | null;
  flag: boolean & tags.Example<false>;
}
typia.json.schema<IMember>();
// nickname: { oneOf: [{ type: "null" }, { type: "string" }] }  <- no example
// flag:     { type: "boolean", example: false }                <- example kept
```

**Root cause, in two parts:**
- **The factory encodes JSON `null` as Go `nil`:** `MetadataTypeTagSchemaFactory` (`packages/typia/native/core/factories/MetadataTypeTagSchemaFactory.go`) returns `nil` for a `null` literal and stores it in the schema map.
- **The writer drops every `nil` map value:** `LiteralFactory` (`literalFactory_writeObject`, `LiteralFactory_OrderedObject`) skips nil-like map values. Many schema writers rely on that to mean "absent", so an explicit JSON `null` cannot be told apart from "no value".
- **Arrays are unaffected:** a `null` element is kept, because `writeArray` writes `nil` as `null`.

## Consequence surface

- **Consumers of tag schema maps:**
  - JSON schema plugin merge (`json_schema_plugin.go`, `json_schema_template.go`), and through it JSON / LLM / OpenAPI schemas;
  - `random` (`minItems`);
  - `check_exclude` (`not.enum`, an array, so unaffected);
  - `llm.evaluation` (`tags.Constant` descriptions);
  - `typia.reflect` metadata JSON, through `encoding/json`.
- **Scope of the loss:** only object-property values lose `null`. Tuple and array elements keep it.

## Invariant

A JSON `null` in a tag's schema content is emitted as `null`. This holds both through `LiteralFactory` and through `encoding/json`. Absent values stay absent.

- **Positive:** `Example<null>`, a `null` field in an object example, and a `null` member of `Examples` all appear in the JSON schema.
- **Negative twin:** absent optional schema fields are still omitted.
- **Regression:** existing tag schema outputs are unchanged, and `pnpm test` stays green.

## Coordination

- Surfaced while fixing #2400. That fix exposed that tuple and `null` example values were rejected by the tag value check, although `Example` and `Examples` declare them. The same pull request (#2402) widens the value check.

### LINKED ISSUE #2404: Downgrading an array schema without items still emits an unconstrained schema
## Problem

#2392 made every upgrader and downgrader emit the open `items: {}` form for an array schema that omits `items`. The **downgraders** still do not. `OpenApiConverter.downgradeSchema` turns `{ type: "array" }` into `{}` for Swagger v2.0, OpenAPI v3.0, and OpenAPI v3.1 alike. That is "any value" instead of "any array".

## Evidence

Reproduced on `master` `63c143d`:

```ts
OpenApiConverter.downgradeSchema({
  version, // "2.0" | "3.0" | "3.1"
  components: {},
  downgraded: {},
  schema: { type: "array" } as any,
});
// => {} for every version; expected { type: "array", items: {} }
```

**Root cause:**
- #2392 added `schema.items === undefined ? {} : ...` **inside** each downgrader's `OpenApiTypeChecker.isArray(schema)` branch, in `OpenApiV3Downgrader`, `OpenApiV3_1Downgrader`, and `SwaggerV2Downgrader`.
- `OpenApiTypeCheckerBase.isArray` requires `items !== undefined`, so an array without `items` never enters that branch, and the guard is dead code.
- The schema also matches no other checker, so `visit` emits nothing for it, and the downgrade collapses to `{}`.

**Why the regression test passed:**
- `test_json_schema_downgrade_items_omitted` asserts `TestValidator.equals("v2.0 open array", v20, { type: "array", items: {} })`, with the actual value first.
- The actual was `{}`, and the one-way `TestValidator.equals` walks only the first argument's keys, so there was nothing to compare. #2401 exposed this.

## Consequence surface

- **Every downgrade path that meets an items-less array:** `OpenApiConverter.downgradeSchema`, `downgradeDocument`, and every schema nested in a downgraded document (properties, items, `oneOf` members, components).
- **What it breaks:** a consumer of the downgraded document loses arrayness, so generated SDKs and validators accept any value.
- **Upgraders are unaffected:** their items-omitted test passes under a symmetric comparison.

## Invariant

Downgrading an array schema without `items` (and without `prefixItems`) emits `{ type: "array", items: {} }`, keeping the array's other keywords, for every target version.

- `test_json_schema_downgrade_items_omitted` passes under a symmetric comparison for v2.0, v3.0, and v3.1.
- A nested items-less array inside an object is also downgraded to the open array.
- Existing downgrade tests stay green.

## Coordination

- Defect in the #2392 fix, found through #2401.
- Implemented in #2402.

### LINKED ISSUE #2405: Native LLM parameters omit the reference description cascade the converter applies
## Problem

When the parameters type is a named reference, `typia.llm.parameters<T>()` and `typia.llm.application<C>()` emit a different root `description` than `LlmSchemaConverter.parameters` does for the same type. The utils converter applies `JsonDescriptor.cascade` to the root: the type's own description plus its namespace parents' descriptions, or `Current Type: {@link Name}` when there is none. The native transform never ported that cascade, so its root carries only the type's own description, or none.

The parity suites `test_llm_parameters_parity_converter_object`, `test_llm_parameters_parity_converter_strict`, and `test_llm_application_parity_converter` exist to hold the two equal. They passed only because the native output came first in a one-way `TestValidator.equals` and lacked the key (#2401).

## Evidence

Reproduced on `master` `63c143d` (difference computed symmetrically):

```ts
interface IParams { /** The article to create. */ input: IBbsArticle.ICreate; }
typia.llm.parameters<IParams>().description;          // undefined
LlmSchemaConverter.parameters({ ...json.schemas<[IParams]>() }).value.description;
// "Current Type: {@link IParams}"
```

- **Only difference:** the root `.description` is the sole difference between the two outputs, including for namespaced `$defs`.
- **Root cause:**
  - `LlmSchemaConverter.parameters` sets the root description through `JsonDescriptor.cascade({ escape: true, key })` whenever the input schema is a reference (`packages/utils/src/converters/LlmSchemaConverter.ts`).
  - Neither native root builder has an equivalent: `llmParametersProgrammer_convert_parameters` (`LlmParametersProgrammer.go`) and `llmApplicationProgrammer_convertParameters` (`LlmApplicationProgrammer.go`).
- **Stale documentation:** the website's "Namespace inheritance" section in `website/src/content/docs/llm/strategy.mdx` shows the cascade applied to a *property* reference (`func.parameters.properties.input.description`). Neither implementation does that, because property references now point into `$defs` rather than being inlined.

## Consequence surface

- **Native paths:** `llm.parameters`, `llm.application` / `llm.controller` (function parameters whose single object parameter is a named type), and `llm.structuredOutput`, which shares `LlmParametersProgrammer`.
- **What the LLM loses:** the parent-namespace prose that `HttpLlm`, which goes through the converter, provides for the same type.
- **Other paths:** the strict-mode and nullable variants take the same path.

## Invariant

- **Parity:** for a named parameters type, the native root `description` equals `JsonDescriptor.cascade(escape: true)`. That is the type's own description, followed by each described namespace parent, or `Current Type: {@link Name}` when there is no description. It must match `LlmSchemaConverter.parameters` exactly.
- **Reachable parents only:** a namespace parent contributes only when its schema is in the same components, because the converter can read only what the collection holds. `typia.llm.parameters<IMember.ICreate>()` names `IMember` only when `ICreate` references it.
- **Anonymous roots:** an anonymous root keeps its own description.
- **Docs:** the guide describes the behavior that actually exists.

- The three parity suites pass under a symmetric comparison.
- Native tests pin all three root cases:
  - an undescribed type;
  - a described type;
  - a namespaced child with a described parent.
- `strategy.mdx` shows real output.

## Coordination

- Found through #2401; implemented in #2402.

### LINKED ISSUE #2406: LLM parameters panic when the root component key contains a slash
## Problem

`typia.llm.parameters<T>()` and `typia.llm.structuredOutput<T>()` panic the compiler when the root type's component key contains a slash, and `typia.llm.application<C>()` leaves such a parameters object as an unresolved `$ref`.

## Evidence

Reproduced on `master` `63c143d`:

```ts
interface IVariant<T extends string> { kind: T; value: number; }
typia.llm.parameters<IVariant<"A/B">>();
// panic: Unreachable code. Failed to find the object schema.
//   LlmParametersProgrammer.WriteParameters (LlmParametersProgrammer.go:87)
```

- **Key:** a generic argument `"A/B"` makes the LLM component key `IVariantA/B`. LLM keys stay raw. Only local `$defs` references encode them as JSON Pointer tokens, which `test_llm_parameters_json_pointer_references` pins.
- **Root cause:** `WriteParameters` and `llmParametersProgrammer_dereference_schema` (`packages/typia/native/core/programmers/llm/LlmParametersProgrammer.go`) read the key as the text after the last `/` of the `$ref`. That gives `B`, which matches no component. `llmSchemaProgrammer_ref_key` in the same package already strips the `#/components/schemas/` prefix correctly.
- **Found by:** the early-warning pass on the #2405 commit in #2402. The new root description reads the key by prefix, so the two lookups in one flow disagreed.

## Consequence surface

- **`llm.parameters` and `llm.structuredOutput`:** a compiler panic, surfaced as a transform failure.
- **`llm.application` and `llm.controller`:** a function whose single parameter is such a type keeps `$ref` as its parameters, which is not an `ILlmSchema.IParameters`.
- **`llm.parse` and `llm.coerce`:** these share `WriteParametersExpression`, so they panic the same way.
- **Unaffected:** JSON schema output escapes component names (`IVariantA_x2F_B-…`), so the native OpenAPI 3.0 downgrader's last-slash lookup never sees a slash.

## Invariant

- A parameters root reached through `$ref` dereferences by its full component key, whatever characters that key contains.

- A regression test generates `llm.parameters`, `llm.structuredOutput`, and `llm.application` parameters for `IVariant<"A/B">`, and asserts each is the dereferenced object.

## Coordination

- Found through #2405; implemented in #2402.

### LINKED ISSUE #2407: An items-less array schema escapes validation and LLM conversion
## Problem

An emended array schema without `items` is still misread by every consumer except the converters. The worst case is that `OpenApiValidator` accepts **any** value for such a property.

## Evidence

Reproduced against `@typia/utils` built from #2402 at `3bb59e4`, after #2404 fixed the downgraders:

```ts
const tags = { type: "array" };            // no items
OpenApiValidator.validate({ components: {}, required: true,
  schema: { type: "object", properties: { tags }, required: ["tags"] },
  value: { tags: "nope" } }).success;       // true
LlmSchemaConverter.schema({ config: { strict: true }, components: {}, $defs: {},
  schema: { type: "array", minItems: 2 } }); // { type: "array", minItems: 2 }
OpenApiTypeChecker.covers({ components: {}, x: { type: "array", items: {} }, y: tags }); // false
```

- **Validator:** `OpenApiStationValidator` matches no branch and falls through to `return true`.
- **LLM converter:** the array branch is skipped. Output keeps no `items`, and under `strict`, `minItems` is never shifted into the description.
- **`covers`:** denies the equivalent open array in both directions.
- **HTTP migration:** `HttpMigrateRouteComposer` rejects such a `pipeDelimited` or `spaceDelimited` query parameter with "requires an array or object schema", and `isNotObjectLiteral` misclassifies it.
- **Root cause:** `OpenApiTypeChecker.isArray` requires `items`. #2392 kept that type contract and made the upgraders emit `items: {}`, and #2404 restores the open form in the downgraders. The other walkers never took the same step, although a hand-written or externally parsed emended schema reaches them too.
- **Found by:** the early-warning pass on the #2404 commit in #2402.

## Consequence surface

- `OpenApiValidator.validate`, and the naming of its expected types;
- `LlmSchemaConverter.schema`, `parameters`, and the strict pre-validation, and through them `HttpLlm.application`;
- `OpenApiTypeChecker.covers` and `escape`, which feed the `oneOf` validator's discrimination;
- `HttpMigration.application` query-parameter checks.

**Out of scope:** `ILlmSchema` walkers. LLM schemas are typia's own output, and the converter now always emits `items`.

## Invariant

- Every emended-schema walker reads `{ type: "array" }` without `items` or `prefixItems` as the open array `{ type: "array", items: {} }`. The schema types keep `items` required, per #2392.

A regression test covers each consumer:

- a non-array is rejected, and a mixed array is accepted;
- the strict and non-strict LLM output carries `items: {}`, with `minItems` shifted under `strict`;
- `covers` holds in both directions;
- a `pipeDelimited` query parameter migrates.

## Coordination

- Follows #2392 and #2404; implemented in #2402.

### LINKED ISSUE #2408: OpenAPI converters drop components whose reference key needs JSON Pointer escaping
## Problem

The OpenAPI converters never decode the JSON Pointer token in a local `$ref`. When a component key needs escaping, such as `A/B` referenced as `#/components/schemas/A~1B`, the converters cannot find the component, and the upgrade or downgrade silently drops or corrupts what referenced it.

## Evidence

Reproduced against `@typia/utils` at #2402 `74033f7`, whose converter code equals `master`:

| Input | Result | Correct |
|---|---|---|
| 3.0 upgrade, operation `parameters: [{ $ref: "#/components/parameters/P~11" }]` | `parameters: []` | the `p` query parameter |
| 3.0 upgrade, `requestBody: { $ref: "#/components/requestBodies/Body~11" }` | request body gone | the JSON body |
| 3.0 upgrade, `responses: { 200: { $ref: "#/components/responses/R~01" } }` | `responses: {}` | the 200 response |
| 2.0 upgrade, the same parameter and response references | both gone | both kept |
| 2.0 downgrade of a component `A/B` | definition keyed `B` | `A/B` |
| 3.0 downgrade of `oneOf: [{ $ref: ".../A~1B" }, { type: "null" }]` | `{ $ref: ".../A~1B" }`, nullability lost | `.../A~1B.Nullable` |

- **Root cause:** about 30 lookups across `SwaggerV2Upgrader`, `OpenApiV3Upgrader`, `OpenApiV3_1Upgrader`, `OpenApiV3_2Upgrader`, `OpenApiV3Downgrader`, and `SwaggerV2Downgrader` take `$ref.split("/").pop()` as the key. RFC 6901 escapes `~` and `/` as `~0` and `~1`, and a URI fragment may also be percent-encoded. `SwaggerV2Downgrader` additionally keyed each definition by `key.split("/").pop()`.
- **Existing decoder:** `LlmReference.readOpenApi` already decodes `#/components/schemas/` references correctly for the validators and the LLM converter, so emended documents are internally consistent. The converters are the odd ones out.
- **Found by:** the early-warning pass on #2406 in #2402, which flagged the native twin. The native twin is unaffected, because native JSON output escapes component names.

## Consequence surface

- **Upgrades:** every reference kind the upgraders resolve (schemas, parameters, headers, request bodies, responses, examples, path items) in 2.0, 3.0, 3.1, and 3.2 documents.
- **Downgrades:** 3.0 and 2.0 nullable references, reference resolution, and definition keys.
- **Downstream:** `HttpLlm` and `HttpMigration` consume the upgraded document, so they lose the same operations' inputs and outputs.

## Invariant

- A converter resolves a local reference by its decoded token.
- It falls back to the raw token for documents that write unescaped keys into references.
- Emitted references keep their escaped token, so the emended document's keys and references stay consistent.

A regression test upgrades 2.0 and 3.0 documents whose parameter, request body, and response keys need escaping, then downgrades an emended document with such a component to 2.0 and 3.0. It asserts:

- nothing is dropped;
- keys stay intact;
- nullability survives.

## Coordination

- Found through #2406; implemented in #2402.

samchon:
## Intent

This pull request resolves the complete accepted cycle: #2400 and #2401, plus #2403, #2404, #2405, #2406, #2407, and #2408, which were published while implementing them (8 issues, 1 pull request).

- **#2400:** a type tag is silently dropped when every tag on the type is malformed. Such tags are now reported the same way as when a valid tag accompanies them.
- **#2403:** a `null` inside a type tag's schema content was dropped on emission. The early-warning pass on the #2400 commit exposed it together with a tuple-example regression, and both are fixed.
- **#2401:** the test suites asserted with `@nestia/e2e`'s one-way `TestValidator.equals`, which misses keys, dates, maps, and sets. Every suite now asserts through one shared oracle, `TestEquality` from `@typia/template/equality`, and a repository guard rejects new `TestValidator.equals` calls.
- **#2404:** exposed by the symmetric oracle. Downgrading an array schema without `items` still collapsed to `{}`, because the guard added in #2392 was unreachable.
- **#2405:** exposed by the symmetric oracle. Native LLM parameters omitted the reference description cascade that `LlmSchemaConverter.parameters` applies, and the parity suites passed only through the one-way comparison.
- **#2406:** found by the early-warning pass on #2405. An LLM parameters root whose component key contains a slash panicked the compiler.
- **#2407:** found by the early-warning pass on #2404. An items-less array still escaped the validator, which accepted any value for it, and also the LLM converter, `covers`, union discrimination, and HTTP migration.
- **#2408:** found by the early-warning pass on #2406. The OpenAPI converters never decoded a `$ref`'s JSON Pointer token, so upgrades dropped parameters, request bodies, and responses reached through escaped keys, and downgrades corrupted keys and lost nullability.

## Scope

| Area | What changes |
| --- | --- |
| Native Go | `MetadataTypeTagFactory` error flushing and tag value domain; `LiteralFactory_Null`; `LlmReferenceDescription` applied to both LLM parameter root builders; full-key root dereference |
| `@typia/utils` | `OpenApiOpenArrayRestorer` ahead of every emended-schema walker's dispatch: downgraders, validator, `oneOf` discrimination, naming rule, type checker, LLM converter, and HTTP migration; `OpenApiReferenceKey` decoding in every converter, schema-walker, and HTTP-composer lookup |
| `@typia/template` | `TestEquality` (`equals`, `subset`, `difference`) under the `equality` entry |
| `tests/*` | every `TestValidator.equals` call migrated; wrong expectations corrected; deliberately partial checks renamed to `subset`; prototype checks compare identity; the inverted invert-suite predicates fixed; test-typia-automated's one-way `primitive_equal_to` delegated to the oracle |
| `test-feature-identity` | the guard against new `TestValidator.equals` calls |
| Docs | "Namespace inheritance" in `llm/strategy.mdx` shows the real root cascade; the development skill states the oracle rule |

## Commits

- `3e2312d` closes #2400.
- `7bf9da2` closes #2403.
- `259fef9` closes #2401. Alone, this commit fails the three LLM parity suites and lacks the #2404 regression test; the next two commits resolve what the symmetric oracle exposed.
- `4e1f3ae` closes #2404.
- `3bb59e4` closes #2405.
- `8b58ef0` closes the oracle holes found after the migration (#2401).
- `bb15576` pins the #2405 cascade on every parameters root.
- `69f9cfe` closes #2406.
- `74033f7` closes #2407; `f780d2e` extends it to union discrimination.
- `715a672` closes the oracle and guard gaps found after `8b58ef0` (#2401).
- `00149f6` closes #2408; `5c8e247` and `9333024` extend it to the schema walkers and HTTP composers.
- `b390a62` applies `pnpm format`.

## Verification

- **Mutation evidence:** recorded per commit in the review ledger. Summary:
  - The guard fails on a planted call.
  - `test_equality_oracle` and `test_total_comparison_shape` fail on a one-way oracle, and the oracle test fails when dates are compared as keyless objects.
  - Disabling the open-array restorer fails both items-omitted tests.
  - Dropping the cascade fails the parameters and application parity suites, and dropping it from the application builder alone fails the application suite.
- **Local results and CI:** recorded in the review ledger on this pull request.
- **Deferred:** the upstream fix of `json_equal_to` belongs to `@nestia/e2e` and is outside this repository.

https://claude.ai/code/session_01De61qv7WJJfAjzKVoSYY2Q


samchon (commented):
**Commit `3e2312d` resolves #2400** and carries its `Close #2400` line.

- **Fix:** `MetadataTypeTagFactory.Analyze` now flushes the collected messages through one path, whether or not a tag object survived. A malformed `validate` is now named `validate`, not `target`, and its message no longer drops a closing quote.
- **Tests:**
  - native test `TestMetadataTypeTagSoleMalformedDiagnostic`;
  - two `test-error` fixtures.
- **Mutation evidence:** restoring the early return fails the native test.
- **Also fixed:** the `llm.evaluation` Go tests from #2399 used diagnostic line numbers as positive twins, and those numbers never pointed at the valid calls, so the twins proved nothing. Each valid twin is now its own project that must compile.

samchon (commented):
**Commit `7bf9da2` resolves #2403**, published during this cycle, and fixes a regression the early-warning pass on `3e2312d` found.

- **Regression:** reporting sole malformed tags broke the documented `tags.Example<["admin", "active"]>`. The tag value check accepted only a single constant or object, although `Example` and `Examples` declare tuple and `null` values. Before #2400, that value check had rejected them when another tag accompanied them, and they were silently dropped when alone. The value check now accepts the declared domain.
- **#2403:** a `null` inside tag schema content was lost on emission, because object writers read a Go `nil` as absent. The new `LiteralFactory_Null` marker keeps a JSON `null` through both the printer and `encoding/json`.
- **Tests:** `test_json_schema_example_tuple_and_null` pins:
  - a sole tuple, and one accompanied by another tag;
  - a `null` example;
  - an object example with a `null` field;
  - an `Examples` record with a `null` member.

  `TestLiteralFactoryNullSurvivesObjects` pins the `encoding/json` side.
- **Mutation evidence:** mutating the value check away and dropping the marker each fail the test.

samchon (commented):
**Commit `259fef9` resolves #2401** and carries its `Close #2401` line.

- **Oracle:** `TestEquality` from `@typia/template/equality`. `equals` is symmetric; `subset` states a deliberately partial check, expected value first; `difference` lists paths.
- **Migration:** every `TestValidator.equals` call and both `_equalsExactly` helpers. About 110 lenient-parse error lists and three other expected-first partial checks became `subset`. Wrong expectations were corrected: the nullable downgrade's `components`, and the JSON Pointer discriminator mapping. Prototype checks compare identity.
- **Guard:** `test_feature_identity_equality_oracle` rejects new calls, with planted spellings as negatives.
- **Exposed:** the parity suites' root description (#2405) and the unreachable items guard (#2404), fixed in the next two commits. On its own, this commit fails those three parity suites.
- **Mutation evidence:**
  - A planted call fails the guard.
  - A one-way oracle fails `test_equality_oracle` and `test_total_comparison_shape`.
  - Comparing dates as keyless objects fails the oracle test.
- **Early-warning pass:** no blocking defect in the migration. It found oracle holes, which `8b58ef0` closes.

samchon (commented):
**Commit `4e1f3ae` resolves #2404.** `OpenApiOpenArrayRestorer` restores `items: {}` before each downgrader dispatches, and the unreachable guards from #2392 are removed.

- **Tests:** `test_json_schema_downgrade_items_omitted_nested` covers a nested property and a described top-level array for 2.0, 3.0, and 3.1.
- **Mutation evidence:** disabling the restorer fails both items-omitted downgrade tests.
- **Early-warning pass:**
  - The diff is correct.
  - It found that the same root cause is live in the validator, the LLM converter, `covers`, and HTTP migration. Published as #2407 and fixed in `74033f7`.
  - The nested test's doc wrongly claimed to exercise the identity check. Corrected in `74033f7`.

samchon (commented):
**Commit `3bb59e4` resolves #2405.** `LlmReferenceDescription` ports `JsonDescriptor.cascade({ escape: true })`, and both native parameters root builders apply it. The guide's "Namespace inheritance" example now prints the real root cascade.

- **Tests:**
  - `TestLlmReferenceDescription` pins undescribed, described, namespaced, empty, missing-parent, inline, and foreign roots.
  - `test_llm_parameters_parity_converter_namespace` pins parity and the verbatim cascade.
- **Mutation evidence:** dropping the cascade fails the parameters and application parity suites. Dropping it from the application builder alone fails the application suite.
- **Early-warning pass:**
  - The port matches utils.
  - It found a pre-existing last-slash key lookup that panics the compiler. Published as #2406 and fixed in `69f9cfe`.
  - It found a misleading `@param` in the example, and cascade coverage missing for structured output and application. Both fixed in `bb15576`.

samchon (commented):
**Commit `8b58ef0` closes the oracle holes (#2401)** that the full local run and the early-warning pass exposed.

- **Invert suites:** nine llm invert cases passed `(key) => key !== "description"` as the exception, which skips every key *but* `description`. With the predicate corrected, three cases failed on legitimate format differences and on empty `$defs`. They now compare against `typia.json.schema` of the same type through `_test_llm_invert`, whose negative probes (a changed constraint, empty `$defs`, a changed member) all fail.
- **Automated JSON cases:** `primitive_equal_to` in test-typia-automated was one-way with the parsed output first, so every generated `json.stringify` case missed a dropped property. It now delegates to `TestEquality`. `test_primitive_equal_to_oracle` pins symmetry, and a one-way mutation fails it.
- **Oracle:** `TestEquality` closes these holes, each in both orders:
  - a missing key read through the prototype;
  - functions ignored anywhere;
  - opaque RegExp, Error, and binary values;
  - set exception and multiplicity;
  - cycles.

  The previous oracle fails the extended table.
- **Guard:** it blanks comments, and also catches `?.`, casts, stored references, and renaming imports.

samchon (commented):
**Commit `bb15576` (#2405 follow-up)** pins the namespace cascade on the structured output and application roots, and drops the example's misleading `@param`. The guide highlight moves with it.

samchon (commented):
**Commit `69f9cfe` resolves #2406**, published during this cycle. The native parameters programmer dereferences the root by its full component key, so `IVariant<"A/B">` no longer panics the compiler or leaves an unresolved `$ref` in application parameters.

- **Tests:** `test_llm_parameters_slash_component_key` pins parameters, structured output, and application.
- **Mutation evidence:** the previous code panics on the test's exact source.

samchon (commented):
**Commit `74033f7` resolves #2407**, published during this cycle. `OpenApiOpenArrayRestorer` moves to `utils/internal`. The validator, the naming rule, the type checker's escape and flattening, the LLM converter's walks, and HTTP migration's checks now restore the open array before dispatching.

- **Tests:** `test_json_schema_items_omitted_consumers` pins:
  - a non-array rejected and a mixed array accepted;
  - strict and non-strict LLM output;
  - `covers` in both directions;
  - a `pipeDelimited` query parameter.
- **Mutation evidence:** disabling the restorer fails this test and both downgrade tests.

samchon (commented):
**Commit `f780d2e` (#2407 follow-up):** `OpenApiOneOfValidator` flattening now restores the open array for discrimination, while keeping the raw member for its identity-based remainder filter.

- **Found by:** the early-warning pass on `74033f7`.
- **Tests:** `test_json_schema_items_omitted_consumers` now covers unions beside an object and beside a string array.
- **Mutation evidence:** removing the restore fails the union case.

samchon (commented):
**Commit `715a672` (#2401 follow-up)** fixes what the early-warning pass on `8b58ef0` found:

- kinds are now read from internal slots, not from `instanceof` or `Symbol.toStringTag`;
- a cycle through a set no longer overflows;
- errors compare their own fields again, and the key exception reaches `name` and `message`;
- a lexer replaces the regex comment blanker, which erased code after a `//` or `/*` inside a string;
- `_test_llm_invert` now sorts union members by keys it actually compares, and keeps a resolved reference's siblings.

**Mutation evidence:**
- The previous oracle fails the extended table.
- The regex blanker fails the new literal cases.

samchon (commented):
**Commit `00149f6` resolves #2408**, published during this cycle. The early-warning pass on `69f9cfe` flagged the native twin, which is unaffected because native JSON escapes names. The utils converters, however, never decoded a `$ref` token.

- **Reproduced on the previous code:**
  - a 3.0 or 2.0 upgrade dropped parameters, request bodies, and responses referenced by an escaped key;
  - the 2.0 downgrade keyed `A/B` as `B`;
  - the 3.0 downgrade lost nullability.
- **Fix:** `OpenApiReferenceKey` decodes the token. It looks up the decoded key first, then the raw token.
- **Tests:** `test_openapi_reference_key_escaped` pins every case.
- **Mutation evidence:** disabling decoding fails the test.

samchon (commented):
**Commit `5c8e247` (#2408 follow-up):** the early-warning pass on `00149f6` found the same raw-token lookups outside the converters.

- **`OpenApiTypeChecker`:** `escape`, `unreference`, `visit`, the recursion check, and the flattening behind `covers` all used raw tokens. `covers` treated `Q~1P` as unknown, so it accepted anything.
- **HTTP composers:** both the migration composer and the LLM application composer used raw tokens. Migration wrapped an object query as a nested one.
- **Description and naming:** `JsonDescriptor` and the validator's type names showed the escaped token.
- **Fix:** `OpenApiReferenceKey` now takes a prefix and decodes through `LlmReference.readToken`, now the only token decoder.
- **Tests:** the regression runs every walker over an escaped key and a plain key and requires the same answers. It also covers 3.1 and 3.2 headers and webhooks.
- **Mutation evidence:** reverting the walkers fails the walker case.
- **Correction:** the commit message's claim that every path agrees is too broad. The validators and the LLM converter stay strict by the #2104 contract; see `9333024`.

samchon (commented):
**Commit `9333024` (#2408 follow-up)** addresses the early-warning pass on `5c8e247`.

- `escape` now describes a literal key such as `A~1B` under the key it was found by, and no longer drops the component's description.
- `unreference` names a foreign reference whole again.
- The lenient-versus-strict boundary is now documented: under #2104, the validators and the LLM converter keep rejecting a malformed reference.
- **Mutation evidence:** without the found key, the literal-key assertion fails.

samchon (commented):
**Self-review round (clean) on `9333024`.**

- **Scope:** the whole `origin/master...9333024` diff.
- **Mechanical test migration:** pinned by the guard.
- **Non-mechanical changes:** read change by change:
  - native tag factory, literal factory, LLM reference description, and parameters programmer;
  - utils converters, walkers, validators, and HTTP composers;
  - `TestEquality` and its guard;
  - `primitive_equal_to` and `_test_llm_invert`;
  - the website guide and the development skill.
- **Findings:** no remaining defect.
- **Accepted as-is:** `escapeSchema`'s "unable to find" message still names only the last token of a foreign reference. That is unchanged from `master`. Only `unreference` now names the whole reference.
- **Evidence at this head:**
  - every CI lane passes;
  - local `GOFLAGS=-p=4 pnpm test` passes (all TypeScript suites plus Go public and native);
  - each fix's mutation evidence is in the commit ledger above.
- **Remaining gate:** the integration gate runs next.

samchon (commented):
**Integration gate passed** on head `9333024`.

- **Base:** `origin/master` is `63c143d`, an ancestor of the head, so the prospective merge tree equals the head tree.
- **Manifest:** `b0322948976ee847`.
- **Where:** a disposable worktree.
- **Steps:**
  1. `pnpm install --frozen-lockfile`: OK.
  2. `pnpm build`: OK.
  3. `GOFLAGS=-p=4 pnpm test`: OK, with every TypeScript suite plus the Go public and native tests.
  4. `pnpm package:tgz`: OK.
  5. Clean packed consumer: the typia, interface, and utils tarballs with ttsc 0.28.1 and TypeScript 7.0.2, built outside the workspace. Running it confirmed:
     - the namespace cascade on `llm.parameters<IMember.ICreate>`;
     - `llm.parameters<IVariant<"A/B">>` compiles without a panic;
     - `OpenApiValidator` rejects a non-array for an items-less array;
     - a 3.0 upgrade keeps a `P~11` parameter reference.
  6. `website`, `pnpm build`: OK.

Issue #2413, opened 2026-09-19
llm.evaluation names never and unknown with the wrong reason


samchon:
## Problem

`typia.llm.evaluation<T>()` rejects unsupported types at compile time with a message naming the reason, but two types get the wrong reason:

```ts
typia.llm.evaluation<{ /** Never? */ n: never }>();
// - $input.n
//   - LLM evaluation does not support optional or undefined properties.

typia.llm.evaluation<{ /** Unknown? */ u: unknown }>();
// - $input.u
//   - LLM evaluation does not support any type.
```

Reproduced on `master` `cbf5dba` with a fresh pack. typia's metadata models `unknown` with the `Any` flag and `never` as an empty schema that is not required, so `LlmEvaluationProgrammer.property` reaches the `any` and the `optional or undefined` branches before it can name the actual type.

The runtime side is exact: a probe of `validate()` against a threshold miss, a gated option below its requirement, an unknown option, an out-of-range score, a missing key, an extra key, a wrong answer kind, a probability above 1, `NaN`, a non-object, `null`, and a `__proto__` key was rejected at the right path in every case.

Coverage gap: `llm_evaluation_rejects_undecidable_types_test.go` pins only `any`. `never` and `unknown` are pinned nowhere, and `tests/test-error` proves rejection without the message.

## Invariant

Every unsupported leaf is named by what it is: `never` as a type no answer can satisfy, and `unknown` as an unbounded type, distinct from `any`.

## Acceptance

- Native tests pin the exact message for `never`, `unknown`, and `any`, each beside an accepting twin.
- `tests/test-error` gains fixtures for `never` and `unknown`.

Issue #2412, opened 2026-09-19
Normalize items-less arrays once at the emended boundary and resolve references by decoded key only


samchon:
## Problem

#2404 and #2407 fixed the items-less array by calling `OpenApiOpenArrayRestorer.restore()` at 12 dispatch sites across the downgraders, the validator, the `oneOf` validator, the naming rule, the type checker's escape and flattening, the LLM converter, and HTTP migration. That patches every consumer of an emended schema instead of establishing the invariant once.

The invariant already has an owner. #2392 decided that the emended `IArray` type keeps `items` required and that the **upgraders normalize** an items-less array into `items: {}`. Every document that enters through `OpenApiConverter.upgradeDocument()` therefore already satisfies the contract. The only way an items-less array reaches a consumer is an emended document built by hand or parsed from JSON, which `upgradeDocument()` returns untouched because `isUpgraded(document)` short-circuits before any normalization.

The canonical fix is to normalize once at that boundary, not to guard 12 dispatches.

Related: #2408's `OpenApiReferenceKey` looks a decoded key up first and then **falls back to the raw token**. OpenAPI 3.x restricts component keys to `^[a-zA-Z0-9.\-_]+$` (the repository's `OpenApiComponentName.GRAMMAR`), under which escaping never changes a key, so decoding is transparent for every spec-compliant document. The fallback serves only a document that puts an unescaped `/` both in a key and in a reference, which is invalid twice over. It is leniency without a contract behind it.

## Invariant

- An emended document that enters through `OpenApiConverter` is normalized once, at the boundary: an array without `items` and `prefixItems` gains `items: {}` there. No consumer dispatch checks for the shape.
- A local `$ref` resolves by its decoded JSON Pointer token only. A component keyed under an unescaped `/` is not resolved by a reference that also fails to escape it.

## Acceptance

- `upgradeDocument()` and `upgradeComponents()` normalize an already-emended input; `OpenApiSchemaSanitizer` owns the walk, next to `omitEmptyRequiredDeep`.
- `OpenApiOpenArrayRestorer` and all 12 `restore()` calls are removed.
- `OpenApiReferenceKey.find` keeps returning the key a component was found under (the `.Nullable` twin needs it), but only the decoded key is tried.
- `test_json_schema_items_omitted_consumers` and `test_json_schema_downgrade_items_omitted_nested` still pass with their inputs routed through `upgradeDocument()`; a test pins that a hand-built emended document is normalized on entry.
[...]

Issue #2411, opened 2026-09-19
Reduce @typia/jev to a converter like its sibling adapters


samchon:
## Problem

`@typia/jev` (#2409, #2410) shipped far more than the repository's adapter packages do. `@typia/vercel`, `@typia/langchain`, and `@typia/mcp` each export plain functions that convert typia output into the target's format and nothing else; the vendor SDK owns transport. `@typia/jev` instead ships:

- `Jev.openrouter()`, a hand-written `fetch` client with a retry policy, a timeout, and three error classes;
- `Jev.typesafe()`, a wrapper over the official SDK client;
- `Jev.Jsonable<S>`, a conditional type re-deriving what JSON can carry, which needed four follow-up commits for typia tags and tuples;
- about 550 lines of source and 1,000 lines of tests.

The package's whole reason to exist is one conversion: `typia.llm.evaluation<T>()` spells the yes/no question `"boolean"`, and Jev's wire format spells it `"noul"`. Choice and score questions are identical, and `validate()` already accepts Jev's native answers.

## Invariant

`@typia/jev` is a converter, shaped like its siblings: it exports `toJevQuestions(questions)` and the two wire types it needs, and nothing else. Transport belongs to `@typesafe-ai/sdk` or to Vercel AI SDK, whose TypeSafe and OpenRouter providers convert the neutral questions themselves.

## Acceptance

- `packages/jev/src/index.ts` exports `toJevQuestions`, `IJevQuestion`, and `IJevNoulQuestion`, following `@typia/vercel`'s plain-function shape.
- `Jev.openrouter`, `Jev.typesafe`, `Jev.Jsonable`, `JevHttpError`, `JevConnectionError`, `JevTimeoutError`, and the retry policy are removed, with their tests.
- `test-jev` keeps the conversion test and the compile-time contract against `@typesafe-ai/sdk`'s `Question` and `systemOne` request.
- The guide, README, example, and ESM experiment show only the conversion.

The package is unpublished, so nothing depends on the removed surface.

Issue #2414, opened 2026-09-19
Scan for TestValidator.equals through the TypeScript AST


samchon:
## Problem

`test_feature_identity_equality_oracle` (#2401) rejects any `TestValidator.equals` call in tracked test sources. To skip comments while keeping calls inside strings honest, it carries an 80-line hand-written lexer for line and block comments, string and template literals, template substitutions, and regular-expression literals, plus a regular expression over the blanked text for the call's spellings.

That is a partial TypeScript scanner. The suite already depends on `ttsc`, which brings the TypeScript compiler, and the compiler's own parser is the canonical way to find a property access on an identifier: a `PropertyAccessExpression` or `ElementAccessExpression` whose expression is the identifier `TestValidator` and whose name is `equals`, plus a binding pattern or import that takes `TestValidator` by another name. Comments and literals never appear in that walk.

The sibling `FeatureIdentity.parse` deliberately uses a regular expression because it matches exactly one declaration form and fails closed. The guard's goal is the opposite, to catch every spelling, which a lexer approximates and a parser decides.

## Invariant

The guard finds every reference to `TestValidator.equals` through the TypeScript AST, so a spelling it did not anticipate is still found and a comment or literal is never mistaken for code.

## Acceptance

- The scan parses each source with `ts.createSourceFile` and walks the AST; the lexer and the spelling regex are removed.
- The planted spellings and the literal cases keep passing as the negative and positive cases.

samchon:
Closing as invalid on the evidence found during implementation.

The premise was that the suite already has the TypeScript compiler's parser through `ttsc`. It does not: the repository compiles with TypeScript 7, whose compiler is native and exposes no JavaScript API (`typescript@7.0.2` ships only `tsc.js`, `getExePath`, and `version`), and `ttsc` exposes no scanner either. The only `typescript@5.9.3` in the tree is nextra's transitive dependency in the website, so adopting it here would add a second TypeScript major to a test workspace for one guard, which is worse than the lexer, not better.

An AST rewrite was implemented and then reverted once the compile errors showed the missing API. The guard keeps its text scan and now states this constraint in place, in `4f0d5a8` on #2415.

samchon:
Correction: the commit is `ce3a6cd`, not `4f0d5a8`.

PR #2415, opened 2026-09-19
fix: converter-only @typia/jev, one boundary and one reader for emended documents, exact never/unknown diagnostics

samchon:
## Intent

This pull request resolves the session re-audit cycle: #2411, #2412, #2413, and #2416 (4 issues, 1 pull request). #2414 was published in the same cycle and closed as invalid on evidence found during implementation; #2416 was found by the early-warning pass on #2412's fix and added to the cycle.

The audit asked three things of everything this session merged: no hardcoding, no monkey patching, and only canonical design; and exact compile-time rejection in `typia.llm.evaluation<T>()`.

- **#2411:** `@typia/jev` shipped an HTTP client, a retry policy, three error classes, an SDK wrapper, and a JSON-shape conditional type. Its siblings export one converting function each. It is now `toJevQuestions()` and two wire types.
- **#2412:** #2404 and #2407 patched an items-less array at 12 dispatch sites. #2392 had placed the invariant at the emended boundary, so `upgradeDocument()`, `upgradeComponents()`, and `downgradeDocument()` now normalize an emended input once, over every schema holder, and the 12 calls are gone. `OpenApiReferenceKey` decodes a reference by RFC 6901 alone: the raw-token fallback is gone, a key holding a space, as Swagger 2.0 allows and the repository's own Semantic Scholar fixture carries, resolves as written, and only a malformed token names no component.
- **#2413:** `never` was reported as "optional or undefined" and `unknown` as "any". typia's metadata folds those types together, so each message now names the class it records and why it is refused, and the Go oracle pins `never`, `unknown`, and `void`.
- **#2416:** the validator and the LLM composers read the same document through a second, stricter reader, so a reference the type checker resolved was malformed to `OpenApiValidator` and `HttpLlm.application()`. One reader now owns the meaning of an OpenAPI reference; the `#/$defs/` reader of an LLM schema keeps #2102's stricter contract, and both share their RFC 6901 escaping through `JsonPointer`.
- **#2414 (closed, invalid):** replacing the guard's lexer with the TypeScript parser assumed a parser the repository does not have: TypeScript 7 is native and ships no JavaScript API. The guard now states that constraint in place.

## Also audited, unchanged

- The runtime `validate()` of `llm.evaluation` was probed against 13 wrong-answer shapes; every one is rejected at its path.
- The Go port of `JsonDescriptor.cascade` duplicates format strings across the JS-descriptor / Go-plugin boundary, which the project skill accepts and a parity test pins.
- The `0.5` default threshold is the public `tags.Probability` contract.
- `TestEquality` stays: `util.isDeepStrictEqual` rejects the JSON-like `undefined`-key and function-key semantics the suites rely on.

## Verification

- Mutation evidence per commit in the review ledger.
- CI and local `pnpm test` results at the settled head in the ledger.

https://claude.ai/code/session_013UyG473gpnxm3YdhDZDNp7


samchon (commented):
## Ledger: 1c05a53 — `@typia/jev` reduced to `toJevQuestions()`

Resolves #2411.

**Landed.** The package is one file, `packages/jev/src/index.ts`, exporting `toJevQuestions(questions)` and the `IJevQuestion` / `IJevNoulQuestion` types. It renames the neutral `"boolean"` question to Jev's `"noul"` and passes choice and score questions through, keyed like the input. Removed: the HTTP client (`Jev.ts`), `JevHttpError.ts`, the retry policy, the `Jsonable<S>` type, `MockFetch`, and every test of them. Kept: `test_jev_questions` (mapping) and `test_jev_typesafe_sdk_contract` (five `Extends<>` cases against `@typesafe-ai/sdk`'s `Question`), so a wire-format drift in the SDK fails the build.

**Why this shape.** The sibling adapters `@typia/vercel`, `@typia/langchain`, and `@typia/mcp` are pure converters exporting plain functions, and the vendor SDK owns transport. `@typia/jev` now follows the same boundary.

**Consumers updated.** README, `website/.../utilization/jev.mdx`, `.../llm/evaluation.mdx`, `examples/src/llm/evaluation.ts`, `experiments/esm/src/test_jev.ts`, the `ILlmEvaluation` JSDoc, and the project skill's package list.

**Mutation evidence.** `toJevQuestions` returning its input unchanged fails `test_jev_questions` and `test_jev_typesafe_sdk_contract`.

samchon (commented):
## Ledger: 95757cd — emended documents normalized once, at entry

Resolves #2412 (first part; fb4b58e completes it).

**Landed.** `OpenApiConverter.upgradeDocument()` and `upgradeComponents()` normalize an already-emended input through `OpenApiSchemaSanitizer` so every `type: "array"` without `items`/`prefixItems` gains `items: {}` once, at the boundary #2392 defines. The twelve `OpenApiOpenArrayRestorer.restore()` calls that #2404 and #2407 placed at consumer dispatches are removed with the restorer itself; the downgraders, validator, type checker, HTTP migration, and LLM composers read the emended type on its word again. `OpenApiReferenceKey` decodes a `$ref` token instead of falling back to the raw token.

**Tests.** `test_openapi_emended_items_omitted_boundary` enters a hand-built emended document and asserts the components, parameter, request body, response, and webhook all gained `items: {}`, that every downgrade keeps the open array, that the validator rejects a non-array, and that migration yields its three routes. The three consumer-level `items_omitted` tests written for #2404/#2407 are deleted as the boundary test subsumes them.

**Early-warning pass on this commit** found three gaps, all repaired in fb4b58e: response headers, `additionalOperations`, and streaming `itemSchema` were not normalized; `downgradeDocument()` did not normalize on entry; and the decoder rejected keys outside the 3.x grammar, which real 2.0 documents in the repository's own fixtures carry.

samchon (commented):
## Ledger: e8fef15 — exact `llm.evaluation` diagnostics for `any`, `unknown`, `void`, `never`

Resolves #2413.

**Landed.** `LlmEvaluationProgrammer.go` names the rejected type by typia's metadata model: `unknown` folds into `Any`, and `never`/`void`/`undefined` fold into `Required=false`, so the two messages now read "does not support any or unknown types, because an evaluation model answers only closed sets" and "does not support optional, undefined, void, or never properties, because every property of the result needs an answer". Previously `never` was reported as "optional or undefined" and `unknown` as "any", which named a type the source never wrote.

**Tests.** The Go test `llm_evaluation_rejects_undecidable_types_test` gains `voided: void`, `impossible: never`, and `opaque: unknown` cases pinning the message text; `tests/test-error` gains `llm.evaluation.never.ts` and `llm.evaluation.unknown.ts` fixtures proving the transform rejects them at compile time.

**Mutation evidence.** Reverting either message string fails the Go test.

samchon (commented):
## Ledger: ce3a6cd — equality-guard rationale

Refs #2414, closed as invalid.

**Landed.** A doc comment only, on `test_feature_identity_equality_oracle.ts`. The issue proposed replacing the guard's hand lexer with a TypeScript AST walk; the attempt compiled against `typescript@7.0.2`, the repository contract, and failed because TypeScript 7 (native) exposes no JavaScript parser API; `typescript@5.9.3` reaches the workspace only transitively through the website. The lexer stays, and the comment records why so the proposal is not repeated.

No behavior change; no test change.

samchon (commented):
## Ledger: fb4b58e — RFC 6901 decoding, every schema holder normalized

Completes #2412 (with 95757cd); also fixes the `@typia/jev` install lines for #2411.

**Landed.**
- `OpenApiReferenceKey` decodes a reference's last token by RFC 6901 alone: percent-decode, `~1`→`/`, `~0`→`~`, every other character is itself, and only a `~` followed by anything but `0`/`1` is malformed and names nothing. The stricter `LlmReference` charset remains the contract of typia's own emitted `$defs`. Every caller handles the `undefined` of a malformed token (downgraders, upgraders, HTTP composers, the type checker's `visit`/`unreference`/`escape`/`isRecursiveReference`, the naming rule).
- `normalizeDocument` now covers response headers, `additionalOperations`, and a streaming `itemSchema`; `downgradeDocument()` normalizes on entry as the boundary in the other direction.
- `OpenApiSchemaSanitizer` is split: `fillOpenArrayDeep` is the boundary rule, `omitEmptyRequiredDeep` the emission rule the migrate composer applies, so version conversion keeps preserving an explicit `required: []` (`test_openapi_converter_empty_required`).
- `JsonDescriptor.cascade` takes its key rather than re-deriving it.

**Tests.** New `test_document_references_resolve` upgrades every fixture of every version and, for each source-defined `#/components/schemas/` reference, asserts a nested `OpenApiTypeChecker.visit` reaches more than the reference itself, and asserts at least one fixture carries a key outside the 3.x grammar (Semantic Scholar's `Autocomplete Paper`), so the check cannot pass on tidy inputs. `test_openapi_reference_key_escaped` gains a space-key case and a malformed-`~2` case; `test_openapi_emended_items_omitted_boundary` gains the header, additional-operation, `itemSchema`, and downgrade-entry cases.

**Mutation evidence.** Rejecting a space in the decoder fails both reference tests; skipping entry normalization, dropping the `additionalOperations` branch, or making `fillOpenArray` a no-op each fails the boundary test.

samchon (commented):
## Ledger: cdf6e89 — one reader for every consumer of an OpenAPI reference

Resolves #2416; refs #2412.

**Found by.** The early-warning pass on fb4b58e: the validator, the `oneOf` validator, `LlmSchemaConverter`, and `LlmParametersComposer` read an inbound `#/components/schemas/` reference through `LlmReference.readOpenApi`, whose URI-fragment charset rejects a literal space, while the walkers, the converters, and HTTP migration read it by RFC 6901. Over `examples/v2.0/semanticscholar.json`, `OpenApiTypeChecker.escape` resolved `Autocomplete Paper`, `OpenApiValidator.validate` called it malformed, and `HttpLlm.application()` dropped `GET /paper/autocomplete` and `GET /paper/search/match`. This predates the cycle; fb4b58e's doc comment had wrongly called the strict reader the contract of typia's own emitted references.

**Landed.**
- `LlmReference.readOpenApi` delegates to `OpenApiReferenceKey.read(reference, "#/components/schemas/")`. `LlmReference.read` for `#/$defs/` keeps #2102's stricter contract, since typia writes those references itself.
- `OpenApiReferenceKey.read` with a prefix requires the reference to follow it, and a `/` that survives percent-decoding is the pointer separator RFC 6901 §6 makes it, so `A%2FB` names no single component, as the strict reader always held.
- New `JsonPointer` holds the RFC 6901 `escape`/`unescape` both readers used to carry as copies.
- The entry-normalization helpers add no key the input left out (`mapValues`, `normalizeMediaType`), matching the upgraders, and the `upgradeDocument` doc no longer claims it strips an empty `required`.

**Tests.** `test_document_references_resolve` now also walks request and response bodies and holds the validator and `LlmSchemaConverter.schema` to the walker's answer for every source-defined reference in every fixture. `test_openapi_reference_key_escaped` gains a "readers agree" matrix over nine spellings (`Plain`, `A~1B`, `A%20B`, `A B`, `100%`, `A%2FB`, `A/B`, `A~2B`, `Missing`) for the type checker, the validator, and the LLM converter. `test_openapi_emended_items_omitted_boundary` asserts `upgradeComponents()` normalization and that no phantom key appears. The OpenAPI rows of `test_llm_schema_json_pointer_references` now expect `C%D`, `A B`, and `A%` to resolve, with `A/B`, `T~N`, and a foreign document still refused.

**Mutation evidence.** Restoring the strict reader in `readOpenApi` fails `references_resolve` ("rejected references", two Semantic Scholar entries) and `reference_key_escaped` ("readers agree" on `A B` and `100%`). Dropping the separator rule fails "readers agree" on `A%2FB` and `A/B`.

samchon (commented):
## Ledger: 41a5514 — absent holders survive entry normalization

Refs #2416.

**Found by.** The early-warning pass on cdf6e89: `mapValues` applied its rule to a hole, so a hand-built emended document whose `responses` record held an `undefined` value threw in `upgradeDocument()`, where the branch it replaced had kept the hole.

**Landed.** The record walk leaves an absent value (`undefined` or `null`) for the reader to judge, as the sanitizer's schema walk leaves an absent schema, and `normalizeMediaType` no longer carries its own guard. The normalization comment no longer credits the upgraders with a shape they do not produce, the reader header links no symbol it does not import, and `test_document_references_resolve`'s untidy-key guard counts keys outside the URI-fragment charset, which is the class the stricter reader refused, rather than the 3.x key grammar, so the guard still bites if the Semantic Scholar fixture were replaced by one with, say, `A:B`.

**Tests.** `test_openapi_emended_items_omitted_boundary` pins that `responses: { 200: undefined, 404: {...} }` normalizes without throwing and keeps the hole.

**Mutation evidence.** Removing the hole guard in `mapValues` fails "absent holder kept" with a `TypeError`.

samchon (commented):
## Ledger: c7d6062 — only the hole the type allows

Refs #2416.

**Found by.** The early-warning pass on 41a5514: `mapValues` kept a `null` as well as an `undefined`, wider than the sanitizer precedent its comment cites and than any type in the emended document allows; the "absent holder kept" assertion compared through an equality that treats a missing key and an `undefined`-valued key alike, so it pinned only "no throw".

**Landed.** The record walk passes only an `undefined` value through; the value type is spelled `Record_[keyof Record_]`; the test pins the kept key by name (`Object.keys`) next to the value.

**Mutation evidence.** Dropping the `undefined` guard still fails "absent holder kept" with a `TypeError`; dropping the `"200"` key from the expected list fails it on `keys`.

samchon (commented):
## Ledger: 31bc12b — redundancies the one reader left behind

Refs #2412, #2416. Found by the solo Self-Review round over the whole base-to-head diff, which a per-commit pass cannot see.

**Landed.** `JsonDescriptor.cascade` drops the `prefix` it stopped reading when its `key` became required (fb4b58e); the two HTTP composers drop the `startsWith(SCHEMAS)` guard that the prefix-requiring reader (cdf6e89) made redundant and resolve a reference through one `find` instead of `read` plus `get`; `unreferenceSchema` names a missing component as `key ?? $ref` instead of re-testing the prefix the reader already required.

No behavior change; `reference`, `migrate`, and `llm_application` tests pass unchanged.

samchon (commented):
## Self-Review: final round over cbf5dba..31bc12b, clean

**Rounds.** Round 1 over the complete base-to-head diff found the `JsonPointer` duplication (folded into fb4b58e), and its early-warning follow-ups surfaced #2416. Round 2 over the new snapshot found the dead `prefix` of `JsonDescriptor.cascade`, the redundant prefix guards in the HTTP composers, and the re-tested prefix in `unreferenceSchema` (31bc12b). Round 3, this one, over cbf5dba..31bc12b: every file of the 52 reread against the four issues' invariants; no sound improvement remains.

**Requirement fidelity, issue by issue.**
- #2411: `packages/jev/src/index.ts` is one function and two types; no client, retry, error class, or `Jsonable` survives; README, guide, example, ESM experiment, `ILlmEvaluation` JSDoc, and the project skill say only what the package does.
- #2412: `OpenApiOpenArrayRestorer` and its 12 calls are gone; `upgradeDocument`, `upgradeComponents`, and `downgradeDocument` normalize every schema holder once; `required: []` is preserved through conversion; the reader decodes by RFC 6901 alone.
- #2413: the two Go messages name the metadata class and its reason; `never`, `unknown`, `void` are pinned in Go and rejected by `tests/test-error` fixtures.
- #2416: one reader (`OpenApiReferenceKey`) for every consumer of an OpenAPI reference; the `#/$defs/` reader keeps #2102's contract; `JsonPointer` holds the shared escaping.
- #2414 stays closed as invalid, with the constraint recorded in the guard.

**Consequence coverage.** Every `OpenApiReferenceKey.read/get/find` caller handles `undefined`; every `LlmReference.readOpenApi` caller keeps its deterministic failure on a malformed or missing reference; `LlmReference.write` still emits canonical `$defs` references, so a lenient inbound reading never produces a non-canonical outbound one (probed: 18 keys round-trip). The Go emitter reads only typia's own keys and is untouched.

**Tests.** Boundary, reader-agreement, fixture-wide resolution, jev conversion and SDK contract, Go message oracle, and `test-error` rejection. Mutation evidence per commit in the ledger above.

**Gates on 31bc12b (manifest 5284dbc095a51219).** CI: all eight checks green on this SHA. Disposable worktree at the merge result (origin/master cbf5dba is an ancestor, so the merge tree equals the head tree): `pnpm install --frozen-lockfile` OK; `pnpm build` OK; `GOFLAGS=-p=4 pnpm test` exit 0; `pnpm package:tgz` OK; ESM experiment against the tarballs: all checks passed including `@typia/jev`; `website` build OK; clean packed consumer (TypeScript 7.0.2, ttsc, `@typesafe-ai/sdk`): compiles, `toJevQuestions` output type-checks as the SDK's `Question` map and `systemOne` request, `validate()` folds native answers into `ITriage`, and a Swagger 2.0 document with a spaced key validates and composes an LLM function with no error.

PR #2423, opened 2026-09-21
fix: harden JEV evaluation before release

### LINKED ISSUE #2417: llm.evaluation accepts incomplete and contradictory probability distributions
## Problem

`typia.llm.evaluation<T>().validate()` accepts a choice or score answer whose optional `probabilities` map is incomplete or inconsistent. At `b686433f244529db302a7489ef45a62f94cfa659`, `_createLlmEvaluation.ts` only rejects unknown keys and non-probability values; it does not require every declared key, a sum of 1, a maximum-probability selected choice, or a score equal to the weighted mean. An accepted answer can therefore contradict the evaluation model's own result.

## Evidence

For a choice `"billing" | "technical"`, `{ type: "choice", choice: "billing", probabilities: { billing: 1 } }` passes despite omitting `technical`; `{ billing: 0.4, technical: 0.6 }` passes while selecting `billing`. For a score `0 | 1 | 2`, `{ type: "score", score: 0, probabilities: { "0": 0, "1": 0, "2": 1 } }` passes although the weighted mean is 2. The owner is `packages/typia/src/internal/_createLlmEvaluation.ts` (`distribution`, `choice`, and `score`). AI SDK 7.0.106's `validateEvaluationAnswers` requires exact distribution keys, sum 1 within its declared precision, selected choice at the maximum, and a score matching the weighted mean.

## Consequence surface

This affects both `experimental_evaluate()` answers and direct Jev/TypeSafe answers passed to `validate()`, for string choices and numeric scores, including nested paths and enum/literal spellings. `probabilities` remains optional: adapters without a distribution must still work when no member-level acceptance minimum requires one. When present, malformed maps must fail on the decision path without throwing. Test exact keys (missing/extra), bounds, non-finite values, sum, ties, maximum choice, weighted score, and valid controls. Preserve prototype-like option names as own keys.

## Invariant and acceptance

A supplied choice or score distribution is complete, finite, normalized, and agrees with the selected answer. A missing optional distribution remains valid when no minimum needs it. `pnpm --filter @typia/test-typia-schema start -- --include llm_evaluation` must cover positive, negative, and boundary controls, followed by the broader package/test gates. This issue concerns answer validation, not compile-time probability annotation coverage (tracked separately).

### LINKED ISSUE #2418: Reject partial probability requirements across evaluation members
## Problem

`typia.llm.evaluation<T>()` silently accepts probability requirements on only some members of a choice, score, or set. At `b686433f244529db302a7489ef45a62f94cfa659`, `LlmEvaluationProgrammer.go` emits a minimum for annotated choice/score members but none for unannotated members; a set silently gives unannotated members `0.5`. This applies equally to `tags.Probability<N>` on literal members and JSDoc `@probability N` on enum members. A partially annotated enum therefore looks calibrated but changes policy by selected member.

## Evidence

`enum Action { /** @probability 0.8 */ escalate = "escalate", reply = "reply" }` compiles. Its `escalate` answer requires a distribution and minimum 0.8; `reply` has no minimum and can pass without a distribution. The equivalent literal union `("escalate" & tags.Probability<0.8>) | "reply"` has the same gap. The compiler owner is `packages/typia/native/core/programmers/llm/LlmEvaluationProgrammer.go`, in its member loop. Existing per-value validation already rejects malformed, duplicate, and out-of-range annotations, but does not check coverage. `@probability 1` is a valid boundary; values above 1 are invalid.

## Consequence surface

Cover string choices, numeric scores, and string-array sets; enum JSDoc and literal type tags; reused enums; property-level `@probability` fallback; zero and one boundaries; malformed/duplicate annotations; generated plan and runtime acceptance. Booleans have one decision and are unaffected. No requirement anywhere remains valid and keeps existing defaults. A property default may fill otherwise unannotated members, while a member-specific requirement overrides it.

## Invariant and acceptance

If any member has a probability requirement, every member must have one after applying the property default. Partial coverage produces a compile-time diagnostic naming the property and missing member. Complete per-member coverage, complete coverage through fallback, and no annotations all compile. Add positive and negative transform tests and run the evaluation runtime suite and native compiler tests. This issue concerns compile-time annotations, not partial answer distributions (#2417).

### LINKED ISSUE #2419: Make evaluation examples standalone and explain score criteria
## Problem
The new evaluation documentation is not a reliable first-use example. At `b686433f244529db302a7489ef45a62f94cfa659`, `website/src/content/docs/llm/evaluation.mdx` teaches choice probability with `tags.Constant`, a harder route than enum-member JSDoc `@probability`; its `Calling Jev directly` block references `triage` and `ticket` without defining either; the main example writes `frustration: 0 | 1 | 2` without showing score-level descriptions or what `criteria` becomes. `ILlmEvaluation.IChoice.criteria` and `IScore.criteria` have only terse field comments. The guide's probability prose also permits gating only some members, which conflicts with the desired complete-coverage compiler rule (#2418).

## Consequence surface
The guide, the `Probability` tag API comment, `ILlmEvaluation` declarations, the TypeSafe example source and its compiled `LocalSource` output, and the website's validator-tag index need to agree. Website guidance should not mention `tags.Constant` or `Constant<...>` anywhere. Show enum-member JSDoc `@probability` on every member when used, with a property default as the alternative. A score example should use documented numeric enum members and display the resulting ordered `criteria`; explain that a provider score indexes the criteria and `validate()` maps it back to the enum numeric value. Every copyable snippet must declare its own inputs and evaluation value.

## Acceptance
- The direct Jev snippet type-checks as a standalone example with declared `ticket`, decision type, and `triage`.
- The score example makes its criteria and mapping explicit in both guide and `ILlmEvaluation` API comments.
- No website page guides users toward `tags.Constant` or `Constant<...>`.
- The Probability example uses enum JSDoc and agrees with #2418.
- Rebuild `examples/` and the website so `LocalSource` reflects the edited source, and check the rendered guide.

This is documentation and example correctness, separate from the answer validator (#2417) and annotation compiler checker (#2418).

### LINKED ISSUE #2420: Preserve declared evaluation rounding when converting AI SDK and Jev answers
## Problem

The AI SDK evaluation contract permits a provider to declare `rounding.probabilityDecimals` and `rounding.scoreDecimals`. Its validator allows the resulting precision loss in a choice/score distribution sum and a score's weighted mean, and `experimental_evaluate()` exposes that declaration as `result.rounding`. OpenRouter's Jev Decisions adapter declares two decimals for both values. TypeSafe's direct Jev answers also use rounded probabilities and scores.

The new distribution checks being added for #2417 currently compare both quantities with a fixed `1e-6` tolerance and `ILlmEvaluation.validate()` takes only `answers`. Thus an AI SDK result already accepted by the SDK can be rejected by typia: three `.33` probabilities sum to `.99` even though they are correctly rounded to two decimals. A checked-in reproduction uses AI SDK's own mock model with `rounding: { probabilityDecimals: 2, scoreDecimals: 2 }`; the SDK accepts it, while typia returns errors at `$input.choice` and `$input.score`.

This is a typia integration error, not a reason to weaken AI SDK's validation or accept arbitrary inaccurate distributions. The declaration must be carried to the converting validator. Without one, the strict tolerance should remain.

## Acceptance criteria

- Extend the provider-neutral `ILlmEvaluation.validate` contract so callers can pass declared rounding precision without depending on AI SDK types. Keep one-argument calls valid and strict.
- Apply AI SDK's declared-precision formula to distribution sums and weighted score comparisons, including invalid-decimal handling, while retaining exact key/range checks, choice argmax, and probability minima.
- Cover accepted rounded AI SDK results and rejected malformed or undeclared-rounded results in runtime tests, with a real `experimental_evaluate()` mock-model integration test.
- Document passing `result.rounding` from AI SDK and explicit known precision for direct Jev; do not silently assume all providers round alike.
- Keep `packages/vercel` free of an unnecessary adapter API: `evaluation.questions` is already assignable to AI SDK 7's question map; prove end-to-end compatibility in tests instead.

Related: #2417.

### LINKED ISSUE #2421: Name evaluation answer conversion distinctly from typia.validate
## Problem

`typia.llm.evaluation<T>()` currently returns `ILlmEvaluation<T>.validate(answers)`. Unlike `typia.validate<T>(value)` and the `validate` methods on `ILlmStructuredOutput` and `ILlmFunction`, this method changes the data domain: it checks an evaluation-model answer map and folds boolean probabilities, choice labels, score indices, and independent set questions into a new `T` value. For example, `{ team: { type: "choice", choice: "billing" } }` becomes `{ team: "billing" }`.

The `validate` name implies the input itself is already a `T` and is only being checked. It also obscures that this specialized method is not a wrapper for typia's general validator. The guide and all examples reinforce that misconception. Because this is a new experimental API being reviewed before its next release, correct the public name now rather than documenting an inconsistent convention permanently.

## Acceptance criteria

- Give the answer-map-to-`T` operation a name that communicates checked conversion; use `decode(answers, rounding?)` consistently in `ILlmEvaluation`, the runtime helper, examples, tests, and website.
- Explain explicitly that `decode` is not `typia.validate`, does not require a second `typia.validate` call for supported decision types, and returns `IValidation<T>` with converted data on success.
- Verify every supported leaf and nested/set reconstruction path still produces valid `T`, and reject at compile time any refinements that conversion cannot guarantee.
- Keep the declared-rounding work in #2420 attached to the renamed method.

### LINKED ISSUE #2422: Reject evaluation types with refinements the answer decoder cannot enforce
## Problem

An evaluation decision type can carry typia validator refinements that its answer decoder never checks. The native `LlmEvaluationProgrammer` reads only probability tags from atomic and literal metadata, but `llmEvaluation_kind` accepts the tagged closed set. A targeted compiler fixture confirmed that this builds without an error:

```ts
typia.llm.evaluation<{
  /** Which team? */
  team: ("a" & tags.MinLength<2>) | "b";
  /** Which level? */
  level: (0 & tags.Minimum<1>) | 2;
}>();
```

If the model selects `"a"` or `0`, the decoder can return `success: true` with a value that `typia.validate<T>()` would reject. This violates the claimed `IValidation<T>` success contract and makes a second general validation appear necessary, even though the evaluation API is intended to be a complete checked conversion boundary.

## Acceptance criteria

- Audit every supported evaluation leaf, member, array, and nested-object metadata path for non-probability constraints and JSDoc constraint tags.
- Reject at compile time any refinement the evaluation decoder does not enforce; retain harmless documentation metadata and the supported probability requirement where justified.
- Add negative compiler fixtures for representative string, numeric, boolean, array, object, and nested constraints, plus positive twins for allowed metadata.
- Prove successful `decode()` output satisfies all supported type requirements without requiring an additional `typia.validate<T>()` call.

Related: #2421.

### LINKED ISSUE #2424: Use the published Vercel AI Gateway Jev model ID in evaluation examples
## Problem

The first AI SDK example in `website/src/content/docs/llm/evaluation.mdx` calls `experimental_evaluate()` with `model: "typesafe-ai/jev-latest"`. Vercel AI Gateway's published Jev model card instead gives `model: "typesafe-ai/jev"` for this exact API. The copyable first example therefore uses an unverified gateway identifier and may fail model resolution before reaching typia's evaluation code.

## Evidence and consequence surface

- Vercel's Jev model card: https://vercel.com/ai-gateway/models/jev
- The guide's gateway example is distinct from TypeSafe's direct `jev-latest` alias. Do not conflate those provider-specific identifiers.
- Search all copyable AI SDK gateway examples for the same identifier and correct them. Preserve the direct TypeSafe alias discussion only where its ownership is clear.

## Acceptance

- The AI SDK gateway example uses the published `typesafe-ai/jev` model ID.
- The guide distinguishes Vercel's gateway identifier from TypeSafe's direct model aliases.
- No unsupported `typesafe-ai/jev-latest` gateway example remains in the website or package examples.

This is a documentation integration defect separate from the compiler and answer-decoding issues in this campaign.

### LINKED ISSUE #2425: Test packed @typia/vercel alongside Jev on AI SDK 7
## Problem

`@typia/vercel` declares `ai >=6.0.0` as a peer, while Jev evaluation through `experimental_evaluate()` needs AI SDK 7. The repository's `tests/test-vercel` and packed ESM consumer both currently install AI SDK 6, so the published adapter's AI SDK 7 compatibility is not exercised. A TypeScript test that imports the two workspace source packages side by side fails because each workspace resolves a different `@ai-sdk/provider-utils` major; that is not a valid published-consumer verdict, since a packed adapter resolves its peer from the consuming app.

## Evidence and consequence surface

- `packages/vercel/package.json` peers on `ai >=6.0.0` and develops against AI SDK 6.
- `experiments/esm/package.json` installs `@typia/vercel` from a staged tarball, but only with `ai ^6.0.116`.
- `experiments/esm/test-adapter-peer-types.mjs` already asserts a single `ai` installation for the packed adapter and compiles its declarations under Node16, NodeNext, and bundler resolution.
- `experiments/esm/src/test_vercel.ts` executes a converted tool; `src/test_jev.ts` checks the evaluation round trip. The relevant CI owner is `.github/workflows/experiments.yml`.

## Invariant

The packed `@typia/vercel` adapter remains usable in an AI SDK 6 consumer and an AI SDK 7 consumer without adding a Jev-specific converter to that package. The AI SDK 7 consumer can also use `@typia/jev` and `typia.llm.evaluation<T>()` directly. Do not infer a published incompatibility from the workspace's intentionally separate development dependencies.

## Acceptance

- Run the packed ESM consumer against AI SDK 6 and 7 in CI, with peer identity and TypeScript declaration checks for each version.
- Execute the Vercel tool and Jev decoder in both runs.
- If the AI SDK 7 run reveals an adapter defect, correct the adapter and retain the test. If it passes, preserve the existing `ai >=6` peer range and document that no adapter API change was needed.

This issue is a missing integration gate, not a claim that the published adapter is currently broken.

### LINKED ISSUE #2426: JEV decoder accepts non-record answer objects rejected by AI SDK
# JEV decoder accepts answer objects that AI SDK rejects as non-records

## Problem

`typia.llm.evaluation<T>().decode()` can return `success: true` for answer maps, individual answers, and probability distributions with an inherited/custom prototype. AI SDK 7's `experimental_evaluate()` rejects each of those as an invalid response. A direct Jev caller can therefore receive a successful `IValidation<T>` from a shape outside the provider-neutral answer contract.

## Evidence

At PR #2423 head `3e16e6786a`, `_createLlmEvaluation.ts` uses `object()` to accept every non-null, non-array JavaScript object. A targeted in-memory transform of that current TypeScript module produced these results for an otherwise valid two-option choice (`a: 0.6`, `b: 0.4`):

| Custom-prototype layer | `decode()` | AI SDK 7.0.106 `experimental_evaluate()` mock |
| --- | --- | --- |
| answer map | `success: true` | `AI_InvalidResponseDataError` |
| `team` answer | `success: true` | `AI_InvalidResponseDataError` |
| `team.probabilities` | `success: true` | `AI_InvalidResponseDataError` |

Each custom object was made by `Object.assign(Object.create({ marker: true }), plainValue)`, leaving the required entries as own properties. AI SDK's `validate-evaluation.ts` restricts records to `Object.prototype` or a null prototype; typia does not. The control with ordinary object literals succeeds in both.

## Consequence surface

The root answer map, boolean/choice/score answers, choice/score probability maps, and optional rounding declaration share typia's object reader. The AI SDK response contract concerns the first five; direct TypeSafe SDK responses are JSON objects and should remain accepted. Null-prototype dictionaries (including maps with `__proto__` keys) are valid AI SDK records and must remain accepted. Error paths should stay `$input` for an invalid root and the decision accessor for invalid answers or distributions. No compiler schema change is required: this is runtime untrusted-answer validation.

During implementation, PR head `d210480700` exposed a second consequence: a proxy can pass the prototype check but throw from `get`, `getOwnPropertyDescriptor`, or `ownKeys`. With an otherwise valid boolean answer, proxies whose `get` or `ownKeys` traps throw make `decode()` throw instead of returning `IValidation`. The same reads occur at the root answer map, boolean/choice/score answers, both distribution kinds, extra answer keys, set-member answers, and optional rounding metadata. The input-reader boundary must contain these exceptions, report an error at the affected decision path (or `$input` for root-map/rounding failure), and not hide ordinary malformed-value diagnostics.

## Invariant and acceptance

Accept ordinary and null-prototype records with exactly the valid question keys and values. Reject custom-prototype objects at each answer-map layer with `IValidation` failures. `decode()` must not throw on malformed record containers or throwing reflection traps; test positive and one-axis negative controls for root, boolean, choice, score, sets, both distribution kinds, extra keys, rounding, and prototype-shaped option names. Keep AI SDK 6/7 packed consumers and direct Jev unaffected. Add a differential AI SDK 7 mock assertion for the plain-record contract and targeted typia runtime regressions for trapping proxies.

Related: #2417 covers distribution contents and consistency; this issue covers the record container itself. Implement in the unified JEV release-readiness PR #2423.

### LINKED ISSUE #2427: JEV decode can accept a probability getter that changes after validation
## Problem

`typia.llm.evaluation<T>().decode()` can return `success: true` even when the validated probability is below an enum member's declared minimum. A plain object getter can return a valid but insufficient number during validation and a string coercing to a larger number when the decoder copies the distribution. The result is then trusted as `T` despite violating the decision threshold. This affects direct Jev/TypeSafe answers and any caller that passes an answer map without AI SDK's preceding validation.

## Evidence

On commit `c497c0f3cbbdef26578a1b52b2949bceab3c6044`, `distribution()` in `packages/typia/src/internal/_createLlmEvaluation.ts` reads `record[key]` in the `isProbability` predicate and reads it again in the `assign(output, key, record[key])` loop. A plain record is accepted by `object()`, so no Proxy is required.

The focused regression was run with `pnpm --filter @typia/test-typia-schema start -- --include llm_evaluation_decode_rejects_malformed_answers` against the original two-read `distribution()` body and failed because the actual `success` was `true`, while the expected value was `false`. The original code validates `0.4`, copies `"1"`, computes an invalid sum as `NaN` (which evades the `> tolerance` comparison), and compares `"1" < 0.5` as false. Thus it returns success despite the only checked probability being below `0.5`:

```ts
enum Team {
  /** @probability 0.5 */ billing = "billing",
  /** @probability 0 */ technical = "technical",
  /** @probability 0 */ sales = "sales",
}
let reads = 0;
const probabilities = { technical: 0.3, sales: 0.3 } as Record<string, unknown>;
Object.defineProperty(probabilities, "billing", {
  enumerable: true,
  get: () => (++reads === 1 ? 0.4 : "1"),
});
const result = typia.llm.evaluation<{
  /** Which team? */
  team: Team;
}>().decode({
  team: { type: "choice", choice: "billing", probabilities },
});
// Actual: result.success === true. Expected: false.
```

An ordinary numeric `billing: 0.4` fails the minimum; `billing: 0.5` with the remainder distributed among the other members succeeds. A fixed own string is rejected. The changing getter is the distinguishing axis.

## Consequence surface and invariant

Review every untrusted field that the decoder inspects more than once: boolean discriminator and probability, selected choice, score number and distribution, distribution keys/values, and rounding declarations. Catch throwing accessors as structured errors under #2426, but also prevent changing accessors from allowing a value different from the validated snapshot to drive conversion. The decoder must validate the exact captured values it uses and return a pathful `IValidation` failure for malformed captures. Preserve ordinary records, null-prototype records, TypeSafe's native `noul` shape, and AI SDK's declared rounding behavior. This does not demand an impossible guarantee that a caller cannot mutate a Proxy after `decode()` has returned.

- The changing-getter reproduction fails on `$input.team`, and the probability getter is read only once. Valid stable numbers still decode.
- A value that changes on successive reads cannot make a score or choice bypass its range, distribution, or selected-option check.
- A getter that throws still returns a pathful failure and never escapes as an exception.
- Run the targeted evaluation runtime cases, AI SDK 7 differential tests, and the PR's test/packed-consumer CI gates.

## Coordination

Implement in the single draft JEV release-readiness PR #2423. Related: #2417 (distribution completeness), #2420 (declared rounding), and #2426 (untrusted record and trap handling). This issue addresses inconsistent multiple reads rather than the record-prototype or exception boundary.

samchon:
## Purpose

One draft PR for the JEV release-readiness campaign. It implements the accepted issues #2417–#2422 and #2424–#2433. The branch includes native compiler checks, a checked answer decoder, JSDoc probability support, AI SDK 7/Vercel consumer coverage, and updated examples and guides.

The base is `b686433f244529db302a7489ef45a62f94cfa659`. All eight GitHub checks passed on the current head, `f0cbff57e1`. The current-head solo PR-diff Self-Review found no further product defect, but the campaign-wide audit is incomplete. This draft is not release approval.

## Acceptance

- Reject incomplete or contradictory choice and score distributions while honoring provider-declared rounding.
- Reject partial member probability declarations, misplaced JSDoc tags, unsupported refinements, and decision types whose required members cannot be decoded.
- Support property and enum-member `@probability`, including the boundary value `1`.
- Expose `decode()` for checked answer-map conversion; distinguish it from `typia.validate<T>()` throughout the examples and documentation.
- Keep website examples free of `tags.Constant`, explain `criteria` and scores, and show direct TypeSafe and AI SDK calls.
- Verify packed AI SDK 6 and 7 consumers, including existing `@typia/vercel` tools and direct evaluation questions.

Local checks are intentionally focused, as requested. The PR's GitHub CI owns the full build and test lanes and does not pause Self-Review. The premerge integration manifest (SHA-256 `59d6df856ffbe3377e5d87fa71ad144d0dd41a84056b13ef7641e43b6d22e895`) is covered by the exact-head CI runs. The campaign-wide discovery census and postmerge verification remain incomplete, so the release recommendation remains **NO-GO**.

Closes #2417, closes #2418, closes #2419, closes #2420, closes #2421, closes #2422, closes #2424, closes #2425, closes #2426, closes #2427, closes #2428, closes #2429, closes #2430, closes #2431, closes #2432, closes #2433.


samchon (commented):
Commit 5c2246eea8 completes the public answer-map decoder naming and documentation for #2421, adds guarded-refinement and probability-boundary checks for #2418/#2422, strengthens distribution negatives for #2417/#2420, and corrects the published AI Gateway Jev model ID for #2424. It also makes the first website example use enum-member JSDoc @probability and gives the direct Jev example an explicit score case for #2419. Targeted 17 evaluation tests and Go tests passed; full CI and solo Self-Review are still pending.

samchon (commented):
Commit 1f81e2dfff aligns the public Jev example and evaluation test narration with the checked decode() operation introduced for #2421. It changes wording only; the prior targeted evaluation suite passed. GitHub CI and the complete solo Self-Review remain the acceptance gates.

samchon (commented):
Commit 3e16e6786a resolves #2425 by expanding the packed ESM consumer gate: it first retains AI SDK 6, then installs AI SDK 7 in the same tarball-based consumer and reruns type identity, declaration compilation, Vercel tool execution, and Jev decoding. The test checks the installed SDK major to prevent a false green run. A source-workspace cross-major probe failed due two provider-utils installations; the version-matched AI SDK 7 source probe passed, so no breaking @typia/vercel peer change was justified. The integration manifest was revised to SHA-256 35c1f25ced50dcf47f508b9bf115c7127967afea5242598cb18dd9f9bae2095c; the new CI gate is pending.

samchon (commented):
Commit 305b411f02 closes #2426: the evaluation decoder now rejects custom-prototype answer maps, answers, and distributions, while accepting plain and null-prototype records. Added a targeted decoder matrix and an AI SDK 7 mock-contract test. Focused in-memory runtime probe passed; current-head GitHub CI and full solo Self-Review are pending.

samchon (commented):
Commit 29c199e57f corrects an early-warning finding from 305b411f02: the AI SDK plain-record rule is limited to answer containers, while the optional rounding argument retains its prior object-like behavior. Added a regression and verified the focused runtime boundary. Current-head CI and a fresh full Self-Review remain pending.

samchon (commented):
Commit a4c64cecde extends #2419 documentation: score is correctly described as a fractional position on the criteria-index scale, and direct TypeSafe SDK examples no longer guess a two-decimal rounding contract. AI SDK callers still pass the provider-declared result.rounding. The correction follows current official TypeSafe Score/API docs; current-head CI and full solo Self-Review remain pending.

samchon (commented):
Commit d210480700 corrects an in-PR #2426 consequence: proxy prototype inspection and revoked-proxy array inspection now return IValidation errors instead of throwing. The focused decoder regression passed, and new or materially changed test cases now use the required scenario-comment structure. This is a Self-Review finding round, not the final clean round; current-head CI and another full round are pending.

samchon (commented):
Commit c497c0f3cb completes the expanded #2426 untrusted-answer boundary: getter, ownKeys, property-descriptor, extra-key, set-member, and rounding traps now return IValidation failures at the affected path. A focused test and in-memory before/after probes passed. The prior prototype-only correction was insufficient; this is a Self-Review finding round, not a clean conclusion. Current-head CI and a fresh full round are pending.

samchon (commented):
Commit bd8c25941d lands #2427: decoder snapshots each untrusted decision field and distribution probability once before using it. A changing getter previously bypassed an enum member minimum; the focused regression failed against the original two-read distribution and passes with the fix. Added choice and score accessor-snapshot controls. Targeted llm_evaluation_decode tests pass; full CI and fresh solo Self-Review remain pending.

samchon (commented):
Commit a92d55a945 adds a compiler provenance control under #2422: a named alias carrying a validator refinement is rejected at $input.allowed, while an annotation-only alias compiles. The focused Go transform test passed. No product behavior changed; full CI and fresh solo Self-Review remain pending.

samchon (commented):
Commit 10996f725f tightens the aliased-refinement regression oracle under #2422. It now requires the exact unsupported trueOnly validation-tag diagnostic at $input.allowed, so an unrelated compile error cannot make the test pass. Focused Go test passed; CI and a fresh solo Self-Review are pending.

samchon (commented):
Commit 3eae7fd3ff closes #2428: declaration-level @probability on reachable type aliases and object declarations now produces decision-path transform diagnostics. Regression fixtures cover alias chains, repeated and nested use, generic substitution, interface inheritance, valid property comments, and unused generic arguments. The same commit corrects #2425 packed AI SDK 7 CJS type-only import resolution after CI exposed TS1541. Focused native tests passed locally; the new PR CI head is pending. This is an implementation ledger, not a completed Self-Review or release approval.

samchon (commented):
Commit cde233a416 continues #2428 and #2425. It rejects enum-declaration probability comments, follows default generic arguments and typeof annotations, and constrains conditional, indexed, mapped, and static-member diagnostics to actual decision paths; focused Go transform matrix passed. It also removes a dangling Action reference from the public Probability example and narrows the packed AI SDK 7 tool-description type check. The preceding CI head passed all lanes except the now-corrected AI SDK 7 consumer type error. CI on this new head and fresh full-surface solo Self-Review now proceed concurrently; no release approval yet.

samchon (commented):
Self-Review follow-up for b7b9558cc1: focused compiler fixtures caught and now pass for unselected wide conditional branches, direct and array infer aliases, and generic indexed access. This commit is not a clean Self-Review verdict; I am restarting the full base-to-head solo review on this exact SHA while CI runs. Release readiness remains pending.

samchon (commented):
Self-Review follow-up for 6239b6a57a: a fresh review of b7b9558cc1 found false diagnostics for non-distributive union conditionals and missed alias annotations through inherited or aliased generic indexed properties. Negative and positive compiler fixtures reproduced the cases, then the focused JEV compiler suite passed after the fix. Starting another full solo base-to-head review on this new exact SHA while CI runs; this is not a release-ready verdict.

samchon (commented):
Self-Review follow-up for b4490d8534: on 6239b6a57a a nested generic in a non-distributive union check could make the AST audit select the wrong arm and miss a used annotated alias. Both true and false branch fixtures now pass, along with focused JEV compiler tests. Restarting the complete solo base-to-head review at this SHA while CI runs; release readiness remains unconfirmed.

samchon (commented):
Self-Review follow-up for f2469e8cca: a fresh check of b4490d8534 reproduced a false alias-level @probability diagnostic for an unselected `T[] extends number[]` arm. The compiler now compares resolved written union and array constituents while leaving unproven nested forms conservative. True- and false-branch fixtures and focused JEV compiler tests pass. Full solo base-to-head review restarts at this exact SHA; CI continues in the background.

samchon (commented):
Self-Review follow-up for 705721320f: on f2469e8cca, distributive array inference and alias-wrapped array/union inference could lose a used alias-level @probability; a union containing `any` could also select only the false arm and miss the true one. Compiler reproductions failed before correction; focused JEV compiler tests pass after written-alias expansion, inferred binding, and unknown-union handling. Restarting full solo base-to-head review on this exact SHA while CI runs. Release readiness remains pending.

samchon (commented):
Self-Review follow-up for e8ef8cbb50: 705721320f diagnosed an alias used only in an unselected inference check, while a nested alias inside a selected union inference still escaped. Positive and negative compiler fixtures reproduced both sides; focused JEV compiler tests pass after scoping alias diagnostics to selected inferred branches and expanding each union constituent. Starting another full solo review on this exact SHA while CI runs. The draft is not release-ready yet.

samchon (commented):
Self-Review follow-up for 9dd23de411: on e8ef8cbb50, tuple and object `infer` paths could lose an annotated alias used by the decoded decision, including a generic interface property and an annotated source interface. A positive tuple control confirms an inferred input discarded by the result remains accepted. Focused compiler tests pass; the full solo base-to-head review restarts at this SHA while CI runs. Release readiness is still pending.

samchon (commented):
Self-review continuation at ecff88474a: numeric indexed access now follows the selected declaration surface, with a negative alias-comment fixture and positive unused-property control. A direct infer path also no longer emits duplicate alias diagnostics. Focused Go compiler tests passed; full CI remains the independent gate. Continuing the solo review while CI runs.

samchon (commented):
Solo Self-Review on ecff88474a found a compiler provenance miss for readonly tuple and readonly array inference. c3abde2efa now follows inferred option declarations through readonly tuple, ReadonlyArray<T>, and readonly T[] syntax. Negative annotated-source tests and positive unused-inference/mutable-pattern controls passed in the focused Go suite. CI is asynchronous; restarting full Self-Review on this exact head.

samchon (commented):
Self-Review continuation at b99ce8174f: cross-form tuple/array conditional inference is now traced (readonly tuple to ReadonlyArray, ReadonlyArray to readonly array syntax, and mutable tuple to Array). Multiple tuple elements retain all inferred declaration surfaces. Parenthesized direct infer is asserted to diagnose once; unused-inference controls remain accepted. Focused Go compiler suite passed. Full CI remains asynchronous and the solo review restarts from this head.

samchon (commented):
Solo Self-Review on b99ce8174f found three more array inference cross-form misses: Array<T> to ReadonlyArray<infer U>, mutable tuple to readonly U[] inference, and readonly tuple to readonly U[] inference. a70ddf52ce now uses a shared written-element matcher while respecting readonly-to-mutable assignability. Negative and positive compiler fixtures and the focused Go suite passed. CI remains asynchronous; restarting review on the new exact head.

samchon (commented):
Solo Self-Review on a70ddf52ce reproduced a false-positive declaration diagnostic: an optional tuple source does not satisfy a required tuple infer pattern, but the syntax-only matcher treated it as selected. 425c76e84e distinguishes optional/rest and named tuple elements, with positive controls for skipped branches and negative controls for genuinely matched optional patterns. Focused Go suite passed; CI remains asynchronous and the next full Self-Review starts at this head.

samchon (commented):
Solo Self-Review on 425c76e84e found another compiler provenance miss for variadic tuple inference. f469bd1386 matches fixed prefix/suffix and rest segments, follows a selected numeric index into the inferred remainder, and compares fixed primitive pattern elements. Negative annotated-selected and positive annotated-unused controls passed the focused Go compiler suite. Full CI is still asynchronous; another exact-head solo review begins now.

samchon (commented):
Self-Review on f469bd1386 found a false-positive boundary: TypeScript permits tuple index 0, but the new tuple provenance fast path only accepted numeric 0 and otherwise traversed unused tuple elements. 39df302b5c handles canonical numeric string indices for direct and inferred tuples. Negative selected-alias and positive unused-tail fixtures, plus focused Go compiler tests, passed. CI is asynchronous; restarting solo review from this exact head.

samchon (commented):
Solo Self-Review on 39df302b5c reproduced false positives when a selected tuple element was hidden behind parentheses or a type alias. It also exposed missed declaration annotations on indexed source interfaces, including inherited property owners. 8a7a054d4d follows the selected tuple surface through wrappers and checks source/owner declarations without scanning unused elements. Focused positive/negative compiler suite passed. CI stays asynchronous; restarting review from this head.

samchon (commented):
Solo Self-Review on 8a7a054d4d found a false-positive declaration diagnostic for a tuple indexed after a generic conditional selected a clean branch. 385a99d529 follows the checker-resolved conditional branch while expanding the indexed source, with selected-annotated negative and unselected-annotated positive fixtures. Focused Go compiler suite passed. CI remains asynchronous; restarting Self-Review at this exact head.

samchon (commented):
Solo Self-Review on 385a99d529 found that conditional indexed-source expansion could select a generic type parameter but stop before applying its concrete tuple argument, falsely traversing an unused annotated member. 280d03889d applies the binding at each expansion step. Selected-annotated negative and unused-annotated positive compiler controls, plus focused Go suite, passed. Full CI stays asynchronous; restarting review from the new head.

samchon (commented):
Commit ff618498de follow-up review: verified numeric tuple indexing over open rest, reachable suffixes, fixed tuple spreads, and alias-wrapped spread/rest declaration provenance with positive and negative compiler fixtures. Focused Go evaluation suite and diff check pass. This is not a clean full Self-Review or release approval; a fresh exhaustive solo review and GitHub CI remain pending.

samchon (commented):
Commit a5bd3940cc follow-up review: a valid array-element inference followed by a conditional reproduced a native compiler panic, and a union-of-array tuple rest caused an indexed-provenance false positive. Both are fixed with selected/unselected regression controls; focused Go evaluation suite and diff check pass. This is not release approval: a fresh exhaustive solo Self-Review and GitHub CI remain pending.

samchon (commented):
Commit 994341a64e follow-up review: normalized written conditional unions (`never | false`, `unknown | true`) to avoid unreachable-branch diagnostics; selected exact positions and suffix reachability for each fixed-tuple union rest arm, including spread-alias declaration provenance. Positive and negative compiler fixtures, focused Go suite, and diff check pass. Full solo Self-Review and CI remain pending; not release approval.

samchon (commented):
Commit c32a259f67 follow-up review: nested nonempty tuple rest alternatives now carry their actual minimum/maximum lengths when deciding whether a later suffix can occupy an index. A naked distributive conditional bound to `never` now skips both unreachable branches. Positive and selected-alias negative compiler controls, focused Go evaluation suite, and diff check pass. This is not a clean full Self-Review or release approval; GitHub CI remains in progress.

samchon (commented):
Commit b946785a5d follow-up review: indexed decision access now audits its key-type declaration as well as its selected value surface. Focused fixtures reproduce the previously missed `@probability` on numeric tuple-key and string property-key aliases; an unselected conditional key is a positive control. Additional array-rest inference branch controls pass. Focused Go suite and diff check pass; full solo Self-Review and CI remain pending.

samchon (commented):
Commit d68b1b6f97 follow-up review: a probability comment on an indexed source property was silently discarded when reused as a decision through I[selected]. The compiler now reports that misplaced annotation on the selected source declaration; an annotated unselected sibling and ordinary direct decision-property comments remain valid. Focused Go evaluation suite and diff check pass. Full solo Self-Review and GitHub CI remain pending; not release approval.

samchon (commented):
Commit e0ba741311 follow-up review: indexed decisions now trace selected getters and index signatures (including inherited/generic declarations) instead of silently discarding their probability annotations. Exact number-vs-string index-signature precedence, including canonical numeric-string keys, has positive/negative compiler and TypeScript type-assertion controls. Focused Go suite and diff check pass. Prior head d68b1b6f97 had all GitHub checks green; this new head is restarting CI. A fresh exhaustive solo Self-Review remains pending, so this is not release approval.

samchon (commented):
Self-review checkpoint at 44b1147e9b: aligned indexed signature provenance with TypeScript inheritance shadowing, numeric-string canonicalization, non-string signature precedence, class/union and broad-key access, accessor annotations, and generic intersections. Added focused positive/negative compiler fixtures. Focused JEV compiler suite passes. Continuing full solo Self-Review while CI runs; PR remains draft.

samchon (commented):
Self-review checkpoint at d7406ceb0c: class implements interfaces are excluded from inherited index-signature provenance, matching TypeScript instance-type semantics. Added positive compiler control and a negative overlapping-template-signature test. TypeScript checker intersects all applicable non-string index signatures, so both annotated contributors must be audited. Focused JEV compiler suite passes; full solo Self-Review and CI continue.

samchon (commented):
Self-review checkpoint at 64cae47862: closed the remaining class implements false-positive path in general declaration traversal, not just indexed access. Added direct and nested class controls, plus numeric-vs-string template index selection controls. Focused JEV compiler suite passes. Full solo Self-Review and CI continue; PR remains draft.

samchon (commented):
Self-review checkpoint at fb6989695a: indexed accessor provenance now follows TypeScript read-type precedence (explicit getter before setter write type), and inherited class/interface members shadowed by derived declarations no longer cause false alias diagnostics. Added selected/unselected compiler controls. Focused JEV compiler suite passes; full solo Self-Review and CI continue.

samchon (commented):
Commit 93f25637ed replaces the 1,500-line manual type-AST provenance walker with a syntactic @probability placement check over the compilation program. TypeScript erases primitive-alias identity, so issue #2428 now explicitly rejects unsupported placement even on unused declarations whenever the program uses llm.evaluation. Focused Go fixtures passed for direct, imported and unused declarations, supported property/enum-member tags, and no-evaluation controls. Full PR CI and solo Self-Review are pending; this is not a release-readiness conclusion.

samchon (commented):
Review finding on 93f25637ed (not a clean Self-Review): a source-property `@probability` can still disappear through generic indexed access (`Select<T,K> = T[K]`), compiling with the default threshold. The read-only early-warning pass also found a computed-name panic, verified from TypeScript-Go's `Node.Text()` behavior. Issues #2429 and #2430 record both classes. Local fixes and failing regressions are uncommitted while I stop the syntax-case patching and settle a sound annotation-placement contract. PR remains draft and is not release-ready.

samchon (commented):
Commit 57014bfb13 keeps property-level @probability JSDoc, documents its property-local semantics across indexed access, and adds executed direct/generic/explicit-threshold controls. It also turns the computed-name placement crash into a normal diagnostic. Resolves #2429 and #2430. Focused TypeScript and Go cases passed locally; full CI and fresh whole-PR Self-Review are in progress. The PR remains draft.

samchon (commented):
Self-Review round on 57014bfb13 (base b686433f2445): NOT CLEAN. The newly documented indexed-access locality still permits an annotated source property to become an unannotated 0.5 decision through generic extraction; the committed test asserts this silent loss. That is not a release-safe resolution of #2429 under the campaign requirement to diagnose wrong probability usage. The PR stays draft. I also found test-shape drift: the new native refinements test file contains two Test functions, contrary to the repository one-case-per-file rule. Correct the contract and tests, then restart full Self-Review on a new immutable head. Focused Go and TS probes passed; CI on this head is still pending.

samchon (commented):
Commit 5c0a0d2 follows the first Self-Review round: splits the native alias-refinement case into its own file, documents new Go cases, caches the program-wide placement scan once per transform factory, and adds executed generic indexed-access tests for boolean, choice, score, and set tags. Focused Go tests and the new TypeScript case pass. The #2429 provenance concern was rechecked: property JSDoc belongs to its declaration, while `T[K]` carries the value type; the guide now gives `tags.Probability` as the portable type-level spelling without removing JSDoc support. CI and a fresh whole-PR Self-Review on this head remain pending; the PR stays draft.

samchon (commented):
Commit e0349f3f27 resolves #2431. The ttsc-typia build and transform callbacks now share one program-scoped declaration-scan cache across source files; the imported-declaration regression runs evaluation transforms in two files and requires diagnostics from both. Focused Go test passed. This is a commit ledger, not a clean Self-Review; a new whole-PR round is in progress on this head while CI runs.

samchon (commented):
Self-Review round on immutable head e0349f3f27: NOT CLEAN. I read the full base-to-head diff and traced runtime, native transform, documentation, package/CI and test consequences. The new two-file imported-declaration test is not sensitive to cache removal: it sees two diagnostics whether the program is scanned once or twice. It also does not exercise the transform CLI path. I will add a direct memoization oracle and transform-path coverage, then restart the entire review on the corrected head. CI remains background evidence, not a prerequisite for review.

samchon (commented):
Commit 721ee22174 closes the Self-Review test-oracle gap on #2431. A concurrent unit regression now counts the program-wide scan and proves one calculation per shared cache, while fresh and uncached maps recalculate. The imported-declaration fixture also exercises both build and transform CLI paths. Both focused Go packages passed. This commit ledger is not a clean Self-Review; the full round restarts on this head.

samchon (commented):
Self-Review round on immutable head 721ee22174: NOT CLEAN. After reading the complete base-to-head diff and tracing the transform dependency collector, I reproduced #2432 with a focused red test: a successful evaluation in src/main.ts is marked dependenciesComplete, but its program-wide @probability placement scan does not register unrelated src/unused.ts as an input. An edit to that file can change the diagnostic without invalidating the cached transform. I will register the scanned sources and keep the test, then restart the whole review. JSDoc property and enum-member support remains part of the contract.

samchon (commented):
Commit 40f96e53d8 resolves #2432. A red-before-green CLI regression showed src/main.ts declared complete dependencies without watching unrelated src/unused.ts even though the program-wide JSDoc scan depends on it. Evaluation transforms now report every scanned program source to the existing dependency collector; the test proves the watch edge and a new diagnostic after unused.ts gains an illegal tag. Focused Go test passed. This is a commit ledger, not the final clean Self-Review.

samchon (commented):
Solo Self-Review, clean round on 40f96e53d8c06cab1759c8e8f57dae1c70984eea (base b686433f244529db302a7489ef45a62f94cfa659): inspected the full base-to-head diff across native transformer, metadata/dependency reporting, decoder, public types, docs, fixtures, package wiring, and CI; checked JSDoc property and enum-member semantics, probability 1, partial coverage, answer records, rounding, score conversion, and Vercel AI SDK 7 consumer boundaries. No confirmed improvement remains in this PR diff. GitHub checks on this exact head are green. This is the PR Self-Review gate only; the campaign-wide fresh issue-discovery and integration/mutation gates remain separate and are not claimed here.

samchon (commented):
Commit a0ed1b80e2 links seven already implemented initial JEV campaign issues to the PR's merge-closing ledger: #2417, #2418, #2419, #2420, #2421, #2422, and #2424. This is an empty bookkeeping commit: no source or tests changed. The issue-to-diff mapping was checked before adding the closing lines. GitHub CI and solo Self-Review must be attributed to the new exact head before merge.

samchon (commented):
Solo Self-Review, clean round on exact head a0ed1b80e2b204ecf50c6cc17ef71cb9fceb94a0 (base b686433f244529db302a7489ef45a62f94cfa659): restarted the full base-to-head diff inspection across every changed native, runtime, public type, test, packed-consumer, workflow, example, and website file. Rechecked upstream AI SDK 7.0.106's installed primary validation source for complete distributions, highest choice, weighted score, plain records, and declared rounding; checked JSDoc member/property support and the complete issue-closing keyword set. No confirmed PR-diff improvement remains. This empty commit preserves the previously reviewed tree (identical tree hash 49b73e348d2b00d585717113f992316a9ed7dc85). CI on this new exact head is still running, and campaign-wide discovery/integration/mutation gates remain separate.

samchon (commented):
Commit d503dfbfb3 removes the last stale 'converting validator' wording from the source JEV example after the public operation became decode(). This is a documentation follow-up to #2421; the website build regenerates the ignored JavaScript example before rendering its code tab. Focused Prettier check passed. The initially suspected missing generated-example defect was disproved by website/package.json's build:prepare → build:examples chain; no issue was published for that false premise. CI and full solo Self-Review restart on this exact new head.

samchon (commented):
Solo Self-Review at d503dfbfb32ac1bcc8d26779c0adf5ca862eadad: reread the complete base-to-head native compiler/runtime, TypeScript API, documentation, website/example, AI SDK integration, and positive/negative test diff; traced property and enum-member JSDoc probability through metadata, compiler diagnostics, and decoding; checked diff whitespace and prohibited website Constant references. No further sound change found in this round. Current-head CI is still running; this is not a release approval.

samchon (commented):
Commit bc00d3ee74 completes #2421 terminology cleanup in two overlooked consumer-facing texts: experiments/esm/README.md now calls the native-answer round trip decode, matching its executed test_jev.ts; the key-encoding regression comment now names decode, matching the test call. No behavior or published API change. The current-head CI and a fresh solo Self-Review are pending.

samchon (commented):
Solo Self-Review clean on exact head bc00d3ee743ca347b676a5281a04d49615834fa2 (base b686433f244529db302a7489ef45a62f94cfa659): reread every changed base-to-head file/hunk across native compiler and CLI, runtime decoder, public types, tests, packed AI SDK 6/7 consumer, example, and website. Verified property and enum-member JSDoc remain accepted, probability 1 and partial-member diagnostics are covered, and the two final decode terminology corrections match executed calls. Diff whitespace clean; no Constant recommendation in website guides. No sound improvement found in this round. Current-head CI and campaign-wide discovery/integration accounting are separate gates; this is not release approval.

samchon (commented):
Commit 24ed4a487c follow-up for #2433: strict JEV-only object metadata now preserves getter/setter decisions and diagnoses required methods, private/protected or symbol members, and call/construct signatures instead of silently emitting a partial decoder. Focused TestLlmEvaluationCompleteObjectShape and go test ./cmd/ttsc-typia -run '^TestLlmEvaluation' -count=1 passed locally; the current-head GitHub CI and renewed solo base-to-head Self-Review are in progress. This is an interim COMMENT review, not approval or release readiness.

samchon (commented):
Commit 27a8af6069 addresses the pushed-commit early warning under #2433: before this fix, both ordinary and intersection /** @internal */ required properties compiled and vanished from the JEV plan. Red focused fixtures reproduced both; the JEV-only strict path now retains them until the existing hidden-property compiler diagnostic, while default metadata consumers remain unchanged. TestLlmEvaluationCompleteObjectShape passes on this head. Current-head GitHub CI and fresh solo Self-Review remain pending; this is a COMMENT review, not approval.

samchon (commented):
Commit 2cc3a23a5f follows #2433's early-warning boundary: a duplicate-key intersection with an @internal required member now has a permanent compile-error regression, and the guide states the supported accessor and rejected non-decision-member contract. The targeted overlapping-intersection test passed. Current-head CI and a fresh complete solo Self-Review are running; this is an interim COMMENT review, not approval.

samchon (commented):
Solo Self-Review round on immutable head 2cc3a23a5f (base b686433f244529db302a7489ef45a62f94cfa659; lock SHA-256 8856f4a8523fa63a193761e392758b5763e65104622cb123c258e7f759ea2e3d): I reread all 68 base-to-head changed files/hunks across public types, native metadata/compiler, decoder, tests, AI SDK 7/Vercel consumers, guides, CI, and lockfile. I traced the strict shape path through direct, inherited, nested, generic, symbol, nominal, callable, constructor, @internal, and overlapping-intersection cases; targeted compiler and metadata tests passed. The stale ignored examples/bin artifact is regenerated by website build:examples before LocalSource embeds it. No new actionable defect survived this PR-diff round. Current-head CI is still pending, and the repository-wide discovery census and mandatory matrix are not complete, so this COMMENT is not release approval or a campaign CLEAN claim.

samchon (commented):
Solo Self-Review finding on head 2cc3a23a5f: repository-wide terminology audit found two unchanged landing surfaces outside that diff, root README and website index, still calling evaluation an answer validator. That phrasing conflicts with the new answer-map-to-T decode contract (#2421). Both are corrected locally; this finding invalidates the preceding clean PR-diff round as the final gate. A new immutable-head full round and current-head CI are required after push.

samchon (commented):
Commit e66b4a5527 completes #2421's landing-page terminology follow-up: root README and website index now describe evaluation as a checked answer decoder, consistent with the renamed API and guides. The prior PR-diff Self-Review found this outside-diff improvement and was invalidated; fresh solo Self-Review and current-head CI are starting. Interim COMMENT only.

samchon (commented):
Commit a02e82d9f5 adds the explicit enum-member @probability 1 compiler boundary to the permanent JEV regression. The focused Go compiler test passed before push; full GitHub CI and a fresh solo Self-Review are pending on this new head.

samchon (commented):
Self-Review round on a02e82d9f5: the complete-object-shape compiler regression did not explicitly cover an inherited ES #private member through a generic base class. A focused probe showed strict JEV metadata already rejects it, so this is a test-coverage improvement, not a new product defect. I added the permanent negative case; its focused canonical Go test passed. Because the test changed, this round is superseded and a fresh whole-diff round is required on the next head. Campaign-wide discovery remains INCOMPLETE.

samchon (commented):
Commit f0cbff57e1 adds a permanent compiler regression for inherited generic ES #private members. JEV already rejects the case; the focused TestLlmEvaluationCompleteObjectShape command passed. This closes a coverage gap under #2433 without changing product behavior. GitHub CI and fresh solo Self-Review now target this new head.

samchon (commented):
Solo Self-Review on immutable head f0cbff57e1c007b4d6ab75a01f7711a24fb0e625: re-read the 70-file base-to-head diff across public interfaces, native metadata/compiler, runtime decoder, tests, packages, packed ESM CI, website guides, examples, and lockfile. Compared the distribution and rounding rules against the installed AI SDK 7.0.106 primary implementation; checked no website tags.Constant examples or stale evaluation.validate calls remain; the worktree and diff checks are clean. No additional sound product finding surfaced in this PR-diff pass. This is not a campaign CLEAN or release approval: the full tracked-file discovery census, mandatory interaction matrix, final integration accounting, and exact-head GitHub CI have not all completed. Keep the PR draft.

samchon:
Exact-head CI result for f0cbff57e1: all eight GitHub checks passed, including Ubuntu compiler/runtime, website build, spelling, Socket, and packed-consumer ESM/declaration checks. This is CI evidence, not campaign completion or release approval; the full-scope census/matrix and final integration gate remain incomplete. PR remains draft and recommendation NO-GO.

samchon:
Integration accounting for exact head f0cbff57e1: frozen manifest SHA-256 59d6df856ffbe3377e5d87fa71ad144d0dd41a84056b13ef7641e43b6d22e895. test.yml run 35627616228 passed root pnpm test (JEV, interface, evaluation schema and native Go suites); experiments.yml run 35627616236 passed packed AI SDK 6/7 ESM, Vercel/JEV, MCP and declaration consumers; build.yml 35627616231, website.yml 35627616220, and format.yml 35627616241 passed. This closes ordinary premerge CI integration accounting only. Full-scope issue-discovery rows and postmerge clean round remain incomplete; PR stays draft / release NO-GO.

samchon (commented):
Final pre-merge gate on f0cbff57e1c007b4d6ab75a01f7711a24fb0e625: complete solo PR-diff Self-Review found no further sound product defect; all eight exact-head GitHub checks passed; premerge integration manifest 59d6df856ffbe3377e5d87fa71ad144d0dd41a84056b13ef7641e43b6d22e895 is covered by the test, build, website, format, packed ESM AI SDK 6/7, declaration, and MCP CI lanes; regression mutations M1-M3 were red/green. origin/master is still the frozen base b686433f244529db302a7489ef45a62f94cfa659, so the prospective squash content is this CI-tested head. The user explicitly requested merging now and previously delegated full build/test to GitHub CI rather than local reruns. The broader postmerge issue-discovery campaign and release approval remain separate.

